PR #45, opened 2014-11-03
Allow setting server span name.

joeslice:
You may acquire new information during the method invocation (i.e., Jersey's @Path annotations of the chosen method), etc.  This allows you to use that information before finally submitting the thread.

Additionally, update some javadoc comments.


joeslice:
Closing this PR.

PR #61, opened 2015-02-05
Demo version of Apache CXF interceptors with Spring feature to install

fedor57:
Hi, Kristof!

I've just finished some new interceptors for Apache CXF client and server as well as Spring feature to configure them. My task was to prove an ability to use Zipkin in our environment.

Well, this is my first Java code ;) I know it's not perfect, it needs to be refactored to better integrate in Brave library structure, add some tests and safety checks. Unfortunately, my test clients/servers are tied hardly to out environment so I will not able to publish them. 

Do you think it's worth to include this in Brave master in some form?

Do you think you can spend some time for taking ownership and refactoring the proposed code? If not my colleagues who occupied with full-time programming will do something some day.

### DIFF
--- a/brave-cxf/README.md
+++ b/brave-cxf/README.md
@@ -1,0 +1,25 @@
+Installs in/out interceptors for Apache CXF client/server request/respone.
+
+Spring ZipkinFeature allows to attach Zipking tracing support by
+adding dependency on brave-cxf and declaring configurable Spring feature that
+installs interceptors.
+
+Installing ZipkinFeature might require selecting relevant Spring contexts, e.g. separately
+for server and client part.
+
+    <bean id="zipkinFeature" class="com.github.kristofa.brave.cxf.ZipkinFeature">
+        <property name="useLogging" value="false"/>
+        <property name="useZipkin" value="true"/>
+        <property name="collectorHost" value="localhost"/>
+        <property name="collectorPort" value="9410"/>
+         <!-- property name="clientServiceName" value="pingsrv2"/> -->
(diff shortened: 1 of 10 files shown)


kristofa:
Hi @fedor57, 

Thanks for this contribution. Integration with new frameworks is definitely valuable and I would love to merge it into master. I'll have a closer look at the code  soon and come back to you.

fedor57:
Thanks, Kristof! 

Actually - I think that the question that has to be resolved is a process of creating human readable span/service names from URLs or other data. I've read in some Brave commit comments about plans to make algorithm standard for better matching of Server and Client span names. I think that using too simple algorithm could significantly spoil product quality of the tracing in some cases. In my tests I saw something like ("api", "/api/service1/method1"). That's why I wrote some processing in SpanAddress, but something similar should be done on the whole Brave library level.

I think there be some solutions like:
- smart default algorithm
- some descriptive url configuration like in url mappings
- or user-defined java url parsers

At the same time I don't understand the problem if client or server description for the same spanId mismatch. I expected that spanId precisely identifies span and we may use either name if they mismatch. In the logically consistent case the span/service name for traced "CS" event might not be defined at all.

But the behaviour in case of mismatch may be dependent on original Zipkin code, right? If so does it conform with Dapper paper? Should Brave just omit sending span names with CS as a workaround? 

What do you think?

fedor57:
Created a new pull request with significantly refactored code

fedor57:
Created a new pull request with significantly refactored code

PR #65, opened 2015-02-20
create TransportProvider abstraction and rewrite SpanCollector to use th...

cbotiza:
I created the ClientTransportProvider abstraction and default, socket based implementation.
Deprecated the socket specific constructor (although the implementations are still valid).
The broader goal is to be able to provide alternative transports, such HTTP/Thrift

Please review. I will soon provide the HTTP transport implementation

### DIFF
--- a/brave-zipkin-spancollector/pom.xml
+++ b/brave-zipkin-spancollector/pom.xml
@@ -42,6 +42,11 @@
           <artifactId>commons-lang3</artifactId>
           <version>3.3.1</version>
       </dependency>
+      <dependency>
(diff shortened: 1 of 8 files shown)


cbotiza:
Created the HTTP Transport implementation as well

kristofa:
I did a bad job by not responding sooner to this pr, sorry for that! 

As far as I know we don't have a http back-end for submitting spans and if so I would like it better if this would be implemented as part of another sub project with its own set if dependencies and not mix it in the existing sub project.

PR #67, opened 2015-03-06
Apache CXF interceptors with Spring feature to install, Root request mode

fedor57:
Hi, Kristofa,

I have updated the library a bit. Some refactoring, safety plus I have added a Root request mode. The functionality is the following:
- disable server part of CXF interceptors
- automatically create SpanId if it does not exist, so all client calls with go to the same trace tree
- two API methods StartRoot() and DoneRoot() to make sure the root span is correctly formed

This is to be used in case current web request is not instrumentated or working from command line utility. I haven't found the same code in Brave. Probably some of the new classes should be moved to brave-client. Please take a look.


brave-impl-spring/pom.xml
```diff
@@ -24,6 +24,7 @@
    </licenses>
   <properties>
     <project.build.sourceEncoding>UTF-8</project.build.sourceEncoding>
+      <maven.test.skip>true</maven.test.skip>
```
kristofa: Tests should not be skipped!

brave-jersey2/pom.xml
```diff
@@ -23,6 +23,7 @@
     <properties>
         <project.build.sourceEncoding>UTF-8</project.build.sourceEncoding>
         <jersey.version>2.13</jersey.version>
+        <maven.test.skip>true</maven.test.skip>
```
kristofa: Tests should not be skipped.

brave-resteasy3-spring/pom.xml
```diff
@@ -24,8 +24,10 @@
   <properties>
     <project.build.sourceEncoding>UTF-8</project.build.sourceEncoding>
     <resteasy.version>3.0.8.Final</resteasy.version>
-    <jetty.version>8.1.9.v20130131</jetty.version>
-     <spring.version>3.2.2.RELEASE</spring.version>
+      <jetty.version>8.1.9.v20130131</jetty.version>
+      <!--<jetty.version>9.3.0.M1</jetty.version>-->
+     <spring.version>4.0.5.RELEASE</spring.version>
+          <maven.test.skip>true</maven.test.skip>
```
kristofa: Tests should not be skipped.

brave-tracefilters/pom.xml
```diff
@@ -24,6 +24,7 @@
    </licenses>
   <properties>
     <project.build.sourceEncoding>UTF-8</project.build.sourceEncoding>
+    <maven.test.skip>true</maven.test.skip>
```
kristofa: Tests should not be skipped.

kristofa:
I won't merge this in its current state:

* There are no tests for any of the classes...
* Tests are disabled for several modules.
* I don't understand `root request mode`. Brave already has functionality, implemented in ServerTracer, to start a new span/trace when no existing span information is received (=when client has no brave/zipkin support). So what is the difference?
* I find it confusing that the implementations of client/server are combined in a single class, for example in `InZipkinInterceptor`.

brave-tracefilters/pom.xml
fedor57: Right.

fedor57:
Yes, I don't ask you to merge it as it is. Maybe if you can include some of the code after refactoring - that would be nice, I'll update the rest at my side so there will be less code divergence from master.

Unfortunately I am not confident enough in all the frameworks to be able to write integration tests. But possibly I will add at least some test myself.

Regarding "root request mode". I've found out that ClientTracer is working incorrectly if nothing has been done with ServerTracer, that may happen by some reasons that may include misconfiguration:
- all client requests fall apart into different traces
- sometimes I had NullReferenceException in Endpoint copy constructor that is assuming that Endpoint has been initialized already.

The main reasons to write a wrapper for me were the following:
- I need an ability for client part to work well automatically in case Zipkin is configured via Spring and no new code were inserted into the client
- There is no public methods on ServerTracer that may check if the root span is already initialized. And no method to create it only if it does not exist. This is a kind of standard Zipkin logic that is not yet implemented clearly in Brave. To include all those details including several calls to ServerSpanThreadBinder and EndPointSubmitted in every client is not a right thing to do.
- The big part for ZipkinConfig is not an interception but getting configuration and initializing SpanCollector. This is not good to be duplicated. And the client config must be short, only on ZipkinFeature with a minimum of settings. Client does not have to know much about instrumentation internals.

Regarding In/Out interceptors. I also had trouble to understand how to do it separately. Apache CXF is a library on a very high level of abstraction with unidirectional paradigm regarding connections stuff. During HTTP call nothing differ much on client and server side, both are similar endpoints that only differ by 1 bit set in Message. Interfaces for interception are the same. Moreover you can't intercept, e.g. "client send" and not get "server send" into the same function. The difference is only understood but calling "isRequestor()" at runtime.

Also if I use ZipkinFeature in a Spring configuration there is some unclear dependency on Spring contexts. Different application use different approaches to load Spring configs, some use inheritance and etc. So you don't know exactly if a particular ZipkinFeature declaration with affect client connections, server connections or both.  

So my other possibility were to create 4 interceptors: two "in" and two "out". Two "in" interceptors will be configured identically and receive the same messages and deciding by message content only if it should somehow process it.

That's why I decided to merge it. Given that both have absolutely the same configuration.

brave-resteasy3-spring/pom.xml
fedor57: Sorry for that, that change has sneaked into my commits. I had a test collision with our production tests possibly because of usage of the same ports or something like this if I remember right. This file changes are not to be included.

PR #73, opened 2015-06-08
Reverts main code to Java 7; Adds animal-sniffer to verify

adriancole:
Java 8 bytecode is supported in many runtimes, and is a reasonable path
for us to take. However, mandating it for all code will limit adoption.

Few who use non-java languages or alternate functional libraries will be
able to leverage java 8 Optional for some time, and it will be more of a
chore to them than nullable. Scala 2.10, other JVM languages and
functional libraries do not support Java 8 bytecode. Even when options
are available, many apps haven't moved their dependency graphs over.
Finally, some CI processes do not run on JDK 8, yet.

For the sake of adoption, it seems best to at least temporarily decouple
our main source tree from requiring Java 8. Otherwise, we'd need another
branch, just for Java 7, or accept some users can't use brave 3 until
their dependency tree, CI processes and deployment are ready for Java 8.

This constrains the main source tree to Java 7 using maven compiler
configuration. It also verifies Java 8 libraries aren't in use in main
code via animal-sniffer. Using maven or animal-sniffer overrides, we can
allow source (such as server code) to use Java 8, while still being
cautious about shared library code.

### DIFF
--- a/brave-impl/src/main/java/com/github/kristofa/brave/ClientRequestAdapter.java
+++ b/brave-impl/src/main/java/com/github/kristofa/brave/ClientRequestAdapter.java
@@ -1,8 +1,6 @@
 package com.github.kristofa.brave;
 
-
 import java.util.Collection;
-import java.util.Optional;
 
 public interface ClientRequestAdapter {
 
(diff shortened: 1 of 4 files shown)


kristofa:
Thanks a lot! Also like the very elaborate and descriptive commit message!

PR #74, opened 2015-06-10
Completes migration off Java 8; Consolidates use of commons-lang3

adriancole:
This completes migration off Java 8, particularly by enabling the
animal-sniffer plugin, left disabled by accident.

This moves use of Optional to javax.annotation.Nullable. The latter
annotation will be read by auto-value's annotation processor in the near
future.

This also consolidates explicit use of commons-lang3, most of which will
also be removed via auto-value generated code.

See #73


kristofa:
Nice job! 
Read about Google AutoValue. Looks interesting.

PR #75, opened 2015-06-10
Removes dependency on commons-lang3 for Base64

adriancole:
This is part of a few changes needed to remove an explicit dependency on
commons-lang3. In this change, we host a local base64 decoder, copied
from okio (under ASL). This decoder is only used once and is 50 loc.

Tests do not have a Java 7 constraint, so can use the built-in Java 8
Base64 codec as opposed to the one from apache commons.


kristofa:
:+1:

kristofa:
It looks like it can't be merged automatically. Can you fix this? Thanks.

adriancole:
ok merge button should work now :)

PR #76, opened 2015-06-21
Flattens Impl classes with AutoValue and Removes commons-lang3

adriancole:
In many cases, immutable class trees end in an only child leave,
    separating an interface from its only implementation. In other cases,
    an Impl suffix was present on types that are unambiguous.
    
    This change flattens some of this, employing google AutoValue. The
    result is a net reduction in complexity and lines of code.
    
    This will break some folks who depend directly on Impl classes or their
    constructors. However, 3.x release is already breaking and accomodating
    these changes is largely mechanical.
    
    Finally, this removes commons-lang3 using AutoValue and some inlining.


adriancole:
other small things in this change: `s/EndPoint/Endpoint`
also, use powermock to freeze time as opposed to subclassing

adriancole:
updated to remove commons-lang3 as a main dep in brave-code.

this leaves only a couple questionable things left: httpcomponents and slf4j. I'm of the (perhaps misguided) opinion that core needs neither. We could probably simply remove the httpcomponents dep, and convert slf4j to jul.

adriancole:
ok and now removing commons-lang3 as a main dep! I think this is the most I feel comfortable having in flight :P

brave-core/src/main/java/com/github/kristofa/brave/ClientTracer.java
```diff
@@ -8,36 +17,129 @@
...
+        return new AutoValue_ClientTracer.Builder();
+    }
+
+    @Override
+    abstract ClientSpanAndEndpoint spanAndEndpoint();
+    abstract Random randomGenerator();
+    abstract SpanCollector spanCollector();
+    abstract List<TraceFilter> traceFilters();
+
+    @AutoValue.Builder
```
kristofa: Nice. Did not know AutoValue also supports Builder pattern.

kristofa: Just read the documentation saw it is new since v1.1
Very nice :)

kristofa:
@adriancole Thanks for this major contribution!

[Google AutoValue](https://github.com/google/auto/tree/master/value) is really nice. And I agree about the interfaces / Impl simplification.
I'm not a fan of powermock but for this `System.currentTimeMillis()` case it is fine.

We have to look at the `httpcomponents` dependency. I can't see why we need a direct dependency on it in `brave-core` (it seems to be pulled in by `libthrift` though)

When it comes it `slf4j` I don't mind actually. At least we do the right thing by not defining compile time dependency on a specific binding.

Thanks again! Ready to merge!

PR #77, opened 2015-06-21
Makes brave-core dependency free

adriancole:
This reworks code and dependencies such that brave-core has a single
dependency on apache-thrift. This decouples us from changes in thrift
and also allows us to consider alternate thrift or otherwise codecs,
such as json or protobuf.

The key changes are local copies of validation code, and a switch from
slf4j to jul. Eventhough slf4j is used when you are using thrift
runtimes, it isn't used in codec. In other words brave-core works
without slf4j.

This also removes the jsr305 dep by using our own Nullable and switching
from PreDestroy to Closable (which are equivalent in Spring Framework).

Finally, this corrects the httpcomponents dependency setup such that it
properly masks and couples to brave-http.

### DIFF
--- a/brave-apache-http-interceptors/pom.xml
+++ b/brave-apache-http-interceptors/pom.xml
@@ -24,29 +24,18 @@
   </properties>
   <dependencies>
     <dependency>
-       <!-- For @javax.annotation.Nullable -->
-       <groupId>com.google.code.findbugs</groupId>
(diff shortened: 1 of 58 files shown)


kristofa:
@adriancole that's quick! 
Going to have a closer look later today.
Thanks for the effort!

kristofa:
:+1:

PR #78, opened 2015-06-23
Enable customising the client provider on subclasses of the Zipkin span collector

wjam:
The use case for this is when looking up the Zipkin collector service is more complicated than a static address and port.


adriancole:
sorry this was overlooked.. are you still interested in this change?

adriancole:
haven't heard back, so closing. happy to re-open if need be.

Issue #66, opened 2015-03-03
Client interceptors depend on server Endpoint set 


fedor57:
Hi,

I have a test with a structure Browser -> Service2.Method2 -> Service1.Method1

On Service2 by some reason interceptors are installed separately for receiving the request for Method2 and sending the request for Service1.Method1 as a client. Those interceptors use ServerAndClientSpanStateImpl. 

If I don't install interceptors for receiving request for Method2 the following happens:
- ClientRequestInterceptor.handle is called by Out interceptor for client request
- it tries to submit "request" binary annotation by clientTracer.submitBinaryAnnotation
- it calls ClientTracerImpl.getClientEndPoint()
- it checks serviceName and if it is not null, tries to create an EndPoint using copy constructor
- the last one assumes that Endpoint was already stored in ThreadLocal because probably we've sent some annotations like "sr" already and so filled the Endpoint
- but in my case the Endpoint was not submitted yet, it has null value, so I get "null reference exception" inside copy constructor
- just adding a test if (endPoint != null) probably will not work well since the serviceName on client will be lost. And it is not clear what should be passed as "spanname", "ip" if I use client only and override serviceName? Is EndPoint capable of storing such a state and Zipkin to process it?

I don't understand what is the source of problem here:
1) incorrect getClientEndPoint implementation?
2) incorrect logic in ClientTracerImpl in an untested scenario, so it incorrectly assumes that EndPoint must be already submitted without testing it?
3) incorrect usage of class ServerAndClientSpanStateImpl in the case when the Method2 is not traced itself so I have to use only ClientState and ClientTracer?

I hope this is 1 or 2. If it's 3 that's bad since I will have to create a different code for client tracing depending on tracing of current method. 

P.S. Also - the whole algorithm for providing service/span names from client and server side is not clear, including the purpose of X-B3-Span-Name header. That happens if they don't match on client/server side. And what are the best practices? I think it's worth to have it all described in one readme.

Thank you for an attention!

kristofa:
Hey @fedor57 ,

So far with all implementations we have the endpoint set as part of the server side integration code.
This does not mean however that we should fail with `NullPointerException` at client side when endpoint has not been set.  This should get fixed.

I also agree that we should be able to start a span from server or client side. Currently it looks like we have a dependency on the server side.

[...]

Issue #59, opened 2015-01-18
Brave 3.0.0


kristofa:
While brave works quite good I think it is time to rework some parts and to remove some legacy code.
Here is a list of improvements I'm thinking about doing:

1.   Introduce `brave server abstraction` similar as the client abstraction @srapp implemented ( #27 ). Just like having the client abstraction having the server abstraction that hides `ServerTracer` will make it less effort to integrate and improve consistency. `ClientTracer` / `ServerTracer` can still be used for non http integrations (eg database). 
2.   Update documentation on how to use brave client / server abstractions. 
3.   Get rid of complexity of dealing with span names. Currently when integrating with http frameworks we use the path part of the URL as the span name. This resulted in inconsistencies ( #32 ) between client/server and lots of effort to clean paths that contain variable content ( #33 ).  Instead I would like to follow the approach of Zipkin / Finagle which uses the http method as span name (GET, PUT, POST,...) and adds the url path unchanged as a binary annotation to the span.
4.  Compatibility with Finagle. Make sure mixing Finagle / Zipkin and Brave services works. Previous point is also related to this. 
5.  Upgrade to Java 7. Java 6 is not supported anymore and Java 7 is getting pretty old already. Require at least Java 7.

These are the main topics I can think of now. Those changes will change the preferred way of integrating with Brave and the visualization of traces in `zipkin-web` that's why version 3.0.0

kristofa:
Feel free to give your input. @joeslice , @henrikno , @K-Jo , @srapp , @wjam , @eirslett , @michaelsembwever , @hzhaofb , @ryantenney.

eirslett:
1. Are you sure you want to use HTTP as the "default", instead of having brave-http-client, brave-http-server etc.? We run a lot of Thrift microservices where we don't need/want HTTP dependencies.
2. Great! The more documentation, the better!
3. Sounds good!
4. That's great! One problem is that you can't mix Finagle clients and Brave clients within the same application - but I guess the most important aspect is that different services within the platform should be able to communicate/trace independent of Zipkin client implementation.
5. Good idea.

Sounds like there will be no breaking changes (to the public API). That's good :-) Since we have a few Brave calls from userland code, not just in commons libraries.

joeslice:
Awesome, looking forward to seeing these changes.  These all sound reasonable @kristofa .  I would add to the list of nice updates:

[...]

PR #80, opened 2015-06-24
Fix NPE with MySQL when EndPoint not submitted yet

wjam:
If the endpoint hasn't been submitted yet, then you cannot reliably trace client requests (e.g. submitBinaryAnnotation relies on the EndPoint being present)


kristofa:
Thanks @wjam. There is another bug related to failure when Endpoint is not set: #66 
I'l try to do a more structural fix for Brave 3.0.0 #59. But for now this makes sense.

Are you using `brave-mysql` in production?

Issue #85, opened 2015-07-22
HttpClientRequestAdapter sends sampled header through as true/false rather than 0/1


wjam:
As commented in commit b71d7434399c4b77e8ee4dc45715553ca5782711, the sampled header should be 0 or 1 rather than true or false for it to be compatible with the Zipkin spec.

In `com.github.kristofa.brave.http.HttpClientRequestAdapter`, the sampled header gets set to true or false rather than 0 or 1.

adriancole:
care to raise a pull request with a fix and a test case?

kristofa:
This is merged. Closing.

PR #86, opened 2015-07-22
Issue #85 - Make client conform to Zipkin standard of 0 or 1

wjam:
This change makes Brave conform to the Zipkin standard of sending 0 or 1 rather than false or true to indicate whether tracing should be enabled or not.

Note that this will be a 'slightly' breaking change in that any downstream service which is on a previous version of Brave will think that the request isn't being traced


adriancole:
:shipit: beautiful!

kristofa:
@wjam The compatibility issue between brave 2.x and brave 3.x was why I did not change the client side yet.  But I think that it is worth doing it now as being compatible with the original Zipkin implementation is important and we have a major version bump anyway... so I guess the time is now.  I merged your changes (had some conflicts as I did some changes as well in the meanwhile).

Thanks!

PR #88, opened 2015-07-29
Add support for Spring RestTemplate interceptors

wjam:
By using this interceptor rather than the Apache httpcomponents one, it allows the use of Spring MockRestServiceServer class to test that Brave has been integrated correctly.

This also removes an unneeded dependency from brave-http.


kristofa:
@wjam Thanks for this! Two remarks:
   *  My current thinking is that `brave-core-spring` is only used to setup/configure Spring dependency injection for Brave. We could put your interceptor in a separate sub project?
   *   You depend on classes in `com.github.kristofa.brave.client` package which I consider as being deprecated.  All future proof re-usable http integration code lives in `com.github.kristofa.brave.http`. See [README.md of brave-http](https://github.com/openzipkin/brave/blob/master/brave-http/README.md) for details. Could your rewrite the interceptor using functionality in `com.github.kristofa.brave.http`?  I already ported `brave-resteasy-spring` which you can use as an example.

wjam:
Updated PR with changes from code review

brave-spring-resttemplate-interceptors/src/test/java/com/github/krisofa/brave/client/BraveClientHttpRequestInterceptorTest.java
```diff
@@ -0,0 +1,128 @@
...
+import static org.mockito.Mockito.inOrder;
+import static org.mockito.Mockito.mock;
+import static org.mockito.Mockito.when;
+
+public class BraveClientHttpRequestInterceptorTest {
+
+    private final ClientTracer clientTracer = mock(ClientTracer.class);
+    private final String serviceName = randomAlphanumeric(20);
+    private final SpanNameProvider spanNameProvider = mock(SpanNameProvider.class);
+    private final BraveClientHttpRequestInterceptor subject = new BraveClientHttpRequestInterceptor(new ClientRequestInterceptor(clientTracer),
```
kristofa: @wjam I think you can improve the test by mocking `ClientRequestInterceptor` and `ClientResponseInterceptor`. Now you are going one layer deeper by mocking `ClientTracer` which means you are also testing / verifying the interceptor which seems unneeded.

kristofa:
@wjam I'll merge this. I think it is good that we have Spring RestTemplate integration. I might do some cleanup in the test I mentioned later today.

PR #95, opened 2015-09-05
Promote the servlet filter from the jersey module to a top level module.

oliver-schoenherr:
I belief we should promote the (hidden) servlet filter server integration to a top level module.

Even in the case of using Spring MVC, i prefer the Servlet Filter, because of the Servlet Filter needs only configured once and not for every Spring Dispatcher Servlet  - like the Spring Handler Interceptor. And the Filter catch unmapped Requests too. The Handler Interceptor executes only if a controller is mapped. And so most of the time we miss 404s.

Maybe we should consider the Servlet Filter implementation as the one and only server integration.


kristofa:
Yes, agree. The servlet filter is not Jersey specific and deserves being in a separate top level project. One remark, can you squash both commits into 1? It looks like build fails with only first commit? Thanks!

PR #98, opened 2015-09-17
Removes support for Annotation.duration and associated apis

adriancole:
While this is a breaking change, I can't find any uses even with global GitHub
search, apart from an experiment and spring-cloud-sleuth, which I'll address.

If we ever want to remove this, it would be best to do so before 3.0 is final.

See https://github.com/openzipkin/zipkin/pull/717


adriancole:
don't merge this until openzipkin/zipkin#717

PR #99, opened 2015-10-08
Fixes accidental builder invisibility

adriancole:
I believe these methods weren't meant to be hidden

### DIFF
--- a/brave-core/src/main/java/com/github/kristofa/brave/ServerTracer.java
+++ b/brave-core/src/main/java/com/github/kristofa/brave/ServerTracer.java
@@ -62,7 +62,7 @@ public abstract class ServerTracer extends AnnotationSubmitter {
          */
         public abstract Builder traceFilters(List<TraceFilter> traceFilters);
 
-        abstract ServerTracer build();
+        public abstract ServerTracer build();
     }
 
     /**


brave-core/src/main/java/com/github/kristofa/brave/ServerTracer.java
```diff
@@ -62,7 +62,7 @@ public Builder state(ServerSpanState state) {
          */
         public abstract Builder traceFilters(List<TraceFilter> traceFilters);
 
-        abstract ServerTracer build();
+        public abstract ServerTracer build();
```
kristofa: Yes, `build()` method should be public.

brave-core/src/main/java/com/github/kristofa/brave/ServerTracer.java
```diff
@@ -47,7 +47,7 @@ public Builder state(ServerSpanState state) {
             return spanAndEndpoint(ServerSpanAndEndpoint.create(state));
         }
 
-        abstract Builder spanAndEndpoint(ServerSpanAndEndpoint spanAndEndpoint);
+        public abstract Builder spanAndEndpoint(ServerSpanAndEndpoint spanAndEndpoint);
```
kristofa: I don't think this needs to be public as it does not need to be set by a user of the api to build a proper `ServerTracer`? `ServerSpanAndEndpoint` is an abstraction on top of `ServerSpanState` which can be configured using the Builder.

adriancole: Agreed (I think ;) )

kristofa:
@adriancole I force pushed with only `build`public. I noticed in [the meanwhile there was already a fix](https://github.com/openzipkin/zipkin-java/commit/c3f786ac7c060edbfe49d899f07b4250b7eac2d8) to use `Brave`to get access to ServerTracer.

kristofa:
If this change is still urgent I`ll make a bug fix release tomorrow morning. Otherwise it will be released later. What do you think?

adriancole:
Not currently urgent. MySQL is awkward and might need some change. Maybe
best to hold any release a day.

PR #101, opened 2015-10-15
Span collector metrics

pbetkier:
Allows to monitor span collector's performance by implementing a single interface. Leaves the developer with free choice of which parameters to monitor, how to call these metrics and which metrics library to use.

What do you think?

### DIFF
--- a/brave-spancollector-scribe/README.md
+++ b/brave-spancollector-scribe/README.md
@@ -1,10 +1,12 @@
-# brave-spancollector-scribe #
+# brave-spancollector-scribe
 
 SpanCollector that is used to submit spans to a Scribe compatible source.
 This can be Scribe itself or the [Zipkin Collector Service](https://github.com/openzipkin/zipkin/tree/master/zipkin-collector-service)
 or another Scribe compatible source like Flume configured with Scribe source.
 
 ![Scribe SpanCollector overview](https://raw.github.com/wiki/kristofa/brave/ZipkinSpanCollector.png)
+
+## Performance
 
 The `ScribeSpanCollector` is build in such a way that it has no or very minimal impact on your application performance:
 
@@ -15,6 +17,17 @@ This approach has again been chosen to minimize the impact on the application. H
 *    The `SpanProcessingThread` does not submit every individual span immediately to the back-end service. It buffers spans and sends them in batches as much as possible.
 However it makes sure that it does not keeps holding onto spans. If the buffer is not full after 10 seconds it sends the received spans in any case.
 
+## Monitoring
 
-If you use this SpanCollector you can reuse the the Zipkin back-end (zipkin-collector-service, Cassandra back-end store, zipkin-query, zipkin-web).
+Monitor `ScribeSpanCollector`'s performance by providing your implementation of `ScribeCollectorMetricsHandler`. It will
+be notified when a span is accepted for processing and when it gets dropped for any reason, so you can update corresponding
+counters in your metrics tool. When a span gets dropped, the reason is written to the application logs.
+The number of spans sent to the target collector can be calculated by subtracting the dropped count from the accepted count.
+
+Refer to `DropwizardMetricsScribeCollectorMetricsHandlerExample` for an example of how to integrate with 
+[dropwizard metrics](https://github.com/dropwizard/metrics). 
+
+## Zipkin integration
+
+If you use this SpanCollector you can reuse the Zipkin back-end (zipkin-collector-service, Cassandra back-end store, zipkin-query, zipkin-web).
 For information on how to set up the Zipkin backend components see [here](https://github.com/openzipkin/zipkin).
--- a/brave-spancollector-scribe/src/main/java/com/github/kristofa/brave/scribe/SpanProcessingThread.java
+++ b/brave-spancollector-scribe/src/main/java/com/github/kristofa/brave/scribe/SpanProcessingThread.java
@@ -41,6 +41,7 @@ class SpanProcessingThread implements Callable<Integer> {
     private final BlockingQueue<Span> queue;
     private final ScribeClientProvider clientProvider;
     private final TProtocolFactory protocolFactory;
+    private final ScribeCollectorMetricsHandler metricsHandler;
     private volatile boolean stop = false;
     private int processedSpans = 0;
     private final List<LogEntry> logEntries;
@@ -52,12 +53,14 @@ class SpanProcessingThread implements Callable<Integer> {
      * @param queue BlockingQueue that will provide spans.
      * @param clientProvider {@link ThriftClientProvider} that provides client used to submit spans to zipkin span collector.
      * @param maxBatchSize Max batch size. Indicates how many spans we submit to collector in 1 go.
+     * @param metricsHandler Handler to be notified of span logging events.
      */
     public SpanProcessingThread(final BlockingQueue<Span> queue, final ScribeClientProvider clientProvider,
-        final int maxBatchSize) {
+        final int maxBatchSize, ScribeCollectorMetricsHandler metricsHandler) {
         if (maxBatchSize <= 0) throw new IllegalArgumentException("maxBatchSize must be positive");
         this.queue = checkNotNull(queue, "Null queue");
         this.clientProvider = checkNotNull(clientProvider, "Null clientProvider");
+        this.metricsHandler = checkNotNull(metricsHandler, "Null metricsHandler");
         protocolFactory = new TBinaryProtocol.Factory();
         this.maxBatchSize = maxBatchSize;
         logEntries = new ArrayList<LogEntry>(maxBatchSize);
@@ -131,6 +134,7 @@ class SpanProcessingThread implements Callable<Integer> {
                 LOGGER.warning("Logging spans failed (couldn't establish connection). " + logEntries.size() + " spans are lost!");
             }
         }
+        metricsHandler.incrementDroppedSpans(logEntries.size());
         return false;
     }
 
--- a/brave-spancollector-scribe/src/test/java/com/github/kristofa/brave/scribe/ScribeReceiver.java
+++ b/brave-spancollector-scribe/src/test/java/com/github/kristofa/brave/scribe/ScribeReceiver.java
@@ -30,11 +30,7 @@ class ScribeReceiver implements Iface {
 
     private static final Logger LOGGER = Logger.getLogger(ScribeReceiver.class.getName());
     private final BlockingQueue<Span> spans = new LinkedBlockingQueue<>();
-    private final int delayMs;
-
-    public ScribeReceiver(final int delayMs) {
-        this.delayMs = delayMs;
-    }
+    private volatile int delayMs;
 
     public void clearReceivedSpans() {
         spans.clear();
@@ -81,4 +77,11 @@ class ScribeReceiver implements Iface {
         return result;
     }
 
+    public void setDelayMs(int delayMs) {
+        if (delayMs < 0) {
+            throw new IllegalArgumentException("Invalid delay of " + delayMs + " ms");
+        }
+
+        this.delayMs = delayMs;
+    }
 }
--- a/brave-spancollector-scribe/pom.xml
+++ b/brave-spancollector-scribe/pom.xml
@@ -48,5 +48,11 @@
         <artifactId>log4j-slf4j-impl</artifactId>
         <scope>test</scope>
     </dependency>
+      <dependency>
+          <groupId>io.dropwizard.metrics</groupId>
+          <artifactId>metrics-core</artifactId>
+          <version>3.1.2</version>
+          <scope>test</scope>
+      </dependency>
   </dependencies>
 </project>
--- a/brave-spancollector-scribe/src/main/java/com/github/kristofa/brave/scribe/EmptyScribeCollectorMetricsHandler.java
+++ b/brave-spancollector-scribe/src/main/java/com/github/kristofa/brave/scribe/EmptyScribeCollectorMetricsHandler.java
@@ -1,0 +1,17 @@
+package com.github.kristofa.brave.scribe;
+
+/**
+ * Empty implementation ignoring all events.
+ */
+class EmptyScribeCollectorMetricsHandler implements ScribeCollectorMetricsHandler {
+
+    @Override
+    public void incrementAcceptedSpans(int quantity) {
+
+    }
+
+    @Override
+    public void incrementDroppedSpans(int quantity) {
+
+    }
+}
--- a/brave-spancollector-scribe/src/main/java/com/github/kristofa/brave/scribe/ScribeCollectorMetricsHandler.java
+++ b/brave-spancollector-scribe/src/main/java/com/github/kristofa/brave/scribe/ScribeCollectorMetricsHandler.java
@@ -1,0 +1,24 @@
+package com.github.kristofa.brave.scribe;
+
+/**
+ * Monitor {@link ScribeSpanCollector} by implementing reactions to these events, e.g. updating suitable metrics.
+ *
+ * See DropwizardMetricsScribeCollectorMetricsHandlerExample in test sources for an example.
+ */
+public interface ScribeCollectorMetricsHandler {
+
+    /**
+     * Called when spans are submitted to {@link ScribeSpanCollector} for processing.
+     *
+     * @param quantity the number of spans accepted.
+     */
+    void incrementAcceptedSpans(int quantity);
+
+    /**
+     * Called when spans become lost for any reason and won't be delivered to the target collector.
+     *
+     * @param quantity the number of spans dropped.
+     */
+    void incrementDroppedSpans(int quantity);
+
+}
--- a/brave-spancollector-scribe/src/main/java/com/github/kristofa/brave/scribe/ScribeSpanCollector.java
+++ b/brave-spancollector-scribe/src/main/java/com/github/kristofa/brave/scribe/ScribeSpanCollector.java
@@ -47,6 +47,7 @@ public class ScribeSpanCollector implements SpanCollector, Closeable {
     private final List<ScribeClientProvider> clientProviders = new ArrayList<>();
     private final List<Future<Integer>> futures = new ArrayList<Future<Integer>>();
     private final Set<BinaryAnnotation> defaultAnnotations = new HashSet<BinaryAnnotation>();
+    private final ScribeCollectorMetricsHandler metricsHandler;
 
     /**
      * Create a new instance with default queue size (= {@link ScribeSpanCollectorParams#DEFAULT_QUEUE_SIZE}) and default
@@ -71,6 +72,7 @@ public class ScribeSpanCollector implements SpanCollector, Closeable {
         checkNotBlank(host, "Null or empty host");
         checkNotNull(params, "Null params");
 
+        metricsHandler = params.getMetricsHandler();
         spanQueue = new ArrayBlockingQueue<Span>(params.getQueueSize());
         executorService = Executors.newFixedThreadPool(params.getNrOfThreads());
 
@@ -80,7 +82,7 @@ public class ScribeSpanCollector implements SpanCollector, Closeable {
             ScribeClientProvider clientProvider = createZipkinCollectorClientProvider(host,
                     port, params);
             final SpanProcessingThread spanProcessingThread = new SpanProcessingThread(spanQueue, clientProvider,
-                    params.getBatchSize());
+                    params.getBatchSize(), metricsHandler);
             spanProcessingThreads.add(spanProcessingThread);
             clientProviders.add(clientProvider);
             futures.add(executorService.submit(spanProcessingThread));
@@ -109,6 +111,7 @@ public class ScribeSpanCollector implements SpanCollector, Closeable {
     @Override
     public void collect(final Span span) {
 
+        metricsHandler.incrementAcceptedSpans(1);
         final long start = System.currentTimeMillis();
 
         if (!defaultAnnotations.isEmpty()) {
@@ -120,6 +123,7 @@ public class ScribeSpanCollector implements SpanCollector, Closeable {
         final boolean offer = spanQueue.offer(span);
         if (!offer) {
             LOGGER.warning("Queue rejected Span, span not submitted: "+ span);
+            metricsHandler.incrementDroppedSpans(1);
         } else {
             final long end = System.currentTimeMillis();
             if (LOGGER.isLoggable(Level.FINE)) {
@@ -172,6 +176,7 @@ public class ScribeSpanCollector implements SpanCollector, Closeable {
             clientProvider.close();
         }
         executorService.shutdown();
+        metricsHandler.incrementDroppedSpans(spanQueue.size());
         LOGGER.info("ScribeSpanCollector closed.");
     }
 
--- a/brave-spancollector-scribe/src/main/java/com/github/kristofa/brave/scribe/ScribeSpanCollectorParams.java
+++ b/brave-spancollector-scribe/src/main/java/com/github/kristofa/brave/scribe/ScribeSpanCollectorParams.java
@@ -1,5 +1,7 @@
 package com.github.kristofa.brave.scribe;
 
+import static com.github.kristofa.brave.internal.Util.checkNotNull;
+
 /**
  * Optional parameters for {@link ScribeSpanCollector}.
  * <p/>
@@ -13,6 +15,7 @@ package com.github.kristofa.brave.scribe;
  * will be thrown.</li>
  * <li>fail on setup: Indicates if {@link ScribeSpanCollector} should fail on creation when connection with collector can't
  * be established or just log error message.</li>
+ * <li>metrics handler: see {@link ScribeCollectorMetricsHandler}.</li>
  * </ul>
  * 
  * @author kristof
@@ -29,6 +32,7 @@ public class ScribeSpanCollectorParams {
     private int nrOfThreads;
     private int socketTimeout;
     private boolean failOnSetup = true;
+    private ScribeCollectorMetricsHandler metricsHandler = new EmptyScribeCollectorMetricsHandler();
 
     /**
      * Create a new instance with default values.
@@ -138,4 +142,11 @@ public class ScribeSpanCollectorParams {
         return failOnSetup;
     }
 
+    public ScribeCollectorMetricsHandler getMetricsHandler() {
+        return metricsHandler;
+    }
+
+    public void setMetricsHandler(ScribeCollectorMetricsHandler metricsHandler) {
+        this.metricsHandler = checkNotNull(metricsHandler, "Null metricsHandler");
+    }
 }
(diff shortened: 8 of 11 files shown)


kristofa:
@pbetkier Thanks for this! This is something that is really useful. Especially measuring when the back-end is not keeping up and we are dropping spans. Unfortunately I wil only be able to look at it in more detail beginning of next week.

kristofa:
@pbetkier I would make the implementation more generic. Instead of having an interface which is tightly coupled to the current Scribe implementation I would make more generic interfaces in a separate sub project: `Counter`, `Gauge`, 'Histogram` and probably a `MetricsFactory` to create the 3 metric types.

Reason is that we expect to have multiple span collector implementations soon (Kafka, http) but those might not follow same design as the current Scribe one. They might for example not batch spans prior to sending, not us an internal queue,... so they could have different metrics to report.

I'm willing to spend some time later this week to come up with an initial implementation. Do you think this makes sense?

pbetkier:
@kristofa What I like about the mechanism that I suggested is the flexibility for the developer: I can decide which parameters to add to my metrics registry and which to ignore, how to call them and perhaps implement a different strategy of notification for some of them, e.g. logging. That would require creating tailored interfaces for each span collector implementation though and a bit of work for the developer to start monitoring things.

I also see the power of an abstraction over metrics that you propose – it would be very easy to setup for the developer and allow him to switch between different span collectors easily. We could even create ready implementations for the most commonly used metrics libraries, like dropwizard metrics.

So it's a trade-off. I like to have control over how things are monitored so I prefer the original solution :) But your idea also makes sense, so go ahead with implementing it you value it more.

adriancole:
My 2p here is that we should only specify the minimum possible. For example the input to the histogram vs the histogram api. I'm even troubled by the queue size guage. This only matters when something else occurs in relation to the queue size. IOTW, we could log the size when that happens (or even allow folks to provide a queue)

In some private implementations of  ML stuff, I have seen this as simple as having classes that perform work leave unimplemented methods for things like incrementing. Ex a simple subtype could bind someone's favorite telemetry instrumentation vs making our own wrappers.

Brave is simple enough that basically I don't see value in cluttering it with an abstraction layer of any sort, solely to provide meters. If we did, would really be best to do the least.

kristofa:
@adriancole Can you clarify a bit more? I'm not sure I understand what you want to say :)

I was thinking about something like this:

```
interface Telemetry {
 Counter counter(String name)
 Gauge gauge(String name)
 Histogram histogram(String name)
}

interface Counter {
  void inc()
  void inc(double value)
}

interface Gauge {
  void set(double value)
}

interface Histogram {
 void observe(double value)
}
```
We would pass a `Telemetry` instance in the `ScribeCollector` and it can create the Counter / Gauge / Histogram instances that are needed.

adriancole:
just wondering how many things you are actually measuring :) Ex. I had guessed there would be less than 10 core metrics in brave. Does this require a type system?

Reason I say that is if you don't then you just expose the actual events ex `incrementSpansReceived(int)` which doesn't require a binding to the namespace scheme an interface hierarchy like above implies.

adriancole:
guess what I'm wondering if we can just expose what we need to record vs creating a type system around the abstract act of recording.

adriancole:
Here's a good reason. Notice you chose double for all the values. You only need to do such a thing when creating abstract systems. spans are always maximum an int (as they are in a collection who's max size is int). histogram with a single idea of double requires a lot of explanation, especially how to merge this thing? We can avoid all of these things by just concretely exposing what's going on as method signatures.

If brave were a complex system, maybe we should have our own telemetry api, just I don't understand how it helps tell the story of what people will be interested in wrt brave, and certainly it limits implementation.

adriancole:
very long winded way of saying that I'm more in favor of the direction of this pull request (notable exception is SpanCollectorGauge type), vs creating a telemetry api only brave uses. :)

kristofa:
@adriancole Yep, understand :) My proposal might be a case of YAGNI :)
Fair enough. I don't have very strong arguments to put against that so I'm fine with that.

Instead of the `Gauge` we could have a method like: `setCurrentQueueSize(int)`?

adriancole:
yeah basically I'd love whatever interface we have to tell the story
of notable events (you should bind to telemetry and probably alarms)!

Ex. if we do it like this, we can put javadoc literally
`onCurrentQueueSize(int)` which says watch out for buildup!

by using methods, we can also link to correlate things we may want to
advise people to alarm on.

Can't do that with abstractions, right? The beauty is that brave is
"experienced enough" to have the advantage of concrete guidance. let's
use it!

adriancole:
ps `onCurrentQueueSize(int)` vs `incrementQueueSize(int)` is a discussion to have :) personally, I like the increment/decrement approach. as long as it is consistent, you can implement a snapshot of "current" with that.

adriancole:
don't mean to drag queue size snapshot into the weeds, as I know it is pros vs cons. just noting things we can toss around.

kristofa:
@pbetkier what do you think about the remarks by @adriancole ?
Can you do these changes?

pbetkier:
I agree with @adriancole, very insightful remarks. He convinced me that we
should replace a gauge with a onCurrentQueueSize method. I'm not sure about
increment / decrement naming though, maybe I'm just not used to that :)

So what do we change in the PR apart from replacing the gauge? Do you agree
with the events that I exposed to monitor, or would you add or remove
something?

I'll be happy to update the PR, but it'll happen after 11 Nov as I don't
have computer access right now.

kristofa:
@pbetkier Start with the ones that you really want to inspect right now. I think the events you defined make sense. It is fine to keep this pr open till you are available to finish it up. Thanks @pbetkier  and @adriancole!

pbetkier:
I've updated my PR by removing the gauge and simplifying the exposed notifications. Please have a look.

I'm still thinking of allowing developers to provide their own queue instead of publishing `onQueueSizeChanged` notifications though. That would simplify the span collector code and make the API more useful in my opinion. I will propose such change in a separate commit so that we have a base for discussion.

brave-spancollector-scribe/src/main/java/com/github/kristofa/brave/scribe/EmptyScribeSpanCollectorNotificationsHandler.java
```diff
@@ -0,0 +1,24 @@
...
+
+import com.twitter.zipkin.gen.Span;
+
+/**
+ * Empty implementation ignoring all notifications.
+ */
+class EmptyScribeSpanCollectorNotificationsHandler implements ScribeSpanCollectorNotificationsHandler {
+
+    @Override
+    public void onQueueSizeChanged(int currentQueueSize) {
```
adriancole: I'd prefer this to be more explicit. Ex instead of the somewhat vague of queue-size changed, make this an event:
onSpanAccepted, onSpanProcessed?

brave-spancollector-scribe/src/main/java/com/github/kristofa/brave/scribe/ScribeSpanCollectorNotificationsHandler.java
```diff
@@ -0,0 +1,35 @@
...
+    void onSpanRejected(Span rejectedSpan);
+
+    /**
+     * Called after sending a batch of spans.
+     *
+     * @param durationInMillis the time it took to send the spans in milliseconds.
+     * @param batchSize the size of the spans batch that was sent.
+     * @param failed true, if sending has resulted in a failure.
+     */
+    void onSpansSent(long durationInMillis, int batchSize, boolean failed);
```
adriancole: for callbacks that are intended for metrics, probably good to think about an example of how these would be exported. Ex. this particular callback seems cluttered with mixed signals. This would be easy to see when we try to explain how to map this to a timeseries metric.

brave-spancollector-scribe/src/test/java/com/github/kristofa/brave/scribe/CountingScribeSpanCollectorNotificationsHandler.java
```diff
@@ -0,0 +1,53 @@
...
+    @Override
+    public void onSpanRejected(Span rejectedSpan) {
+        spansRejected.incrementAndGet();
+    }
+
+    @Override
+    public void onSpansSent(long durationInMillis, int batchSize, boolean failed) {
+        spansSentEvents.incrementAndGet();
+        if (failed) {
+            sendingFailures.incrementAndGet();
```
adriancole: ex. this sends signals of mixed granularity. for example spansRejected is per span and these are ... who knows? I'd prefer signals to be of the same grain, and use atomic addition if we want to increment in bulk

adriancole:
Thanks for the next round. made some comments about a preference towards
same grain as the anchoring concern is spans not "send events". ex a
multithreaded app may arbitrarily have multiple send events, but the
collection tier only cares about spans. We'd probably be best with choosing
a grain that is consistent across the tracing service.

pbetkier:
@adriancole I like your remarks about granularity, this indeed is something worth keeping consistent, especially that such API would be more universal when thinking about other spancollector implementations. So I see the value in introducing `onSpanAccepted`, `onSpanProcessed`. And also `onSpanFailedToProcess` – to report a span dropped due to connection failure.

But I don't agree that we should only care about spans. Queue size and logging latency are technical metrics that give clear information about operational issues and their source. Also helpful for tuning the parameters. While the developer could deduce the current queue size by doing the maths on other notifications, that's just unnecessary complexity for him in my opinion.

So I would add the 3 notifications that I mentioned above and keep `onQueueSizeChanged` (or alternatively make it possible to provide own queue). I would also replace the `onSpansSent` method with a method notifying about logging durations. What do you think?

adriancole:
Let's bear in mind that coding to this api will be a seldom occurrence, so
developer burden for composing things, when they care to, is going to be
limited to a subset of an already small number of developers. It may only
be 3 people!

Again, I think that for each scenario we dream up, we *must* back it with
actual metrics, not hypothetical ones. So, if you really think duration is
going to be one, please list what those are and how you would alert on it.

We owe it to ourselves to make the simplest api possible (ie the least
hooks possible), as often code submitted is written once and never
maintained after.

adriancole:
Basically long story short is that I think we still haven't described well enough the feature.

I don't think flexible ways of adding queues, or conveniently adding arguments in the same method call is a feature.

Knowing when buildup happens, or that our key metric of spans/second dropped is a feature. Doing that without causing a performance problem is hard as we've not created benchmarks, yet, but that's still step 2.

Here's an example of a fairly clean namespace even if there's at least 3 grains, Api calls (PutRecord/s), Records and Bytes: http://docs.aws.amazon.com/kinesis/latest/dev/monitoring-with-cloudwatch.html

Want to give a try at documenting how to apply things in this hook to alerts or action?

brave-spancollector-scribe/src/main/java/com/github/kristofa/brave/scribe/ScribeSpanCollectorNotificationsHandler.java
```diff
@@ -0,0 +1,35 @@
...
+     */
+    void onQueueSizeChanged(int currentQueueSize);
+
+    /**
+     * Called when a span could not be added to the span queue because of it being full
+     * and so the span was rejected.
+     *
+     * @param rejectedSpan the rejected span.
+     */
+    void onSpanRejected(Span rejectedSpan);
```
adriancole: ps probably don't need the Span parameter to increment a counter. ex `incrementDroppedSpans()` or if we want to call out only dropped spans due to queue rejection (vs marshaling, timeouts or other reasons), `incrementRejectedSpans()`

brave-spancollector-scribe/src/main/java/com/github/kristofa/brave/scribe/EmptyScribeSpanCollectorNotificationsHandler.java
```diff
@@ -0,0 +1,24 @@
+package com.github.kristofa.brave.scribe;
+
+import com.twitter.zipkin.gen.Span;
+
+/**
+ * Empty implementation ignoring all notifications.
+ */
+class EmptyScribeSpanCollectorNotificationsHandler implements ScribeSpanCollectorNotificationsHandler {
```
adriancole: nit: prefer Noop or Null prefix

brave-spancollector-scribe/src/main/java/com/github/kristofa/brave/scribe/SpanProcessingThread.java
```diff
@@ -83,8 +86,8 @@ public Integer call() {
                 final Span span = queue.poll(5, TimeUnit.SECONDS);
                 if (span == null) {
                     subsequentEmptyBatches++;
-
                 } else {
+                    notificationsHandler.onQueueSizeChanged(queue.size());
```
adriancole: if we ran a benchmark, this would pop up depending on the type of queue in use. This is why I like increment instead.

brave-spancollector-scribe/src/main/java/com/github/kristofa/brave/scribe/SpanProcessingThread.java
```diff
@@ -105,9 +108,10 @@ public Integer call() {
     private void log(final List<LogEntry> logEntries) {
         final long start = System.currentTimeMillis();
         final boolean success = log(clientProvider.getClient(), logEntries);
```
adriancole: if we were going to use System.currentTimeMillis() to calculate this (which would be more expensive and less precise than System.nanoTime()), we'd probably want to do so conditionally, as if there's a Noop notifier, we'd be creating overhead for no reason.

brave-spancollector-scribe/src/test/java/com/github/kristofa/brave/scribe/CountingScribeSpanCollectorNotificationsHandler.java
```diff
@@ -0,0 +1,53 @@
+package com.github.kristofa.brave.scribe;
+
+import com.twitter.zipkin.gen.Span;
+
+import java.util.concurrent.atomic.AtomicInteger;
+
+class CountingScribeSpanCollectorNotificationsHandler implements ScribeSpanCollectorNotificationsHandler {
```
adriancole: something like this is probably better in the test package, as it is an interesting example, but not actually used, and easy to recreate. I'd expect monitoring bindings to be direct, ex to a codahale or otherwise library, as that allows the most relevant features to be used.

brave-spancollector-scribe/src/test/java/com/github/kristofa/brave/scribe/CountingScribeSpanCollectorNotificationsHandler.java
```diff
@@ -0,0 +1,53 @@
...
+
+    private final AtomicInteger spansRejected = new AtomicInteger();
+    private final AtomicInteger spansSentEvents = new AtomicInteger();
+    private final AtomicInteger sendingFailures = new AtomicInteger();
+    private volatile int lastQueueSize;
```
adriancole: We should make a benchmark project, so stuff like this is of more obvious impact. Wherever we try to track state, we'll likely inherit a performance hit. In this case, it is flushing to main memory on updates to these fields.

brave-spancollector-scribe/src/test/java/com/github/kristofa/brave/scribe/ScribeSpanCollectorTest.java
```diff
@@ -93,7 +82,39 @@ public void testCollectWithDefaultAnnotation() throws TTransportException {
         final BinaryAnnotation binaryAnnotation = binary_annotations.get(0);
         assertEquals(KEY1, binaryAnnotation.getKey());
         assertEquals(VALUE1, new String(binaryAnnotation.getValue()));
+    }
 
+    @Test
+    public void testNotifyOfSpanCollectionEvents() throws TTransportException {
```
adriancole: good start on a feature test. This is where we can explain what the callbacks mean

brave-spancollector-scribe/src/test/java/com/github/kristofa/brave/scribe/ScribeSpanCollectorTest.java
```diff
@@ -93,7 +82,39 @@ public void testCollectWithDefaultAnnotation() throws TTransportException {
...
+
+        // then
+        assertEquals(0, eventsHandler.getLastQueueSize());
+        assertEquals(1, eventsHandler.getTotalSpansSentEvents());
+        assertEquals(0, eventsHandler.getTotalSendingFailures());
+        assertEquals(1, eventsHandler.getLastBatchSize());
+        assertEquals(0, eventsHandler.getTotalSpansRejected());
+    }
+
+    private void collectSingleSpan(ScribeSpanCollector zipkinSpanCollector) {
```
adriancole: I'd probably remove this method and inline it in the test, particularly as you can use try-with-resources on ScribeSpanCollector

pbetkier:
I think the difference in our approaches lies in defining what this feature we're working on is. My original idea was to expose hooks a developer can use to implement his own decisions on what and how to monitor. It'd be up to him if he wants to be alerted by X or Y. Or perhaps he prefers to log stuff instead of using some metrics library. My thinking was that the audience of this library is limited to a few toolmakers and they, seeing how span collector works, are able to make such decisions for themselves. That's why I didn't want to think in terms of incrementing some counters and wanted to provide richer notifications.

But perhaps you're right that we should go further and be opinionated on what and how to monitor, solving this puzzle once and for all. Let's move our focus from the code for a second and try to think what possible failure scenarios do we have and what metrics and alerts do we need for solving them. Once we converge on this, making code changes will be much easier for me.

My view on this:
```
- failure: rejecting incoming spans due to queue being full
    - alert: rejected spans meter > 0 (or some more forgiving threshold)
    - possible reasons:
         - reason: a peak in incoming spans
           observed to learn why: a peak in spans accepted meter
         - reason: increased latency of sending the spans
           observed to learn why: a peak in latency of sending timer

- failure: discarding spans that couldn't be sent due to connection problem
    - alert: discarded spans meter > 0 (or some more forgiving threshold)
    - possible reasons:
         - reason: connection timeout
           observed to learn why: failed requests meter > 0 and plateau near socket timeout value on higher percentiles of sending timer
         - reason: connection broken
           observed to learn why: failed requests meter > 0 and no plateau near socket timeout value visible, probably latency lower than usual on median of sending timer
```

So what I need is a spans rejected meter, spans accepted meter, spans discarded meter, scribe delivery timer and scribe delivery failure meter. I still think it would be nice to view how we're lagging with processing work, even if we're not yet rejecting the spans (e.g. for tuning the configuration parameters): but if not monitoring the queue size directly, then we need spans accepted and spans processed (either delivered or discarded) counters to subtract.

What is your take on this? How would your similar list look like?

brave-spancollector-scribe/src/main/java/com/github/kristofa/brave/scribe/EmptyScribeSpanCollectorNotificationsHandler.java
pbetkier: I prefer Noop as well, but wanted to be consistent with `EmptySpanCollector`.

brave-spancollector-scribe/src/main/java/com/github/kristofa/brave/scribe/SpanProcessingThread.java
pbetkier: touché :)

brave-spancollector-scribe/src/test/java/com/github/kristofa/brave/scribe/CountingScribeSpanCollectorNotificationsHandler.java
pbetkier: I don't understand, this class is in the test sources, used in the tests to verify that the notifications are indeed published. It's doesn't serve as an example on how to bind it to a metrics library in production code.

brave-spancollector-scribe/src/test/java/com/github/kristofa/brave/scribe/CountingScribeSpanCollectorNotificationsHandler.java
pbetkier: You're right, but performance was not my concern here, it's just for checking the notifications logic in the unit tests. This class has no other intention.

brave-spancollector-scribe/src/test/java/com/github/kristofa/brave/scribe/ScribeSpanCollectorTest.java
pbetkier: agreed

brave-spancollector-scribe/src/test/java/com/github/kristofa/brave/scribe/CountingScribeSpanCollectorNotificationsHandler.java
adriancole: ack. somehow I misread this as in the main tree

adriancole:
I'm concerned with this as it reads like guessing. Can you clarify if these
signals (and this work) is related to production outages where other
monitoring tools failed because the root cause was obfuscated in Brave?

@kristofa @eirslett you've certainly had experience. What do you
continually burn your fingers on, which would lead you to want a
brave-specific telemetry api? What are your top-3 failures that eat your
time today, which would eat less if we had brave-specific
meta-instrumentation?

Reason I ask is that unlike some libraries, Brave is 2.5 years old. There
*must* be existing prod support concerns we can leverage vs guessing.

adriancole:
ps Piotr, I bet it isn't obvious through my comments that I'm
interested in this work and your effort towards it. I am happy about
both.. I mainly want your effort behind a feature that is used by
folks besides you, as otherwise, if could just live in a fork.

Feel free to hop on https://gitter.im/openzipkin/zipkin if you want
real-time conversation, noting Eirik and Kristof are in central europe
time.

pbetkier:
I appreciate your remarks :) Indeed it would be great to hear some experiences from the battle-front. We're just introducing tracing on production, so no stories from my side.

kristofa:
@pbetkier @adriancole I wrote a message about our production experiences with Zipkin and what I think we should monitor here: https://gitter.im/openzipkin/zipkin

pbetkier:
@eirslett just a kind reminder, do you have some failure stories to share that would help us shape the metrics API for span collectors? Do you even see the need for monitoring the span collector's behavior?

pbetkier:
Quote from kristofa on gitter: 
> By now I have more experience with using Zipkin (through Finagle) in a production environment as with using Brave but the main issues we encountered so far are:
Cassandra filling up because it is not keeping up with removing expired data. This became obvious by monitoring Cassandra and got fixed by lowering sample rate (=less data) and increasing compaction throughput.
Dropped spans observed by zipkin-collector counter collector.pushback and Finagle counter try_later. This is quite important to look into because if spans are dropped traces do not represent reality which can be worse then having no tracing at all. I guess this can be fixed by having more zipkin-collector instances, tuning/increasing size of Cassandra cluster or lowering sample rate. We increased Cassandra cluster once and also lowered sample rate.
So I would say the most important metrics we want from Brave is counters with successfully delivered spans and dropped spans.

I suggest we go with measuring rejected, delivered and dropped spans then. Notifications API would expose 3 methods called like `incrementSpansDelivered()`. We suggest the developer registered corresponding meters in his registry and mark them when these increments occur.

What's your take on that @adriancole?

adriancole:
Thanks for keeping this alive!

My 2p

I think dropped vs accepted makes sense. I'm still apprehensive about a
separate counter for dropped (for any reason) vs rejected (to queuing
choice). I'd prefer a single dropped counter for now.

There's some nuance to what we indicate as delivered/accepted. Ex we can
count "accepted" as absolute number sampled. Or we can count "delivered" as
absolute number - dropped

I think the term "delivered" is a bit of a lie as we've no idea they were
in fact delivered. I like "accepted" better as it is something the user can
control. Ex sample less means accept less.

pbetkier:
Right, we could drop rejected counter and have one counter for spans dropped for any reason as the details on what happened are available in the logs anyway.

Do I understand your idea with accepted counter correctly – if a span gets rejected by the queue, it is marked both as accepted and dropped?

adriancole:
Right, my take is we increment accepted before any transport attempt is
made, and dropped if any exception occurs.

So 100 accept and 40 drops means 40% known loss.

@kristofa wdyt?

kristofa:
@pbetkier @adriancole That sounds good!

pbetkier:
I've updated my PR implementing the ideas we discussed, please have a look @kristofa @adriancole.

brave-spancollector-scribe/README.md
```diff
@@ -15,6 +17,17 @@ This approach has again been chosen to minimize the impact on the application. H
...
 However it makes sure that it does not keeps holding onto spans. If the buffer is not full after 10 seconds it sends the received spans in any case.
 
+## Monitoring
+
+Monitor `ScribeSpanCollector`'s performance by providing your implementation of `ScribeCollectorMetricsHandler`. It will
+be notified when a span is accepted for processing and when it gets dropped for any reason, so you can update corresponding
+counters in your metrics tool. When a span gets dropped, the reason is written to the application logs.
+The number of spans sent to the target collector can be calculated by subtracting the dropped count from the accepted count.
+
+Refer to `DropwizardMetricsScribeCollectorMetricsHandlerExample` for an example of how to integrate with
```
adriancole: Can link to the actual file

brave-spancollector-scribe/src/test/java/com/github/kristofa/brave/scribe/ScribeReceiver.java
```diff
@@ -81,4 +77,11 @@ public ResultCode Log(final List<LogEntry> messages) throws TException {
         return result;
     }
 
+    public void setDelayMs(int delayMs) {
```
adriancole: Id revert as this is a little more ceremony than I'd expect in a test class

pbetkier: When you say revert you mean replace it with what?

adriancole: Revert I mean just go back to accessing the field (similar to how the code
was originally). Since it is a test class, this sort of defensive
programming seems distracting, IMHO.

pbetkier: I see, I thought you meant removing the setter altogether.

adriancole: I did :) but anyway I don't know how much attention this detail is worth.
Im fine either way

pbetkier: The field is private, originally it was provided in the constructor (it was dead code though, nobody used this feature). I'd prefer to set it via setter where needed to have a common scribe server setup in my newly added test class.

adriancole: I don't feel too strongly, just that setters are boilerplatey. If I were
writing this, I'd do it differently, but at this point we've given this
particular code far too much luxury of attention for its value. Let's move
on.

kristofa:
@pbetkier @adriancole thanks! I think we can merge this.

When we have other collectors (eg Kafka) and `ScribeCollectorMetricsHandler` has been put to use and proven to be valuable we can move the interface to `brave-core` and rename it to for example `CollectorMetricsHandler`. At that point we can use it in all collector implementations. 

But for now this is a good start.

kristofa:
@pbetkier @adriancole brave 3.2.0, which includes this pr, is released and should appear in Maven Central soon. Release notes will follow later, probably tomorrow.

Issue #100, opened 2015-10-09
Pass raw IP to ThreadLocalServerAndClientSpanState


adriancole:
ThreadLocalServerAndClientSpanState requires you use an `InetAddress` class. In many cloud environments, you can lookup your public ip from an environment variable or a metadata service. I think it would be nice if we had a constructor that either took the int form or byte[] form of an IP, avoiding ns lookup.

henrikno:
ThreadLocalServerAndClientSpanState doens't do any lookups, but the Brave builder does call
InetAddressUtilities.getLocalHostLANAddress() which does do a lookup.
InetAddress is only a "data" object for an IP address, unless you invoke one of the lookup methods.
`Brave.Builder` should have a constructor which does not do lookups.

Another problem I see is that it converts the IP address to an int, which means it only supports IPv4. Seems like that is a limitation in Zipkin.

adriancole:
The problem is that the only parameter is an InetAddress, which in many
cases leads a user down paths where they have to do complicated things to
prevent it from doing NS lookups.

adriancole:
IOTW InetAddress makes you do needless conversion, which usually interferes
and has poor relevance for cloud environments, where in many cases your
public ip/port is not readable from the host except via property. The idea
of using something like InetAddress is for type safety, when its only
purpose is carrying an ip address. In this case, this is a cure worse than
the disease. Brave should allow a user to specify an IP without dancing or
monkeying, IOTW.

http://docs.aws.amazon.com/AWSEC2/latest/UserGuide/using-instance-addressing.html#concepts-public-addresses

Regardless, I agree that ipv6 is a limitation of brave.

adriancole:
>
>
> Regardless, I agree that ipv6 is a limitation of brave.
>
sorry I meant zipkin! anyway even when we have ipv6 you wouldn't conflate
them in the same endpoint.

henrikno:
What's wrong with InetAddress.getByAddress(byte[])? Just curious. Is there an easier way of representing a IPv4/IPv6 address (byte[] seems so... raw)?

adriancole:
because zipkin only supports ipv4, for example. the type InetAddress is a lie. I'd rather pass an int or byte[] as that's at least not lying about what it is capable of. In general lying types feel worse to me, and also if you look at the constructor, all it is doing is coercing to an int. Having an alternate constructor that removes the complexity of how to safely use the InetAddress stuff would be of help imho.

adriancole:
[...]

PR #107, opened 2015-11-08
Adds Span.timestamp, duration and address annotations

### LINKED ISSUE #104: Add support for the ServerAddr("sa") and ClientAddr("ca") annotations
When the server-side of a trace is not instrumented, we can log ServerAddr("sa") annotation to indicate who the server is (via endpoint/host).

Case in point. If you are writing a database driver, you might send a "query" annotation, but the host would be the local host (the client). The standard way to indicate the server's address would be to add an "sa" annotation, which would hold the destination.

The ClientAddr similarly stores the client address. Some implementations may store, for example the value of "x-forwarded-for" header.

adriancole:
This updates Brave to the latest thrift specification, which includes
explicit settings for Span timestamp and duration, as well as binary
annotations that indicate the actual IP of a client or server.

This also updates to lowercase convention for Span.name and 
Endpoint.serviceName, also noted in recent thrift documentation.

Fixes #104


adriancole:
@kristofa if this change isn't contested, mind cutting a release after merge (even if subminor?) I'd like to decruft some things.

kristofa:
Thanks for updating to the latest protocol changes!
`Span.duration` is a nice addition. 

I had seen `ca` / `sa` annotations before but never really understood how they were supposed to be used.  I understand now that `ca` helps for knowing ip address of uninstrumented clients like browsers and `sa` for uninstrumented services like databases.  And they make it also easier to build up dependency graph as we only need `sa` / `ca` annotations for building up dependency graph.  Is my understanding correct? Am I missing something?

I'm fine with merging and releasing. Next step would be to actually submit `ca` / `sa` annotations I guess as this pr only supports them in the api?

adriancole:
You understand correctly!  except that it is really only "sa" that's used
in the dep graph right now. If we put "ca", you could imagine the left side
could become huge (with all those browser IPs!)

Agreed next step is to update instrumentation to log these. I can help with
this in the ones I know.

adriancole:
actually, maybe it wouldn't be bad to put browsers in *some* view.. anyway, you get to see the client IP in trace detail screen and that's cool.

eirslett:
> you could imagine the left side
could become huge (with all those browser IPs!)

Maybe the instrumentation should simply log it all as "browser" or "undetermined client"?

adriancole:
@eirslett you mean for the service name in the "ca" when a browser? I was curious what we should use for that.. user-agent seems an odd choice.

adriancole:
Oh I see what you mean. Basically, if we get to the point of logging the
right-hand side, it could be aggregated, too. Ex. android, iphone, browser.
Yeah seems a neat idea.

eirslett:
Yes, I meant that - name of the client.
It makes sense for internal services, like batch clients that make calls to microservices. Maybe a custom header, "X-Trace-ClientName" or something like that? And a fallback to "unknown client" if not specified, or the instrumentation library can use heuristics like looking at the user agent header and define it as "browser".

adriancole:
erg left side

adriancole:
I think some instrumentation would want to do something like that, when the
origin isn't known. maybe it starts as a name function in brave? I will
also punt this to the distrubuted-tracing ml, as this has probably been
discussed by somebody!

adriancole:
https://groups.google.com/forum/#!topic/distributed-tracing/jHjf1n12E4Y

adriancole:
Hey I was thinking. Zipkin already has "unknown" used. Maybe we allow this
to be optional and worst case log "unknown". Tracers could be fitted with a
naming function to prevent falling back to that. Wdyt?

adriancole:
yeah looks like finagle defaults service name to unknown. I'll update the impl to do the same for now.

adriancole:
PTAL: I made serviceName nullable in ClientTracer and ServerTracer. For now, null -> "unknown". This allows us to later add some sort of naming function after we learn what that should be.

I'll add example instrumentation next.

adriancole:
Note my choice of InetSocketAddress is not because I think it is a good choice. I think it is a very bad one, but I don't want to repeat the argument with @henrikno from #100 

Basically, this concedes the usual dangers of using InetSocketAddress, including easy accidental name resolution, permitting unloggable IPv6 addresses, unnecessary allocation (especially since X-Forwarded-For has literally the contents needed), etc. IOTW, I'd rather have some ability to permit this feature without bike shedding. If this were a different project, I'd never use InetSocketAddress for this.

adriancole:
@kristofa I added an example to this PR. If all's well with you, please merge and cut!

kristofa:
@adriancole you are right about `InetAddress`. Recently I had an issue (not brave related) with code that used `InetAddress` which failed on some configurations with `UnknownHostException`. We should at least provide a way around it and in fact I just pushed a commit to Brave master to support passing ip + port in `Brave.Builder` and `ThreadLocalServerAndClientSpanState`.

If you want and still can change this pr to use `int` instead of `InetAddress` that would be good!
I can make a release tomorrow morning my time... so in about 12 hours :)

adriancole:
I'll change this to use ipv4 directly, as it is easier to explain how to do that vs problems mentions. cheers!

adriancole:
bumped

kristofa:
@adriancole Merged and released: brave v3.1.0

adriancole:
thanks!

Issue #90, opened 2015-08-26
Running Brave on custom solution for preserving metadata between threads


pbetkier:
Hello, 

we're planning to adopt Brave at our company and there is one issue we try to tackle to make it possible. 

We already have our own infrastructure for keeping request metadata in thread locals and transferring the information when request processing switches threads. What we'd like to do is to keep span state in our thread locals and make it available to Brave classes that operate on it. From what I see, creating our custom implementation of `ServerAndClientSpanState` that would delegate to our infrastructure and passing it to `ServerTracer` and `ClientTracer` could solve the problem. Unfortunately, `ServerAndClientSpanState` interface is not public.

Could you recommend how to approach running Brave on existing infrastructure? What do you think of making `ServerAndClientSpanState` interface public?

kristofa:
Hello,

Yes I believe you should be able to have your own implementation of `ServerAndClientSpanState`. 

We discussed a while ago that having a common `state` implementation that is shared across implementations (zipkin, brave,...) would make it possible and easier to mix implementations (eg server using brave calling Finagle client for outgoing request).

I'll make the interface public right away.

kristofa:
It is public on current master. Give it a try and let us know if you run into other issues.
As I'm approaching the end of the brave 3.0.0 work I had planned to make another alpha or beta release soon. Hopefully end of this week. It would be good if you can give this release a try.

pbetkier:
Excellent, thanks for such quick help!

pbetkier:
In order to pass my custom implementation to `ServerTracer` and `ClientTracer` I'll need `build()` methods of their builders to be public as well :)

kristofa:
@pbetkier I would propose the following. What if you for now develop against a local copy / branch of brave and when things work out you can submit a pr with required changes?

pbetkier:
I'll do so, that should work better.

adriancole:
ps you might be interested in this topic https://docs.google.com/document/d/10S3wGxV-s2wFeXFE5ZrPAxYUW_HVJ-kUvxW6xG6_A74/edit We discussed it at a working group a couple months ago. There's a follow-up in October if you are interested. https://docs.google.com/document/d/1nFFoIFKsaL5u_XV5k3VpXzFvM2fr2ysYgmKSlFQNJi0/edit

kristofa:
[...]

PR #109, opened 2015-11-19
Makes ThreadLocalServerAndClientSpanState public and polishes use

adriancole:
ServerAndClientSpanState is awkward to use without examples. This shows
use of the initial state value `ServerSpan.EMPTY`, and polishes Brave's
builder, which accidentally implied creating state twice.

This change is paid for by removing `SimpleServerAndClientSpanState`,
which was unusable dead code due to package protection.

This also makes ThreadLocalServerAndClientSpanState public so it can be
used directly in Spring.

See #90


adriancole:
@pbetkier you're probably interested in this.

Note I think we will need follow-up change if we want to make it possible for finagle and brave to collaborate (because finagle would need to create the currently hidden ServerSpan factory method).

Regardless, this is a first step, and helps in cases where you don't need to create your own server spans.

adriancole:
updated to make ThreadLocalServerAndClientSpanState public, which avoids copy/pasting this for use in spring.

brave-core/src/main/java/com/github/kristofa/brave/ThreadLocalServerAndClientSpanState.java
```diff
@@ -50,7 +50,6 @@ public ThreadLocalServerAndClientSpanState(InetAddress ip, int port, String serv
      * @param serviceName Name of the local service being traced. Should be lowercase and not <code>null</code> or empty.
      */
     public ThreadLocalServerAndClientSpanState(int ip, int port, String serviceName) {
-        Util.checkNotNull(ip, "ip address must be specified.");
```
adriancole: int can't be null :P

adriancole:
ps this is the only thing I'm actually blocked on. I found a way around needing to change Random.

kristofa:
:+1:

PR #112, opened 2015-11-20
Adds LocalTracer, designed for in-process activity that explains latency

### LINKED ISSUE #111: Add LocalTracer
I've been playing around with what's needed for local tracing. Most notably, we need more precise timestamps than what System.currentTimeMillis gives. Here's an example

```java
final class LocalTracer {
  private final ClientTracer clientTracer;
  private final ServerAndClientSpanState state;
  private final SpanCollector spanCollector;

  LocalTracer(ClientTracer clientTracer, ServerAndClientSpanState state, SpanCollector spanCollector) {
    this.clientTracer = clientTracer;
    this.state = state;
    this.spanCollector = spanCollector;
  }

  /**
   * Request a new local span, optionally providing the time it started.
   *
   * @param component component responsible for the operation
   * @param operation name of the operation that's begun
   * @param timestamp time the operation started, in epoch microseconds, or null if now.
   * @return metadata about the new span or null if one wasn't started due to sampling policy.
   * @see Constants#LOCAL_COMPONENT
   */
  SpanId startSpan(String component, String operation, @Nullable Long timestamp) {
    SpanId spanId = clientTracer.startNewSpan(operation);
    if (spanId == null) return null; // Don't calculate time if we aren't going to trace.

    com.twitter.zipkin.gen.BinaryAnnotation lc = new com.twitter.zipkin.gen.BinaryAnnotation();
    lc.setKey(Constants.LOCAL_COMPONENT);
    lc.setValue(component.getBytes(Util.UTF_8));
    lc.setAnnotation_type(AnnotationType.STRING);
    lc.setHost(state.getClientEndpoint());

    Span span = state.getCurrentClientSpan();
    synchronized (span) {
      span.setTimestamp(timestamp == null ? System.currentTimeMillis() * 1000 : timestamp);
      span.addToBinary_annotations(lc);
    }
    return spanId;
  }

  /** Log an annotation with a relative timestamp more precise than system time. */
  void submitAnnotation(String value, long timestamp) {
    Span span = state.getCurrentClientSpan();
    if (span == null) return;
    synchronized (span) {
      span.addToAnnotations(new Annotation(timestamp, value).setHost(state.getClientEndpoint()));
    }
  }

  /** Completes the span, which took {@code duration} microseconds. */
  public void finishSpan(long duration) {
    Span span = state.getCurrentClientSpan();
    if (span == null) return;

    synchronized (span) {
      span.setDuration(duration);
      spanCollector.collect(span);
    }

    state.setCurrentClientSpan(null);
    state.setCurrentClientServiceName(null);
  }
}
```

adriancole:
A local span can represent bootstrap, codec, file i/o or other activity 
that notably impacts performance.

Local spans always have a binary annotation "lc" which indicates the 
component name. Usings zipkin's UI or Api, you can query by for spans 
that use a component like this: `lc=spring-boot`

Here's an example of allocating precise duration for a local span:
```java
tracer.startSpan("codec", "encode");
try {
  return codec.encode(input);
} finally {
  tracer.finishSpan();
}
```

Fixes #111
See https://github.com/openzipkin/zipkin/pull/821

Another example: Local span where user does know about time and duration, such as recording events that already occurred:
```java
if (localTracer.startSpan("spring-boot", "bootstrap", startTimestamp) != null) {
  // log historical annotations/binary annotations only if we know the trace was sampled
  localTracer.finishSpan(duration);
}
```

### DIFF
--- a/brave-core/src/main/java/com/github/kristofa/brave/AnnotationSubmitter.java
+++ b/brave-core/src/main/java/com/github/kristofa/brave/AnnotationSubmitter.java
@@ -1,6 +1,7 @@
 package com.github.kristofa.brave;
 
 import com.github.kristofa.brave.internal.Nullable;
+import com.github.kristofa.brave.internal.Util;
 import com.twitter.zipkin.gen.Annotation;
 import com.twitter.zipkin.gen.AnnotationType;
 import com.twitter.zipkin.gen.BinaryAnnotation;
@@ -8,7 +9,6 @@ import com.twitter.zipkin.gen.Endpoint;
 import com.twitter.zipkin.gen.Span;
 
 import java.nio.ByteBuffer;
-import java.nio.charset.Charset;
 
 import static com.github.kristofa.brave.internal.Util.checkNotBlank;
 import static com.github.kristofa.brave.internal.Util.checkNotNull;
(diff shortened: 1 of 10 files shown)


adriancole:
@kristofa PTAL I'm done I think :)

kristofa:
@adriancole Thank you! LGTM!

PR #113, opened 2015-11-22
Kafka SpanCollector implementation

### LINKED ISSUE #102: Kafka Span Transport
We need a kafka span transport, similar to the one in [htrace](https://github.com/apache/incubator-htrace/commit/9ebdafb45df03a4b0d90c77af65e0bd21c09ec1a).

Here's some example code that works against zipkin server infrastructure:

```java
import com.github.kristofa.brave.SpanCollector;
import java.io.ByteArrayOutputStream;
import java.io.Flushable;
import java.util.ArrayList;
import java.util.List;
import java.util.Properties;
import java.util.concurrent.BlockingQueue;
import java.util.concurrent.LinkedBlockingQueue;
import kafka.javaapi.producer.Producer;
import kafka.producer.KeyedMessage;
import kafka.producer.ProducerConfig;
import org.apache.thrift.TException;
import org.apache.thrift.protocol.TBinaryProtocol;
import org.apache.thrift.protocol.TProtocol;
import org.apache.thrift.transport.TIOStreamTransport;

public class KafkaSpanCollector implements SpanCollector, Flushable {
  private final Producer<byte[], byte[]> producer;
  // in real life, this would be bounded.
  private BlockingQueue<com.twitter.zipkin.gen.Span> queue = new LinkedBlockingQueue<>();
  private int limit = 200;

  public KafkaSpanCollector() {
    Properties props = new Properties();
    props.put("metadata.broker.list", System.getenv("METADATA_BROKER_LIST"));
    props.put("request.required.acks", "0");
    props.put("producer.type", "async");
    props.put("serializer.class", "kafka.serializer.DefaultEncoder");
    props.put("compression.codec", "1");
    this.producer = new Producer<>(new ProducerConfig(props));
  }

  @Override
  public void collect(com.twitter.zipkin.gen.Span span) {
    queue.offer(span);
    if (queue.size() >= limit) {
      flush();
    }
  }

  @Override
  public void flush() {
    ByteArrayOutputStream baos = new ByteArrayOutputStream();
    final TProtocol streamProtocol = new TBinaryProtocol.Factory().getProtocol(new TIOStreamTransport(baos));

    List<KeyedMessage<byte[], byte[]>> messsages = new ArrayList<>(queue.size());
    while (!queue.isEmpty()) {
      com.twitter.zipkin.gen.Span span = queue.poll();
      if (span != null) {
        baos.reset();
        try {
          span.write(streamProtocol);
          messsages.add(new KeyedMessage<>("zipkin", baos.toByteArray()));
        } catch (TException e) {
          throw new AssertionError(e);
        }

      }
    }
    if (!messsages.isEmpty()) {
      producer.send(messsages);
    }
  }

  @Override
  public void addDefaultAnnotation(String key, String value) {
  }

  @Override
  public void close() {
    producer.close();
  }
}
```

kristofa:
Supports submitting spans to Kafka.
Fixed #102

### DIFF
--- a/brave-spancollector-kafka/LICENSE
+++ b/brave-spancollector-kafka/LICENSE
@@ -1,0 +1,13 @@
+Copyright 2015 <kristofa@github.com>
+
+  Licensed under the Apache License, Version 2.0 (the "License");
+  you may not use this file except in compliance with the License.
+  You may obtain a copy of the License at
+  
+      http://www.apache.org/licenses/LICENSE-2.0
+  
+  Unless required by applicable law or agreed to in writing, software
+  distributed under the License is distributed on an "AS IS" BASIS,
+  WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+  See the License for the specific language governing permissions and
+  limitations under the License.
--- a/brave-spancollector-kafka/README.md
+++ b/brave-spancollector-kafka/README.md
@@ -1,0 +1,6 @@
+# brave-spancollector-kafka #
+
+SpanCollector that is used to submit spans to Kafka.
+
+Spans are sent to kafka as keyed messages: the key is the topic `zipkin` 
+and the value is a TBinaryProtocol encoded Span.
--- a/brave-spancollector-kafka/src/main/java/com/github/kristofa/brave/kafka/KafkaSpanCollector.java
+++ b/brave-spancollector-kafka/src/main/java/com/github/kristofa/brave/kafka/KafkaSpanCollector.java
@@ -1,0 +1,96 @@
+package com.github.kristofa.brave.kafka;
+
+import com.github.kristofa.brave.SpanCollector;
+
+import java.io.Closeable;
+import java.util.Properties;
+import java.util.concurrent.*;
+import java.util.logging.Level;
+import java.util.logging.Logger;
+
+import com.twitter.zipkin.gen.Span;
+import org.apache.kafka.clients.producer.KafkaProducer;
+import org.apache.kafka.clients.producer.Producer;
+
+/**
+ * SpanCollector which submits spans to Kafka using <a href="http://kafka.apache.org/documentation.html#producerapi">Kafka Producer api</a>.
+ * <p>
+ * Spans are sent to kafka as keyed messages: the key is the topic zipkin and the value is a TBinaryProtocol encoded Span.
+ * </p>
+ */
+public class KafkaSpanCollector implements SpanCollector, Closeable {
+
+    private static final Logger LOGGER = Logger.getLogger(KafkaSpanCollector.class.getName());
+    private static final Properties DEFAULT_PROPERTIES = new Properties();
+
+    static {
+        DEFAULT_PROPERTIES.put("key.serializer", "org.apache.kafka.common.serialization.ByteArraySerializer");
+        DEFAULT_PROPERTIES.put("value.serializer", "org.apache.kafka.common.serialization.ByteArraySerializer");
+    }
+
+    private static Properties defaultPropertiesWith(String bootstrapServers) {
+        Properties props = new Properties();
+        for(String name: DEFAULT_PROPERTIES.stringPropertyNames()) {
+            props.setProperty(name, DEFAULT_PROPERTIES.getProperty(name));
+        }
+        props.setProperty("bootstrap.servers", bootstrapServers);
+        return props;
+    }
+
+    private final Producer<byte[], byte[]> producer;
+    private final ExecutorService executorService;
+    private final SpanProcessingTask spanProcessingTask;
+    private final Future<Integer> future;
+    private final BlockingQueue<Span> queue;
+
+    /**
+     * Create a new instance with default configuration.
+     *
+     * @param bootstrapServers A list of host/port pairs to use for establishing the initial connection to the Kafka cluster.
+     *                         Like: host1:port1,host2:port2,... Does not to be all the servers part of Kafka cluster.
+     */
+    public KafkaSpanCollector(String bootstrapServers) {
+        this(KafkaSpanCollector.defaultPropertiesWith(bootstrapServers));
+    }
+
+    /**
+     * KafkaSpanCollector.
+     *
+     * @param kafkaProperties    Configuration for Kafka producer. Essential configuration properties are:
+     *                           bootstrap.servers, key.serializer, value.serializer. For a
+     *                           full list of config options, see http://kafka.apache.org/documentation.html#producerconfigs.
+     */
+    public KafkaSpanCollector(Properties kafkaProperties) {
+        producer = new KafkaProducer<>(kafkaProperties);
+        executorService = Executors.newSingleThreadExecutor();
+        queue = new ArrayBlockingQueue<Span>(1000);
+        spanProcessingTask = new SpanProcessingTask(queue, producer);
+        future = executorService.submit(spanProcessingTask);
+    }
+
+    @Override
+    public void collect(com.twitter.zipkin.gen.Span span) {
+        if (!queue.offer(span)) {
+            LOGGER.log(Level.WARNING, "Queue rejected span!");
+        }
+    }
+
+    @Override
+    public void addDefaultAnnotation(String key, String value) {
+        throw new UnsupportedOperationException();
+    }
+
+    @Override
+    public void close() {
+        spanProcessingTask.stop();
+        try {
+            Integer nrProcessedSpans = future.get(6000, TimeUnit.MILLISECONDS);
+            LOGGER.info("SpanProcessingTask processed " + nrProcessedSpans + " spans.");
+        } catch (Exception e) {
+            LOGGER.log(Level.WARNING, "Exception when waiting for SpanProcessTask to finish.", e);
+        }
+        executorService.shutdown();
+        producer.close();
+        LOGGER.info("KafkaSpanCollector closed.");
+    }
+}
--- a/brave-spancollector-kafka/src/main/java/com/github/kristofa/brave/kafka/SpanProcessingTask.java
+++ b/brave-spancollector-kafka/src/main/java/com/github/kristofa/brave/kafka/SpanProcessingTask.java
@@ -1,0 +1,61 @@
+package com.github.kristofa.brave.kafka;
+
+import com.twitter.zipkin.gen.Span;
+import org.apache.kafka.clients.producer.Producer;
+import org.apache.kafka.clients.producer.ProducerRecord;
+import org.apache.thrift.TException;
+import org.apache.thrift.protocol.TBinaryProtocol;
+import org.apache.thrift.protocol.TProtocol;
+import org.apache.thrift.transport.TIOStreamTransport;
+
+import java.io.ByteArrayOutputStream;
+import java.util.concurrent.BlockingQueue;
+import java.util.concurrent.Callable;
+import java.util.concurrent.TimeUnit;
+import java.util.logging.Level;
+import java.util.logging.Logger;
+
+/**
+ * Processes spans by sending them, one at a time, to the topic `zipkin`, encoded in {@linkplain TBinaryProtocol}.
+ * <p>
+ * <p/> Note: this class was written to be used by a single-threaded executor, hence it is not thead-safe.
+ */
+class SpanProcessingTask implements Callable<Integer> {
+
+    private static final Logger LOGGER = Logger.getLogger(SpanProcessingTask.class.getName());
+    private final BlockingQueue<Span> queue;
+    private final Producer<byte[], byte[]> producer;
+    private volatile boolean stop = false;
+    private int numProcessedSpans = 0;
+
+    SpanProcessingTask(BlockingQueue<Span> queue, Producer<byte[], byte[]> producer) {
+        this.queue = queue;
+        this.producer = producer;
+    }
+
+    public void stop() {
+        stop = true;
+    }
+
+    @Override
+    public Integer call() throws Exception {
+        final ByteArrayOutputStream baos = new ByteArrayOutputStream(1024);
+        final TProtocol streamProtocol = new TBinaryProtocol.Factory().getProtocol(new TIOStreamTransport(baos));
+        do {
+            final Span span = queue.poll(5, TimeUnit.SECONDS);
+            if (span == null) {
+                continue;
+            }
+            baos.reset();
+            try {
+                span.write(streamProtocol);
+                final ProducerRecord<byte[], byte[]> message = new ProducerRecord<>("zipkin", baos.toByteArray());
+                producer.send(message);
+                numProcessedSpans++;
+            } catch (TException e) {
+                LOGGER.log(Level.WARNING, "TException when writing span.", e);
+            }
+        } while (!stop);
+        return numProcessedSpans;
+    }
+}
--- a/brave-spancollector-kafka/src/test/java/com/github/kristofa/brave/kafka/ITKafkaSpanCollector.java
+++ b/brave-spancollector-kafka/src/test/java/com/github/kristofa/brave/kafka/ITKafkaSpanCollector.java
@@ -1,0 +1,104 @@
+package com.github.kristofa.brave.kafka;
+
+import com.github.charithe.kafka.KafkaJunitRule;
+import com.github.kristofa.brave.SpanCollector;
+import com.twitter.zipkin.gen.Span;
+import kafka.consumer.ConsumerIterator;
+import kafka.consumer.ConsumerTimeoutException;
+import kafka.consumer.KafkaStream;
+import kafka.javaapi.consumer.ConsumerConnector;
+import kafka.producer.ProducerConfig;
+import kafka.serializer.DefaultDecoder;
+import org.apache.kafka.clients.consumer.ConsumerConfig;
+import org.apache.thrift.TDeserializer;
+import org.apache.thrift.TException;
+import org.apache.thrift.protocol.TBinaryProtocol;
+import org.junit.After;
+import org.junit.Before;
+import org.junit.Rule;
+import org.junit.Test;
+import java.util.*;
+import java.util.concurrent.*;
+
+import static org.junit.Assert.assertEquals;
+
+public class ITKafkaSpanCollector {
+
+    private final static String TOPIC = "zipkin";
+
+    @Rule
+    public KafkaJunitRule kafkaRule = new KafkaJunitRule();
+
+    @Test
+    public void submitSingleSpan() throws TException, TimeoutException {
+
+        SpanCollector kafkaCollector = new KafkaSpanCollector("localhost:"+kafkaRule.kafkaBrokerPort());
+        Span span = span(1l, "test_kafka_span");
+        kafkaCollector.collect(span);
+        kafkaCollector.close();
+
+        List<Span> spans = getCollectedSpans(kafkaRule.readMessages("zipkin", 1, new DefaultDecoder(kafkaRule.consumerConfig().props())));
+        assertEquals(1, spans.size());
+        assertEquals(span, spans.get(0));
+    }
+
+    @Test
+    public void submitMultipleSpansInParallel() throws InterruptedException, ExecutionException, TimeoutException, TException {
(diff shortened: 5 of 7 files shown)


kristofa:
@adriancole Kafka SpanCollector implementation.

brave-spancollector-kafka/LICENSE
```diff
@@ -0,0 +1,13 @@
+Copyright 2013 <kristofa@github.com>
```
adriancole: supernit it's almost 2016! :)

brave-spancollector-kafka/README.md
```diff
@@ -0,0 +1,4 @@
+# brave-spancollector-kafka #
+
+SpanCollector that is used to submit spans to Kafka.
+It submits spans to a topic named `zipkin`.
```
adriancole: Spans are sent to kafka as keyed messages: the key is the topic `zipkin` and the value is a TBinaryProtocol encoded Span.

brave-spancollector-kafka/src/main/java/com/github/kristofa/brave/kafka/KafkaSpanCollector.java
```diff
@@ -0,0 +1,100 @@
...
+import java.util.logging.Level;
+import java.util.logging.Logger;
+
+import com.twitter.zipkin.gen.Span;
+import kafka.javaapi.producer.Producer;
+import kafka.producer.ProducerConfig;
+
+/**
+ * SpanCollector which submits spans to Kafka using <a href="http://kafka.apache.org/documentation.html#producerapi">Kafka Producer api</a>.
+ * <p>
```
adriancole: add: <p/>Spans are sent to kafka as keyed messages: the key is the topic zipkin and the value is a TBinaryProtocol encoded Span.

brave-spancollector-kafka/src/main/java/com/github/kristofa/brave/kafka/KafkaSpanCollector.java
```diff
@@ -0,0 +1,100 @@
...
+     *
+     * @param metadataBrokerList Comma separated list of brokers for retrieving metadata and
+     *                           bootstrapping producer. Like: host1:port1,host2:port2,...
+     */
+    public KafkaSpanCollector(String metadataBrokerList) {
+        this(KafkaSpanCollector.defaultPropertiesWith(metadataBrokerList));
+    }
+
+    /**
+     * KafkaSpanCollector.
```
adriancole: scrub unless you have some sort of javadoc that needs text here.

brave-spancollector-kafka/src/main/java/com/github/kristofa/brave/kafka/KafkaSpanCollector.java
```diff
@@ -0,0 +1,100 @@
...
+        executorService = Executors.newSingleThreadExecutor();
+        queue = new ArrayBlockingQueue<Span>(1000);
+        spanProcessingTask = new SpanProcessingTask(queue, producer);
+        future = executorService.submit(spanProcessingTask);
+    }
+
+    @Override
+    public void collect(com.twitter.zipkin.gen.Span span) {
+        if (queue.offer(span) == false) {
+            LOGGER.log(Level.WARNING, "Queue rejected span!");
```
adriancole: is this consistent w/scribe? handling (ex. warning level)? just curious as this hopefully won't fill logs, but it could.

brave-spancollector-kafka/src/main/java/com/github/kristofa/brave/kafka/KafkaSpanCollector.java
```diff
@@ -0,0 +1,100 @@
...
+        producer = new Producer<>(new ProducerConfig(kafkaProperties));
+        executorService = Executors.newSingleThreadExecutor();
+        queue = new ArrayBlockingQueue<Span>(1000);
+        spanProcessingTask = new SpanProcessingTask(queue, producer);
+        future = executorService.submit(spanProcessingTask);
+    }
+
+    @Override
+    public void collect(com.twitter.zipkin.gen.Span span) {
+        if (queue.offer(span) == false) {
```
adriancole: supernit: midly shorter to just say `!queue.offer(span)`

brave-spancollector-kafka/src/main/java/com/github/kristofa/brave/kafka/KafkaSpanCollector.java
adriancole: another option is to increment a local counter and then at close log the warning

brave-spancollector-kafka/src/main/java/com/github/kristofa/brave/kafka/SpanProcessingTask.java
```diff
@@ -0,0 +1,56 @@
...
+import org.apache.thrift.transport.TIOStreamTransport;
+
+import java.io.ByteArrayOutputStream;
+import java.util.concurrent.BlockingQueue;
+import java.util.concurrent.Callable;
+import java.util.concurrent.TimeUnit;
+import java.util.logging.Level;
+import java.util.logging.Logger;
+
+class SpanProcessingTask implements Callable<Integer> {
```
adriancole: nit: javadoc clarifying this is intended for a single thread.

```
/**
 * Processes spans by sending them, one at a time, to the topic `zipkin`, encoded in {@linkplain TBinaryProtocol}.
 *
 * <p/> Note: this class was written to be used by a single-threaded executor, hence it is not thead-safe.
 */
```

brave-spancollector-kafka/src/main/java/com/github/kristofa/brave/kafka/SpanProcessingTask.java
```diff
@@ -0,0 +1,56 @@
...
+
+    private static final Logger LOGGER = Logger.getLogger(SpanProcessingTask.class.getName());
+    private final BlockingQueue<Span> queue;
+    private final Producer<byte[], byte[]> producer;
+    private volatile boolean stop = false;
```
adriancole: // volatile as stop can be called from any thread.

brave-spancollector-kafka/src/main/java/com/github/kristofa/brave/kafka/SpanProcessingTask.java
```diff
@@ -0,0 +1,56 @@
...
+            final Span span = queue.poll(5, TimeUnit.SECONDS);
+            if (span != null) {
+                baos.reset();
+                try {
+                    span.write(streamProtocol);
+                    final KeyedMessage<byte[], byte[]> message = new KeyedMessage<>("zipkin", baos.toByteArray());
+                    producer.send(message);
+                    numProcessedSpans++;
+                } catch (TException e) {
+                    LOGGER.log(Level.WARNING, "TException when writing span.", e);
```
adriancole: generally, libraries should be careful logging at WARNING level.

brave-spancollector-kafka/src/main/java/com/github/kristofa/brave/kafka/SpanProcessingTask.java
```diff
@@ -0,0 +1,56 @@
...
+
+    public void stop() {
+        stop = true;
+    }
+
+    @Override
+    public Integer call() throws Exception {
+        final ByteArrayOutputStream baos = new ByteArrayOutputStream(1024);
+        final TProtocol streamProtocol = new TBinaryProtocol.Factory().getProtocol(new TIOStreamTransport(baos));
+        do {
```
adriancole: nit: while (!stop) is less complex imho, than do while

brave-spancollector-kafka/src/main/java/com/github/kristofa/brave/kafka/SpanProcessingTask.java
```diff
@@ -0,0 +1,56 @@
...
+            final Span span = queue.poll(5, TimeUnit.SECONDS);
+            if (span != null) {
```
adriancole: you can remove a level of indent, increasing readablity, by early exiting.
Ex
```java
if (span == null) { //  no spans waiting to be processed
 continue;
}
``

File: brave-spancollector-kafka/src/test/java/com/github/kristofa/brave/kafka/ITKafkaSpanCollector.java [hunk 12]
```diff
@@ -0,0 +1,83 @@
...
+import org.apache.thrift.protocol.TBinaryProtocol;
+import org.junit.After;
+import org.junit.Before;
+import org.junit.Test;
+import java.util.*;
+import static org.junit.Assert.assertEquals;
+
+public class ITKafkaSpanCollector {
+
+    private final static String TOPIC = "zipkin";
```
adriancole: I'd move the TOPIC constant name zipkin to somewhere in the main tree, package private, then use it everywhere.

brave-spancollector-kafka/src/test/java/com/github/kristofa/brave/kafka/ITKafkaSpanCollector.java
```diff
@@ -0,0 +1,124 @@
...
+        // Prevents that our ConsumerIterator blocks forever.
+        consumerProps.put("consumer.timeout.ms", "5000");
+        ConsumerConnector connector =
+                kafka.consumer.Consumer.createJavaConsumerConnector(new kafka.consumer.ConsumerConfig(consumerProps));
+        Map<String, Integer> topicCountMap = new HashMap<>();
+        topicCountMap.put(TOPIC, 1);
+        Map<String, List<KafkaStream<byte[], byte[]>>> streams = connector.createMessageStreams(topicCountMap);
+
+        TDeserializer deserializer = new TDeserializer(new TBinaryProtocol.Factory());
+        ConsumerIterator<byte[], byte[]> it = streams.get(TOPIC).get(0).iterator();
```
adriancole: fyi tests can be written in java 8 in case that helps..

brave-spancollector-kafka/src/main/java/com/github/kristofa/brave/kafka/KafkaSpanCollector.java
kristofa: The scribe implementation also logs dropped spans as warning messages. I think we should revisit once we have #101 in place.

brave-spancollector-kafka/src/test/java/com/github/kristofa/brave/kafka/KafkaTestServer.java
```diff
@@ -0,0 +1,65 @@
...
+import kafka.utils.ZKStringSerializer$;
+import kafka.zk.EmbeddedZookeeper;
+import org.I0Itec.zkclient.ZkClient;
+import scala.collection.JavaConversions;
+import scala.collection.mutable.Buffer;
+
+import java.util.Collections;
+import java.util.Properties;
+
+public class KafkaTestServer {
```
adriancole: non-blocking note:
maybe just me, but I think `com.github.charithe.kafka.KafkaJunitRule` will lead to less clutter. It is hard to see what these tests are doing. If it is too much work, I could help later, but unfortunately not likely in the next few days.

Ex. It should be much cleaner than this, to use the rule in java: https://github.com/openzipkin/zipkin/blob/master/zipkin-receiver-kafka/src/test/scala/com/twitter/zipkin/receiver/kafka/KafkaProcessorSpec.scala

adriancole:
LG: no non-blocking notes. Merge when you are happy with the code!

brave-spancollector-kafka/src/main/java/com/github/kristofa/brave/kafka/KafkaSpanCollector.java
kristofa: Added line below.

brave-spancollector-kafka/src/main/java/com/github/kristofa/brave/kafka/KafkaSpanCollector.java
kristofa: As we use a String to submit `metadataBrokerList` I think explanation about how to submit hosts + ports is relevant? You think it is obvious enough so we can get rid of javadoc?

adriancole: I was referring to "KafkaSpanCollector." :)

charliek:
Would it make sense to use the new kafka producers for this?  That would eliminate the scala dependency all together, and does not talk to zookeeper directly. In general for apps I work on the api is so similar so converting is not too bad.

Quick overview of the new producers:
http://www.confluent.io/blog/whats-coming-in-apache-kafka-0-8-2/

Maven dependency:
```
<dependency>
	<groupId>org.apache.kafka</groupId>
	<artifactId>kafka-clients</artifactId>
	<version>0.8.2.2</version>
</dependency>
```

kristofa:
@charliek Thanks for the tip. I was not aware there was a new api. I think it is a good idea to migrate to the new producer api. Not needing a dependency on Scala and simplification (sync / async) are good.

Reason why this pr is not merged yet is that I want to get it working with the `docker-zipkin` setup configured to use kafka. I was unsuccessful so far. Once it works with the current producer api we can rework it to use the new api.

charliek:
Sounds good. I've just started digging into zipkin/brave and was excited to see this.

kristofa:
@adriancole Here the information to reproduce the error I get when I try to submit a span to Kafka.

`docker-zipkin docker-compose.yml`:

```
kafka:
  image: openzipkin/zipkin-kafka:1.17.2
  ports:
    - 2181:2181
    - 9092:9092
cassandra:
  image: openzipkin/zipkin-cassandra:1.25.0
  ports:
    - 9042:9042
# The collector process is used for legacy zipkin instrumentation, which log via
# scribe. It can also poll kafka, when the KAFKA_ZOOKEEPER variable is set.
collector:
  image: openzipkin/zipkin-collector:1.25.0
  environment:
    - TRANSPORT_TYPE=kafka
    - STORAGE_TYPE=cassandra
  expose:
    - 9410
  ports:
    - 9410:9410
    - 9900:9900
  links:
    - cassandra:storage
    - kafka
# The query process services the UI, and also exposes a POST endpoint that
# instrumentation can send trace data to.
query:
  image: openzipkin/zipkin-query:1.25.0
  environment:
    # Remove TRANSPORT_TYPE to disable tracing
    - TRANSPORT_TYPE=http
    - STORAGE_TYPE=cassandra
  expose:
    - 9411
  ports:
    - 9411:9411
    - 9901:9901
  links:
    - cassandra:storage
web:
  image: openzipkin/zipkin-web:1.25.0
  environment:
    # Remove TRANSPORT_TYPE to disable tracing
    - TRANSPORT_TYPE=http
  ports:
    - 8080:8080
    - 9990:9990
  links:
    - query
```

Run `docker-compose up` with above file on top of current master of docker-zipkin
Once services are up and running I execute:

```
import com.twitter.zipkin.gen.Annotation;
import com.twitter.zipkin.gen.Endpoint;
import com.twitter.zipkin.gen.Span;
import com.twitter.zipkin.gen.zipkinCoreConstants;
import org.junit.After;
import org.junit.Before;
import org.junit.Test;

import java.util.Random;

public class SubmitSpansTest {

    private final static String SERVICE_NAME = "kafka_service_test";
    private final static String SPAN_NAME = "kafka_request";

    private KafkaSpanCollector collector;
    private Random random;

    @Before
    public void before() {
        collector = new KafkaSpanCollector("192.168.99.100:9092");
        random = new Random();
    }

    @After
    public void after() {
        collector.close();
    }

    @Test
    public void submitSpan() throws InterruptedException {
        long traceId = random.nextLong();
        Span span = span(traceId, traceId, null, SPAN_NAME);
        addAnnotation(span, zipkinCoreConstants.CLIENT_SEND);
        Thread.sleep(50);
        addAnnotation(span, zipkinCoreConstants.CLIENT_RECV);
        collector.collect(span);
    }

    private Span span(long traceId, long spanId, Long parentSpanId, String spanName) {
        Span span = new Span();
        span.setId(spanId);
        span.setTrace_id(traceId);
        if (parentSpanId != null)
        {
            span.setParent_id(parentSpanId.longValue());
        }
        span.setName(spanName);
        return span;
    }

    private void addAnnotation(Span span, String annotationName) {
        Annotation annotation = new Annotation();
        annotation.setTimestamp(currentTimeMicroseconds());
        annotation.setHost(endPoint());
        annotation.setValue(annotationName);
        synchronized (span) {
            span.setTimestamp(annotation.getTimestamp());
            span.addToAnnotations(annotation);
        }
    }

    private long currentTimeMicroseconds() {
        return System.currentTimeMillis() * 1000;
    }

    private Endpoint endPoint() {
        Endpoint e = new Endpoint();
        e.setIpv4(325334);
        e.setPort((short)80);
        e.setService_name(SERVICE_NAME);
        return e;
    }
}
```

Error logs I get from client (test):

```
[2015-11-30 20:43:08,113] ERROR Failed to collate messages by topic, partition due to: Failed to fetch topic metadata for topic: zipkin (kafka.producer.async.DefaultEventHandler:97)
[2015-11-30 20:43:08,403] ERROR Failed to collate messages by topic, partition due to: Failed to fetch topic metadata for topic: zipkin (kafka.producer.async.DefaultEventHandler:97)
[2015-11-30 20:43:09,001] ERROR Failed to send requests for topics zipkin with correlation ids in [0,10] (kafka.producer.async.DefaultEventHandler:97)
```
Error logs I get from Kafka:

```
kafka_1     | 2015-11-30 19:43:09,253 [myid:] - INFO  [ProcessThread(sid:0 cport:-1)::PrepRequestProcessor@645] - Got user-level KeeperException when processing sessionid:0x15159e8ef920001 type:setData cxid:0x29 zxid:0x27 txntype:-1 reqpath:n/a Error Path:/consumers/zipkin/offsets/zipkin/0 Error:KeeperErrorCode = NoNode for /consumers/zipkin/offsets/zipkin/0
kafka_1     | 2015-11-30 19:43:09,257 [myid:] - INFO  [ProcessThread(sid:0 cport:-1)::PrepRequestProcessor@645] - Got user-level KeeperException when processing sessionid:0x15159e8ef920001 type:create cxid:0x2a zxid:0x28 txntype:-1 reqpath:n/a Error Path:/consumers/zipkin/offsets Error:KeeperErrorCode = NoNode for /consumers/zipkin/offsets
```

If I execute the above test code with the default set-up, submitting directly to collector using `ScribeSpanCollector` instead of `KafkaSpanCollector` I can see the Span using zipkin-web.

adriancole:
thanks! I'm really slow this week due to a move. current ETA is 1 week, so
if anyone else wants to try and beat me to it, please do!

charliek:
@kristofa I saw the same issues as you did but was able to work around them by changing the compose setup.  I documented the steps I took to get this working here:  https://gist.github.com/charliek/7958b117400f01b70992

brave-spancollector-kafka/src/main/java/com/github/kristofa/brave/kafka/KafkaSpanCollector.java
charliek: I'm not actually sure this is correct. As written it looks like the Keyed message is given the topic of "zipkin" and uses a null key.  Eventually it may make sense to be able to configure a topic other than zipkin, as opposed to hard coding it.

You wouldn't want to set the key here to "zipkin" as well since that would mean all messages would hash to the same kafka partition.  Using a null key is better as long as order isn't important, because null will hash randomly to each partition however order is only guaranteed per partition.

kristofa:
@charliek Thanks a lot! I'll give your instructions a try soon!

adriancole:
note zipkin (as of 1.26.0) is running kafka 0.9

kristofa:
@charliek I was finally able to try your instructions and they indeed work! Thanks! Next I'll upgrade to kafka 0.9.

kristofa:
@charliek @adriancole Upgraded to `kafka_clients 0.9.0.0` (which support Kafka 0.9.0.0 new features and gets rid of Scala dependency) and using `com.github.charithe:kafka-junit` which worked right away and indeed makes testing code a lot cleaner! Also the docker setup explained by @charliek still works with the 0.9.0.0 client.

charliek:
Glad those instructions worked @kristofa . Thanks for your work on this I'm excited to try it out.

PR #114, opened 2015-12-09
Resolves state conflict between client and local spans

adriancole:
Before, local spans used the same thread local as client ones. This
meant a client span in the same trace would overwite a local span.

Ex. [server -> local -> client] would truncate to [server -> client]

This change adds state to ensure local spans aren't overwritten by
client spans.

This change does not introduce a state stack, eventhough local spans
should be nestable. The reason is that Brave's model isn't currently
designed for that and a nested model would significantly impact the api.

### DIFF
--- a/brave-core/src/main/java/com/github/kristofa/brave/Brave.java
+++ b/brave-core/src/main/java/com/github/kristofa/brave/Brave.java
@@ -27,14 +27,14 @@ public class Brave {
      * Builds Brave api objects with following defaults if not overridden:
      * <p>
      * <ul>
-     * <li>ThreadLocalServerAndClientSpanState which binds trace/span state to current thread.</li>
+     * <li>ThreadLocalServerClientAndLocalSpanState which binds trace/span state to current thread.</li>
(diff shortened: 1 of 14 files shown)


adriancole:
I've tested this with zipkin-java, and it fixed the problem I had earlier, where local components were lost when they made client calls. You can now see local components originating JDBC client calls in the below diagram

screen shot 2015-12-09 at 4 12 36 pm

See https://github.com/openzipkin/zipkin-java/pull/46

kristofa:
@adriancole I went over it and I think it looks good :smile: In my opinion this should probably be the latest extension to the 'state' code we do in its current form. 

I think the code will simplify when we upgrade to a stacked approach.  I try to find some time to have a closer look into the Distributed Context Propagation posts and attached docs in the Distributed Tracing Workshop google group.

kristofa:
@adriancole brave 3.2.0, which includes this pr, is released and should appear in Maven Central soon. Release notes will follow later, probably tomorrow.

PR #118, opened 2016-01-07
Introduces TraceSampler which replaces TraceFilter

adriancole:
TraceFilter worked off a span id, which isn't always the trace id. This
is confusing and could lead to sparse traces. This eliminates use of
TraceFilter, replaced with TraceSampler, which works off the trace id
and is implemented by a benchmarked implementation in zipkin-java.

The use of zipkin-java's TraceIdSampler is shaded, so as to neither
introduce a dependency, nor the perils of copy/paste.

Impact on users is that they should stop using TraceFilter immediately
as Brave will ignore them. Most often, they will want to only provide a
threshold, ex `TraceSampler.create(0.3f)` means retain 30%.

We also no longer publish `brave-tracefilters` and instead an equivalent
`brave-tracesamplers`, which has the same implementation, except based
on trace id as opposed to span id.

Finally, those using zookeeper should set the rate as a percentage, not 
a ratio. For example, `/brave/samplerate` would be 0.2 to retain 20%.


adriancole:
Validated the resulting pom has no zipkin-java dependency, and the classes added are <10K

```
$ jar -tvf /Users/acole/.m2/repository/com/github/kristofa/brave-core/3.3.1-SNAPSHOT/brave-core-3.3.1-SNAPSHOT.jar |grep com/github/kristofa/brave/internal/io/zipkin/
     0 Thu Jan 07 16:12:08 SGT 2016 com/github/kristofa/brave/internal/io/zipkin/
     0 Thu Jan 07 16:12:08 SGT 2016 com/github/kristofa/brave/internal/io/zipkin/internal/
  3124 Thu Jan 07 16:12:08 SGT 2016 com/github/kristofa/brave/internal/io/zipkin/internal/Util.class
   371 Thu Jan 07 16:12:08 SGT 2016 com/github/kristofa/brave/internal/io/zipkin/internal/Nullable.class
   671 Thu Jan 07 16:12:08 SGT 2016 com/github/kristofa/brave/internal/io/zipkin/TraceIdSampler$1.class
   670 Thu Jan 07 16:12:08 SGT 2016 com/github/kristofa/brave/internal/io/zipkin/TraceIdSampler$2.class
  1426 Thu Jan 07 16:12:08 SGT 2016 com/github/kristofa/brave/internal/io/zipkin/TraceIdSampler$ThresholdSampler.class
  1041 Thu Jan 07 16:12:08 SGT 2016 com/github/kristofa/brave/internal/io/zipkin/TraceIdSampler.class
```

brave-core/src/main/java/com/github/kristofa/brave/ClientTracer.java
```diff
@@ -37,7 +36,7 @@ public static Builder builder() {
     abstract ClientSpanAndEndpoint spanAndEndpoint();
     abstract Random randomGenerator();
     abstract SpanCollector spanCollector();
-    abstract List<TraceFilter> traceFilters();
```
adriancole: no more list.. if people have cause to create multiple tiers of filtering, they can make a composite one.

brave-core/src/main/java/com/github/kristofa/brave/TraceSampler.java
```diff
@@ -0,0 +1,43 @@
...
+   *
+   * @param rate minimum sample rate is 0.0001, or 0.01% of traces
+   */
+  public static TraceSampler create(float rate) {
+    return new ZipkinTraceSampler(rate);
+  }
+
+  static final class ZipkinTraceSampler extends TraceSampler {
+
+    private final io.zipkin.TraceIdSampler delegate;
```
adriancole: This wrapping type + shade will protect brave users from a dependency on an emerging library.

brave-tracefilters/src/main/java/com/github/kristofa/brave/tracefilter/ZooKeeperSamplingTraceFilter.java
```diff
@@ -1,170 +0,0 @@
...
-    /**
-     * Creates a new instance. If the initial connection with ZooKeeper can't be established within 2 seconds an unchecked
-     * exception will be thrown as this will probably indicate wrong configuration.
-     * 
-     * @param connectionString ZooKeeper connection string. Should not be <code>null</code> or empty.
-     * @param sampleRateZNode The znode that contains sample rate. Should not be <code>null</code> or empty.
-     * @throws IOException In case we can't connect with ZooKeeper.
-     * @throws InterruptedException In case we can't connect with ZooKeeper.
-     */
-    public ZooKeeperSamplingTraceFilter(final String connectionString, final String sampleRateZNode)
```
adriancole: this is a lot of red, but actually the new class is almost identical. Just that instead of swapping sampleRate, it swaps an instance of traceSampler.

brave-tracesamplers/pom.xml
```diff
@@ -7,9 +7,9 @@
     <version>3.3.1-SNAPSHOT</version>
   </parent>
   
-  <artifactId>brave-tracefilters</artifactId>
+  <artifactId>brave-tracesamplers</artifactId>
   <packaging>jar</packaging>
-  <name>brave-tracefilters</name>
+  <name>brave-tracesamplers</name>
```
adriancole: ps kindof inclined to call this brave-zookeeper or brave-sampler-zookeeper

adriancole: because if we add something inside that doesn't have a ZK dep, then we'd end up with a similar problem as if we put it in core (the problem of unnecessary deps).

brave-core/src/main/java/com/github/kristofa/brave/ClientTracer.java
kristofa: Yes, agree. I done think anyone actually used multiple filters.

brave-tracesamplers/pom.xml
kristofa: Yes, that's a good point. Let's do it!

brave-tracesamplers/pom.xml
```diff
@@ -7,9 +7,9 @@
...
   <description>
     Brave TraceFilter implementations. It contains for example a TraceFilter that gets sample rate from ZooKeeper and is updated automatically
```
kristofa: minor: Description still contains `TraceFilter`.

brave-tracesamplers/pom.xml
adriancole: seems brave-tracesampler-zookeeper follows more similarly to
brave-spancollector-scribe (lowercase type name-impl), so going with
that..

adriancole: done

brave-tracesamplers/pom.xml
adriancole: done

kristofa:
I think this is good to go. :+1: 

This is a breaking change so if we are strict with semantic versioning it should be 4.x. However I don't feel like it should be a major bump. Major bump would be rip out thrift and replace with Java implementation or replace propagation of state with new implementation. wdyt?

adriancole:
I think we've already tainted 3.x with similar changes :) This implies that
the front number is more a marketing one.. I'd recommend we make 4.x count
and just increment minor, particularly as this is an edge-case feature.

kristofa:
Yes, agree :) Previous breaking changes were renames as far as I can remember (like scribe span collector). But still breaking changes. This one has a larger impact. Let's make it 4.x

adriancole:
oh I actually meant save 4.x for larger impact (like opentracing) as this
is an ancillary function.

But if we do want to make this 4.x, then I'd just toss the classes instead
of deprecating. Which do you prefer?

kristofa:
Oh, ok, let's stick to my initial, and your view and leave it at to 3.x. We should just make the release notes clear about what needs to change.

adriancole:
yeah. essentially...

* If not using TraceFilter, do nothing
* If using TraceFilter, your code won't break, but your sample ratio will
be ignored.  Switch to TraceSampler with a percentage.
* If using ZK, you'll need to change your maven dependencies, change to the
ZookeeperTraceSampler, and re-assign your rate in ZK to
a percentage instead of a ratio.

kristofa:
:+1:

PR #119, opened 2016-01-07
Removes SpanCollector.close, deprecates addDefaultAnnotation

adriancole:
Closeable isn't the only way to close things, and in reality we weren't
doing anything with it. This punts the decision on how to close
resources to implementing types by removing `SpanCollector.close`

This also deprecates `SpanCollector.addDefaultAnnotation` as decorating
`SpanCollector.collect` can have the same effect.

Finally, this polishes call sites that make make binary annotations of
type string.

See https://github.com/openzipkin/zipkin/issues/584


adriancole:
rebased

kristofa:
:+1:

PR #120, opened 2016-01-08
Introduces HttpSpanCollector

adriancole:
This introduces HttpSpanCollector, which sends compressed thrifts to
zipkin's POST endpoint. It intends to be similar to OpenTracing
reporter.py, but relevant for Brave. There are a few notes about that.

OpenTracing's report_span seems to allow flushing on the calling thread
whereas Brave's collect should not (as it will block application code).

This implementation optimizes for sending as many spans as possible
within a single POST. Since current zipkin servers don't yet decompress,
spans are not gzipped.

Rather than introduce an async library dependency, this uses the stock
java http client and scheduler service. A future version could allow
this to be pluggable. Current impact is that only a single http-request
will be in-flight at a time.

See https://github.com/opentracing/api-python/blob/master/example/zipkin_like/reporter.py

### DIFF
--- a/brave-core/src/main/java/com/github/kristofa/brave/HttpSpanCollector.java
+++ b/brave-core/src/main/java/com/github/kristofa/brave/HttpSpanCollector.java
@@ -1,0 +1,234 @@
+package com.github.kristofa.brave;
+
+import com.github.kristofa.brave.internal.Nullable;
+import com.google.auto.value.AutoValue;
+import com.twitter.zipkin.gen.Span;
+import java.io.Closeable;
+import java.io.Flushable;
+import java.io.IOException;
+import java.io.InputStream;
+import java.net.HttpURLConnection;
+import java.net.URL;
+import java.util.ArrayDeque;
+import java.util.LinkedList;
+import java.util.Queue;
+import java.util.concurrent.BlockingQueue;
+import java.util.concurrent.Executors;
+import java.util.concurrent.LinkedBlockingQueue;
+import java.util.concurrent.ScheduledExecutorService;
+import okio.Buffer;
+import org.apache.thrift.TException;
+import org.apache.thrift.protocol.TBinaryProtocol;
+import org.apache.thrift.protocol.TList;
+import org.apache.thrift.protocol.TType;
+import org.apache.thrift.transport.TTransport;
+
+import static java.util.concurrent.TimeUnit.SECONDS;
+
+/**
+ * SpanCollector which submits spans to Zipkin, using its {@code POST /spans} endpoint.
+ */
+public final class HttpSpanCollector implements SpanCollector, Flushable, Closeable {
+
+  @AutoValue
+  public static abstract class Config {
+    public static Builder builder() {
+      return new AutoValue_HttpSpanCollector_Config.Builder()
+          .connectTimeout(10 * 1000)
+          .readTimeout(60 * 1000)
+          .flushInterval(1);
+    }
+
+    abstract int connectTimeout();
+
+    abstract int readTimeout();
+
+    abstract int flushInterval();
+
+    @AutoValue.Builder
+    public interface Builder {
+      /** Default 10 * 1000 milliseconds. 0 implies no timeout. */
+      Builder connectTimeout(int connectTimeout);
+
+      /** Default 60 * 1000 milliseconds. 0 implies no timeout. */
+      Builder readTimeout(int readTimeout);
+
+      /** Default 1 second. 0 implies spans are {@link #flush() flushed} externally. */
+      Builder flushInterval(int flushInterval);
+
+      Config build();
+    }
+  }
+
+  private final String url;
+  private final Config config;
+  private final SpanCollectorMetricsHandler metrics;
+  private final BlockingQueue<Span> pending = new LinkedBlockingQueue<>(1000);
+  @Nullable // for testing
+  private final Flusher flusher;
+
+  /**
+   * Create a new instance with default configuration.
+   *
+   * @param baseUrl URL of the zipkin query server instance. Like: http://localhost:9411/
+   * @param metrics Gets notified when spans are accepted or dropped. If you are not interested in
+   * these events you can use {@linkplain EmptySpanCollectorMetricsHandler}
+   */
+  public static HttpSpanCollector create(String baseUrl, SpanCollectorMetricsHandler metrics) {
+    return new HttpSpanCollector(baseUrl, Config.builder().build(), metrics);
+  }
+
+  /**
+   * @param baseUrl URL of the zipkin query server instance. Like: http://localhost:9411/
+   * @param config controls flush interval and timeouts
+   * @param metrics Gets notified when spans are accepted or dropped. If you are not interested in
+   * these events you can use {@linkplain EmptySpanCollectorMetricsHandler}
+   */
+  public static HttpSpanCollector create(String baseUrl, Config config,
+      SpanCollectorMetricsHandler metrics) {
+    return new HttpSpanCollector(baseUrl, config, metrics);
+  }
+
+  // Visible for testing. Ex when tests need to explicitly control flushing, set interval to 0.
+  HttpSpanCollector(String baseUrl, Config config,
+      SpanCollectorMetricsHandler metrics) {
+    this.url = baseUrl + (baseUrl.endsWith("/") ? "" : "/") + "api/v1/spans";
+    this.metrics = metrics;
+    this.config = config;
+    this.flusher = config.flushInterval() > 0 ? new Flusher(this, config.flushInterval()) : null;
+  }
+
+  /**
+   * Queues the span for collection, or drops it if the queue is full.
+   *
+   * @param span Span, should not be <code>null</code>.
+   */
+  @Override
+  public void collect(Span span) {
+    metrics.incrementAcceptedSpans(1);
+    if (!pending.offer(span)) {
+      metrics.incrementDroppedSpans(1);
+    }
+  }
+
+  /**
+   * Calling this will flush any pending spans to the http transport on the current thread.
+   */
+  @Override
+  public void flush() {
+    if (pending.isEmpty()) return;
+    Queue<Span> drained = new ArrayDeque<>(pending.size());
+    pending.drainTo(drained);
+    if (drained.isEmpty()) return;
+
+    // Thrift-encode the spans for transport
+    int spanCount = drained.size();
+    BufferTransport thrifts = new BufferTransport();
+    try {
+      TBinaryProtocol oprot = new TBinaryProtocol(thrifts);
+      oprot.writeListBegin(new TList(TType.STRUCT, drained.size()));
+      while (!drained.isEmpty()) {
+        drained.remove().write(oprot);
+      }
+      oprot.writeListEnd();
+    } catch (TException e) {
+      metrics.incrementDroppedSpans(spanCount);
+      return;
+    }
+
+    // Send the thrifts to the zipkin endpoint
+    try {
+      postSpans(thrifts.buffer);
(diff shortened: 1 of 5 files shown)


adriancole:
FYI @kristofa @marcingrzejszczak in case you're interested in where I'm going with the http adapter.

brave-core/src/main/java/com/github/kristofa/brave/HttpSpanCollector.java
```diff
@@ -0,0 +1,120 @@
...
+   *
+   * @param baseUrl URL of the zipkin query server instance. Like: http://localhost:9411/api/v1
+   * @param metricsHandler Gets notified when spans are accepted or dropped. If you are not
+   * interested in these events you can use {@linkplain EmptySpanCollectorMetricsHandler}
+   */
+  public HttpSpanCollector(String baseUrl, SpanCollectorMetricsHandler metricsHandler) {
+    this.baseUrl = baseUrl;
+    this.metricsHandler = metricsHandler;
+    queue = new LinkedBlockingQueue<>(1000); // if 1000 spans are pending, new spans will be dropped
+    new Thread(new SpanProcessor(), "HttpSpanCollector: " + baseUrl).start();
```
adriancole: single producer thread has pros and cons. The pros are that you can make a lot of assumptions.. and this simplifies the code design. For example, if someone wanted to have multiple collectors threads, they could make a fan-out SpanCollector that instantiates multiple of these!

brave-core/src/main/java/com/github/kristofa/brave/HttpSpanCollector.java
```diff
@@ -0,0 +1,120 @@
...
+    // during each loop, we re-use the list of spans to send
+    final Queue<Span> spansToSend = new ArrayDeque<>();
+
+    @Override
+    public void run() {
+      while (running.get() || !queue.isEmpty()) {
+
+        // Take as many spans as we can from the queue.
+        try {
+          Span first = queue.poll(1, TimeUnit.SECONDS);
```
adriancole: this "quick check, then drain" approach was an idea from htrace

brave-core/src/main/java/com/github/kristofa/brave/HttpSpanCollector.java
```diff
@@ -0,0 +1,120 @@
...
+        } catch (InterruptedException ie) {
+          metricsHandler.incrementDroppedSpans(spansToSend.size());
+          continue;
+        }
+
+        if (spansToSend.isEmpty()) continue;
+
+        // Thrift-encode the spans for transport
+        int encodedSpanCount = spansToSend.size();
+        TMemoryBuffer buffer = new TMemoryBuffer(32 /** initial size */);
```
adriancole: I'll probably change this as it is not efficient (wastes byte arrays vs re-using a buffer, esp as this is single-threaded)

adriancole:
OK complete rewrite. I plan to add more tests tomorrow.

brave-core/src/main/java/com/github/kristofa/brave/HttpSpanCollector.java
```diff
@@ -0,0 +1,198 @@
...
+import java.net.HttpURLConnection;
+import java.net.URL;
+import java.util.ArrayDeque;
+import java.util.LinkedList;
+import java.util.Queue;
+import java.util.concurrent.BlockingQueue;
+import java.util.concurrent.Executors;
+import java.util.concurrent.LinkedBlockingQueue;
+import java.util.concurrent.ScheduledExecutorService;
+import okio.Buffer;
```
adriancole: these are re-packaged and minified. adds 50k I think to the jar file. I think it is worth it, as brave is still under 256k including other imports we've done.

brave-core/src/main/java/com/github/kristofa/brave/HttpSpanCollector.java
```diff
@@ -0,0 +1,198 @@
...
+    Flusher(Flushable flushable, int flushInterval) {
+      this.flushable = flushable;
+      this.scheduler.scheduleWithFixedDelay(this, 0, flushInterval, SECONDS);
+    }
+
+    @Override
+    public void run() {
+      try {
+        flushable.flush();
+      } catch (IOException ignored) {
```
adriancole: ignored bc Flusher doesn't actually throw an IOE.. this accepts the implemented interface for testing purposes, and since that declares IOE, we have to catch it.

brave-core/src/main/java/com/github/kristofa/brave/HttpSpanCollector.java
```diff
@@ -0,0 +1,198 @@
...
+      }
+    }
+  }
+
+  void postSpans(Buffer gzippedThrifts) throws IOException {
+    // intentionally not closing the connection, so as to use keep-alives
+    HttpURLConnection connection = (HttpURLConnection) new URL(url).openConnection();
```
adriancole: wanted to fold-in okhttp, but that'd make too big a jar :P better to make transport an option later (particulary when asked). Less options means less to test, so there's value in incremental development, as tests are still not overwhelmingly long.

brave-core/src/test/java/com/github/kristofa/brave/HttpSpanCollectorTest.java
```diff
@@ -0,0 +1,130 @@
...
+    collector.collect(span(2L, "bar"));
+
+    collector.flush(); // manually flush the spans
+
+    assertThat(metrics.droppedSpans.get()).isEqualTo(2);
+  }
+
+  class TestMetricsHander implements SpanCollectorMetricsHandler {
+
+    final AtomicInteger acceptedSpans = new AtomicInteger();
```
adriancole: these don't need to be atomic, yet, as I've not put any threaded tests in... that'll happen

brave-core/src/main/java/com/github/kristofa/brave/HttpSpanCollector.java
```diff
@@ -0,0 +1,198 @@
...
+        drained.remove().write(oprot);
+      }
+      oprot.writeListEnd();
+    } catch (TException e) {
+      metrics.incrementDroppedSpans(spanCount);
+      return;
+    }
+
+    // Compress the thrifts, as they are usually full of strings
+    Buffer gzippedThrifts = new Buffer();
```
adriancole: ps in the future, I'd likely rework this class to use a soft size boundary. Ex. add spans until we hit a certain payload length. I've other things that are on the mind right now, so punting the idea as I'm sure it would take some number of hours above 2 to get it right.

yurishkuro:
@adriancole @bensigelman 

There is a minor discrepancy between Reporter in api-python and Recorder in api-go. For former takes the official Span object that the user operates on. The latter takes RawSpan, an internal class of the `standard` tracer. Ben's argument for RawSpan was to let the pluggable Recorders operate on a simpler and immutable struct.

bensigelman:
The argument for RawSpan goes beyond immutability in my mind, actually... I haven't been pushing this very hard since it's speculative, but at some point I imagine there will be a desire to build adapters (say, from HTrace) to opentracing. If those can be written at the instrumentation layer, great... but if they can't, maybe there's a way to translate, say, HTrace spans to RawSpans before passing them off to `standard`-compliant opentracing impls.

I.e., RawSpan is immutable, yes, but in my mind it's more about being a general-purpose and wire-friendly format for span data.

If we want to prioritize the above, we wouldn't organize the packages and dependencies the way they are now. And at some point (about when `raw` moved into the standard package) I realized that I didn't personally have enough confidence about how `opentracing` (in general) will be used to argue strongly for what I'm describing above.

adriancole:
This topic is interesting, but I don't want it to get lost in an otherwise
unrelated issue. I opened this:
https://github.com/opentracing/opentracing.github.io/pull/13

brave-core/src/main/java/com/github/kristofa/brave/HttpSpanCollector.java
kristofa: Yes agree that we can improve it later when people would end up having issues issues. One improvement I can see when using a separate tested/proven library is retrying in case connection gets dropped.

kristofa: Configuring read and connect timeouts?

adriancole: >
> Configuring read and connect timeouts?
>
heh.. had that *and pulled it*. don't mind adding back

adriancole: >
> Yes agree that we can improve it later when people would end up having
> issues issues. One improvement I can see when using a separate
> tested/proven library is retrying in case connection gets dropped.
>
makes sense.

kristofa:
The idea about adding the http implementation to `brave-core` and not in a specific `brave-http-spancollector` sub project is because this is the logical default, most approachable implementation?

kristofa:
@adriancole :+1:  I'm fine with merging this. Thanks! 
I guess this concludes what we want to add to v3.4?

adriancole:
>
> The idea about adding the http implementation to brave-core and not in a
> specific brave-http-spancollector sub project is because this is the
> logical default, most approachable implementation?
>
Not just logical default, the "no harm" addition as it has no visible deps.
For example, we could make a brave-hystrix-spancollector which is more
resilient, but adds deps.

adriancole:
>
> @adriancole <https://github.com/adriancole> [image: :+1:] I'm fine with
> merging this. Thanks!
> I guess this concludes what we want to add to v3.4?
>
I think so. I'll shine up some things and merge on my own, since you are ok
with WIP.

brave-core/src/main/java/com/github/kristofa/brave/HttpSpanCollector.java
```diff
@@ -0,0 +1,247 @@
...
+
+import static java.util.concurrent.TimeUnit.SECONDS;
+
+/**
+ * SpanCollector which submits spans to Zipkin, using its {@code POST /spans} endpoint.
+ */
+public final class HttpSpanCollector implements SpanCollector, Flushable, Closeable {
+
+  @AutoValue
+  public static abstract class Config {
```
adriancole: @kristofa this ok for config for now?

adriancole:
@kristofa ps I'm just trying to make sure this works against both zipkin server impls.

brave-core/src/main/java/com/github/kristofa/brave/HttpSpanCollector.java
kristofa: Yes, looks good!

kristofa:
@adriancole Both zipkin server impls?  Do we have 2 http server implementations?

adriancole:
@kristofa yeap zipkin-java-server is a drop-in replacement for zipkin-query-service! We use it in spring projects.

adriancole:
rolled back gzipping as servers don't support decompressing on POST at the moment.

adriancole:
@kristofa ok I think this is ready for merge. Tested against both scala and the java port.

kristofa:
:+1:

brave-core/src/main/java/com/github/kristofa/brave/HttpSpanCollector.java
```diff
@@ -0,0 +1,234 @@
...
+    @Override
+    public void run() {
+      try {
+        flushable.flush();
+      } catch (IOException ignored) {
+      }
+    }
+  }
+
+  void postSpans(Buffer thrifts) throws IOException {
```
adriancole: future extension point, but hidden for now. I'll almost certainly make an okhttp implementation

PR #121, opened 2016-01-09
Renames TraceSampler to be similar to opentracing/api-python

adriancole:
The OpenTracing python api has a zipkin implementation with an almost
identical sampler as `TraceSampler`. This renames things so that it
isn't artificially different.

See https://github.com/opentracing/api-python/blob/master/example/zipkin_like/sampler.py
See https://github.com/openzipkin/zipkin-java/pull/62


adriancole:
@marcingrzejszczak @yurishkuro fyi

adriancole:
@kristofa green! you cool with this?

kristofa:
@adriancole Yes. It is good to stick to same terminology!

PR #124, opened 2016-01-21
Move HttpSpanCollector to its own module

CodingFabian:
similar to Kafka and Scribe, Http should be in its own module, allowing
users to pick a combination which avoids dependencies.
(If I would use script I still would get okio stuff, even when not as external dependency, it would be shaded)

This basically moves https://github.com/openzipkin/brave/commit/2a85ceca5a1b50e8cdfb363a3d82026fea65b205 to its own module

### DIFF
--- a/brave-spancollector-http/src/main/java/com/github/kristofa/brave/http/HttpSpanCollector.java
+++ b/brave-spancollector-http/src/main/java/com/github/kristofa/brave/http/HttpSpanCollector.java
@@ -1,5 +1,6 @@
-package com.github.kristofa.brave;
+package com.github.kristofa.brave.http;
 
+import com.github.kristofa.brave.*;
 import com.github.kristofa.brave.internal.Nullable;
 import com.google.auto.value.AutoValue;
 import com.twitter.zipkin.gen.Span;
--- a/brave-core/pom.xml
+++ b/brave-core/pom.xml
@@ -60,12 +60,6 @@
         <version>3.1.2</version>
         <scope>test</scope>
     </dependency>
-    <dependency>
-        <groupId>com.squareup.okhttp3</groupId>
-        <artifactId>mockwebserver</artifactId>
-        <version>3.0.1</version>
-        <scope>test</scope>
-    </dependency>
   </dependencies>
 
     <build>
@@ -88,7 +82,7 @@
                             <shadeTestJar>false</shadeTestJar>
                             <minimizeJar>true</minimizeJar>
                             <dependencyReducedPomLocation>${project.build.directory}/dependency-reduced-pom.xml</dependencyReducedPomLocation>
-                            <!-- Use of zipkin-java are internal only; don't add dependencies -->
+                            <!-- Use of zipkin-java is internal only; don't add dependency -->
                             <relocations>
                                 <relocation>
                                     <pattern>zipkin</pattern>
--- a/brave-spancollector-http/LICENSE
+++ b/brave-spancollector-http/LICENSE
@@ -1,0 +1,13 @@
+Copyright 2015 <kristofa@github.com>
+
+  Licensed under the Apache License, Version 2.0 (the "License");
+  you may not use this file except in compliance with the License.
+  You may obtain a copy of the License at
+  
+      http://www.apache.org/licenses/LICENSE-2.0
+  
+  Unless required by applicable law or agreed to in writing, software
+  distributed under the License is distributed on an "AS IS" BASIS,
+  WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+  See the License for the specific language governing permissions and
+  limitations under the License.
--- a/brave-spancollector-http/README.md
+++ b/brave-spancollector-http/README.md
@@ -1,0 +1,4 @@
+# brave-spancollector-http #
+
+SpanCollector that is used to submit spans to Zipkins Http endpoint `/spans`.
+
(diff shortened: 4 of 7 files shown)


CodingFabian:
@adriancole I basically moved your commit to its own module to reduce brave-core dependencies to other stuff. I am toying with it and wonderd if I could minimize thrift usage to the generated model, when I noticed the http collector uses it as well.

adriancole:
@CodingFabian there's no hard dependency on okio, and I'd prefer to actually deal with that vs break a fairly basic impl out of the box. Are you having trouble on account of the remapped internal classes?

adriancole:
also, wrt the test dependency, is actually impacting you? or you are just taking it along for the ride?

CodingFabian:
There are two reasons: 
A) design wise, kafka and scribe are "pluggable" extra modules. So I assumed http should be as well.
B) I want to actually reduce "thrift". Before your addtion of `httpSpanCollector` thrift was only used in the generated model classes

The change in dependencies is just "undoing" what you did 2 weeks ago. No special motivation.

adriancole:
summary statement :)

I'm game with removing the okio remapped classes if they are intruding on your search path. I'd rather take that way vs introducing quite a large amount of infra around quite a small amount of code.

adriancole:
A) design wise, kafka and scribe are "pluggable" extra modules. So I assumed http should be as well.
http is pluggable, just that it has no notable dependencies, and is a reasonable default, your point about okio noted and addressable.
 
B) I want to actually reduce "thrift". Before your addtion of httpSpanCollector thrift was only used in the generated model classes
there's more to it than this.. AnnotationSubmitter and a lot of core classes refer to the generated files. I share your opinion, but don't think this change is the pivotal one to "un-thrift" things

adriancole:
this will obviate your bigger complaint and still allow a sensible default https://github.com/openzipkin/brave/pull/125

CodingFabian:
Hi Adrian, this is not the bigger of both concerns :)
I was actually aiming to work on the thrift dependency. When you embed it it blows up the jar significantly.
Okio also does this but has much less impact. Obviously making a thrift-less model is much more work.

adriancole:
So this is about the size of the jar growing, then? Let's isolate the
problems. How much did it grow? Maybe there's a way to address that

CodingFabian:
including thrift doubles the size from less than 200k to more than 400k.
If I remove the HttpSpanCollector, it goes down to 350k, because the whole "Transport" stuff goes away.

adriancole:
If extra classes are added to the jar, that should be sortable as thrift
should be an optional dependency, not a packaged one.

In other words, let's not mistake some sort of shade bug, if exists, as
intent.

HttpSpanReporter referencing classes in an optional dep only hurts if you
try to instantiate it. If we are doing that, it also seems like a bug.

If we can't sort this I promise, I will relent and even help add more
maven.. Just want a chance not to.

CodingFabian:
Thrift is not an optional dependency, as com.twitter.zipkin.gen depends on it, and those are the core model classes used by all of brave.

CodingFabian:
@adriancole I dug further, it looks like we can fix this with some maven-magic.
minimizeJar is somehow now working. it says
```
[DEBUG] Removing org.apache.thrift.async.TAsyncClient
```
But the class is in the shaded jar.
screen shot 2016-01-21 at 10 55 43

My concerns would be solved if that would work. Digging further myself.

CodingFabian:
Maybe it would be worth using zipkin-java classes instead of own copy. Will send a PR for that first.

CodingFabian:
Ok, this can be fixed by taking thrift 0.9.1 or 0.9.2, the 0.9.0 contains .java source files, so its incorrectly packaged. will send a pr including relocation.

adriancole:
Sounds like a real problem in the shade config, but alas I think I'm on a
fools errand here. Say I fixed it ..

well if you were able to undo the general thrift coupling, then this HTTP
reporter has no codec to work with. The same solution ends up at that
point, which would be to move it out.

I want to encourage, not discourage such valiance.

Plus, if in its own module, more folks might know HTTP is an option (and
discover thrift is incidental as json also works).

My advice would be to rebase your change over mine (that undoes okio).
Reason is that this will significantly reduce the complexity of the HTTP
maven module as it wouldn't need shade.

Sound reasonable?

adriancole:
Hah OK well wires crossed a bit. Keep me posted.

CodingFabian:
I fixed the packaging in #126,

the 0.9.0 version of thrift has included java sources, thus the relocation with 0.9.0 would inflate the shaded jar. I chose to upgrade to 0.9.2 because this does not have the problem, and it seems to be the version which was used to generate the model classes: 
https://github.com/openzipkin/brave/blob/master/brave-core/src/main/java/com/twitter/zipkin/gen/Annotation.java#L2

CodingFabian:
I will update this PR as soon as you land #125 - your proposal sounds reasonable.

CodingFabian:
rebased, it is now a very small changeset.

brave-spancollector-http/src/main/java/com/github/kristofa/brave/http/HttpSpanCollector.java
```diff
@@ -1,5 +1,6 @@
-package com.github.kristofa.brave;
+package com.github.kristofa.brave.http;
 
+import com.github.kristofa.brave.*;
```
adriancole: Mind not wildcarding? Pet peeve

adriancole:
One request otherwise, LGTM!

brave-spancollector-http/src/main/java/com/github/kristofa/brave/http/HttpSpanCollector.java
CodingFabian: oh sure, mine as well, was quick change to have smaller diff :)

adriancole:
Heh looks like I accidentally merged this before import change.. Never mind ;)

kristofa:

PR #126, opened 2016-01-21
Package required thrift classes for brave-core

CodingFabian:
The generated model classes require thrift, so thrift is not optional.

libthrift 0.9.0 has a special problem: it includes java source files.
The shade-relocate would extract them and inflate brave-core-shaded to 400k
Newer versions no longer have the problem, and as the model classes were
generated with 0.9.2, I thought this might be a sensible upgrade.

### DIFF
--- a/brave-core/pom.xml
+++ b/brave-core/pom.xml
@@ -78,24 +78,23 @@
         <plugins>
             <plugin>
                 <artifactId>maven-shade-plugin</artifactId>
+                <version>2.4.3</version>
                 <executions>
                     <execution>
                         <phase>package</phase>
                         <goals>
                             <goal>shade</goal>
                         </goals>
-                        <!-- retain the thrift dependency (for now) -->
                         <configuration>
-                            <artifactSet>
-                                <excludes>
-                                    <exclude>org.apache.thrift:libthrift</exclude>
-                                </excludes>
-                            </artifactSet>
                             <shadeTestJar>false</shadeTestJar>
                             <minimizeJar>true</minimizeJar>
                             <dependencyReducedPomLocation>${project.build.directory}/dependency-reduced-pom.xml</dependencyReducedPomLocation>
-                            <!-- Use of okio and zipkin-java are internal only; don't add dependencies -->
+                            <!-- Use of thrift, okio and zipkin-java are internal only; don't add dependencies -->
                             <relocations>
+                                <relocation>
+                                    <pattern>org.apache.thrift</pattern>
+                                    <shadedPattern>com.github.kristofa.brave.internal.thrift</shadedPattern>
+                                </relocation>
                                 <relocation>
                                     <pattern>okio</pattern>
                                     <shadedPattern>com.github.kristofa.brave.internal.okio</shadedPattern>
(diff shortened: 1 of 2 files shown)


adriancole:
Needs a rebase.

There are issues to note about this. Unlike other repackaged deps, we have direct consumers of thrift in brave. For example, zipkin-java is intentionally not exposed to users, because its pre 1.0 status. Thrift types have been, not just via dependencies, but also via api.

Notably Kafka and Scribe. Repackaging thrift here seems it would interfere with those consumers of it. Also, for better or worse the generated types (and indirectly their thrift dependencies) are leaked to users, too at the moment. Repackaging thrift would break them.

Eventhough I like the idea of isolating thrift, I think repackaging needlessly breaks api. It would seem a better fight to remove thrift from base classes completely.

How about upgrading the dep, but not repackaging it?

@kristofa any thoughts on this? Am I overconcerned?

CodingFabian:
I understand your concern, This is for sure not a binary compatible change, however, as far as I can tell the thrift usage is "internal only". I checked a couple of places and as a user I have no access to thrift via the api. But if it already leaks somewhere, this is of course not nice.

As a user of brave-core, I still can repackage it, so I would have no problem to solve this "just for me". But I think this change would make it better for everyone.

PS: I have a branch where I replaced this generated classes with zipkin-java, and it mostly works. There is a single problem because zipkin-java classes are immutable, adding annotations the way brave currently does it is not possible.

For Kafka and Scribe, I think they should maybe declare an explicit dependency? Not sure. It actually proves the way thrift is used is already leaked. It might be still early enough to contain it, by trying to make at least the API of brave-core thrift free and only internally use it (repackage). Other Consumers like Kafka, Scribe and Http might need to then add thrift if they need it.

The other way around could be also possible: Embrace thrift and declare it as real dependency (as it is right now). The concern I have here is that it is unclear which thrift version is compatible with which Java, and it actually feels a bit unnecessary to have it in core.

I am fine with either solution, thanks for hearing me out :)

adriancole:
@CodingFabian by leaked I mean the Span class from Brave might be used directly in other ppls code towards thrift via `write(TProtocol)`, so repackaging would make that not work.. I think you get it. thanks for the discussion.

we have a common idea on what I'd feel is more worthy of the binary break. ex might as well do it with the new classes. See if using a Span.Builder internally helps or not (wrt accumulating annotations)?

CodingFabian:
@adriancole i can push my changes so you can have a look where the problem is caused. It is not the Span side of things, thats fine. it is the ServerStateSpan

adriancole:
sgtm

CodingFabian:
Check out this commit:
https://github.com/CodingFabian/brave/commit/a91a057d8255be75314f848d787be8b30735226f

It *almost* works. you will see compile errors in AnnotationSubmitter. It modifies the current span. So what I used is the copy constructor to make a new span which has the annotation.
But I was unsuccessful in changing ths SpanState management so that I can set a new span. (basically because this autovalue thing only accepts span, and sometimes I need serverspan)

Should be manageable by somebody who is more familiar with the code.

adriancole:
PS conference things swamped me.. On that note, can you send your email to
Adrian dot f dot cole at gmail? Want to pass logistics for the meetup on
monday

PR #127, opened 2016-01-21
Make null handling in interceptors consistent.

CodingFabian:
Takes the same approach as in ServerRequestInterceptor, avoiding JDK 7
Objects.requireNonNull()


CodingFabian:
Makes null handling consistent and allows usage in JRE 6 environments.
(I manually switched animalsniffer to 16 to verify this)

PR #129, opened 2016-02-09
Simplify kafka spancollector

CodingFabian:
Instead of using a byte array output stream, we can ask thrift to directly create the bytes for us by using TSerializer


CodingFabian:
(it is worth noting that TSerializer does exactly the same code :-))

adriancole:
nice!

PR #130, opened 2016-02-09
Remove Scribe spancollector

CodingFabian:
As discussed during the workshop, we should remove the scribe collector from brave. This will make further refactorings on the model easier, because we have to maintain less deprecated code.


adriancole:
@kristofa second opinion wanted. I agree that at this point, we should direct focus towards non-archived transports.

kristofa:
I thought we discussed during the workshop that we were going to remove the need to have thrift as a dependency in `brave-core`?  Do we need to remove scribe-collector because of that? I suppose we will have to copy over the brave-core objects into the thrift once before serialization.

I agree that the future is not scribe but I think there are people still relying on it right now ([although they might use Flume instead of Scribe](https://groups.google.com/forum/#!topic/zipkin-user/v-Yw6G8W0kg)).
So that's why I would still want to keep scribe support around for now.

What do you think?

adriancole:
I'm cool with keeping it... maybe we add Deprecated("Will be killed in
version N") to let folks know its end is near?

kristofa:
Yes, I like that. We can write a note in zipkin-user group indicating the deprecation.

CodingFabian:
@kristofa I was trying to cut down stuff I need to change. I have no hard feelings about scribe, but understood you guys that nobody is using that anyway. if there are, then we should close this pr

adriancole:
I guess it is more or less a when than an if. I will send a general note to
zipkin-user which will tick the box on this for next time. Thanks and hope
this didn't waste too much of your time.

CodingFabian:
git rm ? nah, not that time consuming :)

adriancole:
;)

kristofa:
Closing.

PR #131, opened 2016-02-10
Make brave-core compile using source level 1.6

CodingFabian:
There are very few places in brave core which currently do not compile with java 6.
While I do not want to enforce this for openzipkin/brave, I needed to change these places for our fork, so maybe somebody else finds it useful. There are just a few places where diamond generic types are used. I also corrected a couple of other generic type issues.

### DIFF
--- a/brave-core/src/main/java/com/github/kristofa/brave/ClientResponseInterceptor.java
+++ b/brave-core/src/main/java/com/github/kristofa/brave/ClientResponseInterceptor.java
@@ -2,8 +2,6 @@ package com.github.kristofa.brave;
 
 import static com.github.kristofa.brave.internal.Util.checkNotNull;
 
-import java.util.Objects;
-
 /**
  * Contains logic for dealing with response from client request.
  * This means it will:
--- a/brave-core/src/test/java/com/github/kristofa/brave/ITBrave.java
+++ b/brave-core/src/test/java/com/github/kristofa/brave/ITBrave.java
@@ -6,6 +6,7 @@ import static org.junit.Assert.assertTrue;
 
 import java.util.ArrayList;
 import java.util.Collection;
+import java.util.HashSet;
 import java.util.List;
 import java.util.Random;
 import java.util.concurrent.Callable;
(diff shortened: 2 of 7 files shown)


adriancole:
Interesting.. What's the use case for 1.6? (Asking for a friend:
zipkin-java)

CodingFabian:
Agent based instrumentation using brave-core.
Java 1.6 is widely used and the agent needs to compile against 1.6 to be able to instrument 1.6 runtime code. 

While its possible to use higher source level to compile against lower bytecode level, this is not really working correctly. This will be improved in Java 9. until then it is kind of best practice to keep your source and target level down. Also using animalsniffer on that level to find usage of newer jdk classes (will also be fixed in 9)

Of course 1.6 is pretty restrictive, but in this case it is not so bad.

adriancole:
Ok. Thanks, so basically this is for agents who bytecode manipulate old
libraries. Personally, I have no issue with language level 1.6 scoped to
core, provided we put rationale in an XML comment where it is applied. Else
someone might accidentally "fix" this later.

CodingFabian:
Ok, but I would only do that for brave-core then now, as the changes to the other projects are larger.

brave-core/src/main/java/com/github/kristofa/brave/ClientResponseInterceptor.java
```diff
@@ -2,8 +2,6 @@
 
 import static com.github.kristofa.brave.internal.Util.checkNotNull;
```
adriancole: Hah

brave-core/src/test/java/com/github/kristofa/brave/ITBrave.java
```diff
@@ -107,7 +108,11 @@ public Integer call() throws Exception {
             assertTrue(localSpan.id != 0);
             assertEquals(localSpanName, localSpan.name);
 
-            assertEquals("Span ids should be different.", 3, Stream.of(serverSpan.id, clientSpan.id, localSpan.id).distinct().count());
```
adriancole: Do we need to change the compile level of tests?

CodingFabian: no, but this is kind of tricky with maven and IDEs. I thought I could sacrifice the readability of this test

adriancole: Well, I am surprised if it was only one test!

Projects who are multimodule and leave one in an old source version are
kind of gonna have this coming anyway.

adriancole:
so what's next step, change maven config to 1.6 for core (and place comment on there)?

CodingFabian:
@adriancole I did not want to force my/our compatibility upon others. But maybe there is (silent) consensus. So here we go. I have added the javac and animalsniffer config to brave core and added a comment.

adriancole:
If this affected more, I would hesitate. Having brave work for agents is valuable. Thanks for the change!

kristofa:

PR #137, opened 2016-02-18
Removes thrift dependency from brave-core and cleans model objects

adriancole:
This removes the thrift dependency from brave-core, obviated with
`SpanCodec`, whose default implementation re-uses classes from
zipkin-java.

The types in `com.twitter.zipkin.gen` are modified further:
 * Except Span, types are made immutable. Span still needs to be mutable
 * Immutable types have factory methods and expose final fields
 * Span's fields are marked private to encourage validating setters
 * Span's list fields default to empty vs null
 * Endpoint's port is now nullable, matching zipkin-java


adriancole:
@kristofa @CodingFabian lemme know if you are keen on this, and I'll clean it up (backfill tests on the codec thing)

Note the brave-core jar is now 344Kb, but doesn't have an explicit dep on thrift anymore. If we want to trim it further, I could play more games (like making a separate module with the repackaged thrift stuff in it)

anyway thoughts welcome

brave-spancollector-scribe/src/main/java/com/github/kristofa/brave/scribe/SpanProcessingThread.java
```diff
@@ -118,7 +116,7 @@ private void log(final List<LogEntry> logEntries) {
 
     private boolean log(final Client client, final List<LogEntry> logEntries) {
         try {
-            clientProvider.getClient().Log(logEntries);
```
adriancole: unrelated bug

CodingFabian:
pretty good! In my case i was storing the binary annotation for string directly as string, because if the codec wants to read them as string, you need to create a string out of the bytes first.
On the other hand I also handled the boolean case directly.

adriancole:
ps if you have an existing branch, do push it.. I was just playing around
and haven't invested much.

CodingFabian:
Neither have I, its still in the "working for us" state

adriancole:
Ok, so should I proceed on this? I think what you said about making the
span internally deal with string or boolean (well specifically address
types) sounds good.

adriancole:
actually I'm not sure I want to optimize the internal structure of the gutted classes.. I don't think they'll last too long anyway.

CodingFabian:
the commit is pretty large, so revieweing it is not as easy I I would like :) Let me diff it against our internal version today, and then I ll give feedback within the next hours. ok?

CodingFabian:
The pom.xml still has
```
                        <!-- retain the thrift dependency (for now) -->
                        <configuration>
                            <artifactSet>
                                <excludes>
                                    <exclude>org.apache.thrift:libthrift</exclude>
                                </excludes>
                            </artifactSet>
```
guess that should be removed

adriancole:
good catch

adriancole:
done

CodingFabian:
diff looks close to ours, building your branch now to see if it works

kristofa:
Looks good. I agree that spending more time to clean-up current structures is probably not worth it. It's probably best to switch to the `zipkin-java` in future.

adriancole:
@CodingFabian once you ack, I'll backfill tests on the type I just made and run towards merge. If there's an issue of course, happy to help discuss/remedy.

CodingFabian:
ack! :)

adriancole:
ok all tidied and desc updated

kristofa:
:+1:

kristofa:

PR #139, opened 2016-02-21
Add @Priority annotations to container request/response

### LINKED ISSUE #134: Add @Priority annotation to jersey 2 client and server filters ?
Would not all use-cases require that the filters trigger as early as possible for requests and as late as possible for responses ? In particular, when there are potentially time-consuming operations implemented in a request-filter (e.g. for authentication), I would like to be sure that the client send and server receive annotations are time-stamped before the other filters are executed.
https://jersey.java.net/documentation/latest/filters-and-interceptors.html (10.7) states that the lowest priority is executed first for requests and last for responses. I have achieved thus by subclassing the brave jersey filters, adding a @Priority(0) annotation, but are there good reasons this should not be done in the brave filters themselves ?

jplock:
Ensure the container request is executed first (Priority 0) and the container response is executed last (also Priority 0)

See https://jersey.java.net/documentation/latest/filters-and-interceptors.html#d0e10117 for more details

Resolves #134

### DIFF
--- a/brave-jaxrs2/pom.xml
+++ b/brave-jaxrs2/pom.xml
@@ -37,6 +37,11 @@
             <version>1</version>
         </dependency>
         <dependency>
+            <groupId>javax.annotation</groupId>
+            <artifactId>javax.annotation-api</artifactId>
+            <version>1.2</version>
+        </dependency>
+        <dependency>
             <groupId>javax.servlet</groupId>
             <artifactId>servlet-api</artifactId>
             <version>2.5</version>
--- a/brave-jaxrs2/src/main/java/com/github/kristofa/brave/jaxrs2/BraveClientRequestFilter.java
+++ b/brave-jaxrs2/src/main/java/com/github/kristofa/brave/jaxrs2/BraveClientRequestFilter.java
@@ -5,7 +5,7 @@ import com.github.kristofa.brave.http.HttpClientRequest;
 import com.github.kristofa.brave.http.HttpClientRequestAdapter;
 import com.github.kristofa.brave.http.ServiceNameProvider;
 import com.github.kristofa.brave.http.SpanNameProvider;
-
+import javax.annotation.Priority;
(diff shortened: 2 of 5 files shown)


janolaveide:
What about the client filters ? Same here surely ?

jplock:
@janolaveide I wasn't sure if you could set them on the client filters.  I'll add them anyway.

brave-jaxrs2/pom.xml
```diff
@@ -37,6 +37,11 @@
             <version>1</version>
         </dependency>
         <dependency>
+            <groupId>javax.annotation</groupId>
+            <artifactId>javax.annotation-api</artifactId>
+            <version>1.2</version>
```
adriancole: just curious.. this matches servlet 2.5, right? If so, maybe make a comment that it does (so that someone changing below doesn't invalidate this)

adriancole: or if it has nothing to do with any of the dependencies here, ignore my comment :P

adriancole:
is there any way to test this? (ex show it prevents a bug)

brave-jaxrs2/pom.xml
jplock: I don't think it's related.  I had to add this dependency to get access to the `@Priority` annotation as it wasn't available in any of the other dependencies.

jplock:
I'm trying to think how we could test this.  We'd need to create a set of filters that would inspect the zipkin information show how to ensure its being included in the request I'd imagine.

janolaveide:
Not sure what you want to test, testing that it actually executes first ? I haven't looked at the actual test cases you have in brave, but in my manual testing I add a time-consuming filter with priority > 1 and verifies that brave timing reflects this.

adriancole:
sounds sufficiently tricky to automatically test at the moment. Thanks for manual testing.

kristofa:

PR #142, opened 2016-03-02
Adds HttpSpanCollector.Config.compressionEnabled

adriancole:
HttpSpanCollector.Config.compressionEnabled will gzip spans before
uploading them to a zipkin server. This defaults to false, as server
support requires zipkin-scala 1.34+ or zipkin-java 0.6+.

### DIFF
--- a/brave-spancollector-http/src/test/java/com/github/kristofa/brave/http/HttpSpanCollectorTest.java
+++ b/brave-spancollector-http/src/test/java/com/github/kristofa/brave/http/HttpSpanCollectorTest.java
@@ -1,7 +1,9 @@
 package com.github.kristofa.brave.http;
 
 import com.github.kristofa.brave.SpanCollectorMetricsHandler;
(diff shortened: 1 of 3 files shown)


brave-spancollector-http/src/test/java/com/github/kristofa/brave/http/HttpSpanCollectorTest.java
```diff
@@ -41,7 +43,9 @@ public void dropsWhenQueueIsFull() throws Exception {
     for (int i = 0; i < 1001; i++)
       collector.collect(span(1L, "foo"));
 
-    assertThat(metrics.acceptedSpans.get()).isEqualTo(1001);
```
adriancole: switched this as we already have a test for the accepted counter

kristofa:
I guess it is good to have the option to turn gzip on. Thanks :)

kristofa:

PR #144, opened 2016-03-02
Clean-up README.md

kristofa:
Part of the README.md content was not up to date anymore.
There was information about using Flume as preferred span shipping implementation which is
not the case anymore with HTTP and Kafka support.

Also the overview on how Zipkin/Dapper works doesn't need to be in the brave README.

Also update link to point to OpenZipkin instead of the original Twitter Zipkin link.

### DIFF
--- a/README.md
+++ b/README.md
@@ -3,9 +3,7 @@
 [![Build Status](https://travis-ci.org/openzipkin/brave.svg?branch=master)](https://travis-ci.org/openzipkin/brave)
 [![Maven Central](https://img.shields.io/maven-central/v/com.github.kristofa/brave.svg)](https://maven-badges.herokuapp.com/maven-central/com.github.kristofa/brave)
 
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -3,9 +3,7 @@
 [![Build Status](https://travis-ci.org/openzipkin/brave.svg?branch=master)](https://travis-ci.org/openzipkin/brave)
 [![Maven Central](https://img.shields.io/maven-central/v/com.github.kristofa/brave.svg)](https://maven-badges.herokuapp.com/maven-central/com.github.kristofa/brave)
 
-**Brave moved to OpenZipkin.** [Read more about it here](http://kdevlog.blogspot.de/2015/07/brave-moved-to-openzipkin-org.html).
-
-Java distributed tracing implementation compatible with [Zipkin](https://github.com/twitter/zipkin/).
+Java Distributed Tracing implementation compatible with [Zipkin](https://github.com/openzipkin/zipkin).
```
adriancole: Maybe link to just zipkin.io? Cc @abesto

adriancole:
LG thanks!

PR #147, opened 2016-03-13
initial implementation of grpc support

ewhauser:
Adds basic support for tracing in gRPC applications.


brave-grpc/README.md
```diff
@@ -0,0 +1,20 @@
+# brave-jersey #
+
+Contributed by [Eric Hauser](https://github.com/ewhauser).
```
adriancole: I think this is copy/paste, but usually we don't add this (even if jersey did). we still love you, though!

brave-grpc/README.md
```diff
@@ -0,0 +1,20 @@
+# brave-jersey #
```
adriancole: nit: c&p

brave-grpc/README.md
```diff
@@ -0,0 +1,20 @@
...
+        .addService(ServerInterceptors.intercept(
+            GreeterGrpc.bindService(new GreeterImpl()), new BraveGrpcServerInterceptor(brave)))
+        .build()
+        .start();
+        
+    ManagedChannel channel = ManagedChannelBuilder.forAddress("localhost", serverPort)
+        .intercept(new BraveGrpcClientInterceptor("helloCallingService", brave))
+        .usePlaintext(true)
+        .build();
+```
```
adriancole: this might need a development section, showing how to make intellij compile the protos

brave-grpc/pom.xml
```diff
@@ -0,0 +1,96 @@
...
+        <protobuf.version>3.0.0-beta-2</protobuf.version>
+    </properties>
+
+    <dependencies>
+        <dependency>
+            <groupId>com.github.kristofa</groupId>
+            <artifactId>brave-http</artifactId>
+            <version>3.5.1-SNAPSHOT</version>
+        </dependency>
+        <dependency>
```
adriancole: I think you can remove inject and annotation jars as they aren't needed.

brave-grpc/pom.xml
```diff
@@ -0,0 +1,96 @@
...
+    <properties>
+        <project.build.sourceEncoding>UTF-8</project.build.sourceEncoding>
+        <grpc.version>0.13.2</grpc.version>
+        <protobuf.version>3.0.0-beta-2</protobuf.version>
+    </properties>
+
+    <dependencies>
+        <dependency>
+            <groupId>com.github.kristofa</groupId>
+            <artifactId>brave-http</artifactId>
```
adriancole: this dep is only used to read constants (BraveHttpHeaders). I'd maybe make a comment like:

```
<!-- brave-http contains propagation header keys (BraveHttpHeaders) -->
```

Luckily, brave-http has no additional deps, so is harmless (beyond adding some number of KiB)

brave-grpc/src/main/java/com/github/kristofa/brave/grpc/BraveGrpcClientInterceptor.java
```diff
@@ -0,0 +1,128 @@
...
+                        clientResponseInterceptor.handle(new GrpcClientResponseAdapter(status));
+                        super.onClose(status, trailers);
+                   }
+                }, headers);
+            }
+        };
+
+    }
+
+    private class GrpcClientRequestAdapter<ReqT, RespT> implements ClientRequestAdapter {
```
adriancole: static final class is better, I think. removing private prevents the compiler from having to [generate synthetic methods](http://www.javaworld.com/article/2073578/java-s-synthetic-methods.html). static because AFAICT, there's no shared state with the outer type.

brave-grpc/src/main/java/com/github/kristofa/brave/grpc/BraveGrpcClientInterceptor.java
```diff
@@ -0,0 +1,128 @@
...
+import io.grpc.ForwardingClientCallListener.SimpleForwardingClientCallListener;
+import io.grpc.Metadata;
+import io.grpc.MethodDescriptor;
+import io.grpc.Status;
+import io.grpc.Status.Code;
+
+import java.util.Collection;
+import java.util.Collections;
+
+public class BraveGrpcClientInterceptor implements ClientInterceptor {
```
adriancole: nit: final all the public types (Effective Java Items 16-17)

brave-grpc/src/main/java/com/github/kristofa/brave/grpc/BraveGrpcClientInterceptor.java
```diff
@@ -0,0 +1,128 @@
...
+
+    private String clientService;
```
adriancole: final all the fields, so the class becomes reference immutable!

brave-grpc/src/main/java/com/github/kristofa/brave/grpc/BraveGrpcClientInterceptor.java
```diff
@@ -0,0 +1,128 @@
...
+    private ClientRequestInterceptor clientRequestInterceptor;
+    private ClientResponseInterceptor clientResponseInterceptor;
```
adriancole: Since these fields are read from an anonymous type, you can also choose to remove the private qualifier (regarding the synthetic method concern listed below). Considering this is somewhat mysterious, ignoring this comment is also fine :)

brave-grpc/src/main/java/com/github/kristofa/brave/grpc/BraveGrpcClientInterceptor.java
```diff
@@ -0,0 +1,128 @@
...
+        return new SimpleForwardingClientCall<ReqT, RespT>(next.newCall(method, callOptions)) {
+
+            @Override
+            public void start(Listener<RespT> responseListener, Metadata headers) {
+                clientRequestInterceptor.handle(new GrpcClientRequestAdapter<>(method, headers));
+                final Span currentClientSpan = clientSpanThreadBinder.getCurrentClientSpan();
+                super.start(new SimpleForwardingClientCallListener<RespT>(responseListener) {
+                    @Override
+                    public void onClose(Status status, Metadata trailers) {
+                        clientSpanThreadBinder.setCurrentSpan(currentClientSpan);
```
adriancole: I might move the clientSpanThreadBinder, clientResponseInterceptor fields to a new type ClientCallListenerFactory, and make `ClientCallListenerFactory.create(Span currentClientSpan)`

brave-grpc/src/main/java/com/github/kristofa/brave/grpc/BraveGrpcClientInterceptor.java
```diff
@@ -0,0 +1,128 @@
...
+                        clientResponseInterceptor.handle(new GrpcClientResponseAdapter(status));
+                        super.onClose(status, trailers);
```
adriancole: I think we should revert the current span in a finally block (if nothing else to clear its reference to make it available for gc), @kristofa is that right?

brave-grpc/src/test/java/com/github/kristofa/brave/grpc/BraveGrpcInterceptorsTest.java
```diff
@@ -0,0 +1,126 @@
...
+        HelloReply reply = helloReplyListenableFuture.get();
+        assertThat(reply.getMessage(), is("Hello brave"));
+        validateSpans();
+    }
+
+    private void validateSpans() throws Exception {
+        List<Span> spans = SpanCollectorForTesting.getInstance().getCollectedSpans();
+        assertThat(spans.size(), is(equalTo(2)));
+        assertThat(spans.get(0).getParent_id(), is(equalTo(spans.get(1).getParent_id())));
+        validateSpan(spans.get(0), Arrays.asList("ss", "sr"));
```
adriancole: maybe make a comment explaining that presence of two spans means that propagation between client and server works (ex the headers are propagated)

brave-grpc/src/main/java/com/github/kristofa/brave/grpc/BraveGrpcServerInterceptor.java
```diff
@@ -0,0 +1,122 @@
...
+        public GrpcServerRequestAdapter(MethodDescriptor<ReqT, RespT> method, Metadata requestHeaders) {
+            this.method = checkNotNull(method);
+            this.requestHeaders = checkNotNull(requestHeaders);
+        }
+
+        @Override
+        public TraceData getTraceData() {
+            final String sampled = requestHeaders.get(GrpcHeaders.Sampled);
+            if (sampled != null) {
+                if (sampled.equals("0") || sampled.toLowerCase().equals("false")) {
```
adriancole: sampled = false should probably be tested.

brave-grpc/src/main/java/com/github/kristofa/brave/grpc/BraveGrpcServerInterceptor.java
```diff
@@ -0,0 +1,122 @@
...
+                    return TraceData.builder().sample(false).build();
+                } else {
+                    final String parentSpanId = requestHeaders.get(GrpcHeaders.ParentSpanId);
```
adriancole: span with a parent should probably be tested

adriancole:
@ewhauser great job! I made a fury of mostly stylistic comments, and a suggestion for an additional couple tests. If you need hints on doing those, let me know (I understand those tests aren't necessarily in other modules)

ewhauser:
@adriancole - Changes made with the exception of `ClientCallListenerFactory` suggestion.  Didn't quite follow what you wanted to do there, but happy to adjust.

brave-grpc/src/test/java/com/github/kristofa/brave/grpc/BraveGrpcInterceptorsTest.java
```diff
@@ -0,0 +1,173 @@
...
+        assertThat(reply.getMessage(), is("Hello brave"));
+        List<Span> spans = SpanCollectorForTesting.getInstance().getCollectedSpans();
+        assertThat(spans.size(), is(equalTo(0)));
+    }
+
+    /**
+     * Validating that two spans were generated indicates that a span was generated by both the
+     * server and the client.
+     *
+     * @throws Exception
```
adriancole: supernit: scrub throws

brave-grpc/src/test/java/com/github/kristofa/brave/grpc/BraveGrpcInterceptorsTest.java
```diff
@@ -0,0 +1,173 @@
...
+        Optional<Span> maybeSpan = spans.stream()
+            .filter(s -> s.getAnnotations().stream().anyMatch(a -> "ss".equals(a.value)))
+            .findFirst();
+        assertTrue("Could not find expected server span", maybeSpan.isPresent());
+        Span span = maybeSpan.get();
+        assertThat(span.getTrace_id(), is(equalTo(spanId.getTraceId())));
+        assertThat(span.getParent_id(), is(equalTo(spanId.getSpanId())));
+    }
+
+    @Test
```
adriancole: I'd add a javadoc clarifying what this means.

If sampling status isn't propagated, the server would start a trace even if a client's request wasn't sampled. This tests that sampling status propagates and is honored by the server: both client and server spans are not collected.

brave-grpc/src/test/java/com/github/kristofa/brave/grpc/BraveGrpcInterceptorsTest.java
```diff
@@ -0,0 +1,173 @@
...
+        assertThat(reply.getMessage(), is("Hello brave"));
+        validateSpans();
+    }
+
+    @Test
+    public void usesExistingTraceId() throws Exception {
+        LocalTracer localTracer = brave.localTracer();
+        SpanId spanId = localTracer.startNewSpan("localSpan", "myop");
+        GreeterBlockingStub stub = GreeterGrpc.newBlockingStub(channel);
+        HelloReply reply = stub.sayHello(HELLO_REQUEST);
```
adriancole: add a comment on this line clarifying that this operation should be logged in a child span

adriancole: oh wait.. we are testing parent id, here. I think you should create a SpanId that has parent id set.

adriancole:
thx. other notes weren't terribly important.

brave-grpc/src/test/java/com/github/kristofa/brave/grpc/BraveGrpcInterceptorsTest.java
ewhauser: So what I was doing was starting the trace by creating a local span and then verifying that the local span's id propagated down to the server as the parent id.  Is that not correct?

adriancole: ahh. ok, so I guess just back to my original note: add a comment above `stub.sayHello(HELLO_REQUEST)` saying that this operation should be logged in a child span.

brave-grpc/src/main/java/com/github/kristofa/brave/grpc/BraveGrpcServerInterceptor.java
```diff
@@ -0,0 +1,122 @@
...
+import io.grpc.ServerInterceptor;
+import io.grpc.Status;
+import io.grpc.Status.Code;
+
+import java.util.Collection;
+import java.util.Collections;
+
+public class BraveGrpcServerInterceptor implements ServerInterceptor {
+
+    private ServerRequestInterceptor serverRequestInterceptor;
```
adriancole: nit: final the fields (and also public types in src/main)

brave-grpc/pom.xml
```diff
@@ -0,0 +1,88 @@
...
+                <version>0.5.0</version>
+                <configuration>
+                    <!--
+                      The version of protoc must match protobuf-java. If you don't depend on
+                      protobuf-java directly, you will be transitively depending on the
+                      protobuf-java version that grpc depends on.
+                    -->
+                    <protocArtifact>com.google.protobuf:protoc:3.0.0-beta-2:exe:${os.detected.classifier}</protocArtifact>
+                    <pluginId>grpc-java</pluginId>
+                    <pluginArtifact>io.grpc:protoc-gen-grpc-java:0.13.2:exe:${os.detected.classifier}</pluginArtifact>
```
adriancole: might want to use the grpc version variable here

brave-grpc/pom.xml
```diff
@@ -0,0 +1,88 @@
...
+                <groupId>org.xolstice.maven.plugins</groupId>
+                <artifactId>protobuf-maven-plugin</artifactId>
+                <version>0.5.0</version>
+                <configuration>
+                    <!--
+                      The version of protoc must match protobuf-java. If you don't depend on
+                      protobuf-java directly, you will be transitively depending on the
+                      protobuf-java version that grpc depends on.
+                    -->
+                    <protocArtifact>com.google.protobuf:protoc:3.0.0-beta-2:exe:${os.detected.classifier}</protocArtifact>
```
adriancole: move this comment and extract a version property to the properties section, so that the protoc version and grpc version are close to eachother.

adriancole: oops variables are still not used here and below

adriancole:
will merge tomorrow!

brave-grpc/src/main/java/com/github/kristofa/brave/grpc/BraveGrpcClientInterceptor.java
```diff
@@ -0,0 +1,135 @@
...
+                            clientSpanThreadBinder.setCurrentSpan(null);
+                        }
+                   }
+                }, headers);
+            }
+        };
+
+    }
+
+    private static final class GrpcClientRequestAdapter<ReqT, RespT> implements ClientRequestAdapter {
```
kristofa: Was there a reason why you did not use the Http abstractions like `HttpClientRequestAdapter` instead of `ClientRequestAdapter`?  The advantage of using it is that the conversion span <-> http is done for you and same across implementations. You will probably also end up with less code.

brave-grpc/src/main/java/com/github/kristofa/brave/grpc/BraveGrpcServerInterceptor.java
```diff
@@ -0,0 +1,122 @@
...
+
+            @Override
+            public void close(Status status, Metadata trailers) {
+                serverResponseInterceptor.handle(new GrpcServerResponseAdapter(status));
+                super.close(status, trailers);
+            }
+        }, requestHeaders);
+    }
+
+    private class GrpcServerRequestAdapter<ReqT, RespT> implements ServerRequestAdapter {
```
kristofa: Same here, why not use `HttpServerRequestAdapter` instead of `ServerRequestAdapter`? 
`public TraceData getTraceData()` implementation is a copy of [HttpServerRequestAdapter](https://github.com/openzipkin/brave/blob/master/brave-http/src/main/java/com/github/kristofa/brave/http/HttpServerRequestAdapter.java).

brave-grpc/src/main/java/com/github/kristofa/brave/grpc/BraveGrpcClientInterceptor.java
ewhauser: @kristofa: I thought that they were tied to `HttpServletRequest` but upon inspection it appears that is not the case.  I'll take a look at see if I can adapt.

adriancole:
One thing to be careful about are the tags (binary annotations). Ex I've
noticed these prefixed by product ex "grpc.status"

Such keys could be extracted into a public type, like GrpcKeys. For
example, I have a todo to extract the http ones.

ewhauser:
@kristofa - I refactored to use the http adapters as a separate commit.  Once you give the thumbs up, I'll rebase it.

adriancole:
@kristofa @ewhauser I don't think grpc is a http tracer. It uses http keys for propagation (which is an arbitrary choice), The spans themselves are not the same as http. ex grpc.status_code isn't http status. They are semantically different things, I think 

http://www.grpc.io/grpc-java/javadoc/io/grpc/Status.Code.html

ewhauser:
I think @adriancole's right.  I don't think we want shoehorn the gRPC model into the HTTP one just to save some code.  While gRPC does use HTTP/2 for transport, it's really an implementation of gRPC on top of HTTP:

http://www.grpc.io/docs/guides/wire.html

adriancole:
Tricky topic! Can you hook into the HTTP/2 layer in a standard way in gRPC?
Maybe we can log both layers..

adriancole:
@kristofa @ewhauser I'm pretty much still in favor of commit 1 without any major changes. There's some polishing we'd do, but think this is good start.

kristofa:
@adriancole @ewhauser I'm fine to go with the first commit. If the abstraction doesn't fit we should not use it. :+1:

adriancole:
closed via fd3cdb921964a0d7164c29141fe6aac8099ae4cf with some very minor changes

kristofa:

zinuga:
@adriancole @kristofa i would like to have a "how to use" doc for grpc with zipkin and post it for users. Can you help on that?

adriancole:
The README is the best place to start
https://github.com/openzipkin/brave/tree/master/brave-grpc

PR #149, opened 2016-03-22
Makes Span.toString return json

### LINKED ISSUE #148: default LoggingSpanCollector log is just the pointer id
version 3.5.0
```
INFO  [2016-03-21 15:52:28,854] com.github.kristofa.brave.LoggingSpanCollector: com.twitter.zipkin.gen.Span@3f7c2ac3
```
I think that some kind of toString would be much more help.

adriancole:
Particularly when using `LoggingSpanCollector`, a span's string form
is better value-based (vs identity), and on a single-line.

This reuses the json codec to solve the problem without adding
maintenance.

Fixes #148


adriancole:
cc @a86c6f7964

CodingFabian:
LGTM

kristofa:

PR #151, opened 2016-04-05
Update ClientTracer.java

rafabene:
This allows to create a `ClientRequestInterceptor` as the following code: 
`return new ClientRequestInterceptor(ClientTracer.builder().build());`

It will also be similar to ServerTracer.builder().build()


adriancole:
sounds reasonable. it makes the builder mimic ServerTracer.Builder's access modifiers

kristofa:

PR #156, opened 2016-04-13
Consolidates to a single endpoint per Brave

### LINKED ISSUE #153: ClientTracer.setCurrentClientServiceName should not be used by ClientRequestInterceptor
Brave should use the default endpoint when reporting client side spans (the same way server behaves)
Currently `ClientRequestAdapter.getClientServiceName` method is used to extract the endpoint service name, however the documentation is misleading by saying:

> The service name is expected to be the same as the service we are calling.

adriancole:
This is an api breaking change that aligns Brave with the current
interpretation of Annotation.endpoint as representing the local service
being traced. This dramatically simplifies the codebase and allows us to
enforce endpoint as a reference that cannot be null.

Fixes #153


chemicL:
That's great, thanks @adriancole

kristofa:

PR #159, opened 2016-04-19
Makes it more explicit that gRPC Metadata keys are lowercase

adriancole:
gRPC Metadata keys are implicitly downcased (due in part to being
implemented as HTTP/2 headers). This shows this in a test.


adriancole:
@basvanbeek this should hopefully help make more visible go-kit compatibility

PR #160, opened 2016-04-21
Add support for defining Zipkin's Kafka topic

NegatioN:
There are reasons for using another kafka-topic than the default.
Many Zipkin installations could potentially share Kafka-clusters.
This would simplify segregation in such environments.


adriancole:
looks lovely

kristofa:

PR #165, opened 2016-04-28
Server request/response adapters should add binary annotations

NegatioN:
Problem:
When hitting http://example.com/a/b/c with a browser,
the annotations http.uri and http.responsecode are never recorded.
This means we will not be able to figure out where on the
web site a trace was triggered.

Solution:
Add annotations to server request/response adapters

### DIFF
--- a/brave-http/src/main/java/com/github/kristofa/brave/http/HttpServerRequestAdapter.java
+++ b/brave-http/src/main/java/com/github/kristofa/brave/http/HttpServerRequestAdapter.java
@@ -1,11 +1,9 @@
 package com.github.kristofa.brave.http;
 
 import com.github.kristofa.brave.*;
-import zipkin.TraceKeys;
 
 import java.util.Arrays;
 import java.util.Collection;
-import java.util.Collections;
 
 
 public class HttpServerRequestAdapter implements ServerRequestAdapter {
--- a/brave-http/src/main/java/com/github/kristofa/brave/http/HttpServerRequestAdapter.java
+++ b/brave-http/src/main/java/com/github/kristofa/brave/http/HttpServerRequestAdapter.java
@@ -2,6 +2,7 @@ package com.github.kristofa.brave.http;
 
 import com.github.kristofa.brave.*;
 
+import java.util.Arrays;
 import java.util.Collection;
 import java.util.Collections;
 
@@ -43,7 +44,9 @@ public class HttpServerRequestAdapter implements ServerRequestAdapter {
 
     @Override
     public Collection<KeyValueAnnotation> requestAnnotations() {
-        return Collections.emptyList();
+        KeyValueAnnotation uriAnnotation = KeyValueAnnotation.create(
+                "http.uri", serverRequest.getUri().toString());
+        return Arrays.asList(uriAnnotation);
     }
 
     private SpanId getSpanId(String traceId, String spanId, String parentSpanId) {
--- a/brave-http/src/main/java/com/github/kristofa/brave/http/HttpServerRequestAdapter.java
+++ b/brave-http/src/main/java/com/github/kristofa/brave/http/HttpServerRequestAdapter.java
@@ -1,6 +1,7 @@
 package com.github.kristofa.brave.http;
 
 import com.github.kristofa.brave.*;
+import zipkin.TraceKeys;
 
 import java.util.Arrays;
 import java.util.Collection;
@@ -45,7 +46,7 @@ public class HttpServerRequestAdapter implements ServerRequestAdapter {
     @Override
     public Collection<KeyValueAnnotation> requestAnnotations() {
         KeyValueAnnotation uriAnnotation = KeyValueAnnotation.create(
-                "http.uri", serverRequest.getUri().toString());
+                TraceKeys.HTTP_URL, serverRequest.getUri().toString());
         return Arrays.asList(uriAnnotation);
     }
 
--- a/brave-http/src/main/java/com/github/kristofa/brave/http/HttpServerResponseAdapter.java
+++ b/brave-http/src/main/java/com/github/kristofa/brave/http/HttpServerResponseAdapter.java
@@ -2,6 +2,7 @@ package com.github.kristofa.brave.http;
 
 import com.github.kristofa.brave.KeyValueAnnotation;
 import com.github.kristofa.brave.ServerResponseAdapter;
+import zipkin.TraceKeys;
 
 import java.util.Arrays;
 import java.util.Collection;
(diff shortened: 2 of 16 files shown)


eirslett:
+1 from me! (Because I worked with negatioN on this change :wink: )

brave-http/src/main/java/com/github/kristofa/brave/http/HttpServerResponseAdapter.java
```diff
@@ -18,12 +18,8 @@ public HttpServerResponseAdapter(HttpResponse response)
...
     @Override
     public Collection<KeyValueAnnotation> responseAnnotations() {
-        int httpStatus = response.getHttpStatusCode();
-
-        if ((httpStatus < 200) || (httpStatus > 299)) {
-            KeyValueAnnotation statusAnnotation = KeyValueAnnotation.create("http.responsecode", String.valueOf(httpStatus));
-            return Arrays.asList(statusAnnotation);
-        }
-        return Collections.EMPTY_LIST;
+        KeyValueAnnotation statusAnnotation = KeyValueAnnotation.create(
```
adriancole: ps this is a logic change. It saves size to not log 2xx range responses. https://github.com/openzipkin/zipkin-java/blob/master/zipkin/src/main/java/zipkin/TraceKeys.java#L82

adriancole:
It would be nice to use the standard keys and rationale on this.. I once had a branch that did this, but I got distracted. https://github.com/openzipkin/zipkin-java/blob/master/zipkin/src/main/java/zipkin/TraceKeys.java

brave-http/src/main/java/com/github/kristofa/brave/http/HttpServerRequestAdapter.java
```diff
@@ -43,7 +44,9 @@ public String getSpanName() {
 
     @Override
     public Collection<KeyValueAnnotation> requestAnnotations() {
-        return Collections.emptyList();
+        KeyValueAnnotation uriAnnotation = KeyValueAnnotation.create(
+                "http.uri", serverRequest.getUri().toString());
```
adriancole: "http.url" is the current name, when query parameters are present (which I think this is..)

brave-http/src/main/java/com/github/kristofa/brave/http/HttpServerResponseAdapter.java
```diff
@@ -18,12 +18,8 @@ public HttpServerResponseAdapter(HttpResponse response)
...
+                "http.responsecode", String.valueOf(response.getHttpStatusCode()));
```
adriancole: "http.status_code" if you don't mind.. that way it is the same value java vs otherwise

brave-http/src/main/java/com/github/kristofa/brave/http/HttpServerRequestAdapter.java
eirslett: Ah, I see. We went with Brave's standard, which is http.url:

https://github.com/openzipkin/brave/blob/master/brave-http/src/main/java/com/github/kristofa/brave/http/HttpClientRequestAdapter.java#L47

So we should probably fix that in Brave everywhere?

brave-http/src/main/java/com/github/kristofa/brave/http/HttpServerResponseAdapter.java
eirslett: Same argument as above - we reused Brave's convention (https://github.com/openzipkin/brave/blob/master/brave-http/src/main/java/com/github/kristofa/brave/http/HttpClientResponseAdapter.java#L25) but it should probably be fixed everywhere in Brave.

brave-http/src/main/java/com/github/kristofa/brave/http/HttpServerResponseAdapter.java
eirslett: I don't think it will be a significant change in size to log status code... size is a concern, but I'd rather have fewer traces with more data in each, than more traces with less data in each, does that make sense?

One possible long-term optimization could be to have "debug" annotations that are only logged when you trigger a trace in debug mode (e.g. with the browser extension)

adriancole: hmm sounds like cherry-picking concerns. If consistency is the goal, why is
it ok to change this policy and not the other?

adriancole: Here's the challenge..

Say someone wants to change a documented policy, like to no longer
conditionally log 200 response, something that we took time to
document, put into thrifts, etc. Why not add something nice during the
process.. like using consistent keys?

Then, when we talk about release notes, we can say what the benefit
was that came along with the break.

It is also nicer to folks who maintain docs and constants (like me).
I'd rather to have some help, vs just cleanup work changing the
thrifts etc to say we don't advocate dropping response code on 2xx
anymore. I feel like even doing a global replace for using the
constant keys is a reasonable ask when there's work being added to my
plate. Do you disagree?

eirslett: I think you misunderstood me, maybe... I agree that we should move to using the consistent keys. I was just explaining why we (inadvertently) did it the "wrong way" in the first place ;-)

I think this PR should be updated, so that it also moves to using the consistent keys.

NegatioN: I agree with all the comments here so far as well. Consistent keys > hardcoded strings :)
I didn't know about the TraceKeys, but I'll get on updating the PR as soon as possible.

NegatioN:
@adriancole I think the PR should satisfy the requirements now. :)
Added TraceKeys to the Brave-core project, and re-referenced http.uri and http.responsecode to their respective consistent keys.

brave-http/src/test/java/com/github/kristofa/brave/http/HttpServerRequestAdapterTest.java
```diff
@@ -129,4 +134,15 @@ public void getTraceDataSampledTrueWithParentId() {
         assertEquals(Long.valueOf(IdConversion.convertToLong(PARENT_SPAN_ID)), spanId.getParentSpanId());
     }
 
+    @Test
+    public void uriAnnotation() throws Exception {
+        when(serverRequest.getUri()).thenReturn(new URI("http://youruri.com"));
```
adriancole: small note: might want to add a query parameter to see that it is retained

adriancole:
great. one test update, then 🚢

PR #167, opened 2016-05-04
Adds HttpSpanCollector.tracingEnabled to prevent trace amplification

adriancole:
Now that Cassandra is instrumentable, we've the possibility of creating
a loop. The scenario is that we are tracing the http collector endpoint,
and the span propagates there. This is helpful as it can help diagnose
performance issues in Zipkin's collector.

The problem happens when cassandra's tracer reports back to the same
http endpoint. This creates a loop, unless we have a flag to disable
sampling.


adriancole:
cc @eirslett @michaelsembwever so this is possible unless we make it possible for the the cassandra tracer to write to itself. This is only an issue on the cluster that stores zipkin data.

adriancole:
closed from lack of interest

Issue #166, opened 2016-05-03
Support multiple child spans from the same thread, aka variable trace shapes


adriancole:
Currently, Brave's supports fixed trace shapes defined below. The impact is generally found when people try to create more than one child span, and notice it overwrites the previous child.

https://github.com/openzipkin/brave/blob/3.6.0/brave-core/src/main/java/com/github/kristofa/brave/ServerSpanState.java
https://github.com/openzipkin/brave/blob/3.6.0/brave-core/src/main/java/com/github/kristofa/brave/LocalSpanState.java
https://github.com/openzipkin/brave/blob/3.6.0/brave-core/src/main/java/com/github/kristofa/brave/ClientSpanState.java

Support for variable span shapes has come up, usually in questions. Ex @galderz trying to trace infinispan, @michaelsembwever trying to [make an open tracing bridge](https://github.com/opentracing/opentracing-java/pull/25), and several gitter questions on https://gitter.im/openzipkin/zipkin.

Several options exist, and not all are either/or.
- throw an IllegalStateException as opposed to overwriting the old span. This would make understanding the issue easier.
- Consider using a tree or otherwise to support variable depth. Note this usually comes at a cost of knowing when to close things.
- Consider a different underlying model, such as a Brave copy of [Continuation local storage](https://github.com/DistributedTracing/continuation-local-storage-jvm/pull/1), which has an advantage of progressing that as well.

adriancole:
Note it may be possible to write a custom state manager with the existing interfaces (eventhough I've not tried it).
https://github.com/openzipkin/brave/blob/master/brave-core/src/test/java/com/github/kristofa/brave/example/TestServerClientAndLocalSpanStateCompilation.java

It may also be possible to work around this issue by creating scoped brave instances. Ex. if you are going to make multiple child spans, create a new brave in a try-finally, and for each of them set a current server span before running your task.

adriancole:
ps not sure what the capabilities of this are, but maybe @hyleung has addresses this?

https://github.com/hyleung/ratpack-zipkin/blob/master/ratpack-zipkin/src/main/java/ratpack/zipkin/internal/RatpackServerClientLocalSpanState.java

hyleung:
@adriancole that class is just a `ServiceClientAndLocalSpanState` implementation backed by Ratpack's registry (the default ThreadLocal implementation won't work for Ratpack). I don't believe it would address this issue - at least if i'm understanding it correctly.

adriancole:
[...]

PR #168, opened 2016-05-04
Retrofits SpanId to be effectively portable with Finagle's TraceId

adriancole:
Finagle's TraceId is a propagation value and used in many places, such
as indexing data for collection. It also has a concise fixed-length
serialized form, and similarly concise toString.

The serialized form could be helpful when propagating across systems,
most notably Finagle, but also cassandra, whose baggage includes opaque
values.

See https://github.com/thelastpickle/cassandra-zipkin-tracing/tree/mck/brave-3.5
See #166

Here are the benchmarks. I noticed toString was special-cased in finagle, so I special-cased it here, too.

```
Benchmark                            Mode  Cnt   Score   Error   Units
SpanIdBenchmarks.bytes_brave        thrpt   30  55.749 ± 1.424  ops/us
SpanIdBenchmarks.bytes_finagle      thrpt   30  66.805 ± 4.009  ops/us
SpanIdBenchmarks.fromBytes_brave    thrpt   30  53.761 ± 6.293  ops/us
SpanIdBenchmarks.fromBytes_finagle  thrpt   30  37.689 ± 0.470  ops/us
SpanIdBenchmarks.toString_brave     thrpt   30  17.736 ± 0.143  ops/us
SpanIdBenchmarks.toString_finagle   thrpt   30   2.732 ± 0.031  ops/us
```

### DIFF
--- a/brave-core/src/main/java/com/github/kristofa/brave/SpanId.java
+++ b/brave-core/src/main/java/com/github/kristofa/brave/SpanId.java
@@ -1,55 +1,323 @@
 package com.github.kristofa.brave;
 
 import com.github.kristofa.brave.internal.Nullable;
-import com.google.auto.value.AutoValue;
+import com.twitter.zipkin.gen.Span;
+import java.nio.ByteBuffer;
+
+import static com.github.kristofa.brave.internal.Util.checkNotNull;
 
 /**
- * Identifies a Span.
- * 
- * @author kristof
+ * Contains trace data that's propagated in-band across requests, sometimes known as Baggage.
+ *
+ * <p>This implementation is biased towards a fixed-length binary serialization format that doesn't
+ * have a way to represent an absent parent (root span). In this serialized form, a root span is
+ * when all three ids are the same. Alternatively, you can use {@link #nullableParentId}.
+ *
+ * <p>Particularly, this includes sampled state, and a portable binary representation. The
+ * implementation is a port of {@code com.twitter.finagle.tracing.TraceId}.
  */
-@AutoValue
-public abstract class SpanId {
+public final class SpanId {
 
-    /**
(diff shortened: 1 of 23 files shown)


brave-core/src/main/java/com/github/kristofa/brave/SpanId.java
```diff
@@ -2,54 +2,301 @@
...
- * 
- * @author kristof
+ * Contains trace data that's propagated in-band across requests, sometimes known as Baggage.
+ *
+ * <p>This implementation is biased towards a fixed-length binary serialization format that doesn't
+ * have a way to represent an absent parent (root span). In this serialized form, a root span is
+ * when all three ids are the same. Alternatively, you can use {@link #nullableParentId}.
+ *
+ * <p>Particularly, this includes sampled state, and a portable binary representation. The
+ * implementation is a port of {@code com.twitter.finagle.tracing.SpanId}.
```
kristofa: This is a port of `com.twitter.finagle.tracing.TraceId`, not `SpanId`?

brave-core/src/test/java/com/github/kristofa/brave/TraceDataTest.java
```diff
@@ -9,7 +9,8 @@
 public class TraceDataTest {
```
kristofa: It looks like we can get rid of `TraceData` and use `SpanId` instead?

kristofa:
👍

brave-core/src/test/java/com/github/kristofa/brave/TraceDataTest.java
adriancole: yep, that's a logical next step!

michaelsembwever:
+1

PR #169, opened 2016-05-04
Adds new samplers and benchmarks

adriancole:
This introduces two options for trace sampling:

* BoundaryTraceIdSampler - For high-volume sites, based on Finagle
* CountingTraceIdSampler - Trades lower performance for high accuracy

These include benchmarks and both work well, although the boundary
one is an order of magnitude faster under contention.

Here's the JMH comparison from my macbook pro and 4 threads:

```
SamplerBenchmarks.sampler_boundary            -9223372036854775808  thrpt   15  875.378 ± 28.245  ops/us
SamplerBenchmarks.sampler_boundary             1234567890987654321  thrpt   15  819.325 ± 81.720  ops/us
SamplerBenchmarks.sampler_counting            -9223372036854775808  thrpt   15   22.777 ±  0.540  ops/us
SamplerBenchmarks.sampler_counting             1234567890987654321  thrpt   15   22.223 ±  0.695  ops/us
```

CountingSampler is a better default, as support questions have come up with boundary tracing. A lot of people don't have much traffic and they can be surprised to find nothing being sampled at all!

This counting sampler is a thread-safe version of the old one, which gives an exact sample rate, per 100 decisions. The decisions are shuffled, so people should get fair decisions even before the 100 interval is over. It can do several samples per microsecond under contention, which is good enough for people getting started.

Basically, by switching to this, people can have a better experience, and save the tricky stuff for later.

### DIFF
--- a/brave-core/src/test/java/com/github/kristofa/brave/BoundarySamplerTest.java
+++ b/brave-core/src/test/java/com/github/kristofa/brave/BoundarySamplerTest.java
@@ -1,0 +1,15 @@
+package com.github.kristofa.brave;
+
+import org.assertj.core.data.Percentage;
+
+import static org.assertj.core.data.Percentage.withPercentage;
+
+public class BoundarySamplerTest extends SamplerTest {
+  @Override Sampler newSampler(float rate) {
+    return BoundarySampler.create(rate);
+  }
+
+  @Override Percentage expectedErrorRate() {
+    return withPercentage(10);
+  }
+}
--- a/brave-core/src/test/java/com/github/kristofa/brave/CountingSamplerTest.java
+++ b/brave-core/src/test/java/com/github/kristofa/brave/CountingSamplerTest.java
@@ -1,0 +1,15 @@
+package com.github.kristofa.brave;
+
+import org.assertj.core.data.Percentage;
+
+import static org.assertj.core.data.Percentage.withPercentage;
+
+public class CountingSamplerTest extends SamplerTest {
+  @Override Sampler newSampler(float rate) {
+    return CountingSampler.create(rate);
+  }
+
+  @Override Percentage expectedErrorRate() {
+    return withPercentage(0);
+  }
+}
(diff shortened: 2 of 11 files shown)


adriancole:
CountingSampler is a better default, as support questions have come up with boundary tracing. A lot of people don't have much traffic and they can be surprised to find nothing being sampled at all!

This counting sampler is a thread-safe version of the old one, which gives an *exact* sample rate. It can do several samples per microsecond under contention, which is good enough for people getting started. 

Basically, by switching to this, people can have a better experience, and save the tricky stuff for later.

adriancole:
@kristofa @eirslett if you don't mind.. ptal!

brave-core/src/test/java/com/github/kristofa/brave/BoundarySamplerTest.java
```diff
@@ -0,0 +1,15 @@
...
+
+import static org.assertj.core.data.Percentage.withPercentage;
+
+public class BoundarySamplerTest extends SamplerTest {
+  @Override Sampler newSampler(float rate) {
+    return BoundarySampler.create(rate);
+  }
+
+  @Override Percentage expectedErrorRate() {
+    return withPercentage(10);
```
adriancole: note that usually this is good in 3-5% range over 100K ids. However, sometimes it isn't, so the threshold is high to not flake travis.

brave-core/src/test/java/com/github/kristofa/brave/CountingSamplerTest.java
```diff
@@ -0,0 +1,15 @@
...
+
+import static org.assertj.core.data.Percentage.withPercentage;
+
+public class CountingSamplerTest extends SamplerTest {
+  @Override Sampler newSampler(float rate) {
+    return CountingSampler.create(rate);
+  }
+
+  @Override Percentage expectedErrorRate() {
+    return withPercentage(0);
```
adriancole: note: this is exact, even during the parallel test.

eirslett:
LGTM!

PR #170, opened 2016-05-04
Handling remote endpoint reporting as SERVER_ADDR binary annotation

chemicL:
A recent change in the way Spans for Client Requests are constructed have affected some clients, which reported the CLIENT_SEND annotation with the remote endpoint information. Now the local endpoint will be used, therefore for some clients the recent change might occur as missing non-traced services in the UI.
This PR uses an existing API in the `ClientTracer` for adding a binary annotation with the remote endpoint metadata.
The changes are backwards compatible and the actual extraction of the `Endpoint` is the users' responsibility.

### DIFF
--- a/brave-grpc/src/main/java/com/github/kristofa/brave/grpc/ClientRemoteEndpointExtractor.java
+++ b/brave-grpc/src/main/java/com/github/kristofa/brave/grpc/ClientRemoteEndpointExtractor.java
@@ -1,0 +1,9 @@
+package com.github.kristofa.brave.grpc;
+
+import com.twitter.zipkin.gen.Endpoint;
+import io.grpc.Channel;
+import io.grpc.MethodDescriptor;
+
+public interface ClientRemoteEndpointExtractor {
+    Endpoint remoteEndpoint(Channel channel, MethodDescriptor<?, ?> method);
+}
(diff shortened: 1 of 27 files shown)


adriancole:
based on travis output, looks like some spring tests need to be configured to supply ClientRemoteEndpointExtractor

brave-grpc/src/main/java/com/github/kristofa/brave/grpc/ClientRemoteEndpointExtractor.java
```diff
@@ -0,0 +1,9 @@
+package com.github.kristofa.brave.grpc;
+
+import com.twitter.zipkin.gen.Endpoint;
+import io.grpc.Channel;
+import io.grpc.MethodDescriptor;
+
+public interface ClientRemoteEndpointExtractor {
+    Endpoint remoteEndpoint(Channel channel, MethodDescriptor<?, ?> method);
```
adriancole: why would we do this as opposed to having people override the GrpcClientRequestAdapter?

adriancole:
I like what you are doing, with making it possible to set a server endpoint. However, I don't really like the idea of adding another interface to both http and grpc. It is conceptually simpler, and less complex wiring if we continue using adapters as the things that you modify when you need to "adapt" behavior.

You'll notice `HttpClientRequestAdapter` is already public, so no action needed there.

There's probably a sane default for GRPC, why not just add it? When someone asks to change the grpc adapter, we can make its type public (in a separate change)

By removing the extractor types it will also fix the tests :) as it removes spring wiring that broke them!

chemicL:
Agreed, let's close this PR then and move on to #171
As to the failed tests - I've been just executing `./mvnw test` which skips integration tests.

PR #171, opened 2016-05-05
Remote endpoint reporting as SERVER_ADDR binary annotation

chemicL:
A recent change in the way Spans for Client Requests are constructed have affected some clients, which reported the CLIENT_SEND annotation with the remote endpoint information. Now the local endpoint will be used, therefore for some clients the recent change might occur as missing non-traced services in the UI.
This PR uses an existing API in the ClientTracer for adding a binary annotation with the remote endpoint metadata.
The changes are backwards compatible and the actual extraction of the Endpoint is the users' responsibility by implementing their own request adapters and interceptors.


adriancole:
LGTM unrelated build fail. kicked travis

adriancole:
@kristofa can we enlist you for a release of brave after this? I don't think we broke api between last release. The new annotations and the TraceId thing would be great to have. Particularly, the latter will allow @michaelsembwever and I to fully integrate a cassandra tracer.

PR #177, opened 2016-05-19
Add meaningful service name when calling setClientSent.

tangrui:
1. If providing `zipkinServiceName` as a JDBC connection string properties, it is used.
2. If omit `zipkinServiceName`, the default one will be `mysql-${database}`.
3. Update README file.


adriancole:
very nice job

eirslett:
Great!

PR #178, opened 2016-05-19
Uses OpenZipkin deploy process and migrates to io.zipkin.brave group id

### LINKED ISSUE #161: migrate to zipkin-java'like deployment
Most impactful would be changing the group id from "com.github.kristofa" to "io.zipkin.brave". If we do, then we can use bintray instead of sonatype. This has the advantage of group access controls and nice things like auto-signing of jars. Most importantly, we can use the same process to release zipkin-java and brave.

adriancole:
This re-uses the deployment flow from zipkin-java which has the
following features:

* Makes Brave releaseable by anyone with access to the Bintray org
* Snapshot publishing to http://oss.jfrog.org/artifactory/oss-snapshot-local
* Fully automated release to maven central via release-N.N.N tag
  * Releases also published to https://jcenter.bintray.com

Notably, this migrates the maven group id to `io.zipkin.brave`. This
needs to be communicated on release!

Fixes #161

### DIFF
--- a/RELEASE.md
+++ b/RELEASE.md
@@ -1,0 +1,18 @@
+# Brave Release Process
+
+This repo uses semantic versions. Please keep this in mind when choosing version numbers.
+
+1. **Alert others you are releasing**
+
+   There should be no commits made to master while the release is in progress (about 10 minutes). Before you start
+   a release, alert others on [gitter](https://gitter.im/openzipkin/zipkin) so that they don't accidentally merge
+   anything. If they do, and the build fails because of that, you'll have to recreate the release tag described below.
+
+1. **Push a git tag**
+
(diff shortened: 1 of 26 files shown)


RELEASE.md
```diff
@@ -0,0 +1,18 @@
+# Brave Release Process
+
+This repo uses semantic versions. Please keep this in mind when choosing version numbers.
```
adriancole: @eirslett I bet you'd like this being called out, right? :)

eirslett: You bet!

eirslett:
> Notably, this migrates the maven group id to io.zipkin.brave. This
needs to be communicated on release!

https://maven.apache.org/guides/mini/guide-relocation.html

```xml
<distributionManagement>
    <relocation>
    </relocation>
  </distributionManagement>
```

PR #181, opened 2016-05-20
Update groupId in maven central shield

kristofa:
The Maven Central shield did not display latest brave
version because the groupId was not updated.

/cc @adriancole


eirslett:
LGTM

adriancole:
Thanks for the cleanup!

PR #184, opened 2016-05-25
Moves to zipkin 1.0

adriancole:
zipkin 1.0 is a stable release with semantic version guarantees. This
means we can depend on it directly and not need to shade it.

By not shading it, we have the advantage of other integrations working
without classpath confusion. For example, someone can make a span
collector that writes to zipkin storage directly.

### DIFF
--- a/brave-core/pom.xml
+++ b/brave-core/pom.xml
@@ -24,7 +24,6 @@
   </properties>
 
   <dependencies>
-    <!-- for sampler... don't worry. this dependency is shaded out! -->
     <dependency>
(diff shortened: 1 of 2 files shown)


eirslett:
So where will the release party be?

adriancole:
@eirslett it will be eventually consistent

eirslett:
:shipit: :cocktail: :fireworks: :bowtie: :clap:

Xylus:
:+1: :tada:

adriancole:
judging by the positive emojis, merging :)

PR #186, opened 2016-05-26
Adds LocalSpanCollector

adriancole:
LocalSpanCollector submits spans to a local storage component.

For example, if you are tracing a storage system like cassandra,
you might prefer to write directly to cassandra as opposed to over http.

This code replaces a similar class in zipkin-java, and also supports
cassandra tracing.

### DIFF
--- a/brave-core/src/main/java/com/github/kristofa/brave/FlushingSpanCollector.java
+++ b/brave-core/src/main/java/com/github/kristofa/brave/FlushingSpanCollector.java
@@ -1,0 +1,111 @@
+package com.github.kristofa.brave;
+
+import com.github.kristofa.brave.internal.Nullable;
+import com.twitter.zipkin.gen.Span;
+import java.io.Closeable;
+import java.io.Flushable;
+import java.io.IOException;
+import java.util.ArrayList;
+import java.util.LinkedList;
+import java.util.List;
+import java.util.concurrent.BlockingQueue;
+import java.util.concurrent.Executors;
+import java.util.concurrent.LinkedBlockingQueue;
+import java.util.concurrent.ScheduledExecutorService;
+
+import static java.util.concurrent.TimeUnit.SECONDS;
+
+/**
+ * Extend this class to offload the task of reporting spans to separate thread. By doing so, callers
+ * are protected from latency or exceptions possible when exporting spans out of process.
+ */
+public abstract class FlushingSpanCollector implements SpanCollector, Flushable, Closeable {
+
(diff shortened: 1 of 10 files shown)


adriancole:
cc @michaelsembwever

brave-core/src/main/java/com/github/kristofa/brave/FlushingSpanCollector.java
```diff
@@ -0,0 +1,111 @@
...
+import java.util.concurrent.LinkedBlockingQueue;
+import java.util.concurrent.ScheduledExecutorService;
+
+import static java.util.concurrent.TimeUnit.SECONDS;
+
+/**
+ * Extend this class to offload the task of reporting spans to separate thread. By doing so, callers
+ * are protected from latency or exceptions possible when exporting spans out of process.
+ */
+public abstract class FlushingSpanCollector implements SpanCollector, Flushable, Closeable {
```
adriancole: This used to be AbstractSpanCollector

brave-core/src/main/java/com/github/kristofa/brave/FlushingSpanCollector.java
```diff
@@ -0,0 +1,111 @@
...
+      }
+    }
+  }
+
+  /**
+   * Reports a list of spans over the current transport.
+   *
+   * @throws IOException (or RuntimeException) when thrown, drop metrics will increment accordingly
+   */
+  protected abstract void reportSpans(List<Span> drained) throws IOException;
```
adriancole: Main difference here as opposed to what was AbstractSpanCollector is that we don't require subtypes to encode the drained spans

adriancole:
anybody have feelings about this either way? I'm interested in this as I'd like to help make a cassandra tracer with it (that can self-report), as well I'd like to decommission similar code in zipkin https://github.com/openzipkin/zipkin-java/blob/master/zipkin-server/src/main/java/zipkin/server/brave/LocalSpanCollector.java

brave-core/src/main/java/com/github/kristofa/brave/FlushingSpanCollector.java
eirslett: Will it break anything in the public API?

adriancole: >
> Will it break anything in the public API?
>
Nope. main reason for making this a superclass of AbstractCollector was to
prevent that.

eirslett:
LGTM!

PR #188, opened 2016-06-14
Fix error in sample-rate minimum of BoundrySampler.

NegatioN:
Force usage of float in float-comparison, 0.0001f >= 0.0001 evaluates to false.
Fix docs-error in Sampler to reflect change in default samplers made in Brave 3.6.1

### DIFF
--- a/brave-core/src/main/java/com/github/kristofa/brave/BoundarySampler.java
+++ b/brave-core/src/main/java/com/github/kristofa/brave/BoundarySampler.java
@@ -28,7 +28,7 @@ public final class BoundarySampler extends Sampler {
   public static Sampler create(float rate) {
     if (rate == 0) return Sampler.NEVER_SAMPLE;
     if (rate == 1.0) return ALWAYS_SAMPLE;
-    checkArgument(rate > 0.0001 && rate < 1, "rate should be between 0.0001 and 1: was %s", rate);
+    checkArgument(rate >= 0.0001f && rate < 1, "rate should be between 0.0001 and 1: was %s", rate);
     final long boundary = (long) (rate * 10000); // safe cast as less <= 1
     return new BoundarySampler(boundary);
   }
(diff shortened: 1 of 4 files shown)


NegatioN:
@eirslett PR for sampleRates.
I am unsure about the choice to let Sampler.create() fail with an exception.
Also I don't quite see why the default Sampler should be CountingSampler which has a minimum samplerate of 1%, if it can fail with an exception. If not, then I would argue BoundarySampler should be the default Sampler.

eirslett:
I don't think 1 % is a reasonable minimum sample rate. 0.01 % would be better! Why do we even have a minimum?

NegatioN:
You're right. Maybe a more thorough rework of the class should be done here. The CountingSampler allocates a BitSet of 100 bits which is accessed in a thread-safe manner. I don't see why this BitSet couldn't be larger, other then that it would need more allocated space. Performance of this sampler would naturally be bad at low sample-rates.

Also, is there a particular reason to use floats in the samplers instead of doubles? float precision might end up being a factor if one allows for arbitrarly low sample-rates.

adriancole:
To respond to some rationale questions:

* CountingSampler and 100 bits: There's a javadoc that comments why, it isn't about size, rather accuracy. Right now, accuracy is guaranteed in units of 100 traces. If we enlarged this, it would decrease that accuracy accordingly. Since the counting sampler is made for accuracy, and to cause least confusion a window of 100 is a good default.

* Float vs Double: this is subjective, but a double is distracting as a sample rate. Even twitter don't use a sampling rate so low that you need a double to represent it.

adriancole:
the changes in this PR are valid as they fix bugs.

One could question if we should lower the minimum on the boundary sampler, although it is already quite low. We could take that in another PR I think

NegatioN:
Good to have you back Adrian ;)
I guessed there was something i was missing.
I agree there's no real need for a lower sampling rate than 10k.

adriancole:
good to be back! ps the lowest sampling rate I heard of was 1 out of 7M,
which was a rate I may be misremembering from Twitter's front end (written
in c++). It would be cool if someone asked for us to lower Brave's minimum,
it would hint at a large site and make for an interesting story!

PR #192, opened 2016-06-24
Upgrade libthrift to 0.9.3

klette:
Upgrades libthrift to 0.9.3.

An application in our stack that started experience "gc-madness" last night, the thrift transport started acting out. Seems like we were bitten by the bug fixed by https://github.com/apache/thrift/commit/bb98e97fd3c82117c87d23e3fb6b8bbd800784f2.

There are probably other fixes in there as well that could be useful :)

I don't know if its related to the upgrade, but I had to add a small sleep to `ITScribeSpanCollectorFailOnSetup` to make it run in a less flaky fashion on my machine.


adriancole:
I don't see any reason to not merge this change. Thanks!

PR #194, opened 2016-07-05
Brave cxf

podsiedzik:
Hi @kristofa,

I've added CXF support. It supports both cxf jax-ws and jax-rs servers and clients.
Module depends on `brave-http` and contains cxf interceptors, README file and some tests.
Any hints what else is needed are welcome.

---

Michał Podsiedzik


adriancole:
Thanks for taking time to work on this, Michal, and welcome to Brave! Code-wise, I'd check for error conditions (ex a non-success http response), but the code looks pretty solid.

We do have an important question to ask before digging deep.. should we merge this feature to master?

It usually isn't a good idea to add new module with only the author interested (see [rule of three](https://blog.codinghorror.com/rule-of-three/)). At least one of the regular contributors should be on board with maintaining this, if it is to go to master (as opposed to a separate repository). It ideally should have users besides the author.

Here are some hints for drumming up interest. You might find there are folks who would help use and support this:

CXF was mentioned last year by @fedor57, but hasn't ever been mentioned in zipkin's user group. CXF includes Apache HTrace support (which can export to zipkin), so folks using that only to get to zipkin might be interested. The authors of CXF might be interested. There was also a [blog on CXF](https://dzone.com/articles/knowing-how-all-your) (using finagle) a couple years ago.. maybe the author would be interested.

fedor57:
Yes, I made some classes and was able to intercept CXF requests and responses. But had no energy or skills to make something reusable out of it. Now our project is suspended, here's what has left. You may be interested to look shortly at my code if it can be somewhat useful.

https://github.com/fedor57/brave/tree/fedor-brave-cxf

podsiedzik:
@fedor57 
Hi, I'll definitely check it out.

@adriancole 
I wasn't aware of CXF HTrace support. Maybe we should wait with merging until some other people could participate. Or leave it as it is, as an example hot to integrate CXF with Brave. In the meantime I'll try to add some more tests.

adriancole:
@reta you interested in this? I think you were behind the blog and also HTrace support. We're basically seeing if there are enough people who want to trace CXF into Zipkin (via brave) to warrant merging this.

reta:
Thanks for mentioning this PR, @adriancole! That's right, we have integration with HTrace (which also is fully compatible with Zipkin protocol) as part of CXF distribution. I think it won' hurt to have yet another option with OpenZipkin / Brave on the table so our users could pick the right one. Do you want me to play a little bit with this PR before merging?

adriancole:
yeah please do, Andriy! Let us know what you think

reta:
Hey @podsiedzik @adriancole !

I was fortunate to have some time and play with CXF integration. It is actually very nice, easy to get started and use, definitely good job here! There is one minor thing I would like to suggest just to make integration with traditional JAX-RS applications (which do not use factory bean) easier, is to have a CXF feature provided, very simple one:

```
import org.apache.cxf.Bus;
import org.apache.cxf.annotations.Provider;
import org.apache.cxf.annotations.Provider.Scope;
import org.apache.cxf.annotations.Provider.Type;
import org.apache.cxf.feature.AbstractFeature;
import org.apache.cxf.interceptor.InterceptorProvider;

import com.github.kristofa.brave.Brave;
import com.github.kristofa.brave.cxf.BraveServerInInterceptor;
import com.github.kristofa.brave.cxf.BraveServerOutInterceptor;
import com.github.kristofa.brave.http.SpanNameProvider;

@Provider(value = Type.Feature, scope = Scope.Server)
public class BraveFeature extends AbstractFeature {
    private final SpanNameProvider spanNameProvider;
    private final Brave brave;
    
    public BraveFeature(final Brave brave, final SpanNameProvider spanNameProvider) {
        this.brave = brave;
        this.spanNameProvider = spanNameProvider;
    }
    
    @Override
    protected void initializeProvider(InterceptorProvider interceptorProvider, Bus bus) {
        final BraveServerInInterceptor in = new BraveServerInInterceptor(brave.serverRequestInterceptor(), spanNameProvider);
        final BraveServerOutInterceptor out = new BraveServerOutInterceptor(brave.serverResponseInterceptor());
        interceptorProvider.getInInterceptors().add(in);
        interceptorProvider.getOutInterceptors().add(out);
    }
}
```

In this case it would be very easy to use Brave from JAX-RS application subclasses. Client-side feature would be good as well but not as beneficial as server one.

Another one is a bit more complex issue and in scope of this PR would be nice just to mention it so you guys could be aware (the proper fix may need some thoughts). JAX-RS has support of async API calls (both on the client and on the server). That means that request and response (along with filters and interceptors) are going to be processed in different threads. The effect of this for tracing is that the spans which were started but were never stopped (or closed). The reason is, similarly to HTrace, Brave implementation also uses thread locals to maintain current span. I don't think this should block this PR from merging, but something to address later on. 

Answering on my behalf (but surely all CXF team), we would be more than happy to help you guys out and contribute as well in order to make the integration more robust and JAX-RSy. Please let me know what do you think guys :-)

Thanks, great one!

Best Regards,
    Andriy Redko

adriancole:
eek.. looks like a merge happened. might want to rebase

podsiedzik:
Hi @reta,

I've read your comments regarding async support. I've changed approach a little bit. 
I'm thinking about using cxf `Message` `Exchange` to pass `Span` state between request and response interceptors. Changes are already committed.

@adriancole 
What do you think?

reta:
@podsiedzik  :+1:

brave-cxf/pom.xml
```diff
@@ -0,0 +1,96 @@
...
+    <properties>
+        <project.build.sourceEncoding>UTF-8</project.build.sourceEncoding>
+        <cxf.version>3.1.7</cxf.version>
+    </properties>
+
+    <dependencies>
+        <dependency>
+            <groupId>io.zipkin.brave</groupId>
+            <artifactId>brave-http</artifactId>
+            <version>3.9.1-SNAPSHOT</version>
```
adriancole: watch for drift, or use `${project.version}`

brave-cxf/src/main/java/com/github/kristofa/brave/cxf/BraveServerInInterceptor.java
```diff
@@ -0,0 +1,39 @@
...
+import org.apache.cxf.message.Message;
+import org.apache.cxf.phase.AbstractPhaseInterceptor;
+import org.apache.cxf.phase.Phase;
+
+import com.github.kristofa.brave.Brave;
+import com.github.kristofa.brave.http.HttpServerRequestAdapter;
+import com.github.kristofa.brave.http.SpanNameProvider;
+
+/**
+ * @author Michał Podsiedzik
```
adriancole: please scrub author tags.. unless you want bugs introduced later to be attributed to you :)

brave-cxf/src/main/java/com/github/kristofa/brave/cxf/ServerRequest.java
```diff
@@ -0,0 +1,36 @@
...
+import java.net.URISyntaxException;
+
+import org.apache.cxf.message.Message;
+
+import com.github.kristofa.brave.http.HttpServerRequest;
+
+/**
+ * @author Michał Podsiedzik
+ */
+public class ServerRequest extends MessageWrapper implements HttpServerRequest {
```
adriancole: please make classes like this package private final. they aren't important enough to be public

brave-cxf/src/main/java/com/github/kristofa/brave/cxf/BraveServerInInterceptor.java
```diff
@@ -0,0 +1,39 @@
...
+ */
+public class BraveServerInInterceptor extends AbstractPhaseInterceptor<Message> {
```
adriancole: final all the classes, please, unless you are designing for extension, or have some non-functional reason why they can't be (in that case put a comment why)

brave-cxf/src/main/java/com/github/kristofa/brave/cxf/ClientRequest.java
```diff
@@ -0,0 +1,36 @@
...
+
+    @Override
+    public void addHeader(final String header, final String value) {
+        super.addHeader(header, value);
+    }
+
+    @Override
+    public URI getUri() {
+        try {
+            return new URI((String) message.getExchange().get(Message.ENDPOINT_ADDRESS));
```
adriancole: URI.create so you don't have to catch exception

brave-cxf/src/main/java/com/github/kristofa/brave/cxf/ClientResponse.java
```diff
@@ -0,0 +1,19 @@
+package com.github.kristofa.brave.cxf;
+
+import org.apache.cxf.message.Message;
+
+import com.github.kristofa.brave.http.HttpResponse;
+
+/**
+ * @author Michał Podsiedzik
+ */
+public class ClientResponse extends MessageWrapper implements HttpResponse {
```
adriancole: same comment as ServerRequest

brave-cxf/src/main/java/com/github/kristofa/brave/cxf/ClientResponse.java
```diff
@@ -0,0 +1,19 @@
...
+    public ClientResponse(final Message message) {
```
adriancole: please don't "final" parameters as the key word is distracting especially places like here. Add it when it says something notable (ex it is used in a anonymous class)

brave-cxf/src/main/java/com/github/kristofa/brave/cxf/MessageWrapper.java
```diff
@@ -0,0 +1,64 @@
...
+import java.util.List;
+import java.util.Map;
+import java.util.TreeMap;
+
+import org.apache.cxf.message.Message;
+
+/**
+ * @author Michał Podsiedzik
+ */
+public class MessageWrapper {
```
adriancole: package private. one-line description of why this is here. It seems to be http in nature, so should it be HttpMessageWrapper?

brave-cxf/src/main/java/com/github/kristofa/brave/cxf/MessageWrapper.java
```diff
@@ -0,0 +1,64 @@
...
+    protected final Message message;
+
+    public MessageWrapper(final Message message) {
+        this.message = message;
+    }
+
+    protected URI getUri() {
+        try {
+            return new URI((String) message.getExchange().get(Message.ENDPOINT_ADDRESS));
```
adriancole: URI.create

brave-cxf/src/main/java/com/github/kristofa/brave/cxf/MessageWrapper.java
```diff
@@ -0,0 +1,64 @@
...
+    protected String getHeader(final String header) {
+        List<String> values = getHeaders(message).get(header);
+        if (values != null && !values.isEmpty()) {
+            return values.get(0);
+        }
+        return null;
+    }
+
+    protected Map<String, List<String>> getHeaders(final Message message) {
+        Map<String, List<String>> headers = (Map<String, List<String>>) message.get(Message.PROTOCOL_HEADERS);
```
adriancole: mention in the javadoc of this class that it isn't thread-safe

brave-cxf/src/main/java/com/github/kristofa/brave/cxf/MessageWrapper.java
```diff
@@ -0,0 +1,64 @@
...
+
+    protected URI getUri() {
+        try {
+            return new URI((String) message.getExchange().get(Message.ENDPOINT_ADDRESS));
+        } catch (URISyntaxException e) {
+            throw new RuntimeException(e);
+        }
+    }
+
+    protected void addHeader(final String header, final String value) {
```
adriancole: all of these methods are used inside the same package, so make them package protected

brave-cxf/src/main/java/com/github/kristofa/brave/cxf/MessageWrapper.java
```diff
@@ -0,0 +1,64 @@
...
+    protected String getHttpMethod() {
+        return (String) message.get(Message.HTTP_REQUEST_METHOD);
+    }
+
+    protected int getHttpResponseCode() {
+        Integer code = (Integer) message.get(Message.RESPONSE_CODE);
+        if (code != null) {
+            return code.intValue();
+        } else {
+            // Correct assumption?
```
adriancole: currently, 200 isn't logged as a binary annotation. so change the comment to that and add a test that it doesn't show up when null.

// we don't know the status code. currently 200 doesn't show up anyway, so use that

brave-cxf/pom.xml
```diff
@@ -0,0 +1,96 @@
...
+        </dependency>
+        <dependency>
```
adriancole: make these two test dependencies.

brave-cxf/src/test/resources/log4j2.xml
```diff
@@ -0,0 +1,13 @@
...
+    <Console name="Console" target="SYSTEM_OUT">
+      <PatternLayout pattern="%d{HH:mm:ss.SSS} [%t] %-5level %logger{36} - %msg%n"/>
+    </Console>
+  </Appenders>
+  <Loggers>
+    <Root level="error">
+      <AppenderRef ref="Console"/>
+    </Root>
+  </Loggers>
+</Configuration>
```
adriancole: nit: please add eol before eof to all the files so red doesn't show up here :)

http://stackoverflow.com/questions/10082204/add-a-newline-only-if-it-doesnt-exist

adriancole:
FYI, I'm (finally) working this in. thanks for the efforts here and the patience!

adriancole:
out the door in 3.16

PR #203, opened 2016-07-21
give the flushing span collector flusher thread a name. fixes #202

### LINKED ISSUE #202: give the FlushingSpanCollector#Flusher thread pool scheduler a name
https://github.com/openzipkin/brave/blob/77f33f5ab24a2c29b93e37ab0b45e0df7ba5487e/brave-core/src/main/java/com/github/kristofa/brave/FlushingSpanCollector.java#L73

It make it way easier to debug problems why java processes won't shutdown if all of the threads have names.

a86c6f7964:
fix for #202

### DIFF
--- a/brave-core/src/main/java/com/github/kristofa/brave/FlushingSpanCollector.java
+++ b/brave-core/src/main/java/com/github/kristofa/brave/FlushingSpanCollector.java
@@ -12,6 +12,7 @@ import java.util.concurrent.BlockingQueue;
 import java.util.concurrent.Executors;
 import java.util.concurrent.LinkedBlockingQueue;
 import java.util.concurrent.ScheduledExecutorService;
+import java.util.concurrent.ThreadFactory;
 
 import static java.util.concurrent.TimeUnit.SECONDS;
 
@@ -70,7 +71,12 @@ public abstract class FlushingSpanCollector implements SpanCollector, Flushable,
   /** Calls flush on a fixed interval */
   static final class Flusher implements Runnable {
     final Flushable flushable;
-    final ScheduledExecutorService scheduler = Executors.newScheduledThreadPool(1);
+    final ScheduledExecutorService scheduler = Executors.newSingleThreadScheduledExecutor(new ThreadFactory() {
+      @Override
+      public Thread newThread(final Runnable r) {
+        return new Thread(r, "brave-core-flushing-span-collector-flusher");
+      }
+    });
 
     Flusher(Flushable flushable, int flushInterval) {
       this.flushable = flushable;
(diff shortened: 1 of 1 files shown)


a86c6f7964:
I also threw in setting this thread to be a daemon thread, since it probably does not need to hold up a shutdown.

adriancole:
I'd drop the daemon thread thing. It is a bug if this causes shutdown to hang, because that implies it was never stopped.

brave-core/src/main/java/com/github/kristofa/brave/FlushingSpanCollector.java
```diff
@@ -70,7 +71,14 @@ public void flush() {
   /** Calls flush on a fixed interval */
   static final class Flusher implements Runnable {
     final Flushable flushable;
-    final ScheduledExecutorService scheduler = Executors.newScheduledThreadPool(1);
+    final ScheduledExecutorService scheduler = Executors.newSingleThreadScheduledExecutor(new ThreadFactory() {
+      @Override
+      public Thread newThread(final Runnable r) {
+        final Thread t = new Thread(r, "brave-core-flushing-span-collector-flusher");
```
adriancole: I'd simplify this to the enclosing classname `FlushingSpanCollector` as it isn't likely there's a naming conflict, and using the classname makes lookups easier when there's a problem.

adriancole: actually, that's a bad name, too. let's use the concrete class's simple name. One way to do that is to add an ctor param for the name of the thread pool.

```java
new Flusher(this, flushInterval, getClass().getSimpleName());
```

Since this is package protected (aka internal), this change can be safely done.

adriancole: ex this would resolve to "KafkaSpanCollector" or "HttpSpanCollector"

adriancole:
Thanks for thinking of this! I made some notes.

a86c6f7964:
Okay I made the changes. I think that I got what you were wanting. Usually I use a specific string like I did only because greping through a repo for a verbose string like that is really easy. So I actually find it easier to find the exact place that the thread was created. The name you proposed still makes me have to look through the class, but it will give me the idea where it is.

I think that I also disagree a little about using the simple name, since if I am running brave in a huge project, it would be hard to know that the thread is brave specific, as opposed to having to know that `FlushingSpanCollector` is a brave class, but I am not going to hold anything up.

I removed the daemon. While I agree that it is a bug if you do not shut it down, I feel like it is usually nicer (as a library) to error on the side of ease of use. Maybe throwing warnings for unflushed pipes in shutdown hooks instead of haulting the whole shutdown process.

adriancole:
If you feel strongly about the daemon thread, I'd do it in another PR with
the laundry list of caveats applied (ideally with tests for those things).

Here's an example of how to both be nice and safe at the cost of several
paragraphs of comments to explain how to achieve both:
https://github.com/lightstep/lightstep-tracer-java/blob/788177153b25645b6e680f7c52f3055a26165d0f/common/src/main/java/com/lightstep/tracer/shared/AbstractTracer.java#L113

In the mean time, I'll merge this! thanks for the help

Issue #185, opened 2016-05-25
Add remote addresses to GRPC instrumentation


adriancole:
It looks like there's some support for logging the remote addresses in GRPC via standardized metadata keys.

https://github.com/grpc/grpc-java/blob/master/core/src/main/java/io/grpc/ServerCall.java#L60
https://groups.google.com/forum/#!topic/grpc-io/UodEY4N78Sk

adriancole:
these keys drifted..

PR #204, opened 2016-07-21
Adds the remote client address to server spans for gRPC

ewhauser:
This address the feature request for #185

### DIFF
--- a/brave-grpc/src/main/java/com/github/kristofa/brave/grpc/BraveGrpcServerInterceptor.java
+++ b/brave-grpc/src/main/java/com/github/kristofa/brave/grpc/BraveGrpcServerInterceptor.java
@@ -23,6 +23,7 @@ import io.grpc.ServerInterceptor;
 import io.grpc.Status;
 import io.grpc.Status.Code;
 
+import java.net.SocketAddress;
 import java.util.Collection;
 import java.util.Collections;
 
@@ -42,7 +43,7 @@ public final class BraveGrpcServerInterceptor implements ServerInterceptor {
         return next.startCall(new SimpleForwardingServerCall<ReqT, RespT>(call) {
             @Override
             public void request(int numMessages) {
-                serverRequestInterceptor.handle(new GrpcServerRequestAdapter<>(call.getMethodDescriptor(), requestHeaders));
+                serverRequestInterceptor.handle(new GrpcServerRequestAdapter<>(call, requestHeaders));
                 super.request(numMessages);
             }
 
@@ -56,11 +57,13 @@ public final class BraveGrpcServerInterceptor implements ServerInterceptor {
 
     static final class GrpcServerRequestAdapter<ReqT, RespT> implements ServerRequestAdapter {
 
+        private final ServerCall<ReqT, RespT> call;
         private MethodDescriptor<ReqT, RespT> method;
         private Metadata requestHeaders;
 
-        public GrpcServerRequestAdapter(MethodDescriptor<ReqT, RespT> method, Metadata requestHeaders) {
-            this.method = checkNotNull(method);
+        public GrpcServerRequestAdapter(ServerCall<ReqT, RespT> call, Metadata requestHeaders) {
+            this.call = checkNotNull(call);
+            this.method = checkNotNull(call.getMethodDescriptor());
             this.requestHeaders = checkNotNull(requestHeaders);
         }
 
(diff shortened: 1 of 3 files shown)


brave-grpc/src/main/java/com/github/kristofa/brave/grpc/BraveGrpcServerInterceptor.java
```diff
@@ -90,7 +93,14 @@ public String getSpanName() {
 
         @Override
         public Collection<KeyValueAnnotation> requestAnnotations() {
-            return Collections.emptyList();
+            SocketAddress socketAddress = call.attributes().get(ServerCall.REMOTE_ADDR_KEY);
+            if (socketAddress != null) {
+                KeyValueAnnotation remoteAddrAnnotation = KeyValueAnnotation.create(
```
adriancole: in zipkin, it would be more normal to add a binary annotation of type "ca" (client address). This is done via `ServerTracer.setServerReceived(int ipv4, int port, @Nullable String clientService)`, though maybe it isn't easy to get at here..

ewhauser: The [ServerRequestInterceptor](https://github.com/openzipkin/brave/blob/master/brave-core/src/main/java/com/github/kristofa/brave/ServerRequestInterceptor.java#L49) makes the call to `setServerReceived()`.  Because of the way the [ServerRequestAdapter](https://github.com/openzipkin/brave/blob/master/brave-core/src/main/java/com/github/kristofa/brave/ServerRequestAdapter.java) interface is setup, there is not a hook to set the client address annotation prior to `setServerReceived()` being called.

If you think it makes sense, I can add an additional method to `ServerRequestAdapter` that allows implementors to optionally supply the client info so it can be used by `ServerRequestInterceptor`.  We could also use this to enhance the `HttpServerRequestAdapter` as well using the various methods to pull client IPs ([ServletRequest.html#getRemoteAddr()](i.e. http://docs.oracle.com/javaee/6/api/javax/servlet/ServletRequest.html#getRemoteAddr())and well known proxy headers such as `X-Forwarded-For`)

ewhauser: I'd probably do that under a separate PR since it's a bigger change that affects more than just this issue

adriancole: sounds good.

brave-grpc/src/test/java/com/github/kristofa/brave/grpc/BraveGrpcInterceptorsTest.java
```diff
@@ -150,16 +150,22 @@ public void noSamplesWhenSamplingDisabled() throws Exception {
...
         assertThat(spans.size(), is(equalTo(2)));
-        assertThat(spans.get(0).getTrace_id(), is(equalTo(spans.get(1).getTrace_id())));
-        validateSpan(spans.get(0), Arrays.asList("ss", "sr"));
+
+        Span serverSpan = spans.get(0);
+        assertThat(serverSpan.getTrace_id(), is(equalTo(spans.get(1).getTrace_id())));
+        validateSpan(serverSpan, Arrays.asList("ss", "sr"));
+
+        //Server spans should have the GRPC_REMOTE_ADDR binary annotation
+        Optional<BinaryAnnotation> binaryAnnotation =
```
adriancole: the value of this would be consistent, right? It might be nice to show what it is (ex ip:port)

here's some assertj slickness..

```java
    assertThat(span.binaryAnnotations)
        .filteredOn(b -> b.key.equals(Constants.SERVER_ADDR))
        .extracting(b -> new String(b.value))
        .containsOnly("host:port");
```

adriancole: obv I used the wrong key, here.

adriancole:
one test tidy if you don't mind, then let's merge this! thx

also highly appreciate if you can take a stab at formalizing the "CA" thing as you mentioned (in a different PR)

ewhauser:
Done and done.  Will track "CA" under #205

adriancole:
nicely done!

PR #207, opened 2016-07-26
Adds LocalSpanThreadBinder, for async local spans

adriancole:
When I added LocalSpanState, I forgot to add LocalSpanThreadBinder.
Without this, you can't do things like close local spans with async
callbacks.


adriancole:
I noticed this missing when I tried to create Kafka integration.

Issue #206, opened 2016-07-26
Add service locator for Brave


ewhauser:
Some frameworks that we want to integrate with Brave instantiate control the instantiation of their objects and do not provide capabilities for injecting the Brave instance.  Examples of this include `brave-mysql` and `brave-jdbc`.  The service locator should provide the ability to provide the first instantiated instance of the Brave object and any subsequent instances.  

From @adriancole on #201:

> Finagle uses two contexts: one for finding the tracer(s) (local traceCtx)
> and one for finding the current span (broadcast idCtx).
> 
> At bootstrap, a null-tracer is associated with the local context,
> overridden lazily by service loader, though you can also override it at
> runtime (via what feels like a thread-local).
> 
> In the case of JDBC, where users control the threadpools, we could probably
> get by with a thread local approach.
> 
> ``` scala
> private case class TraceCtx(terminal: Boolean, tracers: List[Tracer]) {
> def withTracer(tracer: Tracer) = copy(tracers=tracer :: this.tracers)
> def withTerminal(terminal: Boolean) =
> if (terminal == this.terminal) this
> else copy(terminal=terminal)
> }
> private object TraceCtx {
> val empty = TraceCtx(false, Nil)
> }
> private[this] val traceCtx = new Contexts.local.Key[TraceCtx]
> ```
> 
> https://github.com/twitter/finagle/blob/fc321f804a22a695ec419902505c8509ffbd594d/finagle-core/src/main/scala/com/twitter/finagle/tracing/Trace.scala

adriancole:
While this doesn't make a service locator, hopefully #379 will make the interaction pattern possible. Let me know if not.

Issue #201, opened 2016-07-21
Add JDBC tracing support


ewhauser:
This issue is to discuss potential options for adding JDBC support to brave.  JDBC does not supply built in hooks for intercepting the execution of SQL calls, so a long held pattern for tracing JDBC calls has been to use a driver spy.  There a large number of these libraries in existence:

https://github.com/p6spy/p6spy
https://github.com/brunorozendo/log4jdbc-log4j2
https://github.com/arthurblake/log4jdbc

Out of these, p6spy seems to be the most attractive option:

* Apache 2 licensed
* Binaries available via Maven Central
* Zero dependencies
* Hooks available for attaching a generic event listener to JDBC actions (https://github.com/p6spy/p6spy/blob/master/src/main/java/com/p6spy/engine/event/JdbcEventListener.java, https://github.com/p6spy/p6spy/blob/master/src/main/java/com/p6spy/engine/logging/LoggingEventListener.java) where Zipkin traces could be added
* Seems actively maintained

The downside to using it seems to be that the user has to knowingly configure their JDBC connection string and supply a `spy.properties` file specific to brave.  Given the lack of hooks in JDBC, there really is not a way to avoid some level of end user configuration.

adriancole:
Sounds reasonable.. how does this actually work, though? is it likely to interfere with other code (magic bytecode weaving etc?)

I like this better than framework-specific choices (like jooq), as there's just so many of them!

ewhauser:
No byte code magic. The driver spy proxies the real JDBC driver. So instead of defining your connection string like:

```
jdbc:mysql://127.0.0.1/db
```

You would do:

```
jdbc:p6spy:mysql://127.0.0.1/db
```

adriancole:
clever!

and how does it access state? I remember the mysql one had an issue where
it needed to read a static variable to collaborate with brave's state..

ewhauser:
Are you talking about how to access the `Brave` instance?  I see that the MySQL one requires static injection of the `ClientTracer` (https://github.com/openzipkin/brave/blob/master/brave-mysql/src/main/java/com/github/kristofa/brave/mysql/MySQLStatementInterceptor.java#L25). Not really sure I see a way around this, although a [ServiceLoader](http://docs.oracle.com/javase/6/docs/api/java/util/ServiceLoader.html) might be a cleaner way to allow pluggable discovery.

adriancole:
[...]

PR #209, opened 2016-07-28
initial working implementation of JDBC tracing via p6spy

ewhauser:
This PR address #201. It is ready for review, but is not ready to be merged.  Outstanding issues:

- [X] ~~p6spy needs to release a version of 3.0.0 to Maven Central (the build will fail until this is resolved)~~
- [ ] address #206 (optional) #

### DIFF
--- a/brave-mysql/src/main/java/com/github/kristofa/brave/mysql/MySQLStatementInterceptor.java
+++ b/brave-mysql/src/main/java/com/github/kristofa/brave/mysql/MySQLStatementInterceptor.java
@@ -7,6 +7,9 @@ import com.mysql.jdbc.ResultSetInternalMethods;
 import com.mysql.jdbc.Statement;
 import com.mysql.jdbc.StatementInterceptorV2;
 
+import zipkin.Constants;
+import zipkin.TraceKeys;
+
 import java.net.Inet4Address;
 import java.net.InetAddress;
 import java.net.URI;
@@ -71,7 +74,7 @@ public class MySQLStatementInterceptor implements StatementInterceptorV2 {
 
     private void beginTrace(final ClientTracer tracer, final String sql, final Connection connection) throws SQLException {
         tracer.startNewSpan("query");
-        tracer.submitBinaryAnnotation("executed.query", sql);
+        tracer.submitBinaryAnnotation(TraceKeys.SQL_QUERY, sql);
 
         try {
             setClientSent(tracer, connection);
@@ -110,7 +113,7 @@ public class MySQLStatementInterceptor implements StatementInterceptorV2 {
                 tracer.submitBinaryAnnotation("warning.count", warningCount);
             }
             if (statementException != null) {
-                tracer.submitBinaryAnnotation("error.code", statementException.getErrorCode());
+                tracer.submitBinaryAnnotation(Constants.ERROR, statementException.getErrorCode());
             }
         } finally {
             tracer.setClientReceived();
--- a/brave-mysql/src/test/java/com/github/kristofa/brave/mysql/MySQLStatementInterceptorTest.java
+++ b/brave-mysql/src/test/java/com/github/kristofa/brave/mysql/MySQLStatementInterceptorTest.java
@@ -10,19 +10,22 @@ import static org.mockito.Mockito.verify;
 import static org.mockito.Mockito.verifyZeroInteractions;
 import static org.mockito.Mockito.when;
 
-import java.sql.DatabaseMetaData;
-import java.sql.SQLException;
-import java.util.Properties;
-
-import org.junit.Before;
-import org.junit.Test;
-import org.mockito.InOrder;
-
 import com.github.kristofa.brave.ClientTracer;
 import com.mysql.jdbc.Connection;
 import com.mysql.jdbc.PreparedStatement;
 import com.mysql.jdbc.ResultSetInternalMethods;
 import com.mysql.jdbc.Statement;
+
+import org.junit.Before;
+import org.junit.Test;
+import org.mockito.InOrder;
+
+import zipkin.Constants;
+import zipkin.TraceKeys;
+
+import java.sql.DatabaseMetaData;
+import java.sql.SQLException;
+import java.util.Properties;
 
 public class MySQLStatementInterceptorTest {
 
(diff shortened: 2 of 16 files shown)


adriancole:
Thanks for raising this. one thing to think about: should we call this brave-p6spy?

The code is centered around p6spy, which is jdbc focused, but a change to p6spy might warrant a change here (ex if an incompatible version update happens). On the other hand, folks may not recognize p6spy as a jdbc library, so naming it brave-p6spy might not help people understand what it does.

so it is a choice between

brave-jdbc
brave-p6spy
brave-jdbc-p6spy

ewhauser:
If we think `brave-jdbc` is too general then I would probably vote for changing it to `brave-jdbc-p6spy`

eirslett:
We could do like brave-http. That's a generic library that's then again used by actual http libraries, the dependency is like this:

application -> brave-okhttp -> brave-http -> brave-core

We could do something similar:

application -> brave-p6spy -> brave-jdbc -> brave-core

brave-jdbc/src/main/java/com/github/kristofa/brave/jdbc/BraveJdbcListener.java
```diff
@@ -0,0 +1,152 @@
...
+import zipkin.Constants;
+import zipkin.TraceKeys;
+
+import java.net.Inet4Address;
+import java.net.InetAddress;
+import java.net.UnknownHostException;
+import java.nio.ByteBuffer;
+import java.sql.SQLException;
+
+public final class BraveJdbcListener extends JdbcEventListener {
```
felixbarny: I think you can extend [`SimpleJdbcEventListener`](https://github.com/p6spy/p6spy/blob/master/src/main/java/com/p6spy/engine/event/SimpleJdbcEventListener.java) to reduce code duplication

brave-jdbc/src/test/java/com/github/kristofa/brave/jdbc/BraveJdbcBenchmarks.java
```diff
@@ -0,0 +1,115 @@
...
+ * driver.
+ */
+@Measurement(iterations = 10, time = 1)
+@Warmup(iterations = 10, time = 1)
+@Fork(1)
+@BenchmarkMode(Mode.Throughput)
+@OutputTimeUnit(TimeUnit.MILLISECONDS)
+@Threads(1)
+@State(Scope.Benchmark)
+public class BraveJdbcBenchmarks {
```
adriancole: thoughts about moving this into the benchmarks module?

brave-jdbc/src/test/java/com/github/kristofa/brave/jdbc/DerbyUtils.java
```diff
@@ -0,0 +1,16 @@
+package com.github.kristofa.brave.jdbc;
+
+import java.io.OutputStream;
+
+public class DerbyUtils {
+
+    //Get rid of the annoying derby.log file
```
adriancole: heh :)

brave-jdbc/src/main/java/com/github/kristofa/brave/jdbc/BraveJdbcListener.java
```diff
@@ -0,0 +1,152 @@
...
+        } catch (UnknownHostException ignored) {
+        }
+        try {
+            port = Integer.parseInt(options.getPort());
+        } catch (Exception ignored) {
+        }
+        serviceName = options.getServiceName();
+    }
+
+    public static void setClientTracer(final ClientTracer tracer) {
```
adriancole: I'd scrub the "final " on the params in this file.. I think most are

brave-jdbc/src/main/java/com/github/kristofa/brave/jdbc/P6BraveOptionsMBean.java
```diff
@@ -0,0 +1,11 @@
+package com.github.kristofa.brave.jdbc;
+
+public interface P6BraveOptionsMBean {
```
adriancole: are these only used for endpoint? If so, add javadoc links to Endpoint#XXX

brave-jdbc/src/test/java/com/github/kristofa/brave/jdbc/BraveJdbcTests.java
```diff
@@ -0,0 +1,135 @@
...
+            assertEquals(count[0], 1);
+            assertEquals(count[1], 1);
+        }, PREPARED_INSERT_SQL);
+        validateTraced(clientTracer, PREPARED_INSERT_SQL);
+    }
+
+    private void validateTraced(ClientTracer clientTracer, String sql) {
+        final InOrder order = inOrder(clientTracer);
+
+        order.verify(clientTracer).startNewSpan("query");
```
adriancole: are you cool with all of these operations having the same name?

adriancole:
we could introduce brave-jdbc (or brave-sql) to normalize how the span looks, if we ever end up having operation names not being "query" or similar. Ex. this is duplicated between mysql and here.

While I think this could be the best way for folks who cannot control libraries on top of JDBC drivers, I hesitate about calling this brave-jdbc. Ex. things like Jooq are nicer when you can control the application layer, as it doesn't require static (or otherwise) collaboration. In either case, the thing that avails you the tracing interceptor is the library, not JDBC. If one day JDBC has an interceptor, well that would be brave-jdbc.

long story short, I like `brave-p6spy` best despite it requiring people to know p6spy is a thing. I think it is probably good that they know this, since that's the thing doing the work anyway.

adriancole:
on the code and docs itself, looks fine and ready to go once minor comments are addressed.

adriancole:
@ewhauser you back from holiday? :D

brave-jdbc/src/main/java/com/github/kristofa/brave/jdbc/BraveJdbcListener.java
ewhauser: I did try this, but that interface doesn't appear to work correctly for insert statements.  The problem appears to be that the `String` of SQL that is passed in when an insert is executed seems to get lost in `SimpleJdbcEventListener` as it is not set on the `StatementInformation`

brave-jdbc/src/test/java/com/github/kristofa/brave/jdbc/BraveJdbcBenchmarks.java
ewhauser: This depends on a bunch of other classes in the test project, so not sure it's worth it to try to bundle those up as shareable so that this class could be in the benchmarks module.

brave-jdbc/src/test/java/com/github/kristofa/brave/jdbc/BraveJdbcTests.java
ewhauser: not sure I follow

ewhauser:
Comments addressed.  Awaiting resolution on #206 (could be addressed separately)

adriancole:
Looks great! thanks for wrapping this up.

PR #210, opened 2016-07-28
add BraveContext which is a registry for all Brave instances

ewhauser:
This addresses issue #206.

### DIFF
--- a/brave-core/src/main/java/com/github/kristofa/brave/Brave.java
+++ b/brave-core/src/main/java/com/github/kristofa/brave/Brave.java
@@ -1,12 +1,13 @@
 package com.github.kristofa.brave;
 
+import static com.github.kristofa.brave.InetAddressUtilities.getLocalHostLANAddress;
+import static com.github.kristofa.brave.InetAddressUtilities.toInt;
+
 import com.github.kristofa.brave.internal.Util;
+
 import java.net.UnknownHostException;
 import java.util.List;
 import java.util.Random;
-
-import static com.github.kristofa.brave.InetAddressUtilities.getLocalHostLANAddress;
-import static com.github.kristofa.brave.InetAddressUtilities.toInt;
 
 public class Brave {
 
@@ -233,5 +234,7 @@ public class Brave {
         serverSpanThreadBinder = new ServerSpanThreadBinder(builder.state);
         clientSpanThreadBinder = new ClientSpanThreadBinder(builder.state);
         localSpanThreadBinder = new LocalSpanThreadBinder(builder.state);
+
+        BraveContext.register(this);
     }
 }
--- a/brave-core/src/main/java/com/github/kristofa/brave/BraveContext.java
+++ b/brave-core/src/main/java/com/github/kristofa/brave/BraveContext.java
@@ -1,0 +1,68 @@
+package com.github.kristofa.brave;
+
+import java.util.Collections;
+import java.util.List;
+import java.util.concurrent.CopyOnWriteArrayList;
+
+/**
+ * The {@link BraveContext} is a registry of all {@link Brave} instances that have been created.
+ */
+public class BraveContext {
+
+    /**
+     * Application's should not be creating large numbers of {@link Brave} instances, so it is preferable
+     * to copy this every time as opposed to taking a read/write lock.
+     */
+    private static final CopyOnWriteArrayList<Brave> instances = new CopyOnWriteArrayList<Brave>();
+
+    private static volatile boolean initialized;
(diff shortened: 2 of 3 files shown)


brave-core/src/main/java/com/github/kristofa/brave/BraveContext.java
```diff
@@ -0,0 +1,68 @@
+package com.github.kristofa.brave;
+
+import java.util.Collections;
+import java.util.List;
+import java.util.concurrent.CopyOnWriteArrayList;
+
+/**
+ * The {@link BraveContext} is a registry of all {@link Brave} instances that have been created.
```
adriancole: is that this is a registry of all instances intentional, or rather a reality of the implementation.

Ex. I understand that if this registers on each new Brave, that this could have many Braves.. wondering if there is a use case for iterating through Braves, or if the use case is just "give me a brave"

adriancole:
I feel like this could be scarier than static injection.. these are hard references to brave and while true that people shouldn't do anything weird that would cause a lot of Braves, it will pin the tree from being gc'ed etc.

It is a good idea to think about nevertheless.. thanks for the spike

eirslett:
I think we used something like this in FINN as well, a ZipkinHolder that held a static reference to the application's Brave instance. We needed it for JDBC, because the Tomcat pool was initialized with configuration that in turn controlled Tomcat's reflection code. Really ugly, really. It should be possible to set up JDBC pools programmatically, and then inject the brave clienttracer. Isn't that possible with p6spy? Maybe we could patch p6spy to add that capability?

adriancole:
Agreed this is a real problem. Just wondering how we can solve it
(particularly as some might instantiate Brave a lot passing the same state
object)

For example, it may be usually the case that there are more Braves than
JDBC pools, but it might be the other way, too.

Iterating through spies might be more meaningful when wiring tracing than
iterating through Braves. Then again, some apps reconstruct their
connection pools late in the app.. Ex when password changes.

Also, since these connection pool threads arent always easy to control, a
static (vs thread pool) association might be most sensible in practice.

So I guess I'm getting at it would be handy to have a brave holder behind
something like Brave.get(). Ex this could be reused for the mysql
interceptor too. Just if we did this, I feel awkward about an iterable
return of Braves and also strong references (since braves may come and go).

Thoughts?

brave-core/src/main/java/com/github/kristofa/brave/Brave.java
```diff
@@ -233,5 +234,7 @@ private Brave(Builder builder) {
         serverSpanThreadBinder = new ServerSpanThreadBinder(builder.state);
         clientSpanThreadBinder = new ClientSpanThreadBinder(builder.state);
         localSpanThreadBinder = new LocalSpanThreadBinder(builder.state);
+
+        BraveContext.register(this);
```
schlosna: might want to consider moving registration to `Builder#build()` to avoid leaking `this`.

brave-core/src/main/java/com/github/kristofa/brave/BraveContext.java
schlosna: would definitely like way to unregister specific `Brave` instance (e.g. as part of service lifecycle hook).

ewhauser:
@adriancole: I'm cool removing the iterable of Brave's and making it last one wins.  I don't really have any use cases where I instantiate multiple instances in a given static context.  The reason behind this was to mimic the Finagle functionality.

Should we move to just tracking the last instantiated instance?

adriancole:
yeah I think last instantiated, while has the same drift concerns when
multiple braves are used, has no leak concern (as only a single ref would
be used). seems like a nice way incremental way to get things going.

brave-core/src/main/java/com/github/kristofa/brave/Brave.java
adriancole: +1

adriancole:
we can probably pick this back up with experience gained in the `CurrentTraceContext` api

adriancole:
Thanks for the work here. I think it makes most sense to add this in Brave 4 https://github.com/openzipkin/brave/pull/379

pradeep12vr:
@adriancole Suggest an example which illustrates the use of openzipkin/brave for distributed application tracing. I am just a beginner and can't seem to find proper resource to start exploring.

adriancole:
how about this? https://github.com/openzipkin/brave-webmvc-example

pradeep12vr:
@adriancole Thanks it was very helpful

Issue #211, opened 2016-08-03
Tracing microservices that use Kafka


jpiechowka:
Is there a documentation or guide on how to setup Brave to trace microservices that use Kafka to communicate? I would like to know how to configure it to work, some example would be perfect.

adriancole:
this is a tricky one, but I have an experiment on this. I'll post it in a
bit. It requires you to use kafka keys for propagation.

piechowka:
Perfect! Thank you very much for this @adriancole

adriancole:
@jpiechowka how's it been going?

jpiechowka:
It works well for me. But we decided to postpone using it on production right now. Priority is to implement tracing in our JavaScript front end and Java back end application.

adriancole:
Now that Kafka 0.11 supports headers, we can likely formalize support for tracing using B3 (like many do for rabbit etc) http://www.mirrorservice.org/sites/ftp.apache.org/kafka/0.11.0.0/RELEASE_NOTES.html

adriancole:
note that SPMC tracing would require consumers to use the incoming span ID as the parent (not share the trace ID as we do in normal and one-way RPC)

adriancole:
https://github.com/openzipkin/brave/tree/master/instrumentation/kafka-clients is going out as Brave 4.6 thanks to @ImFlog

Issue #164, opened 2016-04-28
does brave support async call? 


dragontree101:
i want to use async client, does brave can support async client? 
thanks!

adriancole:
Here's a related issue https://github.com/openzipkin/zipkin/issues/1189

Most importantly, we need to capture what the scenario here is, and what problems are trying to be solved. Currently, as mentioned above, we can "fake" rpc by treating the producer activity as the client side of an RPC and the consumer activity as the server side. That will work for sure between now and such time an alternate method exists.

adriancole:
async is via the "flush" method. Here's example: https://github.com/openzipkin/brave/tree/master/brave/src/test/java/brave/features/async

dragontree101:
thanks very much!

PR #212, opened 2016-08-03
DO NOT MERGE: Adds spike of Kafka integration

adriancole:
This abuses local tracer to send a trace across kafka (where kafka
message key is abused to hold propagated state)

There are a number of tradeoffs to consider, not the least of which the
potential of a dual parent on the consumer side. The current
implementation simulates a client-server span which notably includes
boundary annotations of "ms" "mr".

see #164 #211

### DIFF
--- a/brave-kafka/src/main/java/com/github/kristofa/brave/kafka/TracedKafkaConsumer.java
+++ b/brave-kafka/src/main/java/com/github/kristofa/brave/kafka/TracedKafkaConsumer.java
@@ -1,0 +1,81 @@
+package com.github.kristofa.brave.kafka;
+
+import com.github.kristofa.brave.Brave;
+import com.github.kristofa.brave.SpanId;
+import com.twitter.zipkin.gen.Span;
+import java.io.Closeable;
+import java.io.IOException;
+import java.util.HashMap;
+import java.util.Iterator;
+import java.util.Map;
+import java.util.Properties;
+import kafka.consumer.Consumer;
+import kafka.consumer.ConsumerConfig;
+import kafka.consumer.ConsumerIterator;
+import kafka.consumer.KafkaStream;
+import kafka.javaapi.consumer.ConsumerConnector;
+import kafka.message.MessageAndMetadata;
+import zipkin.Constants;
+
+public final class TracedKafkaConsumer implements Closeable {
+  final Brave brave;
+  final ConsumerConnector delegate;
+
+  TracedKafkaConsumer(Brave brave, Properties properties) {
+    this.brave = brave;
+    this.delegate = Consumer.createJavaConsumerConnector(new ConsumerConfig(properties));
+  }
+
+  Iterator<byte[]> iterator(String topic) {
+    Map<String, Integer> topicCountMap = new HashMap<>();
+    topicCountMap.put(topic, 1);
+    final KafkaStream<byte[], byte[]> stream = delegate.createMessageStreams(topicCountMap)
+        .values()
+        .iterator()
+        .next()
+        .get(0);
+
+    return new Iterator<byte[]>() {
+      ConsumerIterator<byte[], byte[]> delegate = stream.iterator();
+
+      @Override public boolean hasNext() {
+        return delegate.hasNext();
+      }
+
+      @Override public byte[] next() {
+        MessageAndMetadata<byte[], byte[]> result = delegate.next();
+        Span span = SpanId.fromBytes(result.key()).toSpan();
+        span.setName(result.topic());
+        brave.localSpanThreadBinder().setCurrentSpan(span);
+        try {
+          brave.localTracer().submitAnnotation(Constants.WIRE_RECV);
+          brave.localTracer().submitAnnotation("mr");
+          brave.localTracer().submitBinaryAnnotation("kafka.partition", result.partition());
+
+          return result.message();
+        } finally {
+          brave.localTracer().finishSpan(0); // flush (which completes) the messaging span
+
+          // Note: we aren't associating the span that just finished as a thread context
+          // This is potentially a dual-parent problem. Imagine you are a consumer of this iterator.
+          // You may have already been in a trace prior to calling next(). Since zipkin doesn't
+          // support dual parents we have a choice to either ignore the trace that resulted in this
+          // message, or use its trace id as the current trace id.
+          //
+          // Ex. existing trace calls iterator.next() on a kafka producer-backed iterator
+          // This is a join. We could log the joining trace id into the message-trace, or visa-versa
+          //
+          // Ex. no existing trace calls iterator.next() on a kafka producer-backed iterator
+          // We could choose to simply set the Brave's current server span to the span that just finished.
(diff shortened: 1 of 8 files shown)


brave-kafka/src/main/java/com/github/kristofa/brave/kafka/TracedKafkaConsumer.java
```diff
@@ -0,0 +1,81 @@
...
+        brave.localSpanThreadBinder().setCurrentSpan(span);
+        try {
+          brave.localTracer().submitAnnotation(Constants.WIRE_RECV);
+          brave.localTracer().submitAnnotation("mr");
+          brave.localTracer().submitBinaryAnnotation("kafka.partition", result.partition());
+
+          return result.message();
+        } finally {
+          brave.localTracer().finishSpan(0); // flush (which completes) the messaging span
+
```
adriancole: big long note to read here

adriancole:
@coolbetm @dragontree101 @jpiechowka please reply-back if you think this "messaging span" thing is generally useful. No rush, but if there's a critical mass of nods, I don't mind pushing for this being formalized upstream (which would make things like the dependency graph work).

adriancole:
Here's an example scenario asked by @coolbetm
> Consider the following simplish example:
Suppose service A receives a post to its REST
endpoint and produces the body of the post to a kafka
topic. Service B is tailing on that topic and consumes
it, parses it, and produces a derived message to second topic.
Service C is tailing on the second topic, receives the derived
message, and responds by executing a database update.
What do you imagine the zipkin trace should look like (what spans are there
and in what order do they start/end)?

{traceid=1, parent=_, spanid=1 [sr=serviceA, ss=serviceA]} // server span
{traceid=1, parent=1, spanid=2 [ms=serviceA, mr=serviceB]} // messaging span
{traceid=1, parent=2, spanid=3 [lc="something doer"/serviceB]} // local span
{traceid=1, parent=3, spanid=4 [ms=serviceB, mr=serviceC]} // messaging span
{traceid=1, parent=4, spanid=5 [cs=serviceC, cr=serviceC, sa=database]} // client span

jpiechowka:
@adriancole It worked for me :+1:

coolbetm:
Thanks for doing a spike!

This seems useful, but I am not sure it is ready for prime time.  There are two parts to this - the data modeling and the implementation. I have some concerns about both.

#### implementation
WRT the implementation shown here there are just a couple of issues that I think are easy to solve.  

1. Highjacking the key is probably the wrong thing to do.  Some Kafka users make use of the keys to support log compaction in their kafka topics, and the forthcoming Kafka Streams framework makes heavy use of Kafka Message keys to support joins and aggregations.  I think, though, that it should be harmless and *almost* as easy to transparently wrap the Value in an envelope containing the traceId.
1. Much of the world, and certainly the confluent guys with their streams processing framework, is moving on to Kafka 10, where the client API is a bit different.  I suspect the right thing to do there would be to wrap the serializer/deserializer that you pass into the Producer/Consumer  rather than the Producer/Consumer itself, but I'm not 100% sure yet.
1. If we did continue using this method of wrapping the Producer itself, it would have to be more feature complete (for example include all of the overloads for the send method).  I'm sure you know this and just left it out because it was not necessary for the POC, but I'm pointing it out anyway.

It's not yet decided whether my group is going to try using Zipkin to monitor our streams processors.  If the decision is yes, I might be able to get them to let me work on some of this and contribute it back, but that's still an unknown.

#### modeling traces

I'm a Zipkin n00b and it's not yet clear to me how far short (if at all) Zipkin's data model falls of being able to capture the full generality of streams processing architectures.  Even if it can only model a simple linear data flow like the example above it still has value for monitoring and measuring performance, but I would like to understand precisely what the limitations are.

In general, many more complicated scenarios might arise:

- a processor might join two streams to produce a third, and we would love to be able to regard each message produced by that processor as a member of each of two separate traces  (i suspect zipkin's data model will never support this, but am not sure)
- a processor might fork a stream into multiple branches.  We would want the parallel processing in all downstream branches to appear in the same trace (i suspect Zipkin will have no trouble with this, but am not sure)
- it may also be important that no process sending trace data to Zipkin can know whether or not it is the end of the trace.  It's not totally clear to me from the available documentation whether this is an issue.

#### That's all for now.
Those are my thoughts.  I apologize if there are too many of them for one comment.

adriancole:
>
>
>    1. I think, though, that it should be harmless and *almost* as easy to
>    transparently wrap the Value in an envelope containing the traceId.
>
> Practically speaking this is the problem in Kafka. There's outstanding
requests to add metadata keys to the message format, but they haven't
landed. There's two options: key and enveloping. The latter has a huge
drawback you haven't mentioned, which is that it relies on people agreeing
across languages on an envelope format. Folks should be aware that getting
coordination on a value envelope type requires some definition, especially
as there are processors who inspect (peek) at payloads. They'd have to be
rewritten to avoid mistaking a trace envelope for something else. So,
basically we should propose (ideally at Kafka level) an enveloping
workaround for the lack of progress on metadata keys.

I pretty much agree on all the other things you've mentioned, including
learning some new limitations about keys!

>
>    - a processor might join two streams to produce a third, and we would
>    love to be able to regard each message produced by that processor as a
>    member of each of two separate traces (i suspect zipkin's data model will
>    never support this, but am not sure)
>
> htrace does this today (multiple parents), though it requires a primary
parent (ex you have to choose one!). We've discussed many times in zipkin
adding a "links" feature which would allow multiple associations up and
downstream. Basically, it would like different traces together. In the case
of multiple streams converging, they are pretty much guaranteed to be
different traces. So, it isn't a never, just something we need people to be
willing to test. UI would need work, but we could cheaply start by just
building hrefs to the linked traces.


>
>    - a processor might fork a stream into multiple branches. We would
>    want the parallel processing in all downstream branches to appear in the
>    same trace (i suspect Zipkin will have no trouble with this, but am not
>    sure)
>
> agreed. this is normal in zipkin.

>
>    - it may also be important that no process sending trace data to
>    Zipkin can know whether or not it is the end of the trace. It's not totally
>    clear to me from the available documentation whether this is an issue.
>
> no issue at all

>
>
> That's all for now.
>
> Those are my thoughts. I apologize if there are too many of them for one
> comment.
>
nope not too many. really glad about the feedback. I think we need to split
up the modeling thing from the Kafka thing, but I'm glad the Kafka thing
helped smoked out the modeling thing!

adriancole:
let's see if we can nail a design down for zipkin v1 model here: https://github.com/openzipkin/zipkin/issues/1243

adriancole:
related issue: multiple parents aka linked traces https://github.com/openzipkin/zipkin/issues/1244

adriancole:
gonna close this but leave the branch around a while for discussion. I don't think this should be merged.

songxin1990:
@adriancole  kafka kip-82 will add message headers, details see [https://cwiki.apache.org/confluence/display/KAFKA/KIP-82+-+Add+Record+Headers](url)

adriancole:
Great. Lets implement kafka tracing as soon as there is a released version

PR #213, opened 2016-08-04
Add http method to binary annotations

pavolloffay:
Hello, I would like to add HTTP method to the binary annotations.

@kristofa could you please review?


adriancole:
Iirc HTTP method is also the default span name. If we do this, we add 15
bytes to the most popular spans. I'd like another plus one from a higher
scale user before merging.

pavolloffay:
I think it is better to have it explicitly recorded as an HTTP method in binary annotations. What if someone is tracing JMS or even other technologies?

kristofa:
I agree with Adrian. 
If we would trace JMS or other protocols we can add custom annotations like we already do for grpc, mysql.

pavolloffay:
@kristofa @adriancole thanks for a feedback, I'm closing it.

PR #215, opened 2016-08-10
Localtracer - Make builder public

fzakaria:
I was incorporating LocalTracer into some code with a custom LocalSpanAndEndpoint.
I could not create a LocalTracer itself without going through the Brave class which was a bit frustrating.

I'm working on a non-blocking webserver framework and the provided implementation of LocalSpanAndEndpoint does not work - as it relies on thread local.

Instead each request through the framework ([Vert.x](http://vertx.io/)) is given a 'Context' object that is associated to the request and passed through the handler chain (the context object has a Map<String, Object> to store request specific information).

I must create a new LocalSpanAndEndpoint and feed it the request underlying Map at the start of every request - which was wanting me to create a new LocalTracer. 

PS: thanks for getting back to me on the google groups. We are already logging some internal timings to a `MetricScope` class (a directed acyclic graph) with start time/duration. 
I was hoping to try and just convert those tracings but hit the wall against the recursive LocalSpans

### DIFF
--- a/brave-core/src/main/java/com/github/kristofa/brave/LocalTracer.java
+++ b/brave-core/src/main/java/com/github/kristofa/brave/LocalTracer.java
@@ -56,15 +56,15 @@ public abstract class LocalTracer extends AnnotationSubmitter {
     @AutoValue.Builder
     public abstract static class Builder {
 
-        abstract Builder spanAndEndpoint(LocalSpanAndEndpoint spanAndEndpoint);
+        public abstract Builder spanAndEndpoint(LocalSpanAndEndpoint spanAndEndpoint);
 
-        abstract Builder randomGenerator(Random randomGenerator);
+        public abstract Builder randomGenerator(Random randomGenerator);
 
-        abstract Builder spanCollector(SpanCollector spanCollector);
+        public abstract Builder spanCollector(SpanCollector spanCollector);
 
-        abstract Builder traceSampler(Sampler sampler);
+        public abstract Builder traceSampler(Sampler sampler);
 
-        abstract LocalTracer build();
+        public abstract LocalTracer build();
     }
 
     /**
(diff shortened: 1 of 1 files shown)


adriancole:
recursive spans are not supported.. and you'll run into problems where you can't close a parent after you close the child. #166 was opened to support nested spans. Until then, you'd need to make some hacks that violate the contract of LocalTracer. As such, it is almost just as well to create your own LocalTracer (not named localTracer, rather something else).

Otherwise, here's a hack that might help you get by until #166. It totally breaks nested state management in local spans, but it will allow you to create one (assuming you do manual thread state management), and doesn't impact the api.

```java
Span nextLocal;
if (localThreadBinder.getCurrentSpan() == null) { 
  nextLocal = localTracer.startNewSpan(.. // we're the root span
} else { // our parent is a local span
  Span parent = localThreadBinder.getCurrentSpan();

  // Brave's thread state management is fixed, as opposed to a stack:
  // You can only create a child local span from a server span.
  // See https://github.com/openzipkin/brave/issues/166

  // save off the old server span
  Span serverSpan = serverThreadBinder.getCurrentSpan();
  try {
     // fake as if the local span is actually a server one.
     // this will allow us to create a child
     serverTracer.setStateCurrentTrace(
                parent.getTrace_id(),
                parent.getId(),
                parent.getParent_id(),
                parent.getName()
     );
     // this will work it will overwrite the threadlocal of the former local span
     // this means that localTracer.finish() will close the nextLocal, but after that
     // happens, you cannot call it again to close parent!
     nextLocal = localTracer.startNewSpan(... 
  } finally {
    serverThreadBinder.setCurrentSpan(serverSpan); // put the old server span back
  }
}

```

adriancole:
please try https://github.com/openzipkin/brave/blob/master/brave-core/src/main/java/com/github/kristofa/brave/InheritableServerClientAndLocalSpanState.java and let us know if you still need help prior to #166

PR #216, opened 2016-08-12
Newly created local spans should inherit from the current local span

michaelsembwever:
A newly created local span should inherit from any current local span before inheriting from the current server span.

Also add method to clear the current local span.

This is needed to implement a OpenTracing to Zipkin (via Brave) bridge, see https://github.com/opentracing/opentracing-java/pull/25

### DIFF
--- a/brave-core/src/main/java/com/github/kristofa/brave/LocalTracer.java
+++ b/brave-core/src/main/java/com/github/kristofa/brave/LocalTracer.java
@@ -68,6 +68,13 @@ public abstract class LocalTracer extends AnnotationSubmitter {
     }
 
     /**
+     * Clears current span.
+     */
+    public void clearCurrentSpan() {
+        spanAndEndpoint().state().setCurrentLocalSpan(null);
+    }
+
+    /**
      * Request a new local span, which starts now.
      *
      * @param component {@link Constants#LOCAL_COMPONENT component} responsible for the operation
@@ -83,7 +90,13 @@ public abstract class LocalTracer extends AnnotationSubmitter {
     }
 
     private SpanId getNewSpanId() {
-        Span parentSpan = spanAndEndpoint().state().getCurrentServerSpan().getSpan();
+        Span parentSpan = spanAndEndpoint().state().getCurrentLocalSpan();
+        if (parentSpan == null) {
+            ServerSpan serverSpan = spanAndEndpoint().state().getCurrentServerSpan();
+            if (serverSpan != null) {
+                parentSpan = serverSpan.getSpan();
+            }
+        }
         long newSpanId = randomGenerator().nextLong();
         SpanId.Builder builder = SpanId.builder().spanId(newSpanId);
         if (parentSpan == null) return builder.build(); // new trace
--- a/brave-core/src/test/java/com/github/kristofa/brave/LocalTracerTest.java
+++ b/brave-core/src/test/java/com/github/kristofa/brave/LocalTracerTest.java
@@ -1,6 +1,7 @@
 package com.github.kristofa.brave;
 
 import static org.junit.Assert.assertEquals;
+import static org.junit.Assert.assertNotNull;
 import static org.junit.Assert.assertNull;
 import static org.mockito.Mockito.mock;
 import static org.mockito.Mockito.verify;
@@ -50,6 +51,14 @@ public class LocalTracerTest {
                 .build();
     }
 
+    @Test
+    public void testClearCurrentSpan() {
+        state.setCurrentLocalSpan(ServerSpan.create(PARENT_TRACE_ID, PARENT_SPAN_ID, null, "name").getSpan());
+        assertNotNull(state.getCurrentLocalSpan());
+        localTracer.clearCurrentSpan();
+        assertNull(state.getCurrentLocalSpan());
+    }
+
     /**
      * When a local span is started without a timestamp, microseconds and a tick are recorded for
      * duration calculation. A binary annotation is added for search by component.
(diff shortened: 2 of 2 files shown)


brave-core/src/main/java/com/github/kristofa/brave/LocalTracer.java
```diff
@@ -68,6 +68,13 @@ static Builder builder(LocalTracer source) {
     }
 
     /**
+     * Clears current span.
+     */
+    public void clearCurrentSpan() {
```
adriancole: rather than opening a new tool for this job, probably better to direct folks to the existing one. Note that this was only recently noticed.

https://github.com/openzipkin/brave/commit/aca24d9d6ba6f1c00aef3594c3e14e5496296c13

brave-core/src/test/java/com/github/kristofa/brave/LocalTracerTest.java
```diff
@@ -82,6 +91,38 @@ public void startNewSpan() {
     }
 
     /**
+     * When a local span is started without a timestamp, microseconds and a tick are recorded for
```
adriancole: I'd not copy/paste javadoc for this method as its intent is testing something different. For example, its focus isn't microseconds, rather local parent.

brave-core/src/main/java/com/github/kristofa/brave/LocalTracer.java
```diff
@@ -83,7 +90,8 @@ public SpanId startNewSpan(String component, String operation) {
     }
 
     private SpanId getNewSpanId() {
-        Span parentSpan = spanAndEndpoint().state().getCurrentServerSpan().getSpan();
+        Span parentSpan = spanAndEndpoint().state().getCurrentLocalSpan();
```
adriancole: If we were to do this, we should probably keep parity with the logic in ClientTracer, copy/pasting here:

```java
        Span parentSpan = spanAndEndpoint().state().getCurrentLocalSpan();
        if (parentSpan == null) {
            ServerSpan serverSpan = spanAndEndpoint().state().getCurrentServerSpan();
            if (serverSpan != null) {
                parentSpan = serverSpan.getSpan();
            }
        }
```

for example, that addresses the possible null pointer

adriancole:
I see the need, here, but there's an important thing to keep in mind. We are overwriting the old local span. So, for example, when the new local one is finished, the prior isn't available anymore. 

The root cause is discussed here: #166 

OpenTracing's api does not imply propagation, so I'd hesitate before making hooks that will lead people to overwriting state, if OpenTracing is the reason for it.

adriancole:
for example, thread state propagation is explicitly out-of-scope in OT http://opentracing.io/use-cases/#in-process-request-context-propagation

brave-core/src/main/java/com/github/kristofa/brave/LocalTracer.java
michaelsembwever: i'm not sure why i'd want to use localSpanThreadBinder instead of localTracer here?

Given that the other tracers (serverTracer and clientTracer) have this method doesn't it make sense that localTracer also for the sake of consistency has it?

Then the effort to reduce API (/cognitive load) can be done separately, but equally to all three tracers?

brave-core/src/main/java/com/github/kristofa/brave/LocalTracer.java
michaelsembwever: Sure. But that wasn't there before. I will add it.

brave-core/src/test/java/com/github/kristofa/brave/LocalTracerTest.java
michaelsembwever: Sure. But there was nothing in this apidoc that was specific to either test case. hence the re-use looks appropriate.

brave-core/src/main/java/com/github/kristofa/brave/LocalTracer.java
adriancole: This method clearCurrentSpan() only exists on ServerTracer: it is
because of setStateCurrentTrace(), used when a span is received from
headers.

That use case doesn't exist for ClientTracer or LocalTracer.

adriancole:
I humbly ask you to close this change in favor of more complete work in https://github.com/openzipkin/brave/issues/166

I don't think waiting for the real fix is waiting in vane. Last month @jkozlowski looked very deeply into 166 (now on holiday). [Two days ago](https://github.com/DistributedTracing/continuation-local-storage-jvm/pull/1#issuecomment-238862120), @schlosna also offered his help. I think we can get a long-term approach in place, but not today.

In the mean time, if you are ok with the overwriting thing, I'd do a hack which is to wash the local span into a server span. I know it is hairy, but it is more honest about the situation with the current design and could buy you time.

Something like..

```java
Span currentLocal = localThreadBinder.getCurrentSpan();
if (currentLocal != null) { // our parent is a local span
   // Brave's thread state management is fixed, as opposed to a stack.
   // Until we have support for nesting, place the span's parent where it is expected
   // See https://github.com/openzipkin/brave/issues/166
   serverTracer.setCurrentSpan(currentLocal.traceId....
}
```

---
longer details

While subtle, this overwrites a parent on act of creating a child. For example, this natural part of nesting flow wouldn't work:

```java
localTracer.startNewSpan("component", "span1");
localTracer.startNewSpan("component", "span2"); // span2 overwrites span1
span.finishSpan(); // span2 finishes
span.finishSpan(); // there's no span1, so it cannot be reported.
```

It is tempting to say, just use a stack! Probably a stack or some linked data structure is going to be the answer. However, we can't do that in a minor release. `LocalSpanState` and its implementation `ThreadLocalServerClientAndLocalSpanState` have no means to manage a stack.

We had a ton of complaints about breaking api, and when we do, we should make it count.

michaelsembwever:
Closing, as implementing nesting of local spans in Brave requires significant more thought, as is evident by Adrian's comments above and #166

PR #218, opened 2016-08-14
OpenTracing To Zipkin via Brave Bridge

michaelsembwever:
Design:
 - All spans that are explicitly created are done so with the local tracer.
 - CS+CR and SR+SS annotations are done around the inject and extract methods.
 - There is a mismatch between the two APIs in how state of the current span is held. OpenTracing does not hold any such state and passes in the
current span intended to be a parent is an explicit action in the API, while in Brave this state is known and such action not required. OpenTracing's
expectations are honoured here and Brave's internal state is overridden as needed.
 - It is noted that it probably would have been simpler to have implemented against lower APIs in brave, like directly against the collector and spans,
than the top level Brave api.

ref: https://github.com/opentracing/opentracing-java/commit/99a0c6add9b769cb54693756cfc58b6817205b84

### DIFF
--- a/brave-opentracing/.gitignore
+++ b/brave-opentracing/.gitignore
@@ -1,0 +1,14 @@
+*~
+#*
+*#
+.#*
+.factorypath
+.classpath
+.project
+.settings/
+target/
+_site/
+.idea
+*.iml
+*.swp
+/opentracing-brave/nbproject/
--- a/brave-opentracing/.settings.xml
+++ b/brave-opentracing/.settings.xml
@@ -1,0 +1,44 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<!--
+
+    Copyright 2016 The OpenZipkin Authors
+
+    Licensed under the Apache License, Version 2.0 (the "License"); you may not use this file except
+    in compliance with the License. You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+    Unless required by applicable law or agreed to in writing, software distributed under the License
+    is distributed on an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express
+    or implied. See the License for the specific language governing permissions and limitations under
+    the License.
+
+-->
+<settings xmlns="http://maven.apache.org/SETTINGS/1.0.0"
+      xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
+      xsi:schemaLocation="http://maven.apache.org/SETTINGS/1.0.0
+                          http://maven.apache.org/xsd/settings-1.0.0.xsd">
+  <servers>
+    <server>
+      <id>sonatype</id>
+      <username>${env.SONATYPE_USER}</username>
+      <password>${env.SONATYPE_PASSWORD}</password>
+    </server>
+    <server>
+      <id>bintray</id>
+      <username>${env.BINTRAY_USER}</username>
+      <password>${env.BINTRAY_KEY}</password>
+    </server>
+    <server>
+      <id>jfrog-snapshots</id>
+      <username>${env.BINTRAY_USER}</username>
+      <password>${env.BINTRAY_KEY}</password>
+    </server>
+    <server>
+      <id>github.com</id>
+      <username>${env.GH_USER}</username>
+      <password>${env.GH_TOKEN}</password>
+    </server>
+  </servers>
+</settings>
+
--- a/brave-opentracing/README.md
+++ b/brave-opentracing/README.md
@@ -1,0 +1,10 @@
+# OpenTracing API for Java
+
+This library is a Java Bridge for the API to OpenTracing.
+
+## Required Reading
+
+In order to understand OpenTracing API, one must first be familiar with
+the [OpenTracing project](http://opentracing.io) and
+[terminology](http://opentracing.io/spec/) more generally.
+
--- a/brave-opentracing/pom.xml
+++ b/brave-opentracing/pom.xml
@@ -1,0 +1,110 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<!--
+
+    Copyright 2016 The OpenZipkin Authors
+
+    Licensed under the Apache License, Version 2.0 (the "License"); you may not use this file except
+    in compliance with the License. You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+    Unless required by applicable law or agreed to in writing, software distributed under the License
+    is distributed on an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express
+    or implied. See the License for the specific language governing permissions and limitations under
+    the License.
+
+-->
+<project xmlns="http://maven.apache.org/POM/4.0.0"
+        xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
+        xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 http://maven.apache.org/xsd/maven-4.0.0.xsd">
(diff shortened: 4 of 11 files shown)


michaelsembwever:
@adriancole

brave-opentracing/src/main/java/io/opentracing/BraveSpanBuilderImpl.java
```diff
@@ -0,0 +1,142 @@
...
+
+import java.util.Optional;
+
+import com.github.kristofa.brave.Brave;
+import com.github.kristofa.brave.IdConversion;
+import com.github.kristofa.brave.ServerTracer;
+import com.github.kristofa.brave.http.BraveHttpHeaders;
+
+
+final class BraveSpanBuilderImpl extends AbstractSpanBuilder {
```
adriancole: add your notes from the PR description here?

brave-opentracing/README.md
```diff
@@ -0,0 +1,10 @@
+# OpenTracing API for Java
+
+This library is a Java Bridge for the API to OpenTracing.
+
```
adriancole: I'd put in a sub-bullet that the status of this is experimental. Sadly we don't have an "experimental" or "beta" annotation in brave at the moment.

brave-opentracing/src/main/java/io/opentracing/BraveTracerImpl.java
```diff
@@ -0,0 +1,92 @@
...
+
+import com.github.kristofa.brave.Brave;
+import com.github.kristofa.brave.IdConversion;
+import com.github.kristofa.brave.SpanId;
+import com.github.kristofa.brave.http.BraveHttpHeaders;
+
+import io.opentracing.propagation.Format;
+import java.util.HashMap;
+
+public final class BraveTracerImpl extends AbstractTracer {
```
adriancole: Since there's no BraveTracer interface, I think it would be nicer to name this class BraveTracer instead of BraveTracerImpl. Similarly with other types in this PR. We don't tend to follow the *Impl naming convention here unless there's actually a name conflict.

adriancole: This is the only public type (I think) it would be good to put a code example here and also in the README, rather than defer completely to the opentracing website. You can look at LocalTracer for a good example of an explained class.

brave-opentracing/src/main/java/io/opentracing/BraveTracerImpl.java
```diff
@@ -0,0 +1,92 @@
...
+                && null != builder.parentSpanId
+                && null != builder.operationName) {
+
+            brave.serverTracer().setStateCurrentTrace(
+                    builder.traceId,
+                    builder.parentSpanId,
+                    null,
+                    builder.operationName);
+
+            brave.serverTracer().setServerReceived();
```
adriancole: made this comment in the PR that led to this, but the contract of `extract` doesn't imply adding annotations, while this does. For example, if we implement messaging span like we have in the Kafka PR, this becomes invalid.

I can say objectively that this design, while possibly nice for experiments, is unlike all of the other open tracing variants, even the variant in openzipkin's org.

https://github.com/openzipkin/zipkin-go-opentracing

cc @bazvanbeek

For reasons like this, I think we should consider this design an experiment that might be completely removed, changed or obviated by a more semantically implemented variant here or in zipkin-java-opentracing

brave-opentracing/.gitignore
```diff
@@ -0,0 +1,14 @@
...
+.factorypath
+.classpath
+.project
+.settings/
+target/
+_site/
+.idea
+*.iml
+*.swp
+/opentracing-brave/nbproject/
```
adriancole: I don't think this needs its own gitignore.. the top-level one should work

brave-opentracing/.settings.xml
```diff
@@ -0,0 +1,44 @@
...
+      <username>${env.BINTRAY_USER}</username>
+      <password>${env.BINTRAY_KEY}</password>
+    </server>
+    <server>
+      <id>github.com</id>
+      <username>${env.GH_USER}</username>
+      <password>${env.GH_TOKEN}</password>
+    </server>
+  </servers>
+</settings>
```
adriancole: the top-level settings will deploy this.. this file isn't needed.

brave-opentracing/pom.xml
```diff
@@ -0,0 +1,110 @@
...
+        <project.build.resourceEncoding>UTF-8</project.build.resourceEncoding>
+
+        <main.basedir>${project.basedir}</main.basedir>
+    </properties>
+
+    <dependencies>
+        <dependency>
+            <groupId>io.opentracing</groupId>
+            <artifactId>opentracing-impl-java8</artifactId>
+            <version>0.14-SNAPSHOT</version>
```
adriancole: please point a TODO for the pinned version, and a snapshot repository section with a TODO to remove it. These will need to be removed before we merge this.

brave-opentracing/pom.xml
```diff
@@ -0,0 +1,110 @@
...
+    <properties>
+        <project.build.resourceEncoding>UTF-8</project.build.resourceEncoding>
+
+        <main.basedir>${project.basedir}</main.basedir>
+    </properties>
+
+    <dependencies>
+        <dependency>
+            <groupId>io.opentracing</groupId>
+            <artifactId>opentracing-impl-java8</artifactId>
```
adriancole: please use the java7 variant as brave users require java7 code (some have even asked for 6!)

brave-opentracing/pom.xml
```diff
@@ -0,0 +1,110 @@
...
+            <artifactId>mockito-all</artifactId>
+            <version>1.10.19</version>
+            <scope>test</scope>
+        </dependency>
+    </dependencies>
+
+    <build>
+        <plugins>
+            <!-- Ensure main source tree compiles to Java 8 bytecode.    -->
+            <plugin>
```
adriancole: we can remove this once we revert to the java7 variant

adriancole:
summary of major changes needed pre-merge
* revert to java7 opentracing as there's no need to make this expressly incompatible with other brave interfaces
* move to a released version of opentracing so master isn't subject to snapshot drift
* (please) use shorter names when non-ambiguous ex. BraveTracer not BraveTracerImpl etc

Also, I don't think this is a sustainable design, as noted before and also here. It would be best that this design is used by users besides the author. I think @felixbarny asked about brave and opentracing, and perhaps @bazvanbeek.

As mentioned before a much lower maintenance design would be to not try and bind Brave's implicit propagation into OpenTracing which defines no implicit propagation. That would eliminate the problems of needing to do tricks like assume all extracted spans need a "sr" annotation. Personally, I'd be more in favor of helping with maintenance on that type of code, as that design doesn't make other work much harder (like changing how the thread state operates or adding messaging support).

brave-opentracing/src/main/java/io/opentracing/BraveTracerImpl.java
adriancole: mispelled cc @basvanbeek

adriancole:
also cc @schlosna @jkozlowski who were the other two people I recall who asked about OpenTracing. too bad we don't have an issue on this, yet, as it would be less hear-say about who might be interested (shrug)

brave-opentracing/pom.xml
michaelsembwever: this code relies on java8

so it either has to go into its own repository, or brave-opentracing is only built when Java8 is used.
i'm keen to see the later, but it's dependent on the releases being built with Java8 (even though the rest of the reactor is Java7 compatible).

michaelsembwever:
Code is instead at https://github.com/openzipkin/opentracing-brave

PR #219, opened 2016-08-15
Adds InheritableServerClientAndLocalSpanState for multiple local spans

schlosna:
The default ServerClientAndLocalSpanState now supports inheriting server, client, and local span state across threads via InheritableThreadLocal state. The LocalTracer now utilizes a stack to maintain multiple local spans to support variable trace shapes.  

Partly addresses #166

### DIFF
--- a/brave-core/src/main/java/com/github/kristofa/brave/InheritableServerClientAndLocalSpanState.java
+++ b/brave-core/src/main/java/com/github/kristofa/brave/InheritableServerClientAndLocalSpanState.java
@@ -1,0 +1,113 @@
+package com.github.kristofa.brave;
+
+import java.util.Deque;
+import java.util.concurrent.LinkedBlockingDeque;
+
+import com.github.kristofa.brave.internal.Util;
+import com.twitter.zipkin.gen.Endpoint;
+import com.twitter.zipkin.gen.Span;
+
+/**
+ * {@link ServerClientAndLocalSpanState} implementation that keeps trace state using {@link InheritableThreadLocal}
+ * variables and provides local span inheritence from parent to children.
+ * <p>
+ * Important note: when using {@link InheritableServerClientAndLocalSpanState}, tracers must
+ * {@link LocalTracer#finishSpan() finish spans} or clear the local span at
+ * completion of the local trace span to avoid linking spans with incorrect
+ * parents and avoid leaking spans and associated memory.
+ */
+public final class InheritableServerClientAndLocalSpanState implements ServerClientAndLocalSpanState {
+
+    private final InheritableThreadLocal<ServerSpan> currentServerSpan =
+            new InheritableThreadLocal<ServerSpan>() {
+                @Override
+                protected ServerSpan initialValue() {
+                    return ServerSpan.EMPTY;
+                }
+            };
+
+    private final InheritableThreadLocal<Span> currentClientSpan = new InheritableThreadLocal<Span>();
+
+    private final InheritableThreadLocal<Deque<Span>> currentLocalSpan =
+            new InheritableThreadLocal<Deque<Span>>() {
+                @Override
+                protected Deque<Span> initialValue() {
+                    return new LinkedBlockingDeque<Span>();
+                }
+            };
+
+    private final Endpoint endpoint;
+
+    /**
+     * @param endpoint Endpoint of the local service being traced.
+     */
+    public InheritableServerClientAndLocalSpanState(Endpoint endpoint) {
+        this.endpoint = Util.checkNotNull(endpoint, "Endpoint must be specified.");
+    }
+
+    @Override
+    public ServerSpan getCurrentServerSpan() {
+        return currentServerSpan.get();
+    }
+
+    @Override
+    public void setCurrentServerSpan(final ServerSpan span) {
+        if (span == null) {
+            currentServerSpan.remove();
+        } else {
+            currentServerSpan.set(span);
+        }
+    }
+
+    @Override
+    public Endpoint endpoint() {
+        return endpoint;
+    }
+
+    @Override
+    public Span getCurrentClientSpan() {
+        return currentClientSpan.get();
+    }
+
+    @Override
+    public void setCurrentClientSpan(final Span span) {
+        currentClientSpan.set(span);
+    }
+
+    @Override
+    public Boolean sample() {
+        return getCurrentServerSpan().getSample();
+    }
+
+    @Override
+    public Span getCurrentLocalSpan() {
+        return currentLocalSpan.get().peekFirst();
+    }
+
+    /**
+     * Sets the specified local span as the active span at the top of the
+     * stack, or if the specified span is null, the top of the stack is popped.
+     *
+     * @param span Local span.
+     */
+    @Override
+    public void setCurrentLocalSpan(Span span) {
+        Deque<Span> deque = currentLocalSpan.get();
+        if (span == null) {
+            // pop to remove
+            deque.pollFirst();
+        } else {
+            deque.addFirst(span);
+        }
+    }
+
+    @Override
+    public String toString() {
+        return "InheritableServerClientAndLocalSpanState{"
+                + "endpoint=" + endpoint + ", "
+                + "currentLocalSpan=" + currentLocalSpan + ", "
+                + "currentClientSpan=" + currentClientSpan + ", "
+                + "currentServerSpan=" + currentServerSpan
+                + "}";
+    }
+}
--- a/brave-core/src/main/java/com/github/kristofa/brave/LocalTracer.java
+++ b/brave-core/src/main/java/com/github/kristofa/brave/LocalTracer.java
@@ -1,6 +1,7 @@
 package com.github.kristofa.brave;
 
 import com.github.kristofa.brave.SpanAndEndpoint.LocalSpanAndEndpoint;
+import com.github.kristofa.brave.internal.Nullable;
 import com.google.auto.value.AutoValue;
 import com.twitter.zipkin.gen.BinaryAnnotation;
 import com.twitter.zipkin.gen.Span;
@@ -51,6 +52,8 @@ public abstract class LocalTracer extends AnnotationSubmitter {
 
     abstract SpanCollector spanCollector();
 
+    abstract boolean allowNestedLocalSpans();
+
     abstract Sampler traceSampler();
 
     @AutoValue.Builder
@@ -62,9 +65,12 @@ public abstract class LocalTracer extends AnnotationSubmitter {
 
         abstract Builder spanCollector(SpanCollector spanCollector);
 
+        abstract Builder allowNestedLocalSpans(boolean allowNestedLocalSpans);
+
         abstract Builder traceSampler(Sampler sampler);
 
         abstract LocalTracer build();
+
     }
 
     /**
@@ -83,11 +89,40 @@ public abstract class LocalTracer extends AnnotationSubmitter {
     }
 
     private SpanId getNewSpanId() {
-        Span parentSpan = spanAndEndpoint().state().getCurrentServerSpan().getSpan();
+        Span parentSpan = getNewSpanParent();
         long newSpanId = randomGenerator().nextLong();
         SpanId.Builder builder = SpanId.builder().spanId(newSpanId);
         if (parentSpan == null) return builder.build(); // new trace
         return builder.traceId(parentSpan.getTrace_id()).parentId(parentSpan.getId()).build();
+    }
+
+    /**
+     * Request the span that should be considered the new span's parent.
+     *
+     * If {@link #allowNestedLocalSpans()} is enabled, the new span's parent
+     * will be the current local span (if one exists).
+     *
+     * If nested local spans is not enabled or there is no current local span,
+     * the new span's parent will be the current server span (if one exists).
+     *
+     * @return span that should be the new span's parent, or null if one does not exist.
+     */
+    @Nullable
+    Span getNewSpanParent() {
+        ServerClientAndLocalSpanState state = spanAndEndpoint().state();
+        Span parentSpan = null;
+        if (allowNestedLocalSpans()) {
+            parentSpan = state.getCurrentLocalSpan();
+        }
+
+        if (parentSpan == null) {
+            ServerSpan currentServerSpan = state.getCurrentServerSpan();
+            if (currentServerSpan != null) {
+                parentSpan = currentServerSpan.getSpan();
+            }
+        }
+
+        return parentSpan;
     }
 
     /**
@@ -141,7 +176,7 @@ public abstract class LocalTracer extends AnnotationSubmitter {
         } else {
             duration = currentTimeMicroseconds() - span.getTimestamp();
         }
-        finishSpan(duration);
+        internalFinishSpan(span, duration);
     }
 
     /**
@@ -151,6 +186,10 @@ public abstract class LocalTracer extends AnnotationSubmitter {
         Span span = spanAndEndpoint().span();
         if (span == null) return;
 
+        internalFinishSpan(span, duration);
+    }
+
+    private void internalFinishSpan(Span span, long duration) {
         synchronized (span) {
             span.setDuration(duration);
             spanCollector().collect(span);
--- a/brave-core/src/main/java/com/github/kristofa/brave/Brave.java
+++ b/brave-core/src/main/java/com/github/kristofa/brave/Brave.java
@@ -38,6 +38,7 @@ public class Brave {
         private Random random = new Random();
         // default added so callers don't need to check null.
         private Sampler sampler = Sampler.create(1.0f);
+        private boolean allowNestedLocalSpans = false;
 
         /**
          * Builder which initializes with serviceName = "unknown".
@@ -86,6 +87,10 @@ public class Brave {
          */
         public Builder(ServerClientAndLocalSpanState state) {
             this.state = Util.checkNotNull(state, "state must be specified.");
+
+            // the legacy span state doesn't support nested spans per (#166). Only permit nesting on the span
+            // state that has instructions on how to use it properly
+            this.allowNestedLocalSpans = state instanceof InheritableServerClientAndLocalSpanState;
         }
 
         /**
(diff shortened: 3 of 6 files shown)


schlosna:
FYI @adriancole and @kristofa 

This is an initial stab at #166 to allow a stack of local traces, as well as local trace inheritance to child threads.

adriancole:
Thanks for raising this PR @schlosna! So, here's my thought..

The goal seems to be merge this change in Brave 3.x, which allows people with existing Brave usage to get nested local spans. IOTW, this PR buys us time until Brave 4, where we'd ideally drop the entire type chain for more explicitness.

I'd love to get this in, but we need to change some things I think. Particularly, we need the new behavior to be "opt-in" vs overwriting the old implementation. Here are some thoughts on how to get this in without people complaining we caused problems :)

### separate out the fixes from the new implementation.

I noticed you added some tests and fixed some things. if you can, separate out that commit so it is easier to tell what changed (even raise as a separate PR if you like and I can merge it immediately)

### Make the new thread storage optional, and add caveats
This is actually a significant behaviour change, especially the inheritable thread local. Brave currently is consistent about not being inheritable. I'd personally rather be consistent than inconsistent. Ex People who are currently explicitly managing brave state to do so differently only for local spans and not for other spans, right?

Also, in practice we can't guarantee people close span state, and they won't blow up memory if they don't. In current code, the "span leak" problem is limited to a single span, as there's only one holder. Now, we can have a leak due to the queue. In the javadoc for the alternate thread state thing, I'd mention this as it is very important for people to know. (ex that span.finish is a must to avoid a memory leak).

schlosna:
Thanks @adriancole, I can definitely split into smaller commits & PRs, just wanted to get something pushed as I wrapped up last week's project with changes across a variety of libraries, and wanted to get feedback.

I would like to get a change into Brave 3.x that supports multiple local traces. I think its feasible to have something like the `InheritableServerClientAndLocalSpanState` which supports the local span stack, while leaving existing state implementations as-is with `O(1)` memory bounds.

In regards to span leak, I saw https://github.com/openzipkin/brave-opentracing seems to support `AutoCloseable` spans. Another option is a method in `LocalTracer` along the lines of `<T> T trace(Callable<T>)` (and `Runnable` equivalent) that would wrap to ensure proper finishing of spans (when not crossing complex method/async thread boundaries), and becomes much cleaner with Java 8.

adriancole:
@schlosna your plan sounds good on `InheritableServerClientAndLocalSpanState` (esp if all of the thread locals are consistently inheritable as that makes docs easier). Seems tempting to call out that it supports nested spans in the name since that's the cardinal feature. The current design makes naming things based on features pretty .. uh.. less than ideal :)

On the future plans, I think that we really ought to kill the spanstate classes for a single propagable approach. there's nothing about the type of span that requires us to propagate it differently once we have a decent in-process propagator.

OpenTracing Span doesn't include in-process propagation in its definition, and in fact it is explicitly out of scope, so we cannot borrow from that when considering what to do here. The autoclosable there is about finishing the timer and reporting the span. People layer in-process propagation on top of OpenTracing Span in all designs except the Brave bridge experiment.

The design you mention about a traced lambda is how finagle and also zipkin-js work. I really like this design and think we could make a replacement tracer that is consistent and provides for no leaks. One of the struggles is that we tend to have design discussions about #166 on loosely related issues. This makes it hard to keep track of key insights.

Let's see about finishing this issue, which is an evolutionary step as a separate effort than fixing the context propagation story for Brave 4. Sound good?

brave-core/src/main/java/com/github/kristofa/brave/LocalTracer.java
```diff
@@ -129,19 +132,21 @@ public SpanId startNewSpan(String component, String operation, long timestamp) {
...
         if (span == null) return;
 
-        Long startTick = span.startTick;
-        final long duration;
-        if (startTick != null) {
-            duration = Math.max(1L, (endTick - startTick) / 1000L);
+        internalFinishSpan(span, getDuration(span));
+    }
+
+    private long getDuration(Span span) {
```
adriancole: make static and test? makes the extra navigation hit harder when we add quality control

adriancole: ps do we actually need this change? timestamp is guaranteed to be set at construction time if using LocalTracer..

brave-core/src/main/java/com/github/kristofa/brave/LocalTracer.java
```diff
@@ -83,7 +83,10 @@ public SpanId startNewSpan(String component, String operation) {
     }
 
     private SpanId getNewSpanId() {
-        Span parentSpan = spanAndEndpoint().state().getCurrentServerSpan().getSpan();
+        Span parentSpan = spanAndEndpoint().state().getCurrentLocalSpan();
```
adriancole: this is dodgy with the default state backend because people may not have cleared their local span and all of a sudden the traces start having random local spans as their parent. I'd make an internal flag that's set when the state store is InheritableServerClientAndLocalSpanState. Only when that's set accept a local parent.

While seems yakky, this will insure existing users against the behavior change I mentioned. It is super confusing to try and debug spans that have the wrong parent, speaking from experience!

adriancole:
Looks good with two overall notes:

* there's a new synchronized block on reading span.getTimestamp() which appears like racing stripes as in the context of a local span, that's always set. I'd remove it from this change.
* guard the special-casing of nested local spans on the impl that supports it, `InheritableServerClientAndLocalSpanState`, and make sure the javadoc there highlight how very important it is to not leak spans!

adriancole:
basically, on span.getTimestamp() in local spans, it is effectively final as it is set and never changed, so we should be able to reduce the confusion around span.timestamp. If it isn't due to circumstances, I'd almost rather make it final :)

ex.
```java
static final class LocalSpan extends Span {
  final long timestamp;
  final long startTick;
  -- set in ctor and override getter --
}
```

Doing something like this would remove the need to even test concurrency

brave-core/src/main/java/com/github/kristofa/brave/LocalTracer.java
schlosna: I don't want to break back-compat here even if folks having properly finished their spans. I'm curious if folks intentionally do not finish spans or if it is simply due to mistakes. If the former, is that as an intentional filter of sorts? If the latter, I'd probably want to know as that means you're silently losing spans (though you'd see gaps in your traces).

Curious about your thoughts at implementation approach here. I could do an instanceof check on the state to see if its `InheritableServerClientAndLocalSpanState` and only then inherit the local parent span, but as you mentioned that feels really brittle. Slightly better might be an interface that state implementations implement to control their parent span inheritence, something along the lines of:

```
public interface InheritableState {
    /**
     * @return the current parent span to inherit, or null if no parent span should be inherited
     */
    @Nullable
    Span getCurrentParentSpan();
}
```

adriancole: one of the joys of having hidden constructors is we can change them :)

For example, LocalTracer's Builder is hidden. You can add a `Builder allowNestedSpans(boolean allowNestedSpans)` internally to brave, and the only thing that would create it is Brave's instantiation logic who can make an instanceof check. I'd put a TODO referencing #166

adriancole: ex. 

```java
// the legacy span state doesn't support nested spans per (#166). Only permit nesting on the span
// state that has instructions on how to use it properly
boolean nestedSpansAllowed = builder.state instanceof InheritableServerClientAndLocalSpanState;
localTracer = LocalTracer.builder()
        .randomGenerator(builder.random)
        .spanCollector(builder.spanCollector)
        .allowNestedSpans(nestedSpansAllowed)
        .spanAndEndpoint(SpanAndEndpoint.LocalSpanAndEndpoint.create(builder.state))
        .traceSampler(builder.sampler).build();
```

adriancole: Again, users currently don't need to close spans in any of the tracers. For example, they can report them manually, and there's no instructions against that. The key here is to say on the javadocs of the new type that you *must* finish spans, or at least clear the span state. It must also clarify that clearing the span state expects to pop the previous span. For example, right now, it is harmless to clear the span state multiple times. In the new impl, clearing has pop semantics.

Bottom line, I think it is a very clever way to accomplish the goal in Brave 3, and it is good to have a workaround, but we have to be very careful with it as the state store api in Brave 3 wasn't made for stack semantics. By guarding on this, we say "proceed at your own risk" which certainly won't break others.

schlosna: Sounds good, I'll get back to this in about a week.

brave-core/src/main/java/com/github/kristofa/brave/InheritableServerClientAndLocalSpanState.java
```diff
@@ -0,0 +1,123 @@
...
+
+    private final InheritableThreadLocal<ServerSpan> currentServerSpan =
+            new InheritableThreadLocal<ServerSpan>() {
+                @Override
+                protected ServerSpan initialValue() {
+                    return ServerSpan.EMPTY;
+                }
+            };
+
+    private final InheritableThreadLocal<Span> currentClientSpan = new InheritableThreadLocal<>();
```
spmason: The `<>` here isn't compatible with the language-level of this module (Java 6) on the current `master` branch

schlosna: Yeah, I'll rebase this and make it Java 6 compatible next week.

gavin-hu:
hi @adriancole, which version will include this feature,  i am using this feature  : )

adriancole:
just needs some polish, then merge time. 3.11 is probably less than a week
away

adriancole:
@schlosna ping.. back? :)

schlosna:
I'm back, will push updates to address comments in a few hours when I'm fully awake.

adriancole:
hehe thanks for thinking to do that while awake. sleep-coding can be
dangerous.. speaking from experience!

schlosna:
Hey @adriancole, I pushed updates to hit most remaining comments. I think the JavaDoc could still use some love, will do another pass on that in a bit, but interested for feedback on current state of PR.

I went with a combination approach of `InheritableState` interface to allow state to specify how it picks the parent span, as well as `allowNestedLocalSpans` on Brave builder to allow disabling inheritance if desired or when not supported (for back compat).

brave-core/src/main/java/com/github/kristofa/brave/InheritableServerClientAndLocalSpanState.java
```diff
@@ -0,0 +1,137 @@
...
+                @Override
+                protected Deque<Span> initialValue() {
+                    return new LinkedBlockingDeque<Span>();
+                }
+            };
+
+    private final Endpoint endpoint;
+
+    /**
+     * Constructor
```
adriancole: heheh is it? really? :) let's leave out filler javadocs

brave-core/src/main/java/com/github/kristofa/brave/InheritableServerClientAndLocalSpanState.java
```diff
@@ -0,0 +1,137 @@
...
+     *
+     * @param ip Int representation of ipv4 address.
+     * @param port port on which current process is listening.
+     * @param serviceName Name of the local service being traced. Should be lowercase and not <code>null</code> or empty.
+     */
+    public InheritableServerClientAndLocalSpanState(int ip, int port, String serviceName) {
```
adriancole: switch to Endpoint instead. That way people can use ipv6 etc and we don't need to copy/paste a bunch of things

brave-core/src/main/java/com/github/kristofa/brave/InheritableServerClientAndLocalSpanState.java
```diff
@@ -0,0 +1,137 @@
...
+        Util.checkNotBlank(serviceName, "Service name must be specified.");
+        endpoint = Endpoint.create(serviceName, ip, port);
+    }
+
+    /**
+     * {@inheritDoc}
```
adriancole: no filler javadocs por favor.. make each line count!

brave-core/src/main/java/com/github/kristofa/brave/InheritableServerClientAndLocalSpanState.java
```diff
@@ -0,0 +1,137 @@
...
+    public Boolean sample() {
+        return getCurrentServerSpan().getSample();
+    }
+
+    @Override
+    public Span getCurrentLocalSpan() {
+        return currentLocalSpan.get().peekFirst();
+    }
+
+    @Override
```
adriancole: here's where javadoc is helpful. you can mention the pop logic and anything to be careful of.

brave-core/src/main/java/com/github/kristofa/brave/InheritableServerClientAndLocalSpanState.java
```diff
@@ -0,0 +1,137 @@
...
+            // pop to remove
+            deque.pollFirst();
+        } else {
+            deque.addFirst(span);
+        }
+    }
+
+    @Override
+    public Span getCurrentParentSpan() {
+        // inherit previous local span if available
```
adriancole: this should be promoted to javadoc. For example, "Returns a local or server parent, or null"

brave-core/src/main/java/com/github/kristofa/brave/InheritableState.java
```diff
@@ -0,0 +1,16 @@
+package com.github.kristofa.brave;
+
+import com.github.kristofa.brave.internal.Nullable;
+import com.twitter.zipkin.gen.Span;
+
+/**
+ * Interface defining the behavior for inheriting state from {@link ServerClientAndLocalSpanState} implementations.
+ */
+public interface InheritableState {
```
adriancole: hmm... let's not make this public. the current approach is a hack, and if we are adding a new public interface it should be something that allows us to use a real context like the recently extracted and dependency free https://github.com/grpc/grpc-java/blob/master/context/src/main/java/io/grpc/Context.java

brave-core/src/main/java/com/github/kristofa/brave/LocalTracer.java
```diff
@@ -62,9 +65,22 @@ static Builder builder(LocalTracer source) {
...
         abstract Builder traceSampler(Sampler sampler);
 
-        abstract LocalTracer build();
+        abstract LocalTracer autoBuild();
+
+        LocalTracer build() {
+            LocalTracer tracer = autoBuild();
+            if (tracer.allowNestedLocalSpans() && !(tracer.spanAndEndpoint().state() instanceof InheritableState)) {
+                throw new IllegalStateException(
+                        String.format("Cannot allowNestedLocalSpans for state %s that does not implement %s",
```
adriancole: this is over abstraction. Please remove this interface and use only the bridge class you initially defined. I don't want to dig a deeper hole by now making it possible to customize a hack!!

adriancole:
more features more problems. We can barely maintain brave as it is. Please
focus feature work on the future (ex a real context), and resist the
temptation to add extension points to the hack. I don't want us to get
swamped in maintenance or have to now consider how to port this interface
when we define a correct model for context state.

brave-core/src/main/java/com/github/kristofa/brave/InheritableServerClientAndLocalSpanState.java
schlosna: done

brave-core/src/main/java/com/github/kristofa/brave/InheritableServerClientAndLocalSpanState.java
schlosna: done

brave-core/src/main/java/com/github/kristofa/brave/InheritableServerClientAndLocalSpanState.java
schlosna: done

brave-core/src/main/java/com/github/kristofa/brave/InheritableServerClientAndLocalSpanState.java
schlosna: updated JavaDoc

brave-core/src/main/java/com/github/kristofa/brave/InheritableServerClientAndLocalSpanState.java
schlosna: this logic moved to `com.github.kristofa.brave.LocalTracer#getNewSpanParent` and updated JavaDoc

brave-core/src/main/java/com/github/kristofa/brave/InheritableState.java
schlosna: removed this interface completely

brave-core/src/main/java/com/github/kristofa/brave/LocalTracer.java
schlosna: done

schlosna:
Pushed cleanup removing `InheritableState` and addressing other comments.

adriancole:
LOOKS GOOD!

So, I think after we get usage practice with this, we'll want to update javadoc and maybe add instructions to the root README. For now, let's get this out there. Nice work

adriancole:
When you are ready, please squash into a commit with a message you are happy with. I'll merge when you say so.

schlosna:
Thanks @adriancole I squashed and this should be good to go barring any build flakes.

schlosna:
`./mvnw clean install` builds successfully locally for me, but Travis failed twice in brave-grpc:

```
Tests run: 5, Failures: 1, Errors: 0, Skipped: 0, Time elapsed: 1.116 sec <<< FAILURE! - in com.github.kristofa.brave.grpc.BraveGrpcInterceptorsTest
testAsyncUnaryCall(com.github.kristofa.brave.grpc.BraveGrpcInterceptorsTest)  Time elapsed: 0.019 sec  <<< FAILURE!
org.junit.ComparisonFailure: expected:<[2]> but was:<[1]>
```

adriancole:
toying with another branch
https://travis-ci.org/openzipkin/brave/builds/158370900

On Thu, Sep 8, 2016 at 11:53 AM, David Schlosnagle <notifications@github.com

adriancole:
cherry-picked. thanks!

PR #221, opened 2016-08-16
Synchronize span annotation reads/writes

schlosna:
Address possible race condition with inconsistent Span synchronization.


adriancole:
thanks again again!

PR #222, opened 2016-08-16
Consolidate default Brave state construction

schlosna:


adriancole:
this one doesn't look to me as an overall win. It increases navigation significantly only to avoid calling a constructor twice.

schlosna:
I'm fine not merging this, it was mainly for my own convenience as I tested a few different thoughts for the state implementations and change a single location.

adriancole:
thanks for the consideration!

PR #223, opened 2016-08-16
Use ServerSpan.EMPTY constant for initial server span

schlosna:


brave-core/src/main/java/com/github/kristofa/brave/ThreadLocalServerClientAndLocalSpanState.java
```diff
@@ -17,7 +17,7 @@
 
         @Override
         protected ServerSpan initialValue() {
-            return ServerSpan.create(null);
+            return ServerSpan.EMPTY;
```
adriancole: cool especially as it is immutable

Issue #225, opened 2016-08-17
Make sure that Brave and all dependencies are compatible with Java 1.6  


ivansenic:
At the moment classes in Brave are compiled with the target 1.7 which makes library not working on any JVM under 1.7. Now I understand that depending on Java 1.6 is limiting programming options in Brave itself, but with this you are turning away all the people that are still running on Java 1.6 or lower. And I am sure that there are numerous enterprise systems that still use old Java versions (I know several big systems still going with 1.5 :-1: ).

adriancole:
This is a regression. I accidentally screwed this up. Will fix

adriancole:
Here's the thing I broke https://github.com/openzipkin/brave/pull/131

adriancole:
OK this requires extracting a jar in zipkin upstream with only the model and constants. This may be tricky due to Codec stuff.

adriancole:
https://github.com/openzipkin/brave/pull/227 on the way!

adriancole:
very sorry about breaking this. 3.10.0 is on the way to maven central now where this is fixed.

Issue #150, opened 2016-03-29
Feature request: a way to get traceId into an SLF4j MDC


ryangardner:
We've got a lot of log aggregation software already, and it'd be very powerful if we could correlate traceIds in our logs to specific requests in Zipkin (see a slow request in Zipkin? go search for the traceId in our log aggregation and see any messages that tie into it) 

It'd be nice to be able to put the traceId into the MDC every time it is set and clear it out whenever the request is over (or the current one is set to nothing) then clear it out. 

I think in order for this to work some kind of hook needs to be exposed.

I can do it myself with AOP and intercepting calls to the `ServerClientAndLocalSpanState` - but I think that other people would like this functionality as well (It's one of the first things people ask me about in my company when I show brave / zipkin to them)

If you can think of the approach you would prefer for this, I can take a first stab at implementing it and submit a pull request with the changes

adriancole:
relates to https://github.com/openzipkin/brave/pull/110 if you don't mind reading up on that.

ryangardner:
Switching out the spanId generator would make another place to work with - but wouldn't necessarily let you know when the spans are created / ended, but a couple other ways that come to mind to handle this are: 

* Make it possible to decorate behavior of the the ```setStateCurrentTrace``` ```setStateCurrentTrace(..)``` and ```clearCurrentSpan()``` of the ServerTracer methods to have calls to those methods set / unset things in an MDC (this one seems like it would be the least disruptive option)
* Some kind of event system where you can register listeners with the ServerTracer to be notified of trace start / trace end events (this seems like it'd be both overkill and add an unnecessary level of complexity for just this one use case)

adriancole:
ahh I remember this. yeah it is a bit tricky as lifecycle of a span
aren't abstracted 100% (ex we have 3 tracers at the moment). One thing
to add to the toolbox is SpanCollector.collect() is definitely the
end.

spring-cloud-sleuth has an implementation like what you're
mentioning.. I think its scope is a bit broader as it requires ids
even when the trace isn't sampled. anyway prior art is often helpful.
https://github.com/spring-cloud/spring-cloud-sleuth/blob/master/spring-cloud-sleuth-core/src/main/java/org/springframework/cloud/sleuth/log/Slf4jSpanLogger.java

Other interest is opentracing's span is like a tracer, and seems to
have end callbacks. how this goes across threads etc is as yet
undefined. in case you'd like a look
https://github.com/opentracing/opentracing-java/pull/25

[...]

PR #226, opened 2016-08-17
Adds slf4j logging span collectors

schlosna:
Adds 2 implementations of `SpanCollector` for logging purposes:

* `AsyncSlf4jLoggingSpanCollector` collects spans and asynchronously logs to an slf4j logger at INFO level.
* `SynchronousSlf4jLoggingSpanCollector` collects spans and synchronously logs to an slf4j logger at INFO level.

### DIFF
--- a/brave-spancollector-slf4j/LICENSE
+++ b/brave-spancollector-slf4j/LICENSE
@@ -1,0 +1,13 @@
+  Copyright 2016 Palantir Technologies, Inc. All rights reserved.
+
+  Licensed under the Apache License, Version 2.0 (the "License");
+  you may not use this file except in compliance with the License.
+  You may obtain a copy of the License at
+  
+      http://www.apache.org/licenses/LICENSE-2.0
+  
+  Unless required by applicable law or agreed to in writing, software
+  distributed under the License is distributed on an "AS IS" BASIS,
+  WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+  See the License for the specific language governing permissions and
+  limitations under the License.
--- a/brave-spancollector-slf4j/src/main/java/com/github/kristofa/brave/slf4j/AsyncSlf4jLoggingSpanCollector.java
+++ b/brave-spancollector-slf4j/src/main/java/com/github/kristofa/brave/slf4j/AsyncSlf4jLoggingSpanCollector.java
@@ -1,0 +1,84 @@
+/*
+ * Copyright 2016 Palantir Technologies, Inc. All rights reserved.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package com.github.kristofa.brave.slf4j;
+
+import java.io.IOException;
+import java.util.Collections;
+import java.util.LinkedHashSet;
+import java.util.List;
+import java.util.Set;
+import java.util.concurrent.BlockingQueue;
+
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+import com.github.kristofa.brave.FlushingSpanCollector;
+import com.github.kristofa.brave.SpanCollector;
+import com.github.kristofa.brave.SpanCollectorMetricsHandler;
+import com.github.kristofa.brave.internal.Util;
+import com.twitter.zipkin.gen.BinaryAnnotation;
+import com.twitter.zipkin.gen.Span;
+
+/**
+ * {@link SpanCollector} implementation that logs through SLF4J at INFO level.
+ */
+public final class AsyncSlf4jLoggingSpanCollector extends FlushingSpanCollector {
+
+    private final Logger logger;
+    private final Set<BinaryAnnotation> annotations = Collections.synchronizedSet(new LinkedHashSet<BinaryAnnotation>());
+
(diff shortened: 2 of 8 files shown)


schlosna:
@adriancole and @kristofa , I'd like to contribute these 2 slf4j span collectors that are similar to the `LoggingSpanCollector` in `brave-core`, but push directly into slf4j -- one synchronously, the other asynchronously flushing.

brave-spancollector-slf4j/src/main/java/com/github/kristofa/brave/slf4j/AsyncSlf4jLoggingSpanCollector.java
```diff
@@ -0,0 +1,84 @@
+/*
+ * Copyright 2016 Palantir Technologies, Inc. All rights reserved.
```
adriancole: note that at some point we will nuke all of these for "the OpenZipkin or Brave authors" as it is a maintenance burden

brave-spancollector-slf4j/LICENSE
```diff
@@ -0,0 +1,13 @@
+  Copyright 2016 Palantir Technologies, Inc. All rights reserved.
```
adriancole: heh.. reminds me we are overdue enabling the license header plugin.

adriancole:
it would be less maintenance to just make AsyncLoggingSpanCollector in brave core, since we aren't using any SLF4J features.

Mind just doing that? It would be one class+test vs a new module. We have to be conscious of adding modules not requested more than once, but I think the tweak of adding another logger class to core would be fine.

brave-spancollector-slf4j/src/main/java/com/github/kristofa/brave/slf4j/SynchronousSlf4jLoggingSpanCollector.java
```diff
@@ -0,0 +1,71 @@
...
+                span.addToBinary_annotations(ba);
+            }
+
+            getLogger().info(span.toString());
+        }
+    }
+
+    @Override
+    public void addDefaultAnnotation(String key, String value) {
+        getAnnotations().add(BinaryAnnotation.create(key, value, null));
```
adriancole: you can throw unsupported as this isn't called anymore, iirc. Ex none of the other new ones implement this

brave-spancollector-slf4j/src/main/java/com/github/kristofa/brave/slf4j/SynchronousSlf4jLoggingSpanCollector.java
```diff
@@ -0,0 +1,71 @@
...
+    }
+
+    public Logger getLogger() {
+        return logger;
+    }
+
+    public Set<BinaryAnnotation> getAnnotations() {
```
adriancole: why is there a getter?

brave-spancollector-slf4j/src/main/java/com/github/kristofa/brave/slf4j/SynchronousSlf4jLoggingSpanCollector.java
```diff
@@ -0,0 +1,71 @@
...
+
+    @Override
+    public void collect(Span span) {
+        Util.checkNotNull(span, "span must not be null");
+        if (getLogger().isInfoEnabled()) {
+            for (BinaryAnnotation ba : getAnnotations()) {
+                span.addToBinary_annotations(ba);
+            }
+
+            getLogger().info(span.toString());
```
adriancole: if you don't support binary annotations (which many don't), this whole class becomes not much more than a single-method, right?

brave-spancollector-slf4j/src/main/java/com/github/kristofa/brave/slf4j/AsyncSlf4jLoggingSpanCollector.java
```diff
@@ -0,0 +1,84 @@
...
+    }
+
+    private void logSpan(Span span) {
+        Util.checkNotNull(span, "span must not be null");
+        if (getLogger().isInfoEnabled()) {
+            for (BinaryAnnotation ba : annotations) {
+                span.addToBinary_annotations(ba);
+            }
+
+            getLogger().info(span.toString());
```
adriancole: Note: in the near future, brave-core will [fix the java 6 regression](https://github.com/openzipkin/brave/issues/225), which won't have a json codec by default anymore. Span.toString would be just an id string, like SpanId.toString.

adriancole:
@schlosna thanks for thinking about this and cutting a draft. Besides notes mentioned.. are you thinking of logging purposes for like debugging or for normal use (ex log scraping)? Not sure if async is just to make it easier to debug (as less overhead on the caller), or for export to logstash or something.

If you can find a couple other people who want a separate slf4j module (knowing the only difference from j.u.l is that you can accept a Logger ctor parameter), then I'd be open to merging this as a new module. Just feel we owe it to ourselves to make sure we meet rule-of-three when adding new modules.

brave-spancollector-slf4j/src/main/java/com/github/kristofa/brave/slf4j/AsyncSlf4jLoggingSpanCollector.java
adriancole: I take it back.. span.toString is still json. Managed to keep that functionality

devinsba:
We wouldnt use the logging reporter but we could contribute some more slf4j related code to a module or module tree. We have a small amount of code that @ryangardner posted in #150 to decorate the slf4j MDC with parent/trace/span ids

adriancole:
decorating in MDC is a big win. It will also steer any future in-process
propagation work I suspect, as each time a span is associated to the
thread, it should be re-associated with MDC, iirc

adriancole:
@devinsba I think your change about MDC is wanted, but is a different type of change. Mind opening or contributing to a new issue or PR?
@schlosna I'll write up a quick version of this in zipkin-reporter-java, as I think this is a better place to fan-out different logging mechanisms.

adriancole:
@schlosna I'm closing this as I believe zipkin-reporter has similar functionality, or can be extended to. If you still need this still open just ask

PR #229, opened 2016-08-26
Fix p6spy parent version

spmason:
.this is stopping me from building master

cc @kristofa


llinder:
👍 just hit this as well.

adriancole:
we need to set this to a real version prior to release

adriancole:
oh nevermind :) this is just pom version

PR #233, opened 2016-09-05
Bumps to p6spy release candidate

adriancole:


adriancole:
@ewhauser @felixbarny as this api is getting close to final, we should do another review of implementation here. I've noticed there are some new modularity features in p6spy, but haven't assessed if we could benefit from them. If you have some time, please do a sweep before we cut a release of brave.

felixbarny:
I already commented this on the original pull request: I'd suggest extending from `SimpleJdbcEventListener` as it offers more coarse grained callbacks which could help to reduce code duplication.

felixbarny:
As you are adding configuration options, the new `ServiceLoader` feature is probably not interesting for you. It offers to add a `JdbcEventListener` without adding a module.

adriancole:
@felixbarny made the change you mentioned. In the process, I noticed tests never ran. They are failing for reasons I can't dig into right now. We need this to be solid before releasing (or we should temporarily disable it from the main pom)

ewhauser:
@adriancole I'll submit a separate PR that fixes the test name so that master is back in a healthy state and this doesn't block releasing `brave` in the short term.

@felixbarny I had some issues getting all of the use cases working when using the `SimpleJdbcListener`.  I don't remember what they were at the moment, but I will take a pass at this PR once `p6spy` has a new release with p6spy/p6spy#346 available.

adriancole:
fingers crossed

adriancole:
rolled back JdbcEventListener as it tripped up all the tests

PR #236, opened 2016-09-05
Renaming BraveP6SpyTests to BraveP6SpyTest so that it runs under CI

ewhauser:
I'll add an additional comment to verify that it ran under CI once the build completes.


ewhauser:
Verified that tests are running and passing now: https://travis-ci.org/openzipkin/brave/builds/157688482#L3584

adriancole:
great. thanks!

PR #238, opened 2016-09-11
Make the AnnotationSubmitter.currentTimeMicroseconds() pluggable.

michaelsembwever:
Make the AnnotationSubmitter.currentTimeMicroseconds() pluggable.

Even in Java8 with java.time the best precision for timestamp generation is milliseconds. Providing microsecond accuracy in spans requires an implementation to do so to be plugged in.

By making the timestamp generation contextual through the AnnotationSubmitter.Clock interface microsecond accuracy can be achieved.

An example implementation to obtain microsecond accuracy is by using the DataStax CQL java driver
(cassandra-driver-core-*.jar or com.datastax.cassandra:cassandra-driver-core artifact) and using its AtomicMonotonicTimestampGenerator class like:

    AtomicMonotonicTimestampGenerator TIMESTAMP_GENERATOR = new AtomicMonotonicTimestampGenerator();
    ...
    Brave brave = new Brave
            .Builder("my_service_name")
            .clock(() -> { return TIMESTAMP_GENERATOR.next(); })
            .build();

### DIFF
--- a/brave-core/src/main/java/com/github/kristofa/brave/AnnotationSubmitter.java
+++ b/brave-core/src/main/java/com/github/kristofa/brave/AnnotationSubmitter.java
@@ -10,16 +10,43 @@ import static com.github.kristofa.brave.internal.Util.checkNotNull;
 
 /**
  * Used to submit application specific annotations.
- * 
+ *
  * @author kristof
  */
 public abstract class AnnotationSubmitter {
 
-    public static AnnotationSubmitter create(SpanAndEndpoint spanAndEndpoint) {
-        return new AnnotationSubmitterImpl(spanAndEndpoint);
+    /**
+     * This interface is used to make the implementation to AnnotationSubmitter.currentTimeMicroseconds() contextual.
+     * The clock is defined by the subclass's implementation of the `clock()` method.
+     * A DefaultClock implementation is provided that simply returns `System.currentTimeMillis() * 1000`.
+     */
+    public interface Clock {
+        /**
+         * Epoch microseconds used for {@link zipkin.Span#timestamp} and {@link zipkin.Annotation#timestamp}.
+         *
+         * <p>This should use the most precise value possible. For example, {@code gettimeofday} or multiplying
+         * {@link System#currentTimeMillis} by 1000.
+         *
+         * <p>See <a href="http://zipkin.io/pages/instrumenting.html">Instrumenting a service</a> for more.
+         */
+        long currentTimeMicroseconds();
+    }
+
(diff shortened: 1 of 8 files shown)


michaelsembwever:
Example of this in use is @ https://github.com/thelastpickle/cassandra-zipkin-tracing/commit/f22c499d6954798aeef3583c61b6dc17b3781729

adriancole:
Great idea. Mind adding javadoc to clock, at least referring to
http://zipkin.io/pages/instrumenting.html?

Also, Im sure this has been asked before.. can you find an issue here, or
maybe do a gitter search?  We like to follow rule of three and I am sure
this satisfies, but lets at least cc those who asked about it.

Here's one person: @mansu

adriancole:
Ps I am very interested in this, so +1 from me!

Reason is like you mentioned.. some use native libraries, so can get a good
timestamp while they are at it!

Cc @nicmunroe

adriancole:
relates to LocalTracer https://github.com/openzipkin/brave/issues/111

brave-core/src/main/java/com/github/kristofa/brave/AnnotationSubmitter.java
```diff
@@ -10,17 +10,27 @@
...
  * Used to submit application specific annotations.
- * 
+ *
  * @author kristof
  */
 public abstract class AnnotationSubmitter {
 
-    public static AnnotationSubmitter create(SpanAndEndpoint spanAndEndpoint) {
-        return new AnnotationSubmitterImpl(spanAndEndpoint);
+    public interface Clock {
```
adriancole: javadoc suggestion:

```java
/**
 * Epoch microseconds used for {@link zipkin.Span#timestamp} and {@link zipkin.Annotation#timestamp}.
 *
 * <p>This should use the most precise value possible. For example, {@code gettimeofday} or multiplying
 * {@link System#currentTimeMillis} by 1000.
 * 
 * <p>See <a href="http://zipkin.io/pages/instrumenting.html">Instrumenting a service</a> for more.
 */
```

michaelsembwever:
javadocs updated.
there was a +1 from @shakuzen on gitter.

adriancole:
SGTM will merge on green

nicmunroe:
Looking at the javadocs on `AtomicMonotonicTimestampGenerator` I'd be wary of actually using it in practice (I might be overly pessimistic since I've only spent a few minutes looking at it), but I agree that making the impl pluggable is the right way to go here. Individual projects are then free to shoulder the benefits and drawbacks of any given impl as they choose.

PR #240, opened 2016-09-22
Adds Endpoint.ipv6; Supports Endpoint for Brave and peer addresses

adriancole:
This adds Endpoint.ipv6 as a fixed-width byte array (16 bytes). This
formalizes Endpoint.ipv4 == 0 implying there's no ipv4 address. These
are supported in recent versions of Zipkin with the same semantics.

You can use this via the Endpoint constructor of Brave, or via tracer
hooks like ClientTracer.setClientSent or ServerTracer.setServerReceived

This also deprecates Endpoint.create(serviceName,ipv4,port) as it leads
to NullPointerExceptions

See https://github.com/openzipkin/zipkin/issues/306
See https://github.com/openzipkin/zipkin/pull/1309

### DIFF
--- a/brave-core/pom.xml
+++ b/brave-core/pom.xml
@@ -60,7 +60,15 @@
     <dependency>
       <groupId>com.twitter</groupId>
       <artifactId>finagle-core_2.11</artifactId>
-      <version>6.35.0</version>
+      <version>6.38.0</version>
+      <scope>test</scope>
+    </dependency>
+
+    <!-- To avoid java.lang.NoClassDefFoundError: StacktracePrintingMatcher -->
+    <dependency>
+      <groupId>org.hamcrest</groupId>
+      <artifactId>hamcrest-core</artifactId>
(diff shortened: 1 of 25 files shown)


adriancole:
cc @virtuald @devinsba

devinsba:
Unfortunately `ClientRequestInterceptor` will still thrown an NPE if it receives an `Endpoint` with a null port.

https://github.com/openzipkin/brave/blob/master/brave-core/src/main/java/com/github/kristofa/brave/ClientRequestInterceptor.java#L53

Which calls

https://github.com/openzipkin/brave/blob/master/brave-core/src/main/java/com/github/kristofa/brave/ClientTracer.java#L80

adriancole:
my what great eyes you have!!  much appreciated I'll hunt down references
to that and also on the ServerTracer

adriancole:
should be sorted now

Issue #158, opened 2016-04-18
Feature Request: Create a instrumentation for Feign 


rafabene:
"Feign makes writing java http clients easier".

It also supports adding Request Interceptors https://github.com/Netflix/feign#request-interceptors that could be used to instrument Feign

ewhauser:
As an alternative, you can use OkHttp with Feign and use https://github.com/openzipkin/brave/tree/master/brave-okhttp

adriancole:
ps feign isn't the easiest thing to manage right now as interceptor only
affects inbound. Ex you have to play games to catch the other side (by
using an invocation handler or hacking a decoder or the like).

rejigging this part would probably be good for
https://github.com/OpenFeign/feign and emulate how okhttp does it.

The advantage of doing this in feign level is that you can use the template
name or class name of the feign interface as the span name, which makes it
look prettier than "GET"

rafabene:
The greatest advantage for me at this moment is that Feign already provides an integration with Hystrix: https://github.com/OpenFeign/feign/tree/master/hystrix

That's why I wanted to see Feign being supported so I could easily have: Feign + Hystrix for circuit-breaker + Zipking for distributed tracing. 

The implementation that I was able to write to have this integration working is too "complex" to be easily consumed: https://github.com/redhat-helloworld-msa/hola/blob/master/src/main/java/com/redhat/developers/msa/hola/HolaResource.java#L95-L114

adriancole:
we can probably do this, but it is a little tricky, especially when
layering with things like hystrix

in the mean time, you can look at sleuth which already supports
feign+hystrix https://github.com/spring-cloud/spring-cloud-sleuth

adriancole:
In thinking this through, I think that Feign would be best as a "local span".

For example, feign's span would correlate to the method being invoked. Considering that the feign.Client implementation is traced in the case of OkHttp, we'd end up with something like this.

For example, feign's local span would have a child RPC span for each http request attempt:

```
{name: github.contributors // start when the method is invoked
  {name: get ["cs", "error"] }, // new span logged by okhttp
  {name: get ["cs", "cr"] },  // new span logged by okhttp
} // finish when the method returns
```

We could add to it later, like add annotations, etc, but the easiest way to get started is to create a span and assume the client is instrumented.

johnament:
[...]

PR #242, opened 2016-09-25
Rewrites OkHttp interceptor as BraveTracingInterceptor

adriancole:
This is an [OkHttp](https://github.com/square/okhttp) interceptor which
traces client requests. It should be applied as both an application
interceptor and as a network interceptor.

The implementation models the application request as a local span. Each
network request will be a child span. For example, if there's a redirect,
there will be one span for the application request and two child spans
for the associated network requests.

Trace identifiers of each network attempt are propagated to the server
via headers prefixed with `X-B3`. These spans are also reported out of
band, with `cs` (client sent) and `cr` (client receive) annotations,
binary annotations (tags) like `http.url`, and `sa` including the
server's ip and port.

### Configuration
Since this interceptor creates nested spans, you should use nesting-aware
span state like `InheritableServerClientAndLocalSpanState`. If using
asynchronous calls, you must also wrap the dispatcher's executor
service. Regardless, the interceptor must be registered as both an
application and network interceptor. 

Here's how to add tracing to OkHttp:
```java
brave = new Brave.Builder(new InheritableServerClientAndLocalSpanState(localEndpoint))..

// The request dispatcher uses an executor service.. wrap it!
tracePropagatingExecutor = new BraveExecutorService(
    new Dispatcher().executorService(),
    brave.serverSpanThreadBinder()
);

client = new OkHttpClient.Builder()
  .addInterceptor(tracingInterceptor)
  .addNetworkInterceptor(tracingInterceptor)
  .dispatcher(new Dispatcher(tracePropagatingExecutor));
  .build();
```

### Customizing span data
The span associated with the application request is by default named
according to the request tag, falling back to the http method name.

Ex.
```java
// This request's span name will be "get"
request = new Request.Builder()
    .url("https://myhost.com/gummybears").build();

// This request's span name will be "get-gummy-bears"
request = new Request.Builder()
    .url("https://myhost.com/gummybears")
    .tag("get-gummy-bears").build();
```

If you want to change span names or tags, you can override the default
parser.
```java
tracingInterceptor = BraveTracingInterceptor.builder(brave)
  ...
  .parser(new MyOkHttpParser()).build();
```

Be careful when customizing, particularly not to add too much data.
Large span (ex large orders of kilobytes) can be problematic and/or
dropped. Also, be careful that span names have low cardinality (ex no
embedded variables). Finally, prefer names in `zipkin.TraceKeys` where
possible, so that lookup keys are coherent.

For more information, look at our
[instrumentation docs](http://zipkin.io/pages/instrumenting.html)

### Naming the server
When calling an service that isn't traced with Zipkin, such as a cloud
service. You'll want to assign a server name so that it shows up in
Zipkin's service dependency graph. You can set this via
`BraveTracingInterceptor.Builder.serverName()`

```java
tracingInterceptor = BraveTracingInterceptor.builder(brave)
  ...
  .serverName("github").build();

request = new Request.Builder()
    .url("https://api.github.com/repos/square/okhttp/issues")
    ...
```

### DIFF
--- a/brave-okhttp/src/main/java/com/github/kristofa/brave/okhttp/BraveTracingInterceptor.java
+++ b/brave-okhttp/src/main/java/com/github/kristofa/brave/okhttp/BraveTracingInterceptor.java
@@ -1,0 +1,205 @@
+package com.github.kristofa.brave.okhttp;
+
+import com.github.kristofa.brave.Brave;
+import com.github.kristofa.brave.ClientTracer;
+import com.github.kristofa.brave.InheritableServerClientAndLocalSpanState;
+import com.github.kristofa.brave.KeyValueAnnotation;
+import com.github.kristofa.brave.LocalTracer;
+import com.github.kristofa.brave.SpanId;
+import com.github.kristofa.brave.http.BraveHttpHeaders;
+import com.twitter.zipkin.gen.Endpoint;
+import java.io.IOException;
+import java.net.InetSocketAddress;
+import java.nio.ByteBuffer;
+import java.util.List;
+import okhttp3.Connection;
+import okhttp3.Interceptor;
+import okhttp3.OkHttpClient;
+import okhttp3.Request;
+import okhttp3.Response;
+import zipkin.Constants;
+import zipkin.TraceKeys;
+
+import static com.github.kristofa.brave.IdConversion.convertToString;
+import static com.github.kristofa.brave.http.BraveHttpHeaders.Sampled;
+import static com.github.kristofa.brave.internal.Util.checkNotNull;
+
+/**
+ * This is an OkHttp interceptor which traces client requests. It should be applied as both an
+ * {@link OkHttpClient#interceptors() application interceptor} and as a {@link
+ * OkHttpClient#networkInterceptors() network interceptor}.
+ *
+ * <p>The implementation models the application request as a local span. Each network request will
+ * be a child span. For example, if there's a redirect, there will be one for the application
+ * request and two spans for the associated network requests.
+ *
+ * Trace identifiers of each network attempt are propagated to the server via headers prefixed with
+ * `X-B3`. These spans are also reported out of band, with {@link zipkin.Constants#CLIENT_SEND} and
+ * {@link zipkin.Constants#CLIENT_RECV} annotations, binary annotations (tags) like {@link
+ * TraceKeys#HTTP_URL} and {@link zipkin.Constants#SERVER_ADDR the server's ip and port}.
+ *
+ * <h3>Configuration</h3>
+ *
+ * <p>Since this interceptor creates nested spans, you should use nesting-aware
+ * span state like {@link InheritableServerClientAndLocalSpanState}. If using
+ * asynchronous calls, you must also wrap the dispatcher's executor
+ * service. Regardless, the interceptor must be registered as both an
+ * application and network interceptor.
+ *
+ * <p>Here's how to add tracing to OkHttp:
+ * <pre>{@code
+ * brave = new Brave.Builder(new InheritableServerClientAndLocalSpanState(localEndpoint))..
+ *
+ * // The request dispatcher uses an executor service.. wrap it!
+ * tracePropagatingExecutor = new BraveExecutorService(
+ *   new Dispatcher().executorService(),
+ *   brave.serverSpanThreadBinder()
+ * );
+ *
+ * client = new OkHttpClient.Builder()
+ *   .addInterceptor(tracingInterceptor)
+ *   .addNetworkInterceptor(tracingInterceptor)
+ *   .dispatcher(new Dispatcher(tracePropagatingExecutor));
+ *   .build();
+ * }</pre>
+ */
+public final class BraveTracingInterceptor implements Interceptor {
+  public static BraveTracingInterceptor create(Brave brave) {
+    return builder(brave).build();
+  }
+
+  /** Defaults to use {@link OkHttpParser} */
+  public static Builder builder(Brave brave) {
+    return new Builder(brave);
+  }
+
+  public static final class Builder {
+    Brave brave;
+    String serverName = "";
+    OkHttpParser parser = new OkHttpParser();
+
+    Builder(Brave brave) {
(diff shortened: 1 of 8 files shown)


adriancole:
I created this for use in an okhttp-specific elasticsearch impl for zipkin, but @tburch might be interested, too.

https://github.com/openzipkin/zipkin/issues/1312

adriancole:
I'm actually going to rewrite the interceptor and deprecate the old one

adriancole:
actually found a way to do what I wanted..

brave-okhttp/src/main/java/com/github/kristofa/brave/okhttp/BraveTracingInterceptor.java
```diff
@@ -0,0 +1,139 @@
...
+
+  abstract boolean preferTagAsSpanName();
+
+  public Builder toBuilder() {
+    return new AutoValue_BraveTracingInterceptor.Builder(this);
+  }
+
+  @Override
+  public Response intercept(Chain chain) throws IOException {
+    Request request = chain.request();
```
adriancole: this is the same impl as before, just copy/pasted

adriancole:
cc @jakewharton @swankjesse next change after this will take advantage of network interceptors by logging a "ws" (wire send) annotation for each network attempt.

brave-okhttp/src/test/java/com/github/kristofa/brave/okhttp/BraveTracingInterceptorTest.java
```diff
@@ -0,0 +1,285 @@
...
+    server.enqueue(new MockResponse());
+
+    HttpUrl url = server.url("foo");
+    client.newCall(new Request.Builder().url(url).build()).execute();
+
+    assertThat(collectedSpans()).containsExactly(span); // 408 one is ignored
+  }
+
+  @Test
+  public void whenNetworkInterceptor_followupsAreNewSpans() throws Exception {
```
adriancole: new test for when this is a network interceptor. each follow-up (such as a redirect) will show up in zipkin

adriancole:
Here's a sample trace I got from an elasticsearch client I'm working on. You can see the outbound request (which is okhttp) continuing from a server one.

screen shot 2016-09-26 at 8 49 02 pm
screen shot 2016-09-26 at 8 49 20 pm

adriancole:
thinking about an adjustment, which is to make the "application" request a local span. This would make each network attempt a child client span. I think this would be more helpful to explain latency, especially if a tag is set.

Ex. if using feign, you could set the okhttp tag corresponding to the api name called. For example, github#list-users would be a span, and it would have a child named "get" for each network attempt. This would pretty-much obviate any need to make custom feign instrumentation (ex #158)

This needs the special nested-aware span state class, as often the caller of okhttp is already a local span. Lucky thing is that we have that. (InheritableServerClientAndLocalSpanState)

cc @schlosna @spencergibb @rafabene @ewhauser @marcingrzejszczak @shakuzen

adriancole:
ok refactored the thing to do nested spans. very hard to test this, since there's no way in brave to control nanotime or trace id generation. I think I learned more than I wanted to about mockito.

Anyway, will update examples after trying against elasticsearch again

brave-okhttp/src/main/java/com/github/kristofa/brave/okhttp/BraveTracingInterceptor.java
```diff
@@ -0,0 +1,238 @@
...
+    spanName = builder.spanNameProvider;
+    tagIsLocalSpanName = builder.tagIsLocalSpanName;
+    serverName = builder.serverName;
+  }
+
+  @Override
+  public Response intercept(Chain chain) throws IOException {
+    Request request = chain.request();
+    Connection connection = chain.connection();
+    if (connection == null) { // we are an application interceptor, so start a local span.
```
swankjesse: Neat. I've never seen a single interceptor work in both modes, but it makes sense!

adriancole:
Here's how it looks now (using okhttp tags to name application spans)

screen shot 2016-09-27 at 9 02 40 pm

adriancole:
@swankjesse in the image I just pasted, you can see application two okhttp requests resulting from a server. "search-..." are okhttp tag names I assigned to the requests. Each one has a child span "post" corresponding to the actual network request.

I suspect that in the future, we could add things to explain the offset. Ex the first was probably longer due to connection setup and/or dns. For example, we could trace the Dns type or any other hooks that might arise.

brave-okhttp/src/main/java/com/github/kristofa/brave/okhttp/BraveTracingInterceptor.java
```diff
@@ -0,0 +1,238 @@
...
+    OkHttpClientRequest okHttpRequest = new OkHttpClientRequest(request, connection);
+    try {
+      requestInterceptor.handle(new OkHttpClientRequestAdapter(okHttpRequest));
+      Response response = chain.proceed(okHttpRequest.builder.build());
+      responseInterceptor.handle(new HttpClientResponseAdapter(new OkHttpResponse(response)));
+      return response;
+    } catch (IOException | RuntimeException | Error e) {
+      // TODO: revisit https://github.com/openzipkin/openzipkin.github.io/issues/52
+      brave.clientTracer().submitAnnotation(Constants.ERROR);
+      brave.clientTracer().setClientReceived(); // span must be closed!
```
adriancole: probably should clarify the comment. responseInterceptor.handle internally calls setClientReceived last, so on throw we have to do it manually.

adriancole:
@schlosna I'm going to rewrite this so that it is easier to add custom headers as tags

adriancole:
Here's a fun one.. I traced the okhttp crawler example, showing a mix of requests taken from local cache and network responses

screen shot 2016-09-28 at 3 57 49 pm

adriancole:
going out as 3.13.0 and will be used almost immediately in zipkin when tracing its traffic to elasticsearch

schlosna:
Thanks @adriancole, I didn't get a chance to test it today, but sounds like this should cover our need.

PR #244, opened 2016-09-27
Allow BraveOkHttpRequestResponseInterceptor handler injection

schlosna:
While instrumenting an OkHTTP client, realized we'd like to specify the behavior of the `HttpServerRequestAdapter` and `HttpClientRequestAdapter` to provide additional context (e.g. adding an annotation from an HTTP request header such as from `User-Agent`.

Also optimize HTTP client & server request adapters to avoid allocations when trace sampling is not enabled.

Adds toString for Endpoint debugging.

### DIFF
--- a/brave-http/src/main/java/com/github/kristofa/brave/http/HttpClientRequestAdapter.java
+++ b/brave-http/src/main/java/com/github/kristofa/brave/http/HttpClientRequestAdapter.java
@@ -8,50 +8,54 @@ import com.github.kristofa.brave.internal.Nullable;
 import com.twitter.zipkin.gen.Endpoint;
 import zipkin.TraceKeys;
 
-import java.net.URI;
-import java.util.Arrays;
 import java.util.Collection;
+import java.util.Collections;
 
 public class HttpClientRequestAdapter implements ClientRequestAdapter {
 
-    private final HttpClientRequest request;
+    private final HttpClientRequest clientRequest;
     private final SpanNameProvider spanNameProvider;
 
-    public HttpClientRequestAdapter(HttpClientRequest request, SpanNameProvider spanNameProvider) {
-        this.request = request;
+    public HttpClientRequestAdapter(HttpClientRequest clientRequest, SpanNameProvider spanNameProvider) {
+        this.clientRequest = clientRequest;
         this.spanNameProvider = spanNameProvider;
     }
 
-
     @Override
     public String getSpanName() {
-        return spanNameProvider.spanName(request);
+        return spanNameProvider.spanName(clientRequest);
     }
 
     @Override
     public void addSpanIdToRequest(@Nullable SpanId spanId) {
         if (spanId == null) {
-            request.addHeader(BraveHttpHeaders.Sampled.getName(), "0");
+            clientRequest().addHeader(BraveHttpHeaders.Sampled.getName(), "0");
         } else {
-            request.addHeader(BraveHttpHeaders.Sampled.getName(), "1");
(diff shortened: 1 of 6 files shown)


schlosna:
@adriancole would be interested in your thoughts on this.

brave-http/src/main/java/com/github/kristofa/brave/http/HttpClientRequestAdapter.java
```diff
@@ -8,50 +8,54 @@
...
 
-import java.net.URI;
-import java.util.Arrays;
 import java.util.Collection;
+import java.util.Collections;
 
 public class HttpClientRequestAdapter implements ClientRequestAdapter {
 
-    private final HttpClientRequest request;
+    private final HttpClientRequest clientRequest;
```
adriancole: since this code is doing things for acts like saving allocations, might be worth profiling the difference between just making this protected vs making an overridable method.

If all you need is to access the field, I'd probably make it protected, keeps the scope down, as I don't think we should be changing the type of `HttpClientRequest` returned consumed by this adapter without explaining why.

brave-http/src/main/java/com/github/kristofa/brave/http/HttpServerRequestAdapter.java
```diff
@@ -22,36 +25,40 @@ public HttpServerRequestAdapter(HttpServerRequest serverRequest, SpanNameProvide
...
             }
         }
-        return TraceData.builder().build();
+        return EMPTY_MAYBE_TRACE;
     }
 
     @Override
     public String getSpanName() {
-        return spanNameProvider.spanName(serverRequest);
+        return spanNameProvider.spanName(serverRequest());
```
adriancole: same comment on just making this field protected

brave-okhttp/src/main/java/com/github/kristofa/brave/okhttp/BraveOkHttpRequestResponseInterceptor.java
```diff
@@ -29,10 +33,22 @@ public Response intercept(Interceptor.Chain chain) throws IOException {
...
     OkHttpRequest okHttpRequest = new OkHttpRequest(builder, request);
-    clientRequestInterceptor.handle(new HttpClientRequestAdapter(okHttpRequest, spanNameProvider));
+    clientRequestInterceptor.handle(createRequestAdapter(okHttpRequest));
     Response response = chain.proceed(builder.build());
-    clientResponseInterceptor.handle(new HttpClientResponseAdapter(new OkHttpResponse(response)));
+    clientResponseInterceptor.handle(createResponseAdapter(response));
     return response;
   }
 
+  protected SpanNameProvider getSpanNameProvider() {
```
adriancole: Consider the caller can control the constructor, I think this part is over-extension.

brave-okhttp/src/main/java/com/github/kristofa/brave/okhttp/BraveOkHttpRequestResponseInterceptor.java
```diff
@@ -29,10 +33,22 @@ public Response intercept(Interceptor.Chain chain) throws IOException {
...
+    return spanNameProvider;
+  }
+
+  protected HttpClientRequestAdapter createRequestAdapter(HttpClientRequest request) {
```
adriancole: presumably the below here are to support adding the additional headers?

adriancole:
I'll make a test case in the other pull request for adding custom headers. thanks for the idea!

adriancole:
I think this is obviated by #242 except for the cleanups. let me know if not

schlosna:
Closing this out, see #245 and #246 for non-extension related changes

PR #245, opened 2016-09-30
Adds toString for Endpoint debugging

schlosna:


adriancole:
Ideally this would be json so it matches other ones. I *think* zipkin.Endpoint is json string. If there's anything existing converter I might suggest changing to zipkin then to string.  This will make ip addresses far easier to read.

PR #246, opened 2016-09-30
Improve brave-http request/response adapters code consistency

schlosna:
Shows minor performance improvement for arbitrary microbenchmark:

```
Benchmark   Mode  Cnt   Score   Error  Units
Before      avgt   15  65.362 ± 3.308  ns/op
After       avgt   15  48.170 ± 2.641  ns/op
```


adriancole:
Lgtm thanks!

PR #247, opened 2016-09-30
Allow more control over flushing span collector and queue buffer

schlosna:
Allow providing one's own blocking queue for buffering pending spans, and allow specifying the flush interval time unit so that spans can be flushed more frequently than once per second.

### DIFF
--- a/brave-core/src/main/java/com/github/kristofa/brave/AbstractSpanCollector.java
+++ b/brave-core/src/main/java/com/github/kristofa/brave/AbstractSpanCollector.java
@@ -4,6 +4,8 @@ import com.twitter.zipkin.gen.Span;
 import com.twitter.zipkin.gen.SpanCodec;
 import java.io.IOException;
 import java.util.List;
+import java.util.concurrent.BlockingQueue;
+import java.util.concurrent.TimeUnit;
 
 /**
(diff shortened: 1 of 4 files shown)


adriancole:
What's the motivation for a different buffered queue impl? Curious.

schlosna:
Mainly interested in being able to use an ArrayBlockingQueue to avoid some of the GC pressure under load.

adriancole:
how about just changing it to ArrayBlockingQueue? linkedblocking queue was an arbitrary decision, and when it grows to 1000 elements it would be larger than an ArrayBlockingQueue.

IOTW, it is reasonable to make this the default, and less knobs the better

schlosna:
I'll change it to ABQ, though I would still like to be able to specify queue size and finer grained flush frequency, does that seem ok?

adriancole:
seems ok but the whole thing will be obviated when we integrate
zipkin-reporter (imminently)

schlosna:
Just digging into https://github.com/openzipkin/zipkin-reporter-java/issues/1 now. Curious what timeframe you're looking at for integrating the two.

adriancole:
how about now? :)

https://github.com/openzipkin/brave/pull/248

adriancole:
@schlosna if you still want this open considering the reporter change, please say so and I'll click re-open

PR #248, opened 2016-10-01
Deprecates SpanCollector in favor of zipkin.reporter.Reporter

adriancole:
zipkin-reporter was made to allow others to re-use transport logic
without copy/paste. It has all the functionality Brave's SpanCollector
along with new functionality such as being able to send spans sooner
than a polling interval (ex on span count or byte size threshold).

This also allows re-use of third-party transports such as those for
the Amazon cloud.

See https://github.com/openzipkin/zipkin-reporter-java
See https://github.com/openzipkin/zipkin-aws

### DIFF
--- a/brave-core/src/main/java/com/github/kristofa/brave/LocalTracer.java
+++ b/brave-core/src/main/java/com/github/kristofa/brave/LocalTracer.java
@@ -8,6 +8,7 @@ import com.twitter.zipkin.gen.Span;
 import zipkin.Constants;
 
 import java.util.Random;
+import zipkin.reporter.Reporter;
 
 import static zipkin.Constants.LOCAL_COMPONENT;
 
@@ -50,7 +51,7 @@ public abstract class LocalTracer extends AnnotationSubmitter {
 
     abstract Random randomGenerator();
 
-    abstract SpanCollector spanCollector();
+    abstract Reporter<zipkin.Span> reporter();
 
     abstract boolean allowNestedLocalSpans();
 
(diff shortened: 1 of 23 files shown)


adriancole:
cc @schlosna  @llinder @devinsba

adriancole:
note: this code is already used by wingtips, jaeger, finagle, and spring cloud sleuth

brave-core/src/main/java/com/github/kristofa/brave/LoggingReporter.java
```diff
@@ -0,0 +1,38 @@
...
+
+import java.util.logging.Level;
+import java.util.logging.Logger;
+import zipkin.reporter.Reporter;
+
+import static com.github.kristofa.brave.internal.Util.checkNotBlank;
+import static com.github.kristofa.brave.internal.Util.checkNotNull;
+
+/**
+ * Simple {@link SpanCollector} implementation which logs the span through jul at INFO level.
```
shakuzen: Did you mean to still link to `SpanCollector` here instead of `zipkin.reporter.Reporter`?

adriancole:
Yep. Nice catch

brave-core/src/main/java/com/github/kristofa/brave/LoggingReporter.java
adriancole: done

brave-core/src/main/java/com/github/kristofa/brave/LocalTracer.java
```diff
@@ -203,7 +204,7 @@ public void finishSpan(long duration) {
     private void internalFinishSpan(Span span, long duration) {
         synchronized (span) {
             span.setDuration(duration);
-            spanCollector().collect(span);
+            reporter().report(span.toZipkin());
```
schlosna: Is it worth moving the reporting out of the synchronized block (though I wouldn't expect convention on that span)?

adriancole: sure thing. it is worth moving just to avoid distraction

adriancole:
@schlosna lemme know if you want me to merge this and you test with snapshot or if you want to test the branch. either way, we can hold release until you test.

PR #249, opened 2016-10-01
Adds Endpoint.toString that matches json

adriancole:
This adds Endpoint.toString for debugging. It uses json encoding so that
it is easier to compare with spans in zipkin.

Fixes #245


adriancole:
@schlosna I realized this wasn't as easy since we don't format json on endpoint upstream. took the punch to do the work. let me know if this works for you

adriancole:
@schlosna tomorrow I'll update base library to zipkin 1.13 which simplifies this as we just need to convert to zipkin.Endpoint then call toString(). afterwards will shortly release a Brave, assuming you're ok with the Reporter stuff sitting in master.

adriancole:
bumped

PR #251, opened 2016-10-06
Added beans.xml to support brave-jaxrs2 in CDI environments

heiko-braun:
This commit adds `META-INF/beans.xml` to `brave-core`, `brave-http` and `brave-jaxrs2`. This allows it to be used in CDI based environments, which otherwise wouldn't be able to scan the resources.

With regard to the existing implementation, I think this doesn't have any impact.

### DIFF
--- a/brave-core/src/main/resources/META-INF/beans.xml
+++ b/brave-core/src/main/resources/META-INF/beans.xml
@@ -1,0 +1,7 @@
+<?xml version="1.0" encoding="UTF-8"?>
+
+<beans xmlns="http://xmlns.jcp.org/xml/ns/javaee"
+       xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
+       xsi:schemaLocation="http://xmlns.jcp.org/xml/ns/javaee http://xmlns.jcp.org/xml/ns/javaee/beans_1_1.xsd"
+       version="1.2" bean-discovery-mode="all">
+</beans>
--- a/brave-http/src/main/resources/META-INF/beans.xml
+++ b/brave-http/src/main/resources/META-INF/beans.xml
@@ -1,0 +1,7 @@
+<?xml version="1.0" encoding="UTF-8"?>
+
+<beans xmlns="http://xmlns.jcp.org/xml/ns/javaee"
+       xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
+       xsi:schemaLocation="http://xmlns.jcp.org/xml/ns/javaee http://xmlns.jcp.org/xml/ns/javaee/beans_1_1.xsd"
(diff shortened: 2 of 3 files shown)


heiko-braun:
Looks like brave-grpc failed:

```
[INFO] brave-grpc ......................................... FAILURE [  7.032 s]
```

I'll have a look.

heiko-braun:
I am not able to reproduce this error. Maybe an intermittent failure? Would it be possible that someone kicks off another test run?

adriancole:
@heiko-braun sorry about the build flake. ps is there a way to test this change works? ex bring in a test dep that would load this file in a unit test?

adriancole:
wait this is adding beans files to even non java ee modules. I'm not sure if this is something we want to guarantee. What's the impact of not doing this?

heiko-braun:
Thanks for the feedback @adriancole . 

With regard to testing:
Do you think about just loading the file to verify it's packaged correctly, or something more elaborate the verifies the libraries in a CDI environment? I think the later would at quiet some considerable bloat to the test and the project a as a whole.

With regard to your second question:
both `brave-http` and `brave-core` are transitive deps of `brave-jaxrs2`. Discovery in a CDI env requires either a `META-INF/beans.xml` or bean defining annotations on relevant classes. I think the later is what's causing the problem. If you look at `BraveContainerResponseFilter` for instance it uses `@Inject` on the ctor, but the injected param `ServerResponseInterceptor`, which comes from `brave-core` doesn't use any defining annotations.

To overcome this problem, I thought it would be the most simple to just add `META-INF/beans.xml` to all relevant libraries.

adriancole:
This seems more to me that the historically added Inject annotation is the
real problem. There are few things in brave that make sense to wire based
solely on types and wildcarding every jar seems questionable.

Ex Brave is an object graph in itself, so we typically pass that around
after it has been constructed. What if you were to make a configuration
class for Brave in CDI similar to how folks are using spring?

heiko-braun:
That would work too, but it implies we remove the remaining `@Inject` annotations, if I understand you correctly.

adriancole:
yeah. a better punch to take imho. I'd rather mention this in the release
notes (the removal of the inject annotations).

heiko-braun:
I agree. Let close this PR and I'll prepare another one with changes you suggested.

adriancole:
ps @heiko-braun I noticed your work on swarm.. cool stuff! If you get to a point where you want to have it listed on zipkin.io, raise a PR here https://github.com/openzipkin/openzipkin.github.io/blob/master/_data/community_instrumentations.yml

heiko-braun:
@adriancole thanks, I'll do that once our implementation is in a reasonable shape and properly documented.

adriancole:
:thumbsup:

PR #252, opened 2016-10-07
BRAVE-250

gnoremac:
-Added tests to show InheritableThreadLocals will leak across threads

These tests will currently fail. I don't know if it's possible to provide a reliable fix without getting rid of the InheritableThreadLocal. Even if the objects are cloned they will not work in Executors that reuse threads.

A better approach would be to maintain a stack something like

```
class WrappedSpanLocal {
    Span span
    WrappedSpanLocal parent;
}
```

and use ThreadLocals. This would involve updating the BraveExecutors that pass context between the threads to not only pass server spans but client and local spans as well which would break backwards compatibility.

I will continue to think of another solution.

### DIFF
--- a/brave-core/src/test/java/com/github/kristofa/brave/InheritableServerClientAndLocalSpanStateTest.java
+++ b/brave-core/src/test/java/com/github/kristofa/brave/InheritableServerClientAndLocalSpanStateTest.java
@@ -4,13 +4,22 @@ import static org.assertj.core.api.Assertions.assertThat;
 import static org.junit.Assert.assertEquals;
 import static org.junit.Assert.assertNull;
 import static org.junit.Assert.assertSame;
+import static org.junit.Assert.assertTrue;
 import static org.mockito.Mockito.mock;
 import static org.mockito.Mockito.when;
+
+import java.util.concurrent.CountDownLatch;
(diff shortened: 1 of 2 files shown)


adriancole:
cc @schlosna I'm pretty busy next couple days. in case you aren't mind reviewing this?

adriancole:
It looks like you are actually testing the behavior of inheritable thread locals, right? We'd expect inheritable thread locals to propagate to threads.

Can you rephrase your tests on the actual use case you are trying to get to work? or maybe summarize the reason why the inherited state causes a problem? (I know there are a number of reasons why many don't prefer inheritable, but we need to pin this down with a use case)

gnoremac:
Sorry I totally forgot about this issue. Your right my test is really testing thread locals. I guess the issue I'm having is Spans are immutable, most users by default will use a executor that reuses threads which results in weird behaviour as the span moves between threads.

I'm not sure I even understand the use case for InheritableServerClientAndLocalSpanState as it seems like it's very unsafe due to the thread locals. 

Maybe this issue is more of a documentation update issue than a code fix issue as it seems like it's working as intended.

adriancole:
latest brave has a type called `CurrentTraceContext` which can be easily customized to not use inheritable thread locals

PR #254, opened 2016-10-12
Fixes bug where annotation timestamps were recorded improperly

adriancole:
There was a subtle bug where a duration was mistaken for a timestamp.
This corrects that.

### DIFF
--- a/brave-core/src/main/java/com/github/kristofa/brave/AnnotationSubmitter.java
+++ b/brave-core/src/main/java/com/github/kristofa/brave/AnnotationSubmitter.java
@@ -57,8 +57,14 @@ public abstract class AnnotationSubmitter {
     public void submitAnnotation(String value) {
         Span span = spanAndEndpoint().span();
         if (span != null) {
+            Long startTimestamp;
+            Long startTick;
+            synchronized (span) {
+                startTimestamp = span.getTimestamp();
+                startTick = span.startTick;
+            }
             Annotation annotation = Annotation.create(
-                currentTimeMicroseconds(startTick(span)),
+                currentTimeMicroseconds(startTimestamp, startTick),
                 value,
                 spanAndEndpoint().endpoint()
             );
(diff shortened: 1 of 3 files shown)


adriancole:
merging for a quick patch release as without spans will be created with invalid timestamps.

adriancole:
cc @yschimke @schlosna @jplock @hyleung this is something you should update to.

3.14.1 will be out shortly

jplock:
@adriancole I don't think v3.14.1 made it to Maven Central

adriancole:
weird. the build passed, but this part didn't go. Clicked manually here.
thanks for checking!

https://bintray.com/openzipkin/maven/brave/3.14.1#central

brave-core/src/main/java/com/github/kristofa/brave/AnnotationSubmitter.java
```diff
@@ -191,9 +186,9 @@ public void submitBinaryAnnotation(String key, int value) {
         submitBinaryAnnotation(key, String.valueOf(value));
     }
 
-    long currentTimeMicroseconds(@Nullable Long startTick) {
-        return startTick != null
-            ? (System.nanoTime() - startTick) / 1000
```
hyleung: Should this fix (or bug fixes in general) be ported back to previous versions of the library as a patch version (e.g. v3.13.0 -> v3.13.1)? Or is the expectation that users should just bump to the latest minor + patch (in this case v3.14.1)?

adriancole:
>
> Should this fix (or bug fixes in general) be ported back to previous versions of the library as a patch version (e.g. v3.13.0 -> v3.13.1)? Or is the expectation that users should just bump to the latest minor + patch (in this case v3.14.1)?
>
> current expectations are to update, particularly as many minor releases
have only one or two changes (maintaining a branch for each would be quite
a chore)

PR #258, opened 2016-10-27
Supports 128-bit trace IDs, generating them on traceId128Bit(true)

adriancole:
Traditionally, Zipkin trace IDs were 64-bit. Starting with Zipkin 1.14,
128-bit trace identifiers are supported. This can be useful in sites that
have very large traffic volume, persist traces forever, or are re-using
externally generated 128-bit IDs.

If you want Brave to generate 128-bit trace identifiers when starting new
root spans, set `Brave.Builder.traceId128Bit(true)`

When 128-bit trace ids are propagated, they will be twice as long as
before. For example, the `X-B3-TraceId` header will hold a 32-character
value like `163ac35c9f6413ad48485a3953bb6124`.

Before doing this, ensure your Zipkin setup is up-to-date, and downstream
instrumented services can read the longer 128-bit trace IDs.

Note: this only affects the trace ID, not span IDs. For example, span ids
within a trace are always 64-bit.

See https://github.com/openzipkin/zipkin/issues/1298
See https://github.com/openzipkin/zipkin-go-opentracing/pull/31


brave-core/src/test/java/com/github/kristofa/brave/ClientTracerTest.java
```diff
@@ -181,39 +181,19 @@ public void testStartNewSpanSampleNullNotPartOfExistingSpan() {
     }
 
     @Test
-    public void testStartNewSpanSampleTrueNotPartOfExistingSpan() {
```
adriancole: invalid use case as the method tested was never public or used

adriancole:
cc @openzipkin/core @openzipkin/instrumentation-owners fyi this is the second library to support 128-bit trace ids, after zipkin-go-opentracing

jcarres-mdsol:
Cool, I like having a reference implementation 👍

PR #263, opened 2016-11-02
Create bundle Manifests. Fixes #256

### LINKED ISSUE #256: Provide OSGi bundles during brave build
Currently the brave artifacts are no OSGi bundles. I would like to use brave in Apache Karaf but this makes it a lot harder.

If you are interested I can provide a pull request for some initial OSGi support.

cschneider:


cschneider:
The build fails at grpc .. I don't think it is related to my change. The build works locally for me.

adriancole:
Thanks. have you verified the manifests look correct? Also, sometimes exports can be quite huge (ex bloating the jar size). Can you take a look and see what the impact is on file size?

cschneider:
I can not yet fully validate the Manifests as brave relies on zipkin and zipkin is not yet a bundle. I will have to first also create bundles for zipkin.

The jar size should not change much. The maven bundle plugin should only change the Manifest which should not increase the jar size a lot.

adriancole:
>
> The jar size should not change much. The maven bundle plugin should only
> change the Manifest which should not increase the jar size a lot.
>
I think I've seen manifests becoming kilobytes in size in the past, which
is relatively large increase. probably worth simply checking the size
before and after, at least for the core jar

cschneider:
This is the current content of the Manifest for brave core. The size is about 700 bytes.
So that should be fine. 

Manifest-Version: 1.0
Bundle-Description: Brave core.
Bundle-License: http://www.apache.org/licenses/LICENSE-2.0.txt
Bundle-SymbolicName: io.zipkin.brave.core
Archiver-Version: Plexus Archiver
Built-By: cschneider
Bnd-LastModified: 1478097353420
Bundle-ManifestVersion: 2
Import-Package: zipkin,zipkin.internal,zipkin.reporter
Require-Capability: osgi.ee;filter:="(&(osgi.ee=JavaSE)(version=1.6))"
Tool: Bnd-3.2.0.201605172007
Export-Package: com.twitter.zipkin.gen;uses:=zipkin;version="3.15.1",c
 om.github.kristofa.brave;uses:="com.twitter.zipkin.gen,zipkin,zipkin.
 reporter";version="3.15.1"
Bundle-Name: brave-core
Bundle-Version: 3.15.1.SNAPSHOT
Created-By: Apache Maven Bundle Plugin
Build-Jdk: 1.8.0_92

adriancole:
thanks again!

adriancole:
out in brave 3.15.1

Issue #264, opened 2016-11-02
Make BraveServletFilter customizable


remmeier:
I started including brave in our projects and works great. However, we faced some issues when trying to customize the behavior a little bit and ended up copy pasting the complete implementation.

BraveServletFilter.doFilter is to large, the creation of ServletHttpServerRequest and HttpServerRequestAdapter should be moved into a dedicated method that would allow subclassing and modifying it. Or maybe even better would be to be able to set a factory. We modified HttpServerResponseAdapter to provide more than just the TraceKeys.HTTP_STATUS_CODE.

I can do a PR for this.

adriancole:
I've noticed some limitations, too, but I'd be careful about anchoring too
much on the way it currently works, or exposing things that we'd end up
needing to support later.

For example, I made OkHttp's completely independent of the abstraction as
an interim (because it is hard to tell what is actually portable without
multiple implementations).
https://github.com/openzipkin/brave/blob/master/brave-okhttp/src/main/java/com/github/kristofa/brave/okhttp/BraveTracingInterceptor.java

Maybe you can share your own version for now? and show which frameworks it
works with (specifically which use cases it aims to support?)

remmeier:
regarding use cases:

- I added a "getPrincpialName" to a custom HttpServerRequest
- a custom ServerRequestAdapter adds some further binary annotations:
-- the principal
-- values from proprietary request headers (like client id, client version)
-- some metrics collected during the execution that are not traced by themselves

Not quite sure where to place those. Implementing ServerRequestAdapter seemed the best thing at first. But as noted above there is no easy way to add them without copy/pasing the implementation.

I could also imagine wrapping ServerRequestInterceptor and ServerResponseInterceptor and apply my changes there. But currently those classes not interfaces and are to replace as well. The advantage here is that it is a global thing. As long as those interceptors are used by the different filters (okhttp, servlet, etc.), I can reconfigure brave in one location.

yes, I started using okhttp as well. Works great. I think it would be good if it follows the http servlet and makes use of the ServerRequestAdapter/ServerResponseAdapter and forwards the data to ServerRequestInterceptor and ServerResponseInterceptor.

I started a PR. I think a ServerAdapterFactory in brave-core could be a good next step instead of having to use subclassing.

adriancole:
[...]

PR #265, opened 2016-11-03
#264 Make BraveServletFilter customizable

remmeier:
first step to allow customization of BraveServletFilter by subclassing it. New implementations of HttpServerRequestAdapter and HttpServerResponseAdapter can be provided by overriding the
corresponding factory methods. 

A dedicated ServerAdapterFactory class should be considered and move newServerRequestAdapter(...) and newServerResponseAdapter(...) there.

### DIFF
--- a/brave-web-servlet-filter/src/main/java/com/github/kristofa/brave/servlet/BraveServletFilter.java
+++ b/brave-web-servlet-filter/src/main/java/com/github/kristofa/brave/servlet/BraveServletFilter.java
@@ -1,11 +1,8 @@
 package com.github.kristofa.brave.servlet;
 
-import com.github.kristofa.brave.ServerRequestInterceptor;
-import com.github.kristofa.brave.ServerResponseInterceptor;
-import com.github.kristofa.brave.http.HttpResponse;
-import com.github.kristofa.brave.http.HttpServerRequestAdapter;
-import com.github.kristofa.brave.http.HttpServerResponseAdapter;
-import com.github.kristofa.brave.http.SpanNameProvider;
+import java.io.IOException;
+import java.util.ArrayList;
+import java.util.List;
 
 import javax.servlet.Filter;
 import javax.servlet.FilterChain;
@@ -16,7 +13,17 @@ import javax.servlet.ServletResponse;
 import javax.servlet.http.HttpServletRequest;
 import javax.servlet.http.HttpServletResponse;
 import javax.servlet.http.HttpServletResponseWrapper;
-import java.io.IOException;
+
+import com.github.kristofa.brave.ServerAdapterInterceptor;
+import com.github.kristofa.brave.ServerRequestAdapter;
+import com.github.kristofa.brave.ServerRequestInterceptor;
+import com.github.kristofa.brave.ServerResponseAdapter;
+import com.github.kristofa.brave.ServerResponseInterceptor;
+import com.github.kristofa.brave.http.HttpResponse;
+import com.github.kristofa.brave.http.HttpServerRequest;
+import com.github.kristofa.brave.http.HttpServerRequestAdapter;
+import com.github.kristofa.brave.http.HttpServerResponseAdapter;
+import com.github.kristofa.brave.http.SpanNameProvider;
 
 /**
  * Servlet filter that will extract trace headers from the request and send
@@ -29,11 +36,17 @@ public class BraveServletFilter implements Filter {
     private final SpanNameProvider spanNameProvider;
 
     private FilterConfig filterConfig;
+    
+    private List<ServerAdapterInterceptor> adapterInterceptors = new ArrayList<ServerAdapterInterceptor>();
 
     public BraveServletFilter(ServerRequestInterceptor requestInterceptor, ServerResponseInterceptor responseInterceptor, SpanNameProvider spanNameProvider) {
         this.requestInterceptor = requestInterceptor;
         this.responseInterceptor = responseInterceptor;
         this.spanNameProvider = spanNameProvider;
+    }
+    
+    public void addAdapterInterceptor(ServerAdapterInterceptor adapterInterceptor){
+        adapterInterceptors.add(adapterInterceptor);
     }
 
     @Override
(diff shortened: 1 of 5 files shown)


brave-web-servlet-filter/src/main/java/com/github/kristofa/brave/servlet/BraveServletFilter.java
```diff
@@ -52,22 +55,32 @@ public void doFilter(ServletRequest request, ServletResponse response, FilterCha
...
-                filterChain.doFilter(request, statusExposingServletResponse);
-            } finally {
-                responseInterceptor.handle(new HttpServerResponseAdapter(new HttpResponse() {
-                    @Override
-                    public int getHttpStatusCode() {
-                        return statusExposingServletResponse.getStatus();
-                    }
-                }));
-            }
+        	final StatusExposingServletResponse statusExposingServletResponse = new StatusExposingServletResponse(
```
adriancole: it is very hard to tell what changed when formatting is mixed with code change. can you please revert the formatting change and ping back when only the essentials are shown? I'll review once that's done.

remmeier:
one moment...

remmeier:
one test in grpc seems to be a bit shacky... (BraveGrpcInterceptorsTest.testBlockingUnaryCall unrelated to this).

adriancole:
yeah that test is flakey.

Since this module is for something with a very stable api (servlet spec), I
think we should do something that doesn't require exposing so many
intermediate types. I'd suggest following the pattern of OkHttp instead.

https://github.com/openzipkin/brave/blob/master/brave-okhttp/src/main/java/com/github/kristofa/brave/okhttp/OkHttpParser.java

Ex. ServletParser with the same sort of disclaimers presented in the OkHttp
one, and the default behavior that matches the existing codebase. This
provides users with a single type to extend, which makes everything else
simpler on our end.

care to give it a shot?

remmeier:
agree that there are quite a number of intermediate types. But to some degree I also like them. The reason is in our case Brave will get integrated into dozens of projects. All of them make use of HTTP. The header processing and creation of binary annotations will be mostly the same. But the technology stack can differ (resteasy, servlet, etc.). For this reason having classes like HttpServerRequest and ServerRequestAdapter are very valuable to us since they allow the customizations to take place in a single ServerRequestAdapter and ServerResponseAdapter. We don't have to worry about which implementation is used as long as our adapter somehow can be integrated into it.

adriancole:
if the idea for this is to do something generically, you'll notice all the other implementations "new up" these classes or are package protected. In other words, it isn't current practice in Brave, so this would become the different one, right?

remmeier:
I guess this small PR opens up all kinds of questions regarding customizability. Yes, others like BraveGrpcServerInterceptor have all those classes class/package protected. But to some degree I think people will already be happy having access to HttpServerRequest/HttpServerResponse to compute further binary annotations. For more advanced use cases, maybe there should be some way to get the underyling orignal/native request/response (HttpServletRequest, ServerCall, etc.). But I guess it is undeseriable to make everything public. May some generic unwrap(...) function would suffice.

Personally, I don't yet really know in which direction Brave is heading (I'm new here). I'm fine with the current request/response/adapter/interceptor api. But also open for simplications. In either case I think some kind of interface for a "request/response => binary" annotation producer and a generic "http request/response interface" hiding internals of jersey/restesay/okhttp would be great.

adriancole:
I think the challenge here is mostly that we try to support compatibility. Every thing that's added as protected or public means a new compatibility liability/lock-in for the project. So, eventhough the lines of this change are small, it does imply a significant surface area change.. changes we couldn't undo later until a major version bump. That's the main driver in trying to figure out what this would look like for all of your integrations (not just servlet), and also what's the least exposure way for the project to facilitate that (4 types and corresponding protected methods likely isn't least)

something to think about. Sometimes the best way to explore these is to implement them in test cases to prove your use-case works.

remmeier:
so far the existing request/response/adapter interfaces do a great job from my perspective. So new types should not be necessary, it rather comes down to this four methods which would be necessary in all of the implementations to gain customizability. To slim things down, I think newServerRequestAdapter(...) and newServerResponseAdapter(...) are sufficient and we could avoid newHttpServerRequest(...) and newHttpServerResponse(...). It should not be necessary to reimplement those by the application. Maybe something like a unwrap(...) could provide access to the native request/response in the future for more advanced use cases.

The remaining two methods might be provided by a Factory so that applications can replace the Factory instead of subclassing Brave. I could update the PR accordingly and add an example to the tests.

Maybe some classes of Brave could also be placed into internal packages to make it clear from start that those should not be touched.

adriancole:
>
> so far the existing request/response/adapter interfaces do a great job
> from my perspective. So new types should not be necessary, it rather comes
> down to this four methods which would be necessary in all of the
> implementations to gain customizability. To slim things down, I think
> newServerRequestAdapter(...) and newServerResponseAdapter(...) are
> sufficient and we could avoid newHttpServerRequest(...) and
> newHttpServerResponse(...). It should not be necessary to reimplement those
> by the application. Maybe something like a unwrap(...) could provide access
> to the native request/response in the future for more advanced use cases.
>
least change necessary is important. Practically speaking, if you intend to
use this for different frameworks my question remains as to why you are
only doing this for the servlet filter? would we expect a few more PRs for
each of the other ones?

> The remaining two methods might be provided by a Factory so that
> applications can replace the Factory instead of subclassing Brave. I could
> update the PR accordingly and add an example to the tests.
>
they wouldn't be subclassing Brave, they'd be subclassing the filter,
right?

> Maybe some classes of Brave could also be placed into internal packages to
> make it clear from start that those should not be touched.
>
It isn't about "not being touched" It is that these were designed to
implement instrumentation such as the ServletFilter here. Using effective
java, we don't expose arbitrary extension points willy nilly. It just leads
to tech debt. That's why I'm taking time with you now to try and pare this
down and underpin it with an actual use case. Since you noted this was in
support of one adapter for many http frameworks, I'd really like to see
that in practice as you can't do it with the PR you have here for reasons
mentioned about the other instrumentation not being extendible this way.

Remember, you can always copy/paste this <200 line class while you figure
it out. Incomplete extension ideas tend to haunt projects.

remmeier:
I already copy&pasted. So I'm not in a hurry either and totally agree that this has to be done in a careful manner. At some point, naturally, I would like to get rid of the copy/pasting.

I will update the PR to include other filters next to servlet and make use of a factory. I did not want to create to many changes at first before the direction was not discussed.

with "subclassing brave" I meant in generall all brave classes (including the filters). Not just the main Brave class. Sorry for the confusion. When we already talk about the main brave class, I think there potentially ClientTracer, ServerTrace, ClientRequestInterceptor, ClientResponseInterceptor, ServerRequestInterceptor, ServerResponseInterceptor could become interfaces.

adriancole:
> I think there potentially ClientTracer, ServerTrace, ClientRequestInterceptor, ClientResponseInterceptor, ServerRequestInterceptor, ServerResponseInterceptor could become interfaces.
It is very unlikely that we change all of our major types to interfaces. That would be a bit silly

adriancole:
The main reasons for why we don't make all of our public types interfaces have a lot to do with the same reasons I'm hesitant to expose 4 methods to mutate the behavior of ServletFilter. Basically we are in general following effective java which favors less visibility, and final unless needed otherwise. While we do have (and will have) some interface types, it wouldn't be congruent to boiler-plate convert things to interfaces, this in mind. Remember, adding to an interface before java 8 means a breaking change! We not only want to be use-case focused, but also not break our users.

We don't have a HACKING file here, but I've made one in other projects, which helps explains some of the highlights to keep in mind when considering extensibility and change.

https://github.com/OpenFeign/feign/blob/master/HACKING.md

Anyway, let's focus on the actual use cases vs a preference for converting things to interfaces or other arbitrary extension. You're more likely to get support for the change when it is centered on change that has common benefits and is testable to actually work diversely without adding liability to the api.

remmeier:
I committed a next iteration: ClientAdapterInterceptor and ServerAdapterInterceptor introduced to make client and server, request and response adapters customizable.

Most common use case: add further annotations to span

Interceptors makes use of a wrapper approach. The interceptor does not get in touch with the internals of the various filter implementations. Instead, he is enabled to intercept and customize any calls to those implementations. This way the surface area of Brave is kept small.

Two filters (servlet, spring) have been modified to use this approach. Test cases and others to follow once the design is finalized. Naming and everything else up for discussion.

adriancole:
I don't think we want to support a list of interceptors. Someone can delegate/compose to achieve that. Considering the amount of people contributing maintenance to this project, we really need the least surface area possible.

remmeier:
sounds good to me too, Saves a couple of lines.

I also like @llinder proposal of using the builder pattern. Maybe I rather add a method Brave.Builder.setAdapterInterceptor(...). That interceptor is then passed to ServerRequestInterceptor and invoked in ServerRequestInterceptor.handle(...). This way the filters do not have to care about the interceptor anymore.

adriancole:
>
> I also like @llinder <https://github.com/llinder> proposal of using the
> builder pattern. Maybe I rather add a method Brave.Builder.setAdapterInterceptor(...).
> That interceptor is then passed to ServerRequestInterceptor and invoked in
> ServerRequestInterceptor.handle(...). This way the filters do not have to
> care about the interceptor anymore.
>
Depending on how the interceptor is defined, it might need http classes or
concepts which would make it too specific for brave's core builder. It
might be more appropriate to make an http builder (for http-tier
instrumentation). Ex.
HttpInstrumentation.builder(brave).spanNameProvider(foo).requestTags("http.method",
"http.uri")...

remmeier:
currently there is a ClientAdapterInterceptor  and ServerAdapterInterceptor. Maybe we could also consider using a generic Interceptor<T> and could potentially be applied "anywhere". The Interceptor may have an accept(...) method that decides whether the interceptor is applicable to a given source.

interface Intercept<T>{

  boolean accept(Object object);

  T intercept(T interceptedObject);
}

In that case one could implement e.g. a Interceptor<HttpServerRequestAdapter>.

adriancole:
moving the discussion back to the issue so that it isn't limited to only you and me

Issue #268, opened 2016-11-11
com.github.kristofa.brave.internal.Nullable is internal but used by other brave modules


cschneider:
I can not install brave-http in OSGi as it refers to the package com.github.kristofa.brave.internal. It seems that is uses the Nullable anntoation.

So there are two ways to fix this:
1. export the internal package
2. move Nullable to com.github.kristofa.brave as it is used outside brave core

Option 2 is the cleaner solution but it requires bigger changes. What do you think?
I will provide a PR for solution 1 first as it is easy to do.

adriancole:
Can you help me understand why we have to handle this at all? We don't
inspect the annotation at runtime, and nothing else does.

adriancole:
the only thing that uses this is auto-value (at compile time)

adriancole:
hoping OSGi is reasonable enough to accept source retention as a solution :)

https://github.com/openzipkin/brave/pull/275

adriancole:
if Util is a problem, I'd almost rather use shade to copy the 2 checkNotNull methods to where they are used vs export the package internal only to support them. Internal is internal for brave code, and it is a shame there's no solution for this problem beside blanket exporting.

PR #271, opened 2016-11-11
Export package com.github.kristofa.brave.internal

cschneider:
See issue #268

### DIFF
--- a/brave-core/bnd.bnd
+++ b/brave-core/bnd.bnd
@@ -1,4 +1,4 @@
 Export-Package: \
 	com.twitter.zipkin.gen,\
 	com.github.kristofa.brave,\
-	com.github.kristofa.brave.internal
+	com.github.kristofa.brave.internal;braveinternal=true;mandatory:=braveinternal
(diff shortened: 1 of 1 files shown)


adriancole:
looking at #275 there's only a Util class left. I really wish OSGi could be reasonable.

adriancole:
So Util has only checkNotX methods, and it would be nice to not have to export the internal package, and also nice to not have to copy or move the Util class as they are effectively internal to brave's group.

Essentially, it would be nice for us to be able to support external users of brave's "real" public types (i.e. non-internal ones) without eliminating or complicating the checkNotNull story (internal types), which have not been a problem so far.

I'd like to avoid blanket exporting internal as that feels wrong, as the only code that should use this would be code under our control. In other words, I'd like to not make our tacit internal package explicitly scoped for others (and then later have questions if we remove those types or add more into the internal subpackages).

@njbartlett suggested mandatory export attributes is a way out of this

adriancole:
next step: look at impact of the related mail thread and adjust https://mail.osgi.org/pipermail/osgi-dev/2016-November/005880.html

cschneider:
I applied the mandatory attribute like recommended by Tim and it works fine. We do not need special setup on the bundles using this. So looks good to me.

adriancole:
fantastico!

PR #274, opened 2016-11-11
Convenience constructor to BraveServletFilter/JerseyClientTraceFilter

epabst:
This reduces the learning curve for using Brave
and makes it easier to wire these into an application.

### DIFF
--- a/brave-jersey/src/main/java/com/github/kristofa/brave/jersey/JerseyClientTraceFilter.java
+++ b/brave-jersey/src/main/java/com/github/kristofa/brave/jersey/JerseyClientTraceFilter.java
@@ -3,8 +3,10 @@ package com.github.kristofa.brave.jersey;
 import javax.inject.Inject;
 import javax.inject.Singleton;
 
+import com.github.kristofa.brave.Brave;
 import com.github.kristofa.brave.ClientRequestInterceptor;
 import com.github.kristofa.brave.ClientResponseInterceptor;
+import com.github.kristofa.brave.http.DefaultSpanNameProvider;
 import com.github.kristofa.brave.http.HttpClientRequestAdapter;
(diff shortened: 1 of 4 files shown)


epabst:
FWIW, this build failure looks like an intermittent failure, possibly due to a concurrent test.  
"mvn clean install" passed for me.

adriancole:
public constructors are usually a little evil vs factor or builders wrt maintenance. Particularly as there are other various customization requests in-flight (like #265) I'm apprehensive about merging this right now.

adriancole:
ps fwiw I also think that it would be useful to use Brave as a starting point as opposed to elements provided by Brave. My guess is that the reason this wasn't done initially was in support of mocking (ex mocking is easier when you don't have to navigate an object graph like Brave)

epabst:
I really like the idea of making it easy to mock, and yes providing simply a Brave is much simpler for normal usage.  Providing two constructors, one with a Brave, and one with the mockable types may be the sweet spot.

epabst:
@adriancole Updated to only add a single constructor each, that just takes a Brave.

adriancole:
I don't think there's any design we'll end up with where public constructors become more used. How about revising to builder pattern instead (and deprecate the old constructor)?

```java
  public static Builder builder() {
    return new Builder();
  }

  public final static class Builder {
    private ServerRequestInterceptor requestInterceptor;
--snip--

    public Builder requestInterceptor(ServerRequestInterceptor requestInterceptor) {
      this.requestInterceptor = requestInterceptor;
      return this;
    }
--snip--

    public BraveServletFilter build() {
      // conditionally set anything..
      return new BraveServletFilter(this);
    }

    Builder() { // intentionally hidden
    }
  }

  BraveServletFilter(Builder builder) { // intentionally hidden
    this.requestInterceptor = builder.requestInterceptor;
--snip--
```

PR #275, opened 2016-11-12
Moves Nullable annotation to source retention to make OSGi not complain

### LINKED ISSUE #268: com.github.kristofa.brave.internal.Nullable is internal but used by other brave modules

adriancole:
Nullable is used to help direct auto-value, particularly indicating
fields that should result in an NPE when not present.

It seems OSGi complains when there's a annotation in an un-exported
package, and this is the simplest way to dodge the problem.

Fixes #268 (hopefully)

### DIFF
--- a/brave-core/src/main/java/com/github/kristofa/brave/internal/Nullable.java
+++ b/brave-core/src/main/java/com/github/kristofa/brave/internal/Nullable.java
@@ -1,10 +1,12 @@
 package com.github.kristofa.brave.internal;
 
+import java.lang.annotation.RetentionPolicy;
+
 /**
- * Libraries such as Guice and AutoValue will process any annotation named {@code Nullable}.
- * This avoids a dependency on one of the many jsr305 jars.
+ * AutoValue will process any annotation named {@code Nullable}. This avoids a dependency on one of
+ * the many jsr305 jars.
  */
 @java.lang.annotation.Documented
-@java.lang.annotation.Retention(java.lang.annotation.RetentionPolicy.RUNTIME)
+@java.lang.annotation.Retention(RetentionPolicy.SOURCE)
 public @interface Nullable {
 }


cschneider:
I think this works. The declarative services annotations work in the same way. Will check it. On the other hand exporting the package would also not really hurt a lot. At least it would be not any different to the situation outside OSGi.

adriancole:
>
> I think this works. The declarative services annotations work in the same
> way. Will check it. On the other hand exporting the package would also not
> really hurt a lot. At least it would be not any different to the situation
> outside OSGi.
>
Outside OSGi, the word internal indicates a tacit contract which is don't
depend on this api like you would a public one. Hence any project (brave
included) who breaks on account of an internal api break is par for the
course.

OSGi export formalizes internal apis, by making us have to declare them as
peers with others. For example, in javadoc we explicitly exclude internal
packages, which OSGi requires us to explicitly include them.

Bottom-line is that this is IMHO a part of the distraction of OSGi, where
we end up having to formalize things that are plain as day, and why it
irritates me.

adriancole:
complaints aside about the distracting work involved (mostly bore by you) I don't have a problem continuing the effort of OSGi'ing provided we don't end up having to break api or do terrible things to make it work.

cschneider:
I have tested the branch in OSGi and it works fine. So this is the optiomal solution for the problem

adriancole:
thx for testing!

PR #276, opened 2016-11-21
WIP: Introduces TagExtractor, for capturing binary annotations generically

adriancole:
This starts a decoupling between the mechanism to extract a tag, for
example the http path from a request, and actually wanting to add it to
a span. Once complete, it should be possible to add custom dimensions
like `servlet.user` without changing core code.

This is eagerly raised to get opinions on the general approach.. there's a lot to prove out here, and it won't be merged until it proves it can commonly deal with tags across http libraries (at least)

### DIFF
--- a/brave-apache-http-interceptors/src/main/java/com/github/kristofa/brave/httpclient/BraveHttpRequestInterceptor.java
+++ b/brave-apache-http-interceptors/src/main/java/com/github/kristofa/brave/httpclient/BraveHttpRequestInterceptor.java
@@ -2,7 +2,8 @@ package com.github.kristofa.brave.httpclient;
 
 import com.github.kristofa.brave.Brave;
 import com.github.kristofa.brave.ClientRequestInterceptor;
-import com.github.kristofa.brave.http.DefaultSpanNameProvider;
+import com.github.kristofa.brave.ClientRequestAdapter;
+import com.github.kristofa.brave.TagExtractor;
 import com.github.kristofa.brave.http.HttpClientRequestAdapter;
 import com.github.kristofa.brave.http.SpanNameProvider;
 import org.apache.http.HttpRequest;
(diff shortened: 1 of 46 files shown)


adriancole:
Current status: all http libraries now share a common value parser.
Next step: refactor out duplication and make it possible to configure
which tags to attach to the span.

adriancole:
ps expect some of the generic, type etc stuff to simplify before merge. main priority is to prove it out, then I'll pare it down.

adriancole:
moved initialization of TagExtractors to the constructor of each instrumentation type.

next step: weave in the ability to configure them via TagExtractor.Config.

background work: at some point, eliminate code duplication and redundant types

adriancole:
Current status: 
* all http libraries now share a common value parser.
* all http instrumentation can be configured with which keys to add
* all http instrumentation can accept custom value parsers

Next steps: 
* add tests to show this works
* test configuration composition with spring
* pare down and deduplicate types

adriancole:
fyi @epabst when merged, this change will allow all instrumentation types to be created with either `BraveFilter.create(brave)` or `BraveFilter.builder()....build(brave)` syntax. The latter is needed so that we can customize it.

See #274

adriancole:
in order to incorporate something cleanly, I need to be able to decouple the propagation pieces from the *Adapter types. I'm not doing any more on this PR until that's in place.

adriancole:
this is blocked by https://github.com/openzipkin/brave/pull/282

adriancole:
having a 2 week old work in progress gives me the creeps, but this is hugely  important to get it right and leave things better (as it effects almost all instrumentation). It will be a couple more days

adriancole:
a version of this will end up in a PR to address https://github.com/openzipkin/brave/issues/284

PR #278, opened 2016-11-23
Special-case externally provided trace IDs (when sampled left unset)

### LINKED ISSUE #277: Can not filter by “Duration (μs)> =”  in the UI
_From @longinhit on October 24, 2016 9:50_

_Copied from original issue: openzipkin/zipkin#1349_

adriancole:
There are some who want to control trace identifiers, but leave the
sampling decision to the next hop. In this case, they propagate the
required headers (X-B3-TraceId, X-B3-ParentId) and not X-B3-Sampled.

When this happens, the receiver of the headers (the server), should
sample the trace ID before using it.

Before, we weren't implementing this (eventhough Finagle does). The
change here addresses it and also backfills tests accordingly.

See https://github.com/twitter/finagle/blob/develop/finagle-http/src/main/scala/com/twitter/finagle/http/Codec.scala#L341
See https://github.com/twitter/finagle/blob/develop/finagle-core/src/main/scala/com/twitter/finagle/tracing/Tracer.scala#L119
See https://github.com/openzipkin/b3-propagation/pull/8

Fixes #277

### DIFF
--- a/brave-core/src/main/java/com/github/kristofa/brave/ServerRequestInterceptor.java
+++ b/brave-core/src/main/java/com/github/kristofa/brave/ServerRequestInterceptor.java
@@ -34,14 +34,27 @@ public class ServerRequestInterceptor {
         final TraceData traceData = adapter.getTraceData();
 
         Boolean sample = traceData.getSample();
-        if (sample != null && Boolean.FALSE.equals(sample)) {
+        if (Boolean.FALSE.equals(sample)) {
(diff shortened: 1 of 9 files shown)


dangets:
@adriancole Just tested it, looks good.  Duration on root trace is showing up and filtering in the UI works as expected.  Thanks!

adriancole:
good news!

PR #272, opened 2016-11-11
Add jaxrs Feature for Brave

cschneider:
See issue #269

PR #280, opened 2016-11-25
Adds dependency injection instructions and tests

adriancole:
Before, READMEs were all over the place with regards to configuration.
This standardizes them and corrects some dependency and test
parameters. Notably, this adds tests for Jersey 1.x (they never existed
previously).

Once this is in, I can help rebase the JAXRS-2 feature on top (#272)


adriancole:
cc @cschneider @jplock

PR #282, opened 2016-11-30
Adds Propagation, which decouples B3 implementation from framework code

adriancole:
Before, we had the same algorithm for encoding B3 copy pasted a couple
times. This centralizes it into a new type `Propagation`, and starts the
decoupling of how in-band propagation works from what is reported
out-of-band. The design uses the same terminology as OpenTracing.

This will make B3 maintenance easier as it is in one place. It also
decouples grpc from an incidental dependency on brave-http. Finally, it
allows us to refactor how span data is reported (something orthogonal to
how ids are propagated).

Future change will build on this to make interceptors more flexible,
particularly allowing custom tags to be added via composition as opposed
to inheritance.


brave-core/src/main/java/com/github/kristofa/brave/internal/Propagation.java
```diff
@@ -0,0 +1,51 @@
...
+import com.github.kristofa.brave.TraceData;
+
+/**
+ * Injects and extracts {@link SpanId span identifiers} sent in-band across process boundaries.
+ *
+ * <p>Implementations usually wrap Http, Messaging or RPC headers.
+ *
+ * @param <K> Usually, but not always a String
+ */
+public interface Propagation<K> {
```
adriancole: here's the actual propagation type

adriancole:
I've extracted the B3 stuff cleanly now. next step is to use the Propagation type directly instead of via the http request adapters.

brave-core/src/main/java/com/github/kristofa/brave/internal/Propagation.java
```diff
@@ -0,0 +1,61 @@
...
+    @Nullable String get(K key);
+  }
+
+  /**
+   * Calls the setter for each key to send downstream.
+   *
+   * @param spanId possibly null span identifiers. Null means unsampled.
+   * @param setter invoked for each propagation key to add.
+   */
+  void injectSpanId(@Nullable SpanId spanId, Setter<K> setter);
```
adriancole: these two methods obviate `ClientRequestAdapter.addSpanIdToRequest(SpanId)` and `ServerRequestAdapter.getTraceData()`. That makes the adapters' key focus parsing out-of-band data, as opposed to parsing out-of-band data and also hard-wiring propagation logic.

brave-core/src/main/java/com/github/kristofa/brave/internal/Propagation.java
adriancole: this type and methods prefixed internal will stay that way until the out-of-band pull request #276 is done.

brave-core/src/main/java/com/github/kristofa/brave/internal/Propagation.java
```diff
@@ -0,0 +1,61 @@
...
+/**
+ * Injects and extracts {@link SpanId span identifiers} sent in-band across process boundaries.
+ *
+ * <p>Implementations usually wrap Http, Messaging or RPC headers.
+ *
+ * @param <K> Usually, but not always a String
+ */
+public interface Propagation<K> {
+
+  interface Factory {
```
adriancole: while possible to change propagation from B3 to something else, I've left customizing this out of it for now. Mainly I want to finish the other change about specifying keys. Once that's in, it will be more clear which types need to hold this responsibility. Until then, it is status quo (B3 only)

adriancole:
ok tidied it all up, notably borrowing some concepts from OpenTracing (like injector/extractor)

adriancole:
I want to rebase this over retrolambda as it will remove a lot of cruft https://github.com/openzipkin/brave/pull/283/files

adriancole:
ok all rebased and tidied. I will rework my other PR on top of this to verify it operates as planned

adriancole:
put a lot of effort into this, and pretty good, but not merging until I'm sure it won't need to change. eta a couple days

adriancole:
this will arise again as a part of https://github.com/openzipkin/brave/issues/284

PR #287, opened 2016-12-07
Makes implicit change culture transparent

adriancole:
There have been a number of issues where I've needed to clarify why a specific
change isn't ready or shouldn't be merged. The rules of engagement, and their
rationale, shouldn't require me, as I didn't start Brave and may not be around forever!

My hope is that by stating things like how change works in the root folder, we can
avoid some back and forth due to previously unwritten expectations.

Using rule-of-three for this change :) Here are five issues where I've basically restated the contents above manually. I'd really like to use my time and the time of contributors towards shared change, since in both cases it is a scarce resource.

#273 #265 #238  #226 #194

### DIFF
--- a/HACKING.md
+++ b/HACKING.md
@@ -1,0 +1,64 @@
+# Hacking Brave
+Brave is optimized for maintenance vs flexibility. It prefers small
+features that have been asked for repeated times, that are insured with
+tests, and have clear use cases. This limits the lines of code and count
+of modules in Brave's repo.
+
+Code design is opinionated including below:
+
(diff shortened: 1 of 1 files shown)


adriancole:
cc @openzipkin/core

adriancole:
ps the contents of this are a direct port from the same file in Feign, which was created under the exact circumstances as repeat here https://github.com/OpenFeign/feign/blob/master/HACKING.md

adriancole:
will merge this in a day

PR #288, opened 2016-12-07
Backfills http client instrumentation tests

### LINKED ISSUE #286: Add test for when OkHttp requests are queued about the in-flight request/host limit
We need to make sure that thread state isn't lost when there's a backup in the okhttp request queue. This is particularly important as that's exactly the sort of scenario when tracing is helpful!

adriancole:
This backfills http instrumentation tests generically, so that it is
easier to track what supports what (and also how to actually use the
instrumentation!).

Notably, this makes it clear a number of async http clients lose trace
context when their parent is a local span, and that many http clients
don't close spans on exception!

Fixes #286


adriancole:
cc @cschneider

PR #290, opened 2016-12-08
Support local spans in BraveExecutorService, BraveCallable BraveRunnable

### LINKED ISSUE #289: BraveExecutorService doesn't support local spans
#250 and #286 suggest we don't have anything built-in to address async spans with local parents. BraveExecutorService only accepts a serverSpanBinder, which means that it will work for async http clients who's direct parent is a server span, but not a client span. This has been verified to be an issue with new tests added in #288.

Let's fix it, by adding a builder for brave and setting the local span when setting the server one.

adriancole:
This adds new `wrap()` factory methods to `BraveExecutorService`,
`BraveCallable` and `BraveRunnable`. Notably these support attaching
local spans, needed for any async client span that isn't a direct child
of a server span. This also manages the stack properly so that you can
create a new span inside a callable or runnable without tainting the
prior state.

Fixes #289

### DIFF
--- a/brave-core/src/main/java/com/github/kristofa/brave/BraveCallable.java
+++ b/brave-core/src/main/java/com/github/kristofa/brave/BraveCallable.java
@@ -1,9 +1,12 @@
 package com.github.kristofa.brave;
 
+import com.twitter.zipkin.gen.Span;
 import java.util.concurrent.Callable;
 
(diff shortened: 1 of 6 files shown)


adriancole:
cc @schlosna @cschneider

Issue #205, opened 2016-07-23
Provide hook for request adapters to supply client address annotation


ewhauser:
From a review of #204, it was mentioned that frameworks supplying the client IP should be setting the "CA" annotation on the server span.  With the current request adapter pattern, this is not possible because the interceptor adds the annotation.  This issue will track the work of exposing hooks for the frameworks add this information.  Original comment below:  

> The [ServerRequestInterceptor](https://github.com/openzipkin/brave/blob/master/brave-core/src/main/java/com/github/kristofa/brave/ServerRequestInterceptor.java#L49) makes the call to `setServerReceived()`.  Because of the way the [ServerRequestAdapter](https://github.com/openzipkin/brave/blob/master/brave-core/src/main/java/com/github/kristofa/brave/ServerRequestAdapter.java) interface is setup, there is not a hook to set the client address annotation prior to `setServerReceived()` being called.
> 
> If you think it makes sense, I can add an additional method to `ServerRequestAdapter` that allows implementors to optionally supply the client info so it can be used by `ServerRequestInterceptor`.  We could also use this to enhance the `HttpServerRequestAdapter` as well using the various methods to pull client IPs (i.e. [ServletRequest.html#getRemoteAddr()](http://docs.oracle.com/javaee/6/api/javax/servlet/ServletRequest.html#getRemoteAddr%28%29) and well known proxy headers such as `X-Forwarded-For`)

adriancole:
all instrumentation now add client address (unless they have no means to)

Issue #108, opened 2015-11-08
Support logging client IP with ServerRequestInterceptor via X-Forwarded-For


adriancole:
The right-most ipv4 value in the `X-Forwarded-For` header indicates the client IP address. Seems we should be able to log that portably in ServerRequestInterceptor.

For example, in ServerRequestInterceptor, where `setServerReceived()` is invoked, we can call the overload when we've parsed an IP address correctly.

See #107

adriancole:
One way might be to add this to com.github.kristofa.brave.http.HttpServerRequest:

```java

    /**
     * Returns the client's address, or null if the IP is unknown. The port should be 0 if unknown.
     */
    @Nullable InetSocketAddress getClientAddress();
```

adriancole:
all http server instrumentation log client address in Brave 4.3

Issue #284, opened 2016-12-05
Request and Response tracers


adriancole:
There have been a lot of issues raised relating to what I'd call semi-automatic tracing. We've had requests to affect the tags (binary annotations) which are added incidentally to via http request adapters. We've had requests to defer evaluation of tags unless a request is actually traced. This leads to a centralization of concerns about not just http, but RPC tracing in general. I'm working on an upgrade to our approach, calling it Request and Response tracers.

Status quo is actually pretty good. There are interceptors which are available to cover request/response of both client and server operations. They are pretty straight-forward to implement. There are some challenges due to coupling of propagation details and inheritance-based customization.

For example, we can currently add more trace data to RPC spans made with interceptors by subtyping each library's adapters. While straight-forward, adding "another key" means a sweeping change.

Propagation of span ids change much less frequently and for very different reasons than span details. Having both concepts in the same adapter is convenient, but requires a change in one to affect the other. Moreover sometimes propagation+normal span stuff is awkward. Consider gRPC, you cannot navigate from the method descriptor to the headers, as you usually can with http libraries.

I think the best way out is to do a few things.
* make an opentracing-like mechanism for injecting and extracting span identifiers from requests
* make a "tracer" which works on raw request and response types. For example, it accepts a grpc method descriptor, and something underneath knows how to parse tags from it.
* introduce a key-granularity customization hook, where you can teach the tracer how to parse a tag via a lambda

Instrumentation would look like this.. where real backend types (not wrappers), drive the tracer. Underneath, the tracer knows how to handle these types.
```java
requestTracer = ClientRequestTracer.<Request>builder(b.brave)
    .spanNameParser(Request::method)
    .tagsParser(OkHttpParsers.INSTANCE.requestTagsParser).build();

injector = b.brave.propagation().injector(Request.Builder::header);

--snip--

Request request = chain.request();
Request.Builder requestBuilder = request.newBuilder();

SpanId spanId = requestTracer.startNewSpan(request);
injector.inject(spanId, requestBuilder);
try {
  Response response = chain.proceed(requestBuilder.build());
  return responseTracer.finish(response);
} catch (IOException e) {
  throw responseTracer.finish(e);
}
```

Configuration would be a collaboration between what the type defaults are, and any user-level overrides. These overrides would be specified with a tuple of a java type, a tag key and a lambda function.

This makes it easy to read and cheap to override. Ex.
```java
TraceKeys.HTTP_URL, input -> input.url().toString()
```

[...]

Issue #180, opened 2016-05-20
On Brave 4


adriancole:
We've had some feedback that we break api too much :)

Let's leave 3.7+ api stable, and move master to 4.0.0-SNAPSHOT. As a part of that, we can significantly flatten the packages to brave.X and use the zipkin model classes directly instead of wrapping them.

Here are some features for Brave 4:
* repackage into `brave.*` (and coexist with existing code tree)
  * this allows us to change the existing code base so that it might bridge easier.
* no more public everything
  * this will be designed like effective java which will make maintenance easier over time.
* instead of ClientTracer ServerTracer etc, use decorators on a single Span type
  * this simplifies state problems considerably
* Span will have a start/finish function can can be passed around explicitly
  * it can be noop when unsampled
* Allow a simple in-process propagation api for Span, which allows nesting, but don't require it
  * some frameworks will need to use implicit propagation, but some won't
* replace SpanId with TraceContext which carries identifiers even when unsampled
  * this allows brave to be a pass-through
* make B3 batteries included but optional w/Propagation
  * this allows people to use a new header at some point
* Add Client Server and MessageHandler/interceptor (#284)
  * This simplifies the current http interceptors using native types and not requiring http
* Update the [OpenTracing bridge](https://github.com/openzipkin/brave-opentracing)
  * This api fits into OT much more nicely than before, and hopefully will with future versions of OT.

Bridging to brave 3.x is frankly hard because of thread state management, and most interceptors are better off redone as new ones. The focus on interop will be trying to ensure that traces are continued in-process without breaking. This might require a bridge type.

eirslett:
+1

adriancole:
I think any major designs updates like a new interceptor model should be a part of brave 4 since deprecation and trying to portably wire in things you don't have constructor references for is a fools errand.

here's a writeup on new interceptors/handlers https://github.com/openzipkin/brave/issues/284

adriancole:
cc @openzipkin/core @openzipkin/brave I will have something up this week. basically work trying to change all instrumentation seems shorter by actually fixing some of the root apis.

devinsba:
A thought I've been having is would it make sense to make brave an OpenTracing implementation instead of just having the shim. OT doesn't seem completely mature but I don't see 4.x being an immediate release so that may not be an issue.

adriancole:
[...]

PR #295, opened 2016-12-13
Adds Client Address to grpc and servlet-based instrumentation

adriancole:
This adds the client address to servlet-based instrumentation. This does
so without breaking any apis, by directly modifying the span.

The address heuristic is widely used:
* Prefer `X-Forwarded-For` (which returns the origin's IP)
* Fallback to `ServletRequest.getRemoteAddr()`

This also adds the client address for gRPC using
* `ServerCall.REMOTE_ADDR_KEY`

This doesn't use anything expensive to derive the values. For example,
it doesn't use DNS, nor does it attempt to pick a nice, low cardinality
client service name. This intentionally avoids all of this to get a
working implementation out, which people can use today.

The heaviest lifting of this code was done by copying a method from
guava, notably `com.google.common.net.InetAddresses.ipStringToBytes()`
which was cited in the source comments.

See #264 (for servlet customizations)
See #205 (for generic hook for client address, which this doesn't solve)
See #108 (original request for X-Forwarded-For)

### DIFF
--- a/brave-core/src/main/java/com/github/kristofa/brave/internal/InetAddresses.java
+++ b/brave-core/src/main/java/com/github/kristofa/brave/internal/InetAddresses.java
@@ -1,0 +1,178 @@
+package com.github.kristofa.brave.internal;
+
+import java.net.InetAddress;
+import java.nio.ByteBuffer;
+
+/** Utilities for working with IP Addresses. */
+public final class InetAddresses {
+
+  //** Start code from Guava v20 **//
+  private static final int IPV4_PART_COUNT = 4;
+  private static final int IPV6_PART_COUNT = 8;
+
+  /**
+   * Returns the {@link InetAddress#getAddress()} having the given string representation or null if
+   * unable to parse.
+   *
+   * <p>This deliberately avoids all nameservice lookups (e.g. no DNS).
+   *
+   * <p>This is the same as com.google.common.net.InetAddresses.ipStringToBytes(), except internally
+   * Splitter isn't used (as that would introduce more dependencies).
+   *
+   * @param ipString {@code String} containing an IPv4 or IPv6 string literal, e.g. {@code
+   * "192.168.0.1"} or {@code "2001:db8::1"}
+   */
+  @Nullable
+  public static byte[] ipStringToBytes(String ipString) {
+    // PATCHED! adding null/empty escape
+    if (ipString == null || ipString.isEmpty()) return null;
+    // Make a first pass to categorize the characters in this string.
+    boolean hasColon = false;
+    boolean hasDot = false;
+    for (int i = 0; i < ipString.length(); i++) {
+      char c = ipString.charAt(i);
+      if (c == '.') {
+        hasDot = true;
+      } else if (c == ':') {
+        if (hasDot) {
+          return null; // Colons must not appear after dots.
+        }
+        hasColon = true;
+      } else if (Character.digit(c, 16) == -1) {
+        return null; // Everything else must be a decimal or hex digit.
+      }
+    }
+
+    // Now decide which address family to parse.
+    if (hasColon) {
+      if (hasDot) {
+        ipString = convertDottedQuadToHex(ipString);
+        if (ipString == null) {
+          return null;
+        }
+      }
+      return textToNumericFormatV6(ipString);
+    } else if (hasDot) {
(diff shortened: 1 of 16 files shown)


adriancole:
Here's an example span from https://github.com/openzipkin/brave-webmvc-example

```json
{
  "traceId": "f9a761b72c975eda",
  "id": "f9a761b72c975eda",
  "name": "get",
  "timestamp": 1481641768379000,
  "duration": 523860,
  "annotations": [
    {
      "timestamp": 1481641768379000,
      "value": "sr",
      "endpoint": {
        "serviceName": "brave-webmvc-example",
        "ipv4": "192.168.99.1"
      }
    },
    {
      "timestamp": 1481641768902860,
      "value": "ss",
      "endpoint": {
        "serviceName": "brave-webmvc-example",
        "ipv4": "192.168.99.1"
      }
    }
  ],
  "binaryAnnotations": [
    {
      "key": "ca",
      "value": true,
      "endpoint": {
        "serviceName": "",
        "ipv4": "127.0.0.1",
        "port": 60560
      }
    },
    {
      "key": "http.status_code",
      "value": "200",
      "endpoint": {
        "serviceName": "brave-webmvc-example",
        "ipv4": "192.168.99.1"
      }
    },
    {
      "key": "http.url",
      "value": "/a",
      "endpoint": {
        "serviceName": "brave-webmvc-example",
        "ipv4": "192.168.99.1"
      }
    }
  ]
}
```

adriancole:
cc @ewhauser I think you had something similar for grpc, though not sure if you still use servlets anywhere

llinder:
In regards to issue #264, is the intent to forgo changing APIs and instead plug customizations into individual interceptors like this PR does?  I'm mostly asking to anticipate if there will be more work on a generalized approach in the near term or if I should follow the pattern in this PR.

Besides the question I think this is good as well 👍

ewhauser:
@adriancole: I've got an old half finished branch that attempts to do something similar. A couple of notes:

1. RFC-7239 actually deprecates `X-Forwarded-For` as the header https://tools.ietf.org/html/rfc7239 and uses the `Forwarded` header. There are a number of other variations that the RFC suggests to support
2. I added the method `InetSocketAddress getClientAddress()` to `HttpServerRequest` and then set it via `ServerRequestInterceptor` so that any adapter for HTTP could have a hook to supply the client address. With gRPC, the client address is just available.

Nice find on the `ipStringToBytes` prior work.

devinsba:
+1 on that RFC, had never seen that. Very good to know

adriancole:
Thanks for the tips, folks! I'll work them in.

WRT genericness, I found that attempting to change the current apis portably is endless work. We need to actually make brave 4 (#180) and focus efforts on bridging vs perpetual hacking. I've added this to buy time and also to force the issue in brave 4. I'll be raising a minimal brave 4 today or tomorrow.

While tags should be easy to customize generically (per #284), IP addresses likely will end up hard-wired per library. That's because some frameworks use objects that aren't the same as the request object to access them. Sometimes, they aren't even the response object! Ex OkHttp's connection.

It is a good idea to raise this utility class up a bit so that it can be adapted for other libraries. IOTW, while attaching the IP will likely be best hard-wired, there's no reason to not share code cleaning it up.

adriancole:
@ewhauser actually we can't add anything to HttpServerRequest as it is an interface. This will cause everybody's api to break. I will modify the utility class to work with non-servlet, though.

If you have a sample parser for `Forwarded` please raise it in a follow-up. It looks much more complex than the prior..

```
       Forwarded: for="_gazonk"
       Forwarded: For="[2001:db8:cafe::17]:4711"
       Forwarded: for=192.0.2.60;proto=http;by=203.0.113.43
       Forwarded: for=192.0.2.43, for=198.51.100.17
```

adriancole:
ok made a base implementation and shared it across http and grpc

adriancole:
ps for those who are interested in follow-up, the new forwarded header would look like this..

`Forwarded: for=198.51.100.17;by=203.0.113.60;proto=http;host=example.com`

note that for can include ports and ipv6 like "192.0.2.43:47011" and "[2001:db8:cafe::17]:47011"

There's one parser I found. https://github.com/lpinca/forwarded-parse

we'd also want to normalize how this is used (ex do we do some sort of key-based extractor pattern?). These revisions can come in later.

PR #300, opened 2016-12-14
New core api (Brave 4)

adriancole:
Brave Api (v4) (see #180)
==============

Brave is a library used to capture and report latency information about
distributed operations to Zipkin. Most users won't use Brave directly,
rather libraries or frameworks than employ Brave on their behalf.

This module includes tracer creates and joins spans that model the
latency of potentially distributed work. It also includes libraries to
propagate the trace context over network boundaries, for example, via
http headers.

## Basics
The tracer creates and joins spans that model the latency of potentially
distributed work. It can employ sampling to reduce overhead in process
or to reduce the amount of data sent to Zipkin.

Spans returned by a tracer report data to Zipkin when finished, or do
nothing if unsampled. After starting a span, you can annotate events of
interest or add tags containing details or lookup keys.

Spans have a context which includes trace identifiers that place it at
the correct spot in the tree representing the distributed operation.

### Local Tracing

When tracing local code, just run it inside a span.
```java
try (Span span = tracer.newTrace().name("encode").start()) {
  doSomethingExpensive();
}
```

In the above example, the span is the root of the trace. In many cases,
you will be a part of an existing trace. When this is the case, call
`newChild` instead of `newTrace`

```java
try (Span root = tracer.newTrace().name("2pc").start()) {
  try (Span child = tracer.newChild(root.context()).name("prepare").start()) {
    prepare();
  }
  try (Span child = tracer.newChild(root.context()).name("commit").start()) {
    commit();
  }
}
```

Once you have a span, you can add tags to it, which can be used as lookup
keys or details. For example, you might add a tag with your runtime version:

```java
span.tag("clnt/finagle.version", "6.36.0");
```

### RPC tracing
RPC tracing is often done automatically by interceptors. Under the scenes,
they add tags and events that relate to their role in an RPC operation.

Here's an example of a client span:
```java
// before you send a request, add metadata that describes the operation
span = tracer.newTrace().name("get").type(CLIENT);
span.tag("clnt/finagle.version", "6.36.0");
span.tag(TraceKeys.HTTP_PATH, "/api");
span.remoteEndpoint(Endpoint.builder()
    .serviceName("backend")
    .ipv4(127 << 24 | 1)
    .port(8080).build());

// when the request is scheduled, start the span
span.start();

// if you have callbacks for when data is on the wire, note those events
span.annotate(Constants.WIRE_SEND);
span.annotate(Constants.WIRE_RECV);

// when the response is complete, finish the span
span.finish();
```

## Sampling
Sampling may be employed to reduce the data collected and reported out
of process. When a span isn't sampled, it adds no overhead (noop).

Sampling is an up-front decision, meaning that the decision to report
data is made at the first operation in a trace, and that decision is
propagated downstream.

By default, there's a global sampler that applies a single rate to all
traced operations. `Tracer.Builder.sampler` is how you indicate this,
and it defaults to trace every request.

### Custom sampling

You may want to apply different policies depending on what the operation
is. For example, you might not want to trace requests to static resources
such as images, or you might want to trace all requests to a new api.

Most users will use a framework interceptor which automates this sort of
policy. Here's how they might work internally.

```java
Span newTrace(Request input) {
  SamplingFlags flags = SamplingFlags.NONE;
  if (input.url().startsWith("/experimental")) {
    flags = SamplingFlags.SAMPLED;
  } else if (input.url().startsWith("/static")) {
    flags = SamplingFlags.NOT_SAMPLED;
  }
  return tracer.newTrace(flags);
}
```

## Propagation
Brave has customizable support for in-band propagation of the trace
context. For example, it has built-in support for [B3](https://github.com/openzipkin/b3-propagation),
which has implementations in many languages and frameworks.

Most users will use a framework interceptor which automates propagation.
Here's how they might work internally.

```java
// configure a function that injects a trace context into a request
injector = Propagation.B3_STRING.injector(Request.Builder::addHeader);

// before a request is sent, add the current span's context to it
injector.inject(span.context(), request);

// configure a function that extracts the trace context from a request
extractor = Propagation.B3_STRING.extractor(Request::getHeader);

// when a server receives a request, it joins or starts a new trace
contextOrFlags = extractor.extract(request);
span = contextOrFlags.context() != null
    ? tracer.joinSpan(contextOrFlags.context())
    : tracer.newTrace(contextOrFlags.samplingFlags());
```

## Performance
Brave has been built with performance in mind. Using the core Span api,
you can record spans in sub-microseconds. When a span is sampled, there's
effectively no overhead (as it is a noop).

Unlike previous implementations, Brave 4 only needs one timestamp per
span. All annotations are recorded on an offset basis, using the less
expensive and more precise `System.nanoTime()` function.

## Upgrading from Brave 3
Brave 4 was designed to live alongside Brave 3. Using `TracerAdapter`,
you can navigate between apis, buying you time to update as appropriate.

### Creating a Brave 3 instance
To wire up a mixed-version application, first create a (Brave 4) Tracer.
Then, use `TracerAdapter` to create a (Brave 3) .. Brave.

```java
Tracer brave4 = Tracer.newBuilder()...build();
Brave brave3 = TracerAdapter.newBrave(brave4);
```

### Converting between types
`TracerAdapter.toSpan` is used to navigate between span types.

If you have a reference to a Brave 3 Span, you can convert like this:
```java
brave3Span = brave3.localSpanThreadBinder().getCurrentLocalSpan();
brave4Span = TracerAdapter.toSpan(brave4, brave3Span);
```

You can also attach Brave 4 Span to a Brave 3 thread binder like this:
```java
brave3Span = TracerAdapter.toSpan(brave4Span.context());
brave3.localSpanThreadBinder().setCurrentSpan(brave3Span);
```

### Dependencies
`io.zipkin.brave:brave`(Brave 4) is an optional dependency of
`io.zipkin.brave:brave-core`(Brave 3). To use `TracerAdapter`, you need
to declare an explicit dependency, and ensure both libraries are of the
same version.

### Notes
* Never mutate `com.twitter.zipkin.gen.Span` directly. Doing so will
  break the above integration.
* The above only works when `com.github.kristofa.brave.Brave` was
  instantiated via `TracerAdapter.newBrave`

## 128-bit trace IDs

Traditionally, Zipkin trace IDs were 64-bit. Starting with Zipkin 1.14,
128-bit trace identifiers are supported. This can be useful in sites that
have very large traffic volume, persist traces forever, or are re-using
externally generated 128-bit IDs.

If you want Brave to generate 128-bit trace identifiers when starting new
root spans, set `Tracer.Builder.traceId128Bit(true)`

When 128-bit trace ids are propagated, they will be twice as long as
before. For example, the `X-B3-TraceId` header will hold a 32-character
value like `163ac35c9f6413ad48485a3953bb6124`.

Before doing this, ensure your Zipkin setup is up-to-date, and downstream
instrumented services can read the longer 128-bit trace IDs.

Note: this only affects the trace ID, not span IDs. For example, span ids
within a trace are always 64-bit.

## Acknowledgements
Brave 4's design lends from past experience and similar open source work.
Quite a lot of decisions were driven by portability with Brave 3, and the
dozens of individuals involved in that. There are notable new aspects
that are borrowed or adapted from others.

### Stateful Span object
Brave 4 allows you to pass around a Span object which can report itself
to Zipkin when finished. This is better than using thread contexts in
some cases, particularly where many async hops are in use. The Span api
is derived from OpenTracing, narrowed to more cleanly match Zipkin's
abstraction. As a result, a bridge from Brave 4 to OpenTracing v0.20.2
is relatively little code. It should be able to implement future
versions of OpenTracing as well.

### Recorder architecture
Much of Brave 4's architecture is borrowed from Finagle, whose design
implies a separation between the propagated trace context and the data
collected in process. For example, Brave's MutableSpanMap is the same
overall design as Finagle's. The internals of MutableSpanMap were adapted
from [WeakConcurrentMap](https://github.com/raphw/weak-lock-free).

### Propagation Api
OpenTracing's Tracer type has methods to inject or extract a trace
context from a carrier. While naming is similar, Brave optimizes for
direct integration with carrier types (such as http request) vs routing
through an intermediate (such as a map). Brave also considers propagation
a separate api from the tracer.

### Public namespace
Brave 4's pubic namespace is more defensive that the past, using a package
accessor design from [OkHttp](https://github.com/square/okhttp).

### DIFF
--- a/brave-core/src/main/java/com/github/kristofa/brave/Brave.java
+++ b/brave-core/src/main/java/com/github/kristofa/brave/Brave.java
@@ -151,6 +151,18 @@ public class Brave {
             return this;
         }
 
+        /** Internal hook */
+        Builder spanFactory(SpanFactory spanFactory) {
+            this.spanFactory = spanFactory;
+            return this;
+        }
+
+        /** Internal hook */
+        Builder recorder(Recorder recorder) {
+            this.recorder = recorder;
+            return this;
+        }
+
         /**
          * @deprecated use {@link #reporter(Reporter)}
          */
--- a/brave/README.md
+++ b/brave/README.md
@@ -1,0 +1,267 @@
+Brave Api (v4)
+==============
+
+Brave is a library used to capture and report latency information about
+distributed operations to Zipkin. Most users won't use Brave directly,
+rather libraries or frameworks than employ Brave on their behalf.
+
+This module includes tracer creates and joins spans that model the
+latency of potentially distributed work. It also includes libraries to
+propagate the trace context over network boundaries, for example, via
+http headers.
+
+## Basics
+The tracer creates and joins spans that model the latency of potentially
+distributed work. It can employ sampling to reduce overhead in process
+or to reduce the amount of data sent to Zipkin.
+
+Spans returned by a tracer report data to Zipkin when finished, or do
+nothing if unsampled. After starting a span, you can annotate events of
+interest or add tags containing details or lookup keys.
+
+Spans have a context which includes trace identifiers that place it at
+the correct spot in the tree representing the distributed operation.
+
+### Local Tracing
+
+When tracing local code, just run it inside a span.
+```java
+try (Span span = tracer.newTrace().name("encode").start()) {
+  doSomethingExpensive();
+}
+```
+
+In the above example, the span is the root of the trace. In many cases,
+you will be a part of an existing trace. When this is the case, call
+`newChild` instead of `newTrace`
+
+```java
+try (Span root = tracer.newTrace().name("2pc").start()) {
+  try (Span child = tracer.newChild(root.context()).name("prepare").start()) {
+    prepare();
+  }
+  try (Span child = tracer.newChild(root.context()).name("commit").start()) {
+    commit();
+  }
+}
+```
+
+Once you have a span, you can add tags to it, which can be used as lookup
+keys or details. For example, you might add a tag with your runtime version:
+
+```java
+span.tag("clnt/finagle.version", "6.36.0");
+```
+
+### RPC tracing
+RPC tracing is often done automatically by interceptors. Under the scenes,
+they add tags and events that relate to their role in an RPC operation.
+
+Here's an example of a client span:
+```java
+// before you send a request, add metadata that describes the operation
+span = tracer.newTrace().name("get").type(CLIENT);
+span.tag("clnt/finagle.version", "6.36.0");
+span.tag(TraceKeys.HTTP_PATH, "/api");
+span.remoteEndpoint(Endpoint.builder()
+    .serviceName("backend")
+    .ipv4(127 << 24 | 1)
+    .port(8080).build());
+
+// when the request is scheduled, start the span
+span.start();
+
+// if you have callbacks for when data is on the wire, note those events
+span.annotate(Constants.WIRE_SEND);
+span.annotate(Constants.WIRE_RECV);
+
+// when the response is complete, finish the span
+span.finish();
+```
+
+## Sampling
+Sampling may be employed to reduce the data collected and reported out
+of process. When a span isn't sampled, it adds no overhead (noop).
+
+Sampling is an up-front decision, meaning that the decision to report
+data is made at the first operation in a trace, and that decision is
+propagated downstream.
+
+By default, there's a global sampler that applies a single rate to all
+traced operations. `Tracer.Builder.sampler` is how you indicate this,
+and it defaults to trace every request.
+
+### Custom sampling
+
+You may want to apply different policies depending on what the operation
+is. For example, you might not want to trace requests to static resources
+such as images, or you might want to trace all requests to a new api.
+
+Most users will use a framework interceptor which automates this sort of
+policy. Here's how they might work internally.
+
+```java
+Span newTrace(Request input) {
+  SamplingFlags flags = SamplingFlags.NONE;
+  if (input.url().startsWith("/experimental")) {
+    flags = SamplingFlags.SAMPLED;
+  } else if (input.url().startsWith("/static")) {
+    flags = SamplingFlags.NOT_SAMPLED;
+  }
+  return tracer.newTrace(flags);
+}
+```
+
+## Propagation
+Propagation is needed to ensure activity originating from the same root
+are collected together in the same trace. The most common propagation
+approach is to copy a trace context from a client sending an RPC request
+to a server receiving it.
+
+For example, when an downstream Http call is made, its trace context is
+sent along with it, encoded as request headers:
+
+```
+   Client Span                                                Server Span
+┌──────────────────┐                                       ┌──────────────────┐
+│                  │                                       │                  │
+│   TraceContext   │           Http Request Headers        │   TraceContext   │
+│ ┌──────────────┐ │          ┌───────────────────┐        │ ┌──────────────┐ │
+│ │ TraceId      │ │          │ X─B3─TraceId      │        │ │ TraceId      │ │
+│ │              │ │          │                   │        │ │              │ │
+│ │ ParentSpanId │ │ Extract  │ X─B3─ParentSpanId │ Inject │ │ ParentSpanId │ │
+│ │              ├─┼─────────>│                   ├────────┼>│              │ │
+│ │ SpanId       │ │          │ X─B3─SpanId       │        │ │ SpanId       │ │
+│ │              │ │          │                   │        │ │              │ │
+│ │ Sampled      │ │          │ X─B3─Sampled      │        │ │ Sampled      │ │
+│ └──────────────┘ │          └───────────────────┘        │ └──────────────┘ │
+│                  │                                       │                  │
+└──────────────────┘                                       └──────────────────┘
+```
+
+The names above are from [B3 Propagation](https://github.com/openzipkin/b3-propagation),
+which is built-in to Brave and has implementations in many languages and
+frameworks.
+
+Most users will use a framework interceptor which automates propagation.
+Here's how they might work internally.
+
+```java
+// configure a function that injects a trace context into a request
+injector = Propagation.B3_STRING.injector(Request.Builder::addHeader);
+
+// before a request is sent, add the current span's context to it
+injector.inject(span.context(), request);
+
+// configure a function that extracts the trace context from a request
+extractor = Propagation.B3_STRING.extractor(Request::getHeader);
+
+// when a server receives a request, it joins or starts a new trace
+contextOrFlags = extractor.extract(request);
+span = contextOrFlags.context() != null
+    ? tracer.joinSpan(contextOrFlags.context())
+    : tracer.newTrace(contextOrFlags.samplingFlags());
+```
+
+## Performance
+Brave has been built with performance in mind. Using the core Span api,
+you can record spans in sub-microseconds. When a span is sampled, there's
+effectively no overhead (as it is a noop).
+
+Unlike previous implementations, Brave 4 only needs one timestamp per
+span. All annotations are recorded on an offset basis, using the less
+expensive and more precise `System.nanoTime()` function.
+
+## Upgrading from Brave 3
+Brave 4 was designed to live alongside Brave 3. Using `TracerAdapter`,
+you can navigate between apis, buying you time to update as appropriate.
+
+### Creating a Brave 3 instance
+If your code uses Brave 3 apis, all you need to do is use `TracerAdapter`
+to create a (Brave 3) .. Brave. You don't have to change anything else.
+
+```java
+Tracer brave4 = Tracer.newBuilder()...build();
+Brave brave3 = TracerAdapter.newBrave(brave4);
+```
+
+### Converting between types
+Those coding directly to both apis can use `TracerAdapter.toSpan` to
+navigate between span types.
+
+If you have a reference to a Brave 3 Span, you can convert like this:
+```java
+brave3Span = brave3.localSpanThreadBinder().getCurrentLocalSpan();
+brave4Span = TracerAdapter.toSpan(brave4, brave3Span);
+```
+
+You can also attach Brave 4 Span to a Brave 3 thread binder like this:
+```java
+brave3Span = TracerAdapter.toSpan(brave4Span.context());
+brave3.localSpanThreadBinder().setCurrentSpan(brave3Span);
+```
+
+### Dependencies
+`io.zipkin.brave:brave`(Brave 4) is an optional dependency of
+`io.zipkin.brave:brave-core`(Brave 3). To use `TracerAdapter`, you need
+to declare an explicit dependency, and ensure both libraries are of the
+same version.
+
+### Notes
+* Never mutate `com.twitter.zipkin.gen.Span` directly. Doing so will
+  break the above integration.
+* The above only works when `com.github.kristofa.brave.Brave` was
+  instantiated via `TracerAdapter.newBrave`
+
+## 128-bit trace IDs
+
+Traditionally, Zipkin trace IDs were 64-bit. Starting with Zipkin 1.14,
+128-bit trace identifiers are supported. This can be useful in sites that
+have very large traffic volume, persist traces forever, or are re-using
+externally generated 128-bit IDs.
+
+If you want Brave to generate 128-bit trace identifiers when starting new
+root spans, set `Tracer.Builder.traceId128Bit(true)`
+
+When 128-bit trace ids are propagated, they will be twice as long as
+before. For example, the `X-B3-TraceId` header will hold a 32-character
+value like `163ac35c9f6413ad48485a3953bb6124`.
+
+Before doing this, ensure your Zipkin setup is up-to-date, and downstream
+instrumented services can read the longer 128-bit trace IDs.
+
+Note: this only affects the trace ID, not span IDs. For example, span ids
+within a trace are always 64-bit.
+
+## Acknowledgements
+Brave 4's design lends from past experience and similar open source work.
+Quite a lot of decisions were driven by portability with Brave 3, and the
+dozens of individuals involved in that. There are notable new aspects
+that are borrowed or adapted from others.
+
+### Stateful Span object
+Brave 4 allows you to pass around a Span object which can report itself
+to Zipkin when finished. This is better than using thread contexts in
+some cases, particularly where many async hops are in use. The Span api
+is derived from OpenTracing, narrowed to more cleanly match Zipkin's
+abstraction. As a result, a bridge from Brave 4 to OpenTracing v0.20.2
+is relatively little code. It should be able to implement future
+versions of OpenTracing as well.
+
+### Recorder architecture
+Much of Brave 4's architecture is borrowed from Finagle, whose design
+implies a separation between the propagated trace context and the data
+collected in process. For example, Brave's MutableSpanMap is the same
+overall design as Finagle's. The internals of MutableSpanMap were adapted
+from [WeakConcurrentMap](https://github.com/raphw/weak-lock-free).
+
+### Propagation Api
+OpenTracing's Tracer type has methods to inject or extract a trace
+context from a carrier. While naming is similar, Brave optimizes for
+direct integration with carrier types (such as http request) vs routing
+through an intermediate (such as a map). Brave also considers propagation
+a separate api from the tracer.
+
+### Public namespace
+Brave 4's pubic namespace is more defensive that the past, using a package
+accessor design from [OkHttp](https://github.com/square/okhttp).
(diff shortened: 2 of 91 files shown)


adriancole:
@openzipkin/brave fyi In order to keep line count down, I'm intentionally not adding instrumentation implementations to this PR (eventhough I have done some of them).

adriancole:
ps here's an example of what an instrumentation would look like (note the handlers aren't in this PR, neither is the tag parsing stuff)

```java
public final class HttpClient4Interceptors {
  // still not happy with configuring the parsers, yet.. this is a different PR though
  public static class Config extends ClientHandler.Config<HttpClientContext, HttpClientContext> {

    @Override protected Parser<HttpClientContext, String> spanNameParser() {
      return c -> c.getRequest().getRequestLine().getMethod();
    }

    @Override protected Parser<HttpClientContext, zipkin.Endpoint> responseAddressParser() {
      return new ServerAddressParser("");
    }

    @Override protected TagsParser<HttpClientContext> requestTagsParser() {
      return TagsParser.builder(HttpClientContext.class)
          .addParser(TraceKeys.HTTP_URL, input -> input.getRequest().getRequestLine().getUri())
          .build();
    }

    @Override protected TagsParser<HttpClientContext> responseTagsParser() {
      return TagsParser.builder(HttpClientContext.class)
          .addParser(TraceKeys.HTTP_STATUS_CODE, input -> {
            int httpStatus = input.getResponse().getStatusLine().getStatusCode();
            return httpStatus < 200 || httpStatus > 299 ? String.valueOf(httpStatus) : null;
          }).build();
    }
  }

  // all this still will be set via builder
  Span.Factory spanFactory = Tracer.builder().build(); // add reporter etc
  ClientHandler<HttpClientContext, HttpClientContext> clientHandler =
      ClientHandler.create(new Config());
  TraceContextInjector<HttpMessage> contextInjector =
      Propagation.B3_STRING.injector(HttpMessage::setHeader);

  public HttpRequestInterceptor requestInterceptor(final TraceContext parent) {
    return (request, context) -> {
      Span span = spanFactory.newSpan(parent);
      try {
        clientHandler.handleSend(HttpClientContext.adapt(context), span);
        contextInjector.inject(span.context(), request);
        context.setAttribute(ClientHandler.CONTEXT_KEY, span);
      } catch (RuntimeException e) {
        throw clientHandler.handleError(e, span);
      }
    };
  }

  public HttpResponseInterceptor responseInterceptor() {
    return (response, context) -> {
      Span span = (Span) context.getAttribute(ClientHandler.CONTEXT_KEY);
      if (span == null) return;
      clientHandler.handleReceive(HttpClientContext.adapt(context), span);
    };
  }
}
```

adriancole:
@felixbarny @yschimke fyi

brave/src/main/java/brave/RealSpan.java
```diff
@@ -0,0 +1,219 @@
...
+    return addAnnotation(relativeTimestamp(), value);
+  }
+
+  @Override public Span addAnnotation(long timestamp, String value) {
+    if (value == null) throw new NullPointerException("value == null");
+    setIfLessThanCurrentTimestamp(timestamp);
+    synchronized (lock) {
+      if (annotations == null) annotations = new ArrayList<>();
+      annotations.add(timestamp);
+      annotations.add(value);
```
devinsba: Why are these not tied together by some object? It would seem more natural and less error prone if `annotations` was a `List<Pair>` or something along those lines.

adriancole:
>
> +      annotations.add(timestamp);
> +      annotations.add(value);
>
> Why are these not tied together by some object? It would seem more natural
> and less error prone if annotations was a List<Pair> or something along
> those lines.
>
it is only an internal detail, as this collection isn't used by users. That
said.. wrapping won't hurt benchmarks much and could make a prettier
toString with less effort. I can create a class for this.

adriancole:
>> Why are these not tied together by some object? It would seem more natural
>> and less error prone if annotations was a List<Pair> or something along
>> those lines.
done! benchmarks don't care, and agreed it is less surprising.

adriancole:
Added comparison benchmarks with prior api. Span recording is 15-45% less overhead than before, depending on contention and verbosity of the span.

brave/src/main/java/brave/SpanRecorder.java
```diff
@@ -0,0 +1,92 @@
...
+  }
+
+  void addAnnotation(long timestamp, String value) {
+    // don't report server-side timestamp on shared-spans
+    if (shared && value.equals(Constants.SERVER_RECV)) {
+      recordTimestampAndDuration = false;
+    }
+    String maybeAddressType = END_ANNOTATION_TO_ADDRESS_TYPE.get(value);
+    if (maybeAddressType != null) addressType = maybeAddressType;
+    span.addAnnotation(Annotation.create(timestamp, value, localEndpoint));
```
adriancole: @devinsba I refactored everything so there's no intermediate types used. Current impl is more efficient and less code. Thanks for the prod that led to this. Incidentally, the code is more similar to [zipkin-finagle](https://github.com/openzipkin/zipkin-finagle) now.

```
SpanCreationBenchmarks.no_contention:no_contention_elaborateSpan_brave3       thrpt   15  0.572 ± 0.032  ops/us
SpanCreationBenchmarks.no_contention:no_contention_elaborateSpan_brave4       thrpt   15  0.739 ± 0.035  ops/us
SpanCreationBenchmarks.no_contention:no_contention_simpleRootSpan_brave3      thrpt   15  1.393 ± 0.075  ops/us
SpanCreationBenchmarks.no_contention:no_contention_simpleRootSpan_brave4      thrpt   15  2.067 ± 0.097  ops/us
```

adriancole:
cleaned up the api a bit. for example. I'm using the verb "note" vs the self-referencing and laborious to type "addAnnotation" which literally means add a note

adriancole:
switched verb from "note" to "annotate", as it is still less arduous than addAnnotation, but also using the exact terminology

adriancole:
gonna rejig this to stop encouraging users to have to know zipkin core annotations. Ex in zipkin 2 model we say span.kind = SERVER.. so we can just make that a start option similar to stackdriver trace https://github.com/GoogleCloudPlatform/cloud-trace-java/blob/master/sdk/core/src/main/java/com/google/cloud/trace/core/StartSpanOptions.java#L26

That way, we don't have to worry about implicit stuff at all, and we can have a configuration which in zipkin 1 mode (default) backdates "sr" "ss" etc, but in zipkin 2 mode doesn't have to

adriancole:
added client/server type, paid for by removing the multi-tags api.

`tags(Map)` was racing stripes on a honda. Benchmarks show adding many tags in an uncontended lock vs lock per key is scarcely measurable and certainly not worth making a special api over considering most spans have at most a handful of tags. Any efforts on tagging should focus on conditional value production as it is producing of values that's the expensive part.

adriancole:
updated internals to act like the finagle tracer, basically allow you to act on a span, given its context.

next step: attempt to backport SpanId so that it can be used as a lookup key. then try to rewire the old tracers and `ServerClientAndLocalSpanState` to use the new recorder. That would make a bridge between the old and the new api.

adriancole:
haven't forgotten about this. other zipkin work and refactoring to support this is the reason it has lingered so long. ex #311 will help build the bridge as we can join on span.context() between the old and the new api

adriancole:
retrofitting has been a lot of small changes and deprecations. this won't be done for christmas, but won't be long.

latest effort is discouraging the builders that were made public on ClientTracer and ServerTracer (I remember grudgingly making them public; the excess work in upgrades and refactors are the reason why). #316

adriancole:
next refactoring is around the way that the reference to the current span is obtained #318. This is needed because the new api won't initially include a thread state management api. It will need to hook into the existing one, and having only one makes the job easier.

adriancole:
Added new recorder implementation. This is similar to the one in finagle, but uses a weak map similar to the one @raphw created here https://github.com/raphw/weak-lock-free

brave/src/main/java/brave/internal/recorder/MutableSpanMap.java
```diff
@@ -0,0 +1,134 @@
...
+
+/**
+ * Similar to Finagle's deadline span map, except this is GC pressure as opposed to timeout driven.
+ * This means there's no bookkeeping thread required in order to flush orphaned spans.
+ *
+ * <p>Spans are weakly referenced by their owning context. When the keys are collected, they are
+ * transferred to a queue, waiting to be reported. A call to modify any span will implicitly flush
+ * orphans to Zipkin. Spans in this state will have a "brave.flush" annotation added to them.
+ *
+ * <p>The internal implementation is derived from WeakConcurrentMap by Rafael Winterhalter. See
```
adriancole: @raphw hope this credit works.. the implementation is fairly hacked up, but the principles hold I hope :)

raphw: Thanks for the credit. I appreciate it! I wonder though, why do you rely on the hash code for equality comparison? In this case, you could save yourself from using the weak reference alltogether. The weakness is required for comparison to avoid false positives if a hash code is equal for two instances.

adriancole: ah whoops. hashCode for equals was a mistake. will fix (and backfill a test for it)

adriancole: ok done and backfilled. thanks for the eagle eyes :P

raphw: Another small correction, it should be:
```java
@Override public boolean equals(Object other) {
  TraceContext context = get(), other = ((RealKey) other).get();
  if (context == null) {
    return other == null;
  } else {
    return context.equals(other);
  }
}
```
This way, the reference queue has a chance to clear out the reference. Alternatively, you can do:
```java
@Override public boolean equals(Object other) {
  if (this == other) {
    return true;
  }
  TraceContext context = get();  
  return context != null && context.equals(((RealKey) other).get());
}
```

adriancole: very appreciated. will backfill test on this one, too :P

adriancole:
getting very close. I pared down this PR to what's needed to re-plumb the existing brave 3 interfaces under the new "recorder" internal architecture. For example, this doesn't do any propagation in or out of process.

Should be ready to merge this weekend

brave-core/src/main/java/com/github/kristofa/brave/Brave.java
```diff
@@ -257,17 +257,22 @@ public AnnotationSubmitter serverSpanAnnotationSubmitter() {
     }
 
     private Brave(Builder builder) {
-        SpanFactory spanFactory = builder.spanFactoryBuilder.build();
```
adriancole: changes here yank the old stuff out (while still keeping deprecated code working). was very hard work, but portability is important!

brave/README.md
```diff
@@ -0,0 +1,50 @@
...
+```java
+try (Span span = tracer.newTrace().name("encode").start()) {
+  doSomethingExpensive();
+}
+```
+
+When tracing remote code, note events of interest that explain latency
+```java
+// before you send a request, add metadata that describes the operation
+span = tracer.newTrace().name("get).type(CLIENT);
```
jplock: nit: missing a closing quote

brave/README.md
```diff
@@ -0,0 +1,117 @@
...
+## Sampling
+Sampling is an up-front decision, meaning that the decision to report
+data is made at the first operation in a trace, and that decision is
+propagated downstream.
+
+By default, there's a global sampler that applies a single rate to all
+traced operations. `Tracer.Builder.sampler` is how you indicate this,
+and it defaults to trace every request.
+
+### Custom sampling
```
adriancole: @robinst here's how custom sampling would work under the covers in brave 4

adriancole:
updated the design, calling specifically out how to do custom sampling (which you could do in Brave 3 using the TraceData type)

### Custom sampling

You may want to apply different policies depending on what the operation
is. For example, you might not want to trace requests to static resources
such as images, or you might want to trace all requests to a new api.
Most users will use a framework interceptor which automates this sort of
policy. Here's how they might work internally.

```java
Span newTrace(Request input) {
  SamplingFlags flags = SamplingFlags.NONE;
  if (input.url().startsWith("/experimental")) {
    flags = SamplingFlags.SAMPLED;
  } else if (input.url().startsWith("/static")) {
    flags = SamplingFlags.NOT_SAMPLED;
  }
  return tracer.newTrace(flags);
}
```

adriancole:
made the new api an optional dependency (which means no risk to existing code).

next step: finish tests of the new adapter and write up docs for it

adriancole:
reworked until brave4 can underpin brave3 it without referencing any internals from brave4. This is important as it will keep things from breaking later.

brave/README.md
robinst: Sounds good!

adriancole:
@devinsba you once asked about opentracing. The current revision can back the opentracing 0.20.7 api in <500 lines of code including imports.

Interop between brave3 brave4 and opentracing 0.20.7 works as the below code all produces the same resulting span:

```java
  @Test public void startWithOpenTracingAndFinishWithBrave() {
    io.opentracing.Span openTracingSpan = opentracing.buildSpan("encode")
        .withTag(Constants.LOCAL_COMPONENT, "codec")
        .withStartTimestamp(1L).start();

    brave.Span braveSpan = ((BraveSpan) openTracingSpan).unwrap();

    braveSpan.annotate(2L, "pump fake");
    braveSpan.finish(3L);

    checkSpanReportedToZipkin();
  }

  @Test public void startWithBraveAndFinishWithOpenTracing() {
    brave.Span brave4Span = brave4.newTrace().name("encode")
        .tag(Constants.LOCAL_COMPONENT, "codec")
        .start(1L);

    io.opentracing.Span openTracingSpan = BraveSpan.wrap(brave4Span);

    openTracingSpan.log(2L, "pump fake");
    openTracingSpan.finish(3L);

    checkSpanReportedToZipkin();
  }

  @Test public void startWithLocalTracerAndFinishWithTracer() {
    SpanId spanId = brave3.localTracer().startNewSpan("codec", "encode", 1L);

    brave.Span span = toSpan(brave4, spanId);

    span.annotate(2L, "pump fake");
    span.finish(3L);

    checkSpanReportedToZipkin();
  }

  @Test public void startWithTracerAndFinishWithLocalTracer() {
    brave.Span brave4Span = brave4.newTrace().name("encode")
        .tag(Constants.LOCAL_COMPONENT, "codec")
        .start(1L);

    com.twitter.zipkin.gen.Span brave3Span = toSpan(brave4Span.context());
    brave3.localSpanThreadBinder().setCurrentSpan(brave3Span);

    brave3.localTracer().submitAnnotation("pump fake", 2L);
    brave3.localTracer().finishSpan(2L);

    checkSpanReportedToZipkin();
  }
```

adriancole:
added propagation classes. I've only added example tests for OpenTracing at the moment, though I have ones for http and grpc saved off. Will add later

brave/src/main/java/brave/propagation/TraceContext.java
```diff
@@ -0,0 +1,147 @@
...
+   */
+  public interface Extractor<C> {
+
+    /**
+     * Returns either a trace context or sampling flags parsed from the carrier. If nothing was
+     * parsable, sampling flags will be set to {@link SamplingFlags#EMPTY}.
+     *
+     * @param carrier holds the propagated keys. For example, an incoming message or http request.
+     */
+    TraceContextOrSamplingFlags extract(C carrier);
```
adriancole: This is the direct replacement for the class TraceData

adriancole:
ok updated readme after adding grpc and async feature tests

brave/README.md
```diff
@@ -0,0 +1,153 @@
...
+  if (input.url().startsWith("/experimental")) {
+    flags = SamplingFlags.SAMPLED;
+  } else if (input.url().startsWith("/static")) {
+    flags = SamplingFlags.NOT_SAMPLED;
+  }
+  return tracer.newTrace(flags);
+}
+```
+
+## Propagation
```
adriancole: Here's the new propagation library

adriancole:
OK the code is ready. I'm backfilling one last unit test

adriancole:
added upgrade instructions

brave/README.md
```diff
@@ -0,0 +1,240 @@
...
+## Performance
+Brave has been built with performance in mind. Using the core Span api,
+you can record spans in sub-microseconds. When a span is sampled, there's
+effectively no overhead (as it is a noop).
+
+Unlike previous implementations, Brave 4 only needs one timestamp per
+span. All annotations are recorded on an offset basis, using the less
+expensive and more precise `System.nanoTime()` function.
+
+## Upgrading from Brave 3
```
adriancole: Upgrade instructions

Issue #294, opened 2016-12-13
Report of trace not continuing between CXF Web Services


adriancole:
If someone can validate this, it would be most appreciated. Basically to know if this is a code concern (bug or usage bug), configuration one (propagation state disconnect due to something not being wrapped), or a misunderstood normal behavior (like the trace ids being different is normal)

cc @cschneider @reta @podsiedzik

Copying this from a gitter post from @auntscrewy 
> Here is my junit test. I want to call a service that calls a service. I believe I would want the traceId to be the same. Right?
```java
package com.github.kristofa.brave.cxf3.integration;
import static org.assertj.core.api.Assertions.assertThat;
import static org.junit.Assert.assertEquals;
import javax.jws.WebMethod;
import org.apache.cxf.endpoint.Server;
import org.apache.cxf.jaxws.JaxWsProxyFactoryBean;
import org.apache.cxf.jaxws.JaxWsServerFactoryBean;
import org.junit.After;
import org.junit.Before;
import org.junit.Test;
import com.github.kristofa.brave.Brave;
import com.github.kristofa.brave.cxf3.BraveClientInInterceptor;
import com.github.kristofa.brave.cxf3.BraveClientOutInterceptor;
import com.github.kristofa.brave.cxf3.BraveServerInInterceptor;
import com.github.kristofa.brave.cxf3.BraveServerOutInterceptor;
import com.github.kristofa.brave.cxf3.ReporterForTesting;
public class ITBraveCxfInterceptorsJws
{
Server server;
FooService client;
static ReporterForTesting reporter = new ReporterForTesting();
Brave brave = new Brave.Builder().reporter(reporter).build();
BraveServerInInterceptor serverInInterceptor = BraveServerInInterceptor.create(brave);
BraveServerOutInterceptor serverOutInterceptor = BraveServerOutInterceptor.create(brave);
BraveClientInInterceptor clientInInterceptor = BraveClientInInterceptor.create(brave);
BraveClientOutInterceptor clientOutInterceptor = BraveClientOutInterceptor.create(brave);
public interface FooService
{
    @WebMethod
    String foo();
}

public static class DefaultFooService implements FooService
{
    private static int cnt = 0;
    // ReporterForTesting reporter = new ReporterForTesting();
    Brave brave = new Brave.Builder().reporter(reporter).build();
    BraveClientInInterceptor clientInInterceptor = BraveClientInInterceptor.create(brave);
[...]

PR #304, opened 2016-12-19
Attempt to fix trace id propagation. Issue #294

podsiedzik:
@adriancole 

Possible solution:
* during sync processing are passed by ThreadLocal
* during async jax-rs processing, current thread will is cleaned, to pass ServerSpan from Exchange to current Thread BraveCxfExecutorService could be used.


brave-cxf3/src/main/java/com/github/kristofa/brave/cxf3/BraveCxfExecutorService.java
```diff
@@ -0,0 +1,157 @@
...
+  public <T> T invokeAny(Collection<? extends Callable<T>> tasks) throws InterruptedException, ExecutionException {
+    return delegate.invokeAny(wrap(tasks));
+  }
+
+  @Override
+  public <T> T invokeAny(Collection<? extends Callable<T>> tasks, long timeout, TimeUnit unit) throws InterruptedException, ExecutionException, TimeoutException {
+    return delegate.invokeAny(wrap(tasks), timeout, unit);
+  }
+
+  private <T> T getFromExchange(String key) {
```
adriancole: Is it possible to add a filter before execution which does this? ex it copies the span from the current message into the server span thread binder? That way, the normal brave executor service would work (unless I'm missing something)

brave-cxf3/README.md
```diff
@@ -65,7 +61,6 @@ CXF JAX-RS client setup:
 JAXRSClientFactoryBean clientFactory = new JAXRSClientFactoryBean();
 clientFactory.setAddress("http://localhost:9001/");
 clientFactory.setServiceClass(FooService.class);
-clientFactory.getInInterceptors().add(BraveClientInInterceptor.create(brave));
-clientFactory.getOutInterceptors().add(BraveClientOutInterceptor.create(brave));
+clientFactory.getFeatures().add(BraveClientFeature.create(brave));
```
adriancole: we'd end up needing to document whatever the thread concern is here.

brave-cxf3/src/main/java/com/github/kristofa/brave/cxf3/BraveCxfExecutorService.java
```diff
@@ -0,0 +1,157 @@
...
+import java.util.Collection;
+import java.util.List;
+import java.util.concurrent.*;
+
+/**
+ * Wraps around {@link ExecutorService} and passes {@link ServerSpan}
+ * from cxf {@link org.apache.cxf.message.Exchange} context (if exists)
+ * to current {@link ServerSpanThreadBinder} during execution.
+ */
+public class BraveCxfExecutorService implements ExecutorService, Closeable {
```
reta: I don't think you need to mimic executor here. The `wrap` methods family is enough, they everyone to wrap callable and runnable and are executor-agnostic.

brave-cxf3/src/main/java/com/github/kristofa/brave/cxf3/BraveAsyncJaxRsThreadCleaner.java
```diff
@@ -0,0 +1,44 @@
...
+import javax.ws.rs.container.ResourceInfo;
+import javax.ws.rs.container.Suspended;
+import javax.ws.rs.core.Context;
+import java.io.IOException;
+import java.lang.annotation.Annotation;
+
+/**
+ * Cleans current server thread binder in async scenarios.
+ */
+class BraveAsyncJaxRsThreadCleaner implements ContainerRequestFilter {
```
reta: Why do you need this guy? `BraveServerInInterceptor` should do the job, right?

brave-cxf3/src/main/java/com/github/kristofa/brave/cxf3/BraveCxfExecutorService.java
```diff
@@ -0,0 +1,157 @@
...
+
+  private final ExecutorService delegate;
```
reta: Now, if you make this class a contextual one (implementing ContextProvider), JAX-RS resource may be able to inject it using `@Context` whenever tracing capabilities are required. Client will use the `wrap` methods in case of own thread pools or whatnot.

brave-cxf3/src/main/java/com/github/kristofa/brave/cxf3/BraveCxfExecutorService.java
podsiedzik: I don't think so, because in async scenario RequestFilter would be executed on Thread1 and Response will be finished on Thread2, thus T1 wouldn't be cleaned.
Eventually we could get rid of BraveAsyncJaxRsThreadCleaner and BraveCxfExecutorService could pass value from T1 to T2, and finally clear T1.

brave-cxf3/src/main/java/com/github/kristofa/brave/cxf3/BraveAsyncJaxRsThreadCleaner.java
podsiedzik: It could do it, but I wanted clean dependency. If someone uses only jaxws, jaxrs dependency would be excessive. But we can merge those two if it's not a problem.

brave-cxf3/src/main/java/com/github/kristofa/brave/cxf3/BraveCxfExecutorService.java
podsiedzik: You're right but it may lead to other problems. When FooService is singleton, Executor will be injected as ThreadLocalProxy it means that Executor will be available only 
on current Thread. When someone would want to use it to execute any other async processing inside it will fail.

```java
  public class FooService {

    @Context
    private Executor executor;

    @POST
    @Path("/foo")
    public void foo(@Suspended AsyncResponse response) {
      // executor is taken from ThreadLocal
      executor.execute(() -> {
      	// executor is not available.
        response.resume("foo " + client.bar());
      });
    }
  }
```
But taking into account your other comment. 
We could change BraveCxfExecutorService to AsyncContext (or something like this) and stop implementing any interface, we could also clean ServerSpan there.

```java
  public class FooService {

    @Context
    private AsyncContext asyncContext;

    @POST
    @Path("/foo")
    public void foo(@Suspended AsyncResponse response) {   
      asyncContext.wrap(() -> {
        executor.execute(() -> {
          response.resume("bar");
        });
      });
    }
  }
```

reta: You actually would better do that:
```
public void foo(@Context TracerContext context, @Suspended AsyncResponse response) {   
        executor.execute(context.wrap(() -> {
          response.resume("foo");
        });
      });
    }
```

In this case the executor is application specific, while `tracerContext` wraps callable / runnable into scope. Plus, it will work with any executor :-) What do you think?

adriancole:
ps thanks for working these things out together.. please ping me
directly when you want a general review (I don't have much insight
into cxf ) cc @cschneider

adriancole:
ps thanks for the help, folks. Once this is in, I'll cut a release.

reta:
@podsiedzik What do you think about converting `BraveCxfExecutorService` to contextual object? All the necessary scaffolding will be provided (`wrap` methods), plus the additional simplification includes removing `getFromExchange` method. Context providers are created from `Message` instance, so you could pass saved span directly to the contextual object instance directly.

podsiedzik:
@reta

Yes, I tought about it yesterday, but didn't mange to commit :)
After your reviews I've changed:

* Created `TracerContext` with `TracerContextProvider`, provider gets server span from message exchange and passes it to TracerContext.
* TracerContext clears current server span and wraps provided Runnable and passes ServerSpan to its context.
* BraveCxfExecutorService and BraveAsyncJaxRsThreadCleaner are removed.

brave-cxf3/src/main/java/com/github/kristofa/brave/cxf3/TracerContext.java
```diff
@@ -0,0 +1,67 @@
...
+   * @param command {@link Callable}
+   * @return wrapped {@link Callable}
+   */
+  public <V> Callable<V> wrap(Callable<V> command) {
+    if (command == null) {
+      return null;
+    }
+
+    // we should clean this thread, ServerSpan will finish on another one
+    serverSpanThreadBinder.setCurrentSpan(null);
```
reta: You probably should not be doing that. May be something like this:
```
final ServerSpan spanToUse = serverSpanThreadBinder.getCurrentSpan();
if (spanToUse == null) {
    spanToUse = serverSpan;
}
```

The span for current thread should not be affected anyhow. You can call `wrap` as many times as you want, but span should be propagated to all wrapped runnables/callables. What do you think?

reta:
@podsiedzik Overall looks much simpler and better (as for me), should be good once we figure out how to keep current thread unaffected (posted a comment about that). Thanks!

brave-cxf3/src/main/java/com/github/kristofa/brave/cxf3/TracerContext.java
podsiedzik: So `TracerContext#wrap` implementation should look almost the same as before (`BraveCxfExecutorService#warp`)?

```java
  public Runnable wrap(Runnable command) {
    if (command == null) {
      return null;
    }

    final ServerSpan previous = serverSpanThreadBinder.getCurrentServerSpan();

    return () -> {
      try {
        if (serverSpan != null) {
          serverSpanThreadBinder.setCurrentSpan(serverSpan);
        }
        command.run();
      } finally {
        if (serverSpan != null) {
          serverSpanThreadBinder.setCurrentSpan(previous);
        }
      }
    };
  }
```

So current thread wouldn't be affected, but how to clean it then?

reta: This part is loosing the current span (aka `previous`):
```
if (serverSpan != null) {
     serverSpanThreadBinder.setCurrentSpan(serverSpan);
}
```

So if the `previous` span is not null, than user either created one by himself (not good but possible) or it is a sync call, nothing to do here.  This is why I suggested something like that:
```

if (previous == null) {
    serverSpanThreadBinder.setCurrentSpan(serverSpan);
}
```
The `previous` should be always null in case of async call.

However, no matter what `previous`  one is, you are restoring it in the finally block:
```
serverSpanThreadBinder.setCurrentSpan(previous);
```

Hope it makes sense.

podsiedzik: Yes it does.
I get it now. It should look like this, right? Also added special handling of `ServerSpan#EMPTY` in `getCurrentServerSpan`.


```java
  public Runnable wrap(Runnable command) {
    if (command == null) {
      return null;
    }

    // get server span bind to current thread
    final ServerSpan current = getCurrentServerSpan();

    return () -> {
      try {
        // When current span is not null, it means it was bind manually by developer
        // or this wrapper is used in sync scenario.
        // Bind server span from Exchange only when current span is null.
        if (current == null) {
          serverSpanThreadBinder.setCurrentSpan(serverSpan);
        }
        command.run();
      } finally {
        serverSpanThreadBinder.setCurrentSpan(current);
      }
    };
  }

  private ServerSpan getCurrentServerSpan() {
    // gets existing server span or reinitializes to ServerSpan#EMPTY using ThreadLocal#initialValue
    final ServerSpan current = serverSpanThreadBinder.getCurrentServerSpan();

    // when value was just initialized
    if (ServerSpan.EMPTY == current) {
      serverSpanThreadBinder.setCurrentSpan(null);
      return null;
    }
    return current;
  }
```

But it also means, that we still need find out way to clean server span from thread in async scenarios, 
because code fragment below (taken from `BraveServerInInterceptor`) doesn't do it's job.

```java
if (!message.getExchange().isSynchronous()) {
  threadBinder.setCurrentSpan(null);
}
```

That's why I've previously added 'BraveAsyncJaxRsThreadCleaner':

```java
class BraveAsyncJaxRsThreadCleaner implements ContainerRequestFilter {		
		
  @Context		
  private ResourceInfo resourceInfo;		
  private final ServerSpanThreadBinder serverThreadBinder;		
		
  BraveAsyncJaxRsThreadCleaner(Brave brave) {		
    this.serverThreadBinder = brave.serverSpanThreadBinder();		
  }		
		
  @Override		
  public void filter(ContainerRequestContext containerRequestContext) throws IOException {		
    if (isAsyncResponse()) {		
      serverThreadBinder.setCurrentSpan(null);		
    }		
  }		
		
  protected boolean isAsyncResponse() {		
    for (final Annotation[] annotations : resourceInfo.getResourceMethod().getParameterAnnotations()) {		
      for (final Annotation annotation : annotations) {		
        if (annotation.annotationType().equals(Suspended.class)) {		
          return true;		
        }		
      }		
    }		
    return false;		
  }		
}
```

Should I revert it's removal or do you have other idea what to do in such case?

reta: @podsiedzik You are very right, I remember now, we hit the same issue at the time, `exchange.isSynchronous()` seems to be unreliable in case of JAX-RS. Yes, I think returning `BraveAsyncJaxRsThreadCleaner` would make sense. How it would co-exist with `BraveServerInInterceptor`? If it will be called before, the current span will be lost, right?

What would be an ideal solution is to have `BraveJaxRsRequestFilter`

podsiedzik: @reta 

In current setup `BraveServerInInterceptor` logic must be executed before `TraceContextProvider#createContext` is invoked. In that time, Exchange doesn't have any info about resource method and it's parameters' annotations (@Suspended). 
We don't know whether this is async or not.

If `TracerContext#wrap` could resolve ServerSpan from `PhaseInterceptorChain` and not from field assigned by constructor (as before), then order of interceptor invokation could be different. 
`BraveServerInInterceptor` could be called after resource method info is put into `Exchange`, 
so it would know whether to clean thread or not, and there wouldn't be any need for extra filter classes. What do you think?

reta: I see, thanks for clarification. I would suggest to branch off JAX-RS and JAX-WS parts: using request/response container filters in JAX-RS and interceptors in JAX-WS. In this case 95% of the code could be shared (using abstract classes f.e.) but the end solution would be very well aligned with the JAX-RS / JAX-WS. A little bit more work though, what to you think, @podsiedzik ?

podsiedzik: @reta

I've separated jax-rs and jax-ws a little bit, they are still based on cxf interceptors.

To summarize what was changed as a whole:

* traceId is propagated
* is async scenarios
	* `TracerContext` must be used to propagate traceId to other thread
	* initial thread is cleaned
* cxf server features, separate for jax-ws and jax-rs
	* `BraveJaxRsServerFeature`
	* `BraveJaxWsServerFeature`
	* `cxf-rt-frontend-jaxws` and `cxf-rt-frontend-jaxrs` deps are in provided scope, if someone want to use only jax-ws, there's no dependency for jax-rs stuff.
* cxf feature for clients `BraveClientFeature`

reta:
@podsiedzik Looks good to me, great job. @adriancole what do you think?

brave-cxf3/src/main/java/com/github/kristofa/brave/cxf3/AbstractBraveServerInInterceptor.java
```diff
@@ -0,0 +1,45 @@
...
+import com.github.kristofa.brave.http.HttpServerRequestAdapter;
+import com.github.kristofa.brave.http.SpanNameProvider;
+import org.apache.cxf.interceptor.Fault;
+import org.apache.cxf.message.Message;
+import org.apache.cxf.phase.AbstractPhaseInterceptor;
+import org.apache.cxf.phase.Phase;
+
+import static com.github.kristofa.brave.cxf3.BraveCxfConstants.BRAVE_SERVER_SPAN;
+
+public abstract class AbstractBraveServerInInterceptor extends AbstractPhaseInterceptor<Message> {
```
adriancole: don't think this needs to be public

brave-cxf3/src/main/java/com/github/kristofa/brave/cxf3/BraveClientFeature.java
```diff
@@ -0,0 +1,27 @@
...
+
+import com.github.kristofa.brave.Brave;
+import org.apache.cxf.Bus;
+import org.apache.cxf.feature.AbstractFeature;
+import org.apache.cxf.interceptor.InterceptorProvider;
+
+/**
+ * Configures cxf client with brave interceptors.
+ */
+public class BraveClientFeature extends AbstractFeature {
```
adriancole: eventhough you've hid the constructor, finaling public types is a nice reinforcement

brave-cxf3/src/main/java/com/github/kristofa/brave/cxf3/BraveJaxRsServerFeature.java
```diff
@@ -0,0 +1,36 @@
...
+import org.apache.cxf.endpoint.Server;
+import org.apache.cxf.feature.AbstractFeature;
+import org.apache.cxf.jaxrs.provider.ServerProviderFactory;
+
+import java.util.Arrays;
+
+/**
+ * Configures cxf jax-rs server with brave interceptors.
+ */
+public class BraveJaxRsServerFeature extends AbstractFeature {
```
adriancole: same on final

brave-cxf3/src/main/java/com/github/kristofa/brave/cxf3/BraveJaxRsServerFeature.java
```diff
@@ -0,0 +1,36 @@
...
+  protected final Brave brave;
```
adriancole: is this intentionally protected?

brave-cxf3/src/main/java/com/github/kristofa/brave/cxf3/BraveJaxRsServerInInterceptor.java
```diff
@@ -0,0 +1,65 @@
...
+import org.apache.cxf.jaxrs.interceptor.JAXRSInInterceptor;
+import org.apache.cxf.jaxrs.model.OperationResourceInfo;
+import org.apache.cxf.message.Message;
+
+import javax.ws.rs.container.Suspended;
+import java.lang.annotation.Annotation;
+
+import static com.github.kristofa.brave.internal.Util.checkNotNull;
+
+public class BraveJaxRsServerInInterceptor extends AbstractBraveServerInInterceptor {
```
adriancole: nit final

brave-cxf3/src/main/java/com/github/kristofa/brave/cxf3/BraveJaxWsServerFeature.java
```diff
@@ -0,0 +1,28 @@
...
+
+import com.github.kristofa.brave.Brave;
+import org.apache.cxf.Bus;
+import org.apache.cxf.endpoint.Server;
+import org.apache.cxf.feature.AbstractFeature;
+
+/**
+ * Configures cxf jax-ws server with brave interceptors.
+ */
+public class BraveJaxWsServerFeature extends AbstractFeature {
```
adriancole: nit final

brave-cxf3/src/main/java/com/github/kristofa/brave/cxf3/BraveJaxWsServerFeature.java
```diff
@@ -0,0 +1,28 @@
...
+  protected final Brave brave;
```
adriancole: same.. remove protected

brave-cxf3/src/main/java/com/github/kristofa/brave/cxf3/TracerContext.java
```diff
@@ -0,0 +1,100 @@
...
+import com.github.kristofa.brave.ServerSpan;
+import com.github.kristofa.brave.ServerSpanThreadBinder;
+import org.apache.cxf.message.Message;
+
+import java.util.concurrent.Callable;
+
+/**
+ * Used to handle {@link ServerSpan} retrieved from cxf message exchange.
+ */
+public class TracerContext {
```
adriancole: final

brave-cxf3/src/main/java/com/github/kristofa/brave/cxf3/TracerContext.java
```diff
@@ -0,0 +1,100 @@
...
+
+  /**
+   * Wraps provided {@link Runnable and ensures same {@link ServerSpan} context as the thread from which the Runnable was executed.
+   * {@link ServerSpan} is taken from cxf {@link org.apache.cxf.message.Message} {@link org.apache.cxf.message.Exchange}
+   *
+   * @param command {@link Runnable}
+   * @return wrapped {@link Runnable}
+   */
+  public Runnable wrap(Runnable command) {
+    if (command == null) {
```
adriancole: why permit null?

brave-cxf3/src/main/java/com/github/kristofa/brave/cxf3/TracerContext.java
```diff
@@ -0,0 +1,100 @@
...
+
+  /**
+   * Wraps provided {@link Callable} and ensures same {@link ServerSpan} context as the thread from which the Callable was executed.
+   * {@link ServerSpan} is taken from cxf {@link org.apache.cxf.message.Message} {@link org.apache.cxf.message.Exchange}
+   *
+   * @param command {@link Callable}
+   * @return wrapped {@link Callable}
+   */
+  public <V> Callable<V> wrap(Callable<V> command) {
+    if (command == null) {
```
adriancole: why permit null?

brave-cxf3/src/main/java/com/github/kristofa/brave/cxf3/TracerContext.java
adriancole: maybe rename this to CurrentSpanContext?

brave-cxf3/README.md
adriancole: Add a section how to use TracerContext/CurrentSpanContext, especially when you would need to use it

brave-cxf3/src/main/java/com/github/kristofa/brave/cxf3/TracerContext.java
```diff
@@ -0,0 +1,100 @@
...
+        }
+        return command.call();
+      } finally {
+        serverSpanThreadBinder.setCurrentSpan(current);
+      }
+    };
+  }
+
+  private ServerSpan getCurrentServerSpan() {
+    // gets existing server span or reinitializes to ServerSpan#EMPTY using ThreadLocal#initialValue
```
adriancole: looking at this again.. this smells of working around an upstream behavior.. Are we inconsistently setting the thread local value to EMPTY or null?

brave-cxf3/src/main/java/com/github/kristofa/brave/cxf3/TracerContext.java
```diff
@@ -0,0 +1,100 @@
...
+   * @param command {@link Callable}
+   * @return wrapped {@link Callable}
+   */
+  public <V> Callable<V> wrap(Callable<V> command) {
+    if (command == null) {
+      return null;
+    }
+
+    // get server span bind to current thread
+    final ServerSpan current = getCurrentServerSpan();
```
adriancole: probably you want to say previous

brave-cxf3/src/main/java/com/github/kristofa/brave/cxf3/TracerContext.java
podsiedzik: So if I understood correctly, `CurrentSpanContext` would look like this?

```java
public final class CurrentSpanContext {

  private final ServerSpanThreadBinder serverSpanThreadBinder;
  private final Message message;

  CurrentSpanContext(Brave brave, Message message) {
    this.serverSpanThreadBinder = brave.serverSpanThreadBinder();
    this.message = message;
  }

  public void attach() {
    // get server span bind to current thread
    final ServerSpan current = serverSpanThreadBinder.getCurrentServerSpan();

    // when no server span is attached used this from exchange
    if (current == null || ServerSpan.EMPTY == current) {
      final ServerSpan fromExchange = getFromExchange();
      serverSpanThreadBinder.setCurrentSpan(fromExchange);
    }
  }

  private <T> T getFromExchange() {
    return (T) message.getExchange().get(BraveCxfConstants.BRAVE_SERVER_SPAN);
  }
}
```

and used like this:

```java
public class FooEndpoint {
    private final Executor executor;

    public FooService(Executor executor) {
      this.executor = executor;
    }

    @POST
    @Path("/foo")
    public void foo(@Suspended AsyncResponse response, @Context CurrentSpanContext context) {
      executor.execute(() -> {
      	// attaches server span from exchange
        context.attach();
        
        response.resume("foo");
      });
    }
  }
```

adriancole:
>
> and used like this:
>
almost .. fixing below

> public class FooEndpoint {
>     private final ExecutorService executor;
>
>     public FooService(ExecutorService executor, Brave brave) { // or inject a brave'd one directly
>       this.executor = BraveExecutorService.wrap(executor, brave);
>
>     }
>
>     @POST
>     @Path("/foo")
>     public void foo(@Suspended AsyncResponse response, @Context CurrentSpanContext context) {
>
>       	// attaches server span from exchange
        context.attach();

      executor.execute(() -> {
>         response.resume("foo");
>       });
>     }
>   }
>
>
So the context is attached prior to executor.execute().. unless you
intentionally don't want the server context to be available for other
things.
For example, if you attach it at the entry of the method, it doesnt' matter
if you use serverTracer directly or a brave executor service.. it will work!

It seems that it should also be possible to do the same with an
interceptor, too. Ex. AttachSpanInterceptor which is configured to run
before a suspended method.. Anyway if it isn't the above is what I was
thinking about and would be what we document..

reta:
@adriancole Sorry, but I would like to argue (please feel free to challenge/disagree) that using something like `wrap` for Callable / Runnable is more clean and convenient approach that `context.attach`. Let us take these two examples:
```
      context.attach();
      executor.execute(() -> {
         response.resume("foo");
       });
```
And
```
      executor.execute(context.wrap(() -> {
         response.resume("foo");
       }));
```

There are quite a few arguments but I would like to present just a couple most important ones (as for me). First of all, ` context.attach()` essentially becomes an implicit precondition for `executor.execute` however there is no link between these two calls. Secondly, `context.attach()` should be followed to `context.detach()` at some point, this is probably what most of the people would expect, right? And lastly, let say I don't want to submit the execution block immediately but rather queue it? That would require the queuing mechanism to be aware of context in order to submit the callable / runnable when time comes. The `context.wrap` essentially returns a value, which could be passed around and used anywhere without the fear forgetting to call something before. 

What do you think? 
Thanks!

Best Regards,
    Andriy Redko

adriancole:
>
> @adriancole <https://github.com/adriancole> Sorry, but I would like to
> argue (please feel free to challenge/disagree) that using something like
> wrap for Callable / Runnable is more clean and convenient approach that
> context.attach. Let us take these two examples:
>
fair play. I'm going to respond in the line of re-using vs creating new
Brave concepts.

>       context.attach();
>       executor.execute(() -> {
>          response.resume("foo");
>        });
>
>       executor.execute(context.wrap(() -> {
>          response.resume("foo");
>        }));
>
>  don't forget!

      context.attach();
      brave.serverTracer().submitAnnotation("foo"); // or any brave 4
variant that comes in the future!!

 There are quite a few arguments but I would like to present just a couple
> most important ones (as for me). First of all, context.attach()
> essentially becomes an implicit precondition for executor.execute however
> there is no link between these two calls.
>
So is annotating the method, right? Seems odd that we have to annotate the
method and also do anything else to associate the span with the current
context. Why is it impossible for us to make a filter do this?

> Secondly, context.attach() should be followed to context.detach() at some
> point, this is probably what most of the people would expect, right?
>
Look carefully at the implementation. We are only adding the server span
when it wasn't previously associated with the thread. If there is some
in/out interceptor prior, we can accomplish the same implicitly. Moreover,
brave3 is filled with code that doesn't unattach the server span from the
thread. In fact that's the default! Look at how Brave's ServerTracer is
written, and you'll notice that this is status quo.

And lastly, let say I don't want to submit the execution block immediately
> but rather queue it? That would require the queuing mechanism to be aware
> of context in order to submit the callable / runnable when time comes. The
> context.wrap essentially returns a value, which could be passed around
> and used anywhere without the fear forgetting to call something before.
>
Right, but as mentioned there's already a mechanism for wrapping a runnable
or callable. If instead we invested in this, we would be able to centralize
the "new" concern from existing ones. Remember, executors are not the only
async mechanism in java, neither should someone need to fire a thread or
use the Callable function in order to schedule something. For example,
there are existing RxJava decorators which would work out of the box if we
focused on the new concern which is copying the span from the context to
the thread when it isn't there already.

The long story short is that I'm requesting we don't add more technical
debt or new apis that pin to very specific things that we can't easily
rewire later. When you create a new form of BraveExecutorService, I can't
simply rewire it to Brave 4. Neither can I help as easily as this is now a
bespoke mechanism vs what is used by the many other libraries we have. So
basically, I'm requesting to take a step back from the narrow view and see
if there's a way to make this play nice with Brave, such that you can get
free maintenance and so that others moving from other frameworks that use
Brave to CXF don't need to learn new concepts. If this implies we adjust
core code, let's do that!

>

brave-cxf3/src/main/java/com/github/kristofa/brave/cxf3/TracerContext.java
adriancole: PS @reta in brave 4, this could become nice and explicit because Span is always not null (ex it can be a noop)

```java
public void foo(@Suspended AsyncResponse response, @Context brave.Span currentSpan) {
   // use any utility which can act on a span!
}
```

This is why I'm hesitant to adding more things to the api, especially ones that make things that didn't exist yet in Brave 3 explicit. For example, it would be lovely to have an interceptor for today which is wired with a feature and its only role is to attach the brave 3 current span when there's a suspended response. That would mean the users don't have to use any apis which would be almost immediately deprecated!

brave-cxf3/src/main/java/com/github/kristofa/brave/cxf3/TracerContext.java
```diff
@@ -0,0 +1,100 @@
...
+        serverSpanThreadBinder.setCurrentSpan(current);
+      }
+    };
+  }
+
+  private ServerSpan getCurrentServerSpan() {
+    // gets existing server span or reinitializes to ServerSpan#EMPTY using ThreadLocal#initialValue
+    final ServerSpan current = serverSpanThreadBinder.getCurrentServerSpan();
+
+    // when value was just initialized
```
adriancole: ps I looked into the code, and actually we can't depend on any semantics around null. For example, ServerSpan.EMPTY is indeed the initial value, but both of our implementations revert to this on `setServerSpan(null)` (via threadlocal.remove which reverts to initial value which is EMPTY!)

in #321 I clarified this by explicitly setting to ServerSpan.EMPTY (since that's what ends up happening anyway).

Long story short, we have no means to tell whether something was explicitly set to EMPTY or by accident as eventhough javadoc say don't accept null, we did, and also we don't control all implementations of span state :(

reta:
@adriancole Fair enough, I think you much better understand the trade-offs you are willing to make and where Brave is going. Thanks for response.

podsiedzik:
@adriancole @reta 

CurrentSpanContext was introduced only for one reason, to pass ServerSpan from Exchange to new thread.
We use Exchange because current thread (Thread1) is cleaned in ServerInInterceptor (starts server span).
It's cleaned because, ServerOutInterceptor (finishes server span) is invoked on another thread (Thread2).

If we don't care about cleaning Thread1 then there's no reason to have CurrentSpanContext at all.

reta:
@podsiedzik  I think cleaning the request thread (Thread1) is very important otherwise be ready for side-effects. @adriancole, what do you think?

adriancole:
So I like the idea of having something to associate/disassociate the span
from a thread. I'm just wondering if we can do this in a way which is
automatic (via feature or filter they could disable), as opposed to needing
to programmatically call something.

If we have to programmatically call something, then I'd prefer some hook
that doesn't duplicate an api. This is the brainstorm

reta:
@adriancole This is an excellent point you brought, in fact. I think we can discuss as separate issue, that would be great enhancement by itself. 

With JAX-RS (this is not CXF-specific issue per se), in case of async call there is no explicit demarcation boundary when request processing is over (in this case we could pretty much keep the current span till this moment). This unfortunately forces to look for workarounds, like the @Context f.e. :(

podsiedzik:
@reta maybe we should look at custom cxf `Invoker` that would always clean Thread1 after method invocation. ServerSpan from Thread1 to Thread would be passes with BraveExecutorService?
Or is it to much? :)

reta:
@podsiedzik I think with customer Invoker it would be a bit easier to solve the span hand-off issue, however this is indeed too much (in my opinion). Invokers are not composable and are quite implementation-specific. In case of JAX-RS, f.e, although internals may differ a bit, the user experience and expectations would be similar across CXF or Jersey, just add BraveFeature or BraveProvider and you are good to go.

adriancole:
fyi just pushed brave 3.17, but no worries. I can push another release whenever this is ready.

podsiedzik:
Hi @reta


So what we should do with this issue, any new ideas? There're pros and cons for every previous solution. 

Precondition `executor` is instance of `BraveExecutorService`

* Using `attach` as adriancole proposed:
	* ads new api
	
Usage:

```java
    @POST
    @Path("/foo")
    public void foo(@Suspended AsyncResponse response, @Context CurrentSpanContext context) {
      context.attach();
      executor.execute(() -> {
        response.resume("foo");
      });
      context.detach();
    }
```

* Using `wrap` as @ proposed:
	* ads new api
	
Usage:
	
```java
    @POST
    @Path("/foo")
    public void foo(@Suspended AsyncResponse response, @Context CurrentSpanContext context) {
      executor.execute(context.wrap(() -> {
        response.resume("foo");
      }));
    }
```

After second thought I've started to like idea about custom invoker.

* Using custom `BraveInvoker` that delegates to proper `Invoker`
	* invoker used only to clean Thread1
	* no new api
	* all code is same for jaxrs and jaxws
	* limited to cxf only but for me it's not a problem because it's brave-cfx not brave-jaxrs

Usage:

```java
    @POST
    @Path("/foo")
    public void foo(@Suspended AsyncResponse response) {
      executor.execute(() -> {
        response.resume("foo");
      });
    }
```

BraveInvoker could look like this:

```java
public class BraveInvoker implements Invoker {
  final Brave brave;
  final Invoker invoker;

  public BraveInvoker(Brave brave, Invoker invoker) {
    this.brave = brave;
    this.invoker = invoker;
  }

  @Override
  public Object invoke(Exchange exchange, Object o) {
    try {
      return invoker.invoke(exchange, o);
    } finally {
      if (exchange.get(BraveCxfConstants.METHOD_INVOKED) == null) {
        brave.serverTracer().clearCurrentSpan();
        exchange.put(BraveCxfConstants.METHOD_INVOKED, true);
      }
    }
  }
}
```

cxf ServerFeature could look like this:

```java
  @Override
  public void initialize(Server server, Bus bus) {
    //...

    Invoker invoker = server.getEndpoint().getService().getInvoker();
    if (!(invoker instanceof BraveInvoker)) {
      server.getEndpoint().getService().setInvoker(new BraveInvoker(brave, invoker));
    }   
  }
```

`ServerInInterceptor`
* starts new span
* attaches to Thread1
* ads it to Exchange

existing `BraveExecutorService`
*  attaches to Thread2

`BraveInvoker`
* clears Thread1 after method inovocation

`ClientOutInterceptor`
* starts client span

`ClientInInterceptor`
* finishes client span
	
`ServerOutInterceptor`
* takes span from Exchange and finishes it 


What do you think?

reta:
Hey @podsiedzik !

Thanks a lot for summing up the alternatives! It is really a tough call to make, but let me try to express my opinion on that. 

So the first one I think is what @adriancole is suggesting, in my opinion it requires from developer / user to follow some kind of agreement (context.attach() / context.detach()), which is not necessarily bad. However, it still requires `executor` to be dedicated Brave executor, right?  

The second one is somewhat functional approach, when we wrap callable / runnable and pass it as a value. It is going to work with any executor and does not require additional instrumentation, right?

The last one is CXF specific indeed and actually requires only dedicated Brave `executor`, so the current span will be handed off to another thread. I honestly do not know if using `Invoker` is good or bad, but is assumes more knowledge from developer / user for sure, for example what delegating `invoker` should be? We have quite a lot of them ....

Personally, I would go with 1st or 2nd approach (@adriancole's call here). The good part of having context object is that it could allow developer / user to create new child spans or even get Brave instance so one can do whatever she wants. Otherwise, getting Brave instances in JAX-RS resources would be difficult. 

Does it make sense? What do you think, @podsiedzik ?
Thanks!

podsiedzik:
@reta

1) 
	Yes it would require BraveExecutorService
2) 
	It would work with any executor
3) 
	Yes it would require BraveExecutorService,
	Developer doesn't have to choose Invoker, in `ServerFeature` we would get existing invoker and wrap it in `BraveInvoker`
  
  Personaly I don't like idea of Context object, I would prefere that brave feature would be as hidden as possible, so it doesn't obscures real business logic and whatnot.
  When I would like to use something very custom with Brave, I still have possibility to inject `Brave` instance and use it there.

reta:
@podsiedzik All valid arguments. As it is all about trade-offs, I think you could pick the solution you are more comfortable with and fits better to Brave usage patterns. Just one note regarding invokers, as I mentioned before, they are not composable. You are making an implicit assumption that your feature will be picked up last, but depending on initialization order, BraveInvoker may be overridden / replaced by another invoker, the same way you do through feature f.e. Just to keep it in mind :-)

suspended:
Don't tag me. Thanks

adriancole:
here's a summary of a design I have in mind for the "brave 4" current span api. please take a look!  https://github.com/openzipkin/brave/issues/166#issuecomment-277634798

adriancole:
I know that this code is moving into CXF, but fyi I'm starting to backfill tests for all HTTP implementations, including CXF. I'm ignoring this as a known issue here https://github.com/openzipkin/brave/pull/362

adriancole:
obviated by https://github.com/openzipkin/brave/issues/354

PR #308, opened 2016-12-22
Support Apache HttpAsyncClient "client received"

robinst:
Before, BraveHttpResponseInterceptor would call
ClientResponseInterceptor and it would call setClientReceived. But it
would not submit the end annotation because the current client span was
null. This is because the response interceptor is not called on the same
thread as the request interceptor.

This can be solved by storing the client span as an attribute on the
HttpContext instead.

### DIFF
--- a/brave-apache-http-interceptors/pom.xml
+++ b/brave-apache-http-interceptors/pom.xml
@@ -42,5 +42,10 @@
         <artifactId>brave-http-tests</artifactId>
         <scope>test</scope>
     </dependency>
+      <dependency>
+          <groupId>org.apache.httpcomponents</groupId>
+          <artifactId>httpasyncclient</artifactId>
+          <scope>test</scope>
+      </dependency>
   </dependencies>
 </project>
--- a/pom.xml
+++ b/pom.xml
@@ -188,6 +188,11 @@
             <version>${httpcomponents.version}</version>
         </dependency>
         <dependency>
+            <groupId>org.apache.httpcomponents</groupId>
+            <artifactId>httpasyncclient</artifactId>
+            <version>4.1.2</version>
+        </dependency>
+        <dependency>
             <groupId>org.apache.logging.log4j</groupId>
             <artifactId>log4j-core</artifactId>
             <version>${log4j.version}</version>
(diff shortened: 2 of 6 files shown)


adriancole:
yep this will do it. Thanks for the help!

brave-apache-http-interceptors/pom.xml
```diff
@@ -42,5 +42,10 @@
         <artifactId>brave-http-tests</artifactId>
         <scope>test</scope>
     </dependency>
+      <dependency>
+          <groupId>org.apache.httpcomponents</groupId>
+          <artifactId>httpasyncclient</artifactId>
+          <scope>test</scope>
```
adriancole: build seems to break on a snapshot repo https://github.com/openzipkin/brave/issues/309

wonder if it is actually only this PR. can you try adding the version either here as `<version>${httpcomponents.version}</version>` or doing the same in the root pom under dependency management?

pom.xml
```diff
@@ -188,6 +188,11 @@
             <version>${httpcomponents.version}</version>
         </dependency>
         <dependency>
+            <groupId>org.apache.httpcomponents</groupId>
+            <artifactId>httpasyncclient</artifactId>
+            <version>4.1.2</version>
```
adriancole: actually can you switch to the same version pattern `<version>${httpcomponents.version}</version>` or is this unrelated to the other versions?

adriancole: ahh I see it is unrelated. this might be obvious to those who know this library, but not intuitive. mind adding an XML comment that this has a different versioning scheme vs the other files?

adriancole:
nuked the build cache which might let https://github.com/openzipkin/brave/issues/309 pass

adriancole:
thanks. point I mentioned wasn't important :P will be released in the next day or two, once #304 is done

pom.xml
robinst: Yeah, I was also confused because I tried to use `${httpcomponents.version}` first. I'll raise a PR tomorrow :).

robinst:
Awesome, thanks!

PR #310, opened 2016-12-22
Moves offset-logic into the default clock type

adriancole:
Before, we repeated offset logic in each of the 3 tracers plus an
internal tick holder. This centralizes it into the default clock, which
dramatically simplifies the code and eliminates the need to expose a
tick anywhere. If anyone doesn't like offset-based approach, they can
easily replace the clock, too.

### DIFF
--- a/brave-core/src/test/java/com/github/kristofa/brave/LocalTracerTest.java
+++ b/brave-core/src/test/java/com/github/kristofa/brave/LocalTracerTest.java
@@ -23,8 +23,10 @@ import com.twitter.zipkin.gen.Span;
 import zipkin.reporter.Reporter;
 
 @RunWith(PowerMockRunner.class)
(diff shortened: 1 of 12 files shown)


brave-core/src/test/java/com/github/kristofa/brave/LocalTracerTest.java
```diff
@@ -158,96 +157,6 @@ public void finishSpan_lessThanMicrosRoundUp() {
...
-        localTracer.finishSpan(500L);
-
-        verify(mockReporter).report(finished.toZipkin());
-        verifyNoMoreInteractions(mockReporter);
-
-        assertEquals(500L, finished.getDuration().longValue());
-    }
-
-    @Test
-    public void finishSpan_unsampled() {
```
adriancole: these two tests didn't really prove anything

PR #312, opened 2016-12-22
Add comment for httpasyncclient version in pom

robinst:


adriancole:
how responsible! thanks

PR #316, opened 2016-12-24
Deprecates ServerTracer and ClientTracer.builder

adriancole:
ServerTracer and ClientTracer.builder were partially exposed, which will
interfere with our ability to upgrade Brave. This deprecates them so
that the we can backport code to Brave 4's api.

Users should prefer the accessors in Brave until then,

Ex. Brave.clientTracer()


brave-core/src/test/java/com/github/kristofa/brave/LocalTracerTest.java
```diff
@@ -182,59 +178,33 @@ public void startSpan_with_inheritable_nested_local_spans() {
     }
 
     @Test
-    public void startSpan_nested_local_spans_disabled() {
```
adriancole: removed this test as it is impossible to configure this since the nested flag is not exposed

PR #320, opened 2016-12-28
Locks down internal methods that create spans

adriancole:
This locks down methods that create `c.t.z.gen.Span`, by forcing them
through the internal package. By doing so, we know if a span was created
in a deprecated way or not. We also remove temptation to use internal
hooks accidentally (by virtue of not reading docs).

This change works by making a package accessor `InternalSpan`, in the
same way as OkHttp works. In the rare cases we need to create spans for
test purposes, we need to set the reference using `initializeForTests()`

The majority of the code impact is old tests which don't need to create
spans explicitly, but would take effort to port to not do so.

### DIFF
--- a/brave-core/src/main/java/com/twitter/zipkin/gen/Span.java
+++ b/brave-core/src/main/java/com/twitter/zipkin/gen/Span.java
@@ -1,6 +1,7 @@
 package com.twitter.zipkin.gen;
 
 import com.github.kristofa.brave.SpanId;
+import com.github.kristofa.brave.internal.InternalSpan;
 import com.github.kristofa.brave.internal.Util;
 import java.io.Serializable;
 import java.util.ArrayList;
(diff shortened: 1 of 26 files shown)


brave-core/src/main/java/com/twitter/zipkin/gen/Span.java
```diff
@@ -12,15 +13,39 @@
 
 /** This is an internal type representing a span in the trace tree. */
 public class Span implements Serializable {
-
-  /** Internal. Do not create spans directly */
-  public static Span create(SpanId spanId) {
```
adriancole: this change avoids exposing this into a released version of brave

brave-core/src/main/java/com/twitter/zipkin/gen/Span.java
```diff
@@ -32,10 +57,13 @@ public static Span create(SpanId spanId) {
...
 
-  /** @deprecated internally we call {@link Span#Span(SpanId)} because it sets the identity */
+  /**
+   * Span is an internal type, don't create new instances manually.
+   *
+   * @deprecated internally we call {@link Span#Span(SpanId)} because it sets the identity
+   */
   @Deprecated
   public Span() {
-    assert false : "do not construct spans except via their context";
```
adriancole: removed this assertion so tests can pass. deprecation and other assertions still make this deprecation pretty loud

brave-core/src/main/java/com/twitter/zipkin/gen/Span.java
```diff
@@ -49,15 +77,6 @@ public Span() {
     if (context.debug()) debug = true;
   }
 
-  /**
-   * Internal api that returns the trace context this span was created with.
-   *
-   * @since 3.17
-   */
-  public SpanId context() {
```
adriancole: this change also prevents this from getting into a non-internal package

PR #324, opened 2017-01-02
Makes it possible to specify a recorder

adriancole:
Brave 3's recorder is internal. If we make it possible to supply this directly, we can bridge Brave3 to Brave4 using optional dependencies (As opposed to inheritance or otherwise a strict dependency)


adriancole:
example use in tests..

```java
@RunWith(Enclosed.class)
public class MaybeAddClientAddressTest {

  public static class Brave4 extends Brave3 {
    public Brave4() {
      brave = new Brave.Builder().recorder(
          Brave4Recorder.create(Tracer.builder().reporter(spans::add).build())).build();
    }
  }

  public static class Brave3 {
    List<Span> spans = new ArrayList<>();
    Brave brave = new Brave.Builder().reporter(spans::add).build();
```

PR #329, opened 2017-01-05
Exposes Brave.clock

adriancole:
This exposes Brave.clock so that it can be used to record timestamps
out-of-band.


adriancole:
as requested by @devinsba

PR #337, opened 2017-01-16
Fixes incorrect translation of IPv6 ::1 as 0.0.0.1

adriancole:
In perusing self-trace data, I noticed a client address of 0.0.0.1. This
fixes that.


adriancole:
cc @llinder this is what I was talking about

brave-core/src/main/java/com/github/kristofa/brave/internal/MaybeAddClientAddress.java
```diff
@@ -52,7 +52,16 @@ public final void accept(T input) {
           break;
         }
       }
-      if (maybeIpv4Compat) {
+      boolean maybeLocalhost = true;
+      for (int i = 10; i < 15; i++) {
```
adriancole: out of view of this patch, but visually directly about we end a loop on i<10

PR #345, opened 2017-01-18
Adds issue template to hopefully reduce the Q/A issues opened

adriancole:
Adding an issue template will make what's appropriate more explicit. The
goal here is to set expectations.

The text here was modified from OkHttp


adriancole:
cc @dragontree101

PR #347, opened 2017-01-21
Documents one-way span modeling

adriancole:
This documents the simplest possible api for one-way RPC tracing, which is to use `flush` instead of `finish`

---
Sometimes you need to model an asynchronous operation, where there is a
request, but no response. In normal RPC tracing, you use `span.finish()`
which indicates the response was received. In one-way tracing, you use
`span.flush()` instead, as you don't expect a response.

Here's how a client might model a one-way operation
```java
// start a new span representing a client request
oneWaySend = tracer.newSpan(parent).kind(Span.Kind.CLIENT);

// Add the trace context to the request, so it can be propagated in-band
Propagation.B3_STRING.injector(Request::addHeader)
                     .inject(oneWaySend.context(), request);

// fire off the request asynchronously, totally dropping any response
request.execute();

// start the client side and flush instead of finish
oneWaySend.start().flush();
```

And here's how a server might handle this..
```java
// pull the context out of the incoming request
contextOrFlags =
    Propagation.B3_STRING.extractor(Request::getHeader).extract(request);

// convert that context to a span which you can name and add tags to
oneWayReceive = tracer.joinSpan(contextOrFlags.context())
    .name("process-request")
    .kind(SERVER)
    ... add tags etc.

// start the server side and flush instead of finish
oneWayReceive.start().flush();

// you should not modify this span anymore as it is complete. However,
// you can create children to represent follow-up work.
next = tracer.newSpan(oneWayReceive.context).name("step2").start();
```

### DIFF
--- a/brave-core/src/test/java/com/github/kristofa/brave/ClientTracerTest.java
+++ b/brave-core/src/test/java/com/github/kristofa/brave/ClientTracerTest.java
@@ -53,6 +53,7 @@ public class ClientTracerTest {
     assertThat(spans).matches(s -> s.isEmpty() || s.contains(BASE_SPAN));
   }
 
+  @Test
   public void setClientSent_doesntFlush() {
     brave.clientSpanThreadBinder().setCurrentSpan(span);
     brave.clientTracer().setClientSent();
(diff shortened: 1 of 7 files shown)


adriancole:
Thanks to @dragontree101 for helping test these ideas

PR #350, opened 2017-01-25
Brave 4.x modernization

### LINKED ISSUE #264: Make BraveServletFilter customizable

adriancole:
This module a redo of all major instrumentation libraries since Brave 3.
Artifacts have the naming convention "brave-instrumentation-XXX": for
example, the directory "servlet" includes the artifact "brave-instrumentation-servlet".

Notably, this adds a more `HttpTracing` component to configure http client
and server tracing. It also adds support for async servlet and apache http
client, and tests edge cases like multiple servlet versions. Finally, it
tests and benchmarks every http component, so that people are aware of
overhead is involved in tracing.

Here's an example of configuring OkHttp. Note that all instrumentation
have the same naming policy TracingXXX, where XXX is usually the same
as the type returned.

```java
                 .localServiceName("my-service")
                 .reporter(reporter)
                 .build();
httpTracing = HttpTracing.newBuilder(tracing).serverName("github").build();
okhttp = TracingCallFactory.create(httpTracing, new OkHttpClient());
```

### Http tracing
Most instrumentation are based on http communication. For this reason,
we have specialized handlers for http clients and servers. All of these
are configured with `HttpTracing`.

The `HttpTracing` class holds a reference to a tracing component and also
includes instructions on what to put into http spans.

By default, the following is added for both http clients and servers:
* Span.name as the http method in lowercase: ex "get"
* Tags/binary annotations:
  * "http.path", which does not include query parameters.
  * "http.status_code" when the status us not success.
* Remote IP and port information

Naming and tags are configurable in a library-agnostic way. For example,
the same `HttpTracing` component configures OkHttp or Apache HttpClient
identically.

For example, to change the span and tag naming policy for clients, you
can do something like this:

```java
    .clientParser(new HttpClientParser() {
      @Override public <Req> String spanName(HttpAdapter<Req, ?> adapter, Req req) {
        return adapter.method(req).toLowerCase() + " " + adapter.path(req);
      }

      @Override
      public <Req> void requestTags(HttpAdapter<Req, ?> adapter, Req req, brave.Span span) {
        span.tag(TraceKeys.HTTP_URL, adapter.url(req)); // the whole url, not just the path
      }
    })
    .serverName("remote-service") // assume both libraries are calling the same service
    .build();

apache = TracingHttpClientBuilder.create(httpTracing).build();
okhttp = TracingCallFactory.create(httpTracing, new OkHttpClient());
```

---
Work accomplished in support of this
 - [x] Rough in a tentative client and server tracer api
 - [x] Integrate each library such that it sets spans in scope so `Tracer.currentSpan()` works
 - [x] Integrate each staticly initialized library such that it uses `Tracer.current()`
 - [x] Backport client and server integration tests
 - [x] add Async HttpClient and Servlet (which we never had before)
 - [x] add Tests for servlet 2.5 and grpc compat (which we never had before)
 - [x] ensure we can swap out propagation format when needed. Ex https://github.com/TraceContext/tracecontext-spec
 - [x] revise tagging approach so that it solves #276 #264
 - [x] Add tests for http client redirects
 - [x] Revisit exception parsing
 - [x] Revisit remote address parsing

Fixes #276
Fixes #264

### DIFF
--- a/brave/src/main/java/brave/http/HttpTracing.java
+++ b/brave/src/main/java/brave/http/HttpTracing.java
@@ -1,0 +1,70 @@
+package brave.http;
+
+import brave.Tracing;
+import com.google.auto.value.AutoValue;
+import zipkin.Endpoint;
+
+@AutoValue
+public abstract class HttpTracing {
+  public static HttpTracing create(Tracing tracing) {
+    return newBuilder(tracing).build();
+  }
+
+  public static Builder newBuilder(Tracing tracing) {
+    return new AutoValue_HttpTracing.Builder()
+        .tracing(tracing)
+        .serverName("")
+        .clientParser(new HttpClientParser())
+        .serverParser(new HttpServerParser());
+  }
+
+  public abstract Tracing tracing();
+
+  public abstract HttpClientParser clientParser();
+
+  /**
+   * Used by http clients to indicate the name of the destination service.
+   *
+   * Defaults to "", which will not show in the zipkin UI or end up in the dependency graph.
+   *
+   * <p>When present, a link from {@link Tracing.Builder#localServiceName(String)} to this name will
+   * increment for each traced client call.
+   *
+   * <p>As this is endpoint-specific, it is typical to create a scoped instance of {@linkplain
+   * HttpTracing} to assign this value.
+   *
+   * For example:
+   * <pre>{@code
+   * github = TracingHttpClientBuilder.create(
+   *   httpTracing.toBuilder().serverName("github").build()
+   * ).build();
+   * }</pre>
+   *
+   * @see zipkin.Constants#SERVER_ADDR
+   * @see brave.Span#remoteEndpoint(Endpoint)
+   */
+  public abstract String serverName();
+
+  public abstract HttpServerParser serverParser();
+
+  public abstract Builder toBuilder();
+
+  @AutoValue.Builder
+  public static abstract class Builder {
+    public abstract Builder tracing(Tracing tracing);
+
+    public abstract Builder clientParser(HttpClientParser clientParser);
+
+    public abstract Builder serverName(String serverName);
+
+    public abstract Builder serverParser(HttpServerParser serverParser);
+
+    public abstract HttpTracing build();
+
+    Builder() {
+    }
+  }
+
+  HttpTracing() {
+  }
+}
--- a/instrumentation/httpasyncclient/src/main/java/brave/httpasyncclient/TracingHttpAsyncClientBuilder.java
+++ b/instrumentation/httpasyncclient/src/main/java/brave/httpasyncclient/TracingHttpAsyncClientBuilder.java
@@ -1,0 +1,257 @@
+package brave.httpasyncclient;
+
+import brave.Span;
+import brave.Tracer;
+import brave.Tracer.SpanInScope;
+import brave.Tracing;
+import brave.http.HttpClientHandler;
+import brave.http.HttpTracing;
+import brave.propagation.TraceContext;
+import java.io.IOException;
+import java.util.concurrent.Future;
+import org.apache.http.Header;
+import org.apache.http.HttpException;
+import org.apache.http.HttpHost;
+import org.apache.http.HttpMessage;
+import org.apache.http.HttpRequest;
+import org.apache.http.HttpRequestInterceptor;
+import org.apache.http.HttpResponse;
+import org.apache.http.HttpResponseInterceptor;
+import org.apache.http.client.methods.HttpRequestWrapper;
+import org.apache.http.client.protocol.HttpClientContext;
+import org.apache.http.concurrent.FutureCallback;
+import org.apache.http.impl.nio.client.CloseableHttpAsyncClient;
+import org.apache.http.impl.nio.client.HttpAsyncClientBuilder;
+import org.apache.http.nio.ContentDecoder;
+import org.apache.http.nio.ContentEncoder;
+import org.apache.http.nio.IOControl;
+import org.apache.http.nio.protocol.HttpAsyncRequestProducer;
+import org.apache.http.nio.protocol.HttpAsyncResponseConsumer;
+import org.apache.http.protocol.HttpContext;
+import zipkin.Endpoint;
+
+public final class TracingHttpAsyncClientBuilder extends HttpAsyncClientBuilder {
+
+  public static HttpAsyncClientBuilder create(Tracing tracing) {
+    return new TracingHttpAsyncClientBuilder(HttpTracing.create(tracing));
+  }
+
+  public static HttpAsyncClientBuilder create(HttpTracing httpTracing) {
+    return new TracingHttpAsyncClientBuilder(httpTracing);
+  }
+
+  final Tracer tracer;
+  final TraceContext.Injector<HttpMessage> injector;
+  final HttpClientHandler<HttpRequest, HttpResponse> handler;
+  final String remoteServiceName;
+
+  TracingHttpAsyncClientBuilder(HttpTracing httpTracing) { // intentionally hidden
+    if (httpTracing == null) throw new NullPointerException("httpTracing == null");
+    this.tracer = httpTracing.tracing().tracer();
+    this.remoteServiceName = httpTracing.serverName();
+    this.handler = new HttpClientHandler<>(new HttpAdapter(), httpTracing.clientParser());
+    this.injector = httpTracing.tracing().propagation().injector(HttpMessage::setHeader);
+  }
+
+  @Override public CloseableHttpAsyncClient build() {
+    super.addInterceptorFirst((HttpRequestInterceptor) (request, context) -> {
+      TraceContext parent = (TraceContext) context.getAttribute(TraceContext.class.getName());
+      Span span = parent == null ? tracer.newTrace() : tracer.newChild(parent);
+
+      HttpHost host = HttpClientContext.adapt(context).getTargetHost();
+      if (!span.isNoop() && host != null) {
+        parseServerAddress(host, span);
+        handler.handleSend(HttpRequestWrapper.wrap(request, host), span);
+      } else {
+        handler.handleSend(request, span);
+      }
+
+      injector.inject(span.context(), request);
+      context.setAttribute(Span.class.getName(), span);
+      context.setAttribute(SpanInScope.class.getName(), tracer.withSpanInScope(span));
+    });
+    super.addInterceptorLast((HttpRequestInterceptor) (request, context) -> {
+      SpanInScope scope = (SpanInScope) context.getAttribute(SpanInScope.class.getName());
+      if (scope != null) {
+        context.removeAttribute(SpanInScope.class.getName());
+        scope.close();
+      }
+    });
+    super.addInterceptorLast((HttpResponseInterceptor) (response, context) -> {
+      Span span = (Span) context.getAttribute(Span.class.getName());
+      handler.handleReceive(response, span);
+    });
+    return new TracingHttpAsyncClient(super.build());
+  }
+
+  void parseServerAddress(HttpHost host, Span span) {
+    Endpoint.Builder builder = Endpoint.builder().serviceName(remoteServiceName);
+    if (!builder.parseIp(host.getAddress())) builder.parseIp(host.getHostName());
+    builder.port(host.getPort());
+    span.remoteEndpoint(builder.build());
+  }
+
+  static final class HttpAdapter extends brave.http.HttpAdapter<HttpRequest, HttpResponse> {
+    @Override public String method(HttpRequest request) {
+      return request.getRequestLine().getMethod();
+    }
+
+    @Override public String path(HttpRequest request) {
+      String result = request.getRequestLine().getUri();
+      int queryIndex = result.indexOf('?');
+      return queryIndex == -1 ? result : result.substring(0, queryIndex);
+    }
+
+    @Override public String url(HttpRequest request) {
+      if (request instanceof HttpRequestWrapper) {
+        HttpRequestWrapper wrapper = (HttpRequestWrapper) request;
+        HttpHost target = wrapper.getTarget();
+        if (target != null) return target.toURI() + wrapper.getURI();
+      }
+      return request.getRequestLine().getUri();
+    }
+
+    @Override public String requestHeader(HttpRequest request, String name) {
+      Header result = request.getFirstHeader(name);
+      return result != null ? result.getValue() : null;
+    }
+
+    @Override public Integer statusCode(HttpResponse response) {
+      return response.getStatusLine().getStatusCode();
+    }
+  }
+
+  final class TracingHttpAsyncClient extends CloseableHttpAsyncClient {
+    private final CloseableHttpAsyncClient delegate;
+
+    public TracingHttpAsyncClient(CloseableHttpAsyncClient delegate) {
+      this.delegate = delegate;
+    }
+
+    @Override public <T> Future<T> execute(HttpAsyncRequestProducer requestProducer,
+        HttpAsyncResponseConsumer<T> responseConsumer, HttpContext context,
+        FutureCallback<T> callback) {
+      Span span = tracer.currentSpan();
+      context.setAttribute(TraceContext.class.getName(), span != null ? span.context() : null);
+      return delegate.execute(
+          new TracingAsyncRequestProducer(requestProducer, context),
+          new TracingAsyncResponseConsumer<>(responseConsumer, context),
+          context,
+          callback
+      );
+    }
+
+    @Override public void close() throws IOException {
+      delegate.close();
+    }
+
+    @Override public boolean isRunning() {
+      return delegate.isRunning();
+    }
+
+    @Override public void start() {
+      delegate.start();
+    }
+  }
+
+  final class TracingAsyncRequestProducer implements HttpAsyncRequestProducer {
+    final HttpAsyncRequestProducer requestProducer;
+    final HttpContext context;
+
+    TracingAsyncRequestProducer(HttpAsyncRequestProducer requestProducer, HttpContext context) {
+      this.requestProducer = requestProducer;
+      this.context = context;
+    }
+
+    @Override public void close() throws IOException {
+      requestProducer.close();
+    }
+
+    @Override public HttpHost getTarget() {
+      return requestProducer.getTarget();
+    }
+
+    @Override public HttpRequest generateRequest() throws IOException, HttpException {
+      return requestProducer.generateRequest();
+    }
+
+    @Override public void produceContent(ContentEncoder encoder, IOControl io) throws IOException {
+      requestProducer.produceContent(encoder, io);
+    }
+
+    @Override public void requestCompleted(HttpContext context) {
+      requestProducer.requestCompleted(context);
+    }
+
+    @Override public void failed(Exception ex) {
+      Span currentSpan = (Span) context.getAttribute(Span.class.getName());
+      if (currentSpan != null) {
+        handler.handleError(ex, currentSpan);
+        context.removeAttribute(Span.class.getName());
+      }
+      requestProducer.failed(ex);
+    }
+
+    @Override public boolean isRepeatable() {
+      return requestProducer.isRepeatable();
+    }
+
+    @Override public void resetRequest() throws IOException {
+      requestProducer.resetRequest();
+    }
+  }
+
+  final class TracingAsyncResponseConsumer<T> implements HttpAsyncResponseConsumer<T> {
+    final HttpAsyncResponseConsumer<T> responseConsumer;
+    final HttpContext context;
+
+    TracingAsyncResponseConsumer(HttpAsyncResponseConsumer<T> responseConsumer,
+        HttpContext context) {
+      this.responseConsumer = responseConsumer;
+      this.context = context;
+    }
+
+    @Override public void responseReceived(HttpResponse response)
+        throws IOException, HttpException {
+      responseConsumer.responseReceived(response);
+    }
+
+    @Override public void consumeContent(ContentDecoder decoder, IOControl ioctrl)
+        throws IOException {
+      responseConsumer.consumeContent(decoder, ioctrl);
+    }
+
+    @Override public void responseCompleted(HttpContext context) {
+      responseConsumer.responseCompleted(context);
+    }
+
+    @Override public void failed(Exception ex) {
+      Span currentSpan = (Span) context.getAttribute(Span.class.getName());
+      if (currentSpan != null) {
+        handler.handleError(ex, currentSpan);
+        context.removeAttribute(Span.class.getName());
+      }
+      responseConsumer.failed(ex);
+    }
+
+    @Override public Exception getException() {
+      return responseConsumer.getException();
+    }
+
+    @Override public T getResult() {
+      return responseConsumer.getResult();
+    }
+
+    @Override public boolean isDone() {
+      return responseConsumer.isDone();
+    }
+
+    @Override public void close() throws IOException {
+      responseConsumer.close();
+    }
+
+    @Override public boolean cancel() {
+      return responseConsumer.cancel();
+    }
+  }
+}
--- a/instrumentation/jaxrs2/README.md
+++ b/instrumentation/jaxrs2/README.md
@@ -1,0 +1,108 @@
+# brave-instrumentation-jaxrs2
+This module contains JAX-RS 2.x compatible tracing filters and a feature
+to automatically configure them.
+
+The `TracingClientFilter` adds trace headers to outgoing requests.
+`TracingContainerFilter` extracts trace state from incoming requests.
+Both report to Zipkin how long each request takes, along with relevant
+tags like the http url.
+
+## Configuration
+
+### Dependency Injection
+Tracing always needs an instance of type `brave.http.HttpTracing`
+configured. Make sure this is in place before proceeding.
+
+Then, configure `TracingFeature`, which sets up filters automatically.
+
+For example, this is how configuration looks with Jersey 2.x:
+
+
+In your web.xml:
+
+```xml
+<init-param>
+    <param-name>jersey.config.server.provider.packages</param-name>
+    <param-value>my.existing.packages,brave.jaxrs2</param-value>
+</init-param>
+```
+
+### Manual
+
+For server side setup, an Application class could look like this:
+
+```java
+public class MyApplication extends Application {
+
+  public Set<Object> getSingletons() {
+    Tracing tracing = Tracing.newBuilder().build();
+    TracingFeature tracingFeature = TracingFeature.create(tracing);
+    return new LinkedHashSet<>(Arrays.asList(tracingFeature));
+  }
+
+}
+```
+
+For client side setup, you just have to register the BraveTracingFeature
+with your `WebTarget` before you make your request:
+
+```java
+WebTarget target = target("/mytarget");
+target.register(TracingFeature.create(tracing));
+```
+
+## Exception handling
+`ContainerResponseFilter` has no means to handle uncaught exceptions.
+Unless you provide a catch-all exception mapper, requests that result in
+unhandled exceptions will leak until they are eventually flushed.
+
+Besides using framework-specific code, the only approach is to make a
+custom `ExceptionMapper`, similar to below, which ensures a request is
+sent back even when something unexpected happens.
+
+```java
+@Provider
+public static class CatchAllExceptions implements ExceptionMapper<Exception> {
+
+  @Override public Response toResponse(Exception e) {
+    if (e instanceof WebApplicationException) {
+      return ((WebApplicationException) e).getResponse();
+    }
+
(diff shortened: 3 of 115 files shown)


adriancole:
waiting a day for feedback on https://github.com/openzipkin/brave/issues/166#issuecomment-277634798 before proceeding to implement with the propagation design in mind

brave-apache-http-interceptors/src/main/java/brave/httpclient4/HttpClient4Instrumentation.java
```diff
@@ -0,0 +1,126 @@
...
+import org.apache.http.HttpResponseInterceptor;
+import org.apache.http.client.methods.HttpGet;
+import org.apache.http.client.protocol.HttpClientContext;
+import org.apache.http.impl.client.CloseableHttpClient;
+import org.apache.http.impl.client.HttpClients;
+import zipkin.TraceKeys;
+
+import static com.github.kristofa.brave.internal.Util.checkNotNull;
+
+public final class HttpClient4Instrumentation {
```
adriancole: we've been doing this wrong :) if we want folks to use tracer.currentSpan we need to implement HttpProcessor so that we can scope the span before any of the interceptors fire.

that way, people can write an interceptor that uses the current span without needing to know a magic attribute. This is particularly important for slf4j or log4j2 mdc logging. If we set the span ID before any interceptors go, anything that happens will end up with the same ids in the log. The only way I can see how to do this is implementing a TracingHttpProcessor

cc @shakuzen

adriancole: actually this won't work because there's a lot of logic inside the HttpClientBuilder which is only invoked if HttpProcessor isn't overridden. IOTW, I don't think we'll be able to neatly do implicit propagation. Rather, we'd need to document the attribute we use internally if someone wants to access the current span.

brave-grpc/src/test/java/brave/grpc/ITBraveClientInterceptor.java
```diff
@@ -0,0 +1,293 @@
...
+import zipkin.Constants;
+import zipkin.Endpoint;
+import zipkin.Span;
+import zipkin.storage.InMemoryStorage;
+
+import static java.util.Arrays.asList;
+import static org.assertj.core.api.Assertions.assertThat;
+import static org.assertj.core.api.Assertions.failBecauseExceptionWasNotThrown;
+
+public class ITBraveClientInterceptor {
```
adriancole: Added this which is a port of the http test

adriancole:
added jaxrs2 client impl and tests

adriancole:
ok all client instrumentation has been retouched except the database ones.. I'm kindof going to save "best for last" on that one. Moving on to server instrumentation

note: pay little attention to the "Config" types defined, they haven't been thought through and are subject to change. This is mainly getting all tests passing and exercising the "current span" apis

adriancole:
added some of the server instrumentation cc @kmaci3k

note: this PR needs one change to the "current span" PR it depends on #369, which is a replacement for BraveExecutorService (a couple http client libraries need it and tests are disabled until that's in)

adriancole:
all instrumentation except jdbc, mysql and cxf have been redone. I will probably do cxf just to prove the api approach works

brave-http-tests/src/main/java/brave/http/StrictCurrentTraceContext.java
```diff
@@ -0,0 +1,27 @@
+package brave.http;
+
+import brave.propagation.CurrentTraceContext;
+import brave.propagation.TraceContext;
+
+import static org.assertj.core.api.Assertions.assertThat;
+
+class StrictCurrentTraceContext implements CurrentTraceContext {
```
adriancole: Added this as the default implementation being an inheritable thread local makes it hard to test if propagation works by accident (ex only because inheritable thread local is used), or if because it was coded correctly. You'll notice this also is defensive about being closed on the same thread as it was opened.

adriancole:
rebased on top of #369

adriancole:
NEXT STEPS:
* implement sparkjava and database instrumentation with brave 4 apis
* take a closer look at instrumentation approach and generalize anything copy/pasted >3 times
* revise tagging approach so that it solves #276 #264

adriancole:
UPDATE: implemented sparkjava and database instrumentation with brave 4 apis

noticed we should revisit #210 as the database drivers need a reference to a tracer post-initialization (static reference)

adriancole:
ok rebased over some build things that were tripping up yesterday. will resume in a bit

adriancole:
ok current code solves the JDBC problem via #379 

Moving on to other todos

adriancole:
https://github.com/openzipkin/brave/pull/381 is needed so that we can wire in propagation handlers

adriancole:
rebased over #381 but not refactored to take advantage of it (beyond making propagation type not explicit)

brave-apache-http-interceptors/src/main/java/brave/httpclient4/HttpClientTracing.java
```diff
@@ -0,0 +1,200 @@
...
+    return new HttpClientBuilder() {
+
+      /**
+       * protocol exec is the second in the execution chain, so is invoked before a request is
+       * provisioned. We provision and scope a span here, so that application interceptors can see
+       * it via {@link Tracer#currentSpan()}.
+       */
+      @Override protected ClientExecChain decorateProtocolExec(ClientExecChain exec) {
+        return (route, request, context, execAware) -> {
+          Span next = tracer.nextSpan();
```
pavolloffay: It is a nice approach :+1: But It seems that this is run in a different thread.

adriancole:
Ahah but if it did the test would fail!

This is not the async client, so it is blocking (same thread). Theres
another safety which is the strict context impl. Havent done async in this
library yet

On 22 Apr 2017 23:16, "Pavol Loffay" <notifications@github.com> wrote:

*@pavolloffay* commented on this pull request.
------------------------------

In brave-apache-http-interceptors/src/main/java/brave/httpclient4/
HttpClientTracing.java
<https://github.com/openzipkin/brave/pull/350#discussion_r112811778>:

> +  /**
+   * Preferred way for constructing http clients, as it allows use of
{@link Tracer#currentSpan}
+   *
+   * <p>Implicitly requires a minimum version of httpclient 4.3
+   */
+  public HttpClientBuilder httpClientBuilder() {
+    return new HttpClientBuilder() {
+
+      /**
+       * protocol exec is the second in the execution chain, so is
invoked before a request is
+       * provisioned. We provision and scope a span here, so that
application interceptors can see
+       * it via {@link Tracer#currentSpan()}.
+       */
+      @Override protected ClientExecChain
decorateProtocolExec(ClientExecChain exec) {
+        return (route, request, context, execAware) -> {
+          Span next = tracer.nextSpan();

It is a nice approach 👍 But It seems that this is run in a different
thread.

—
You are receiving this because you authored the thread.
Reply to this email directly, view it on GitHub
<https://github.com/openzipkin/brave/pull/350#pullrequestreview-34159439>,
or mute the thread

pavolloffay:
@adriancole thanks, my fault I didn't use brave API properly. It's pity that this cannot be used for an async client.

adriancole:
@adriancole <https://github.com/adriancole> thanks, my fault I didn't use
brave API properly. It's pity that this cannot be used for an async client


Async has different dep (different version number too annoyingly). Imho it
is better to align these with the primary jar.. so not a big deal to do
another type for async. Also that keeps complexity in the async type. Our
old interceptors didnt work well for async anyway.

brave-apache-http-interceptors/src/main/java/brave/httpclient4/HttpClientTracing.java
```diff
@@ -0,0 +1,139 @@
...
+        };
+      }
+
+      /**
+       * main exec is the first in the execution chain, so last to execute. This creates a concrete
+       * http request, so this is where the timing in the span occurs.
+       *
+       * <p>This ends the span (and scoping of it) created by {@link #decorateMainExec(ClientExecChain)}.
+       */
+      @Override protected ClientExecChain decorateMainExec(ClientExecChain exec) {
```
adriancole: @llinder @devinsba @hyleung so when introducing support for custom parsing, I chose to keep generic parameters "raw" meaning they represent the actual native types. This is mostly to make the below code readable and without the clutter of `new FooRequestAdapter(request)` and `new FooResponseAdapter(response)`.

You'll also notice the adapter (maybe converter?) below is very straightforward.

brave-http-tests/src/main/java/brave/http/ITHttpClient.java
```diff
@@ -0,0 +1,250 @@
...
+
+    assertThat(collectedSpans())
+        .extracting(s -> s.name)
+        .containsExactly("get");
+  }
+
+  @Test
+  public void supportsPortableCustomization() throws Exception {
+    close();
+    client = newClient(server.getPort(), new HttpClientParser<Req, Resp>(){
```
adriancole: @llinder @devinsba @hyleung the tradeoffs come out here.. `HttpClientParser` is generically parameterized, but.. it can be used for any type. For example, the spanName below gives you the ability to adapt the raw type into a generic http model, but you can also access the raw request directly should you need to.

While the adapt (maybe we should call it convert or translate) is a bit explicit, it has the advantage of more concise and direct code in the instrumentation, and also a minor point which is that you don't need to allocate objects for every request and response (because this is converting an existing object vs wrapping it)

adriancole:
started readying for final lap.. this included removing instrumentation that is very old (like jersey v1 as they can be asked for later).. I've moved all new instrumentation under the instrumentation folder and using brave-instrumentation-X which is similar naming to zipkin-js

here's what's left.
* rework client code to be like server code
* cleanup READMEs
* add javadoc
* get others to look at it.

brave-http-tests/src/main/java/brave/http/ITHttpClient.java
llinder: The pattern looks good to me and I can't think of anything I would change just looking at it.  Nice work so far!

adriancole:
ok reworked the server code. other pending todos still todos

instrumentation/servlet/src/main/java/brave/servlet/HttpServletHandler.java
```diff
@@ -0,0 +1,62 @@
...
+import brave.http.HttpServerParser;
+import brave.internal.Nullable;
+import java.nio.ByteBuffer;
+import javax.servlet.http.HttpServletRequest;
+import javax.servlet.http.HttpServletResponse;
+import zipkin.Endpoint;
+
+import static brave.internal.InetAddresses.ipStringToBytes;
+
+public final class HttpServletHandler // public for others like sparkjava to use
```
adriancole: @hyleung @llinder so all server impls will need to declare their own HttpAdapter, unless they share types with another library. server impls that have special knowledge of the client address may choose to make their own HttpServerHandler to accomodate that.

I hope looking at the servlet example here might help decide if extending similarly is worth it for ratpack. If you notice gaps, do tell regardless!

instrumentation/servlet/src/main/java/brave/servlet/HttpServletHandler.java
```diff
@@ -0,0 +1,62 @@
...
+    extends HttpServerHandler<HttpServletRequest, HttpServletResponse> {
+
+  public HttpServletHandler(HttpServerParser<HttpServletRequest, HttpServletResponse> parser) {
+    super(new HttpServletAdapter(), parser);
+  }
+
+  @Nullable public Endpoint.Builder remoteEndpoint(HttpServletRequest request) {
+    byte[] addressBytes = ipStringToBytes(request.getHeader("X-Forwarded-For"));
+    if (addressBytes == null) addressBytes = ipStringToBytes(request.getRemoteAddr());
```
adriancole: ex we are overriding here because servlet has a `getRemoteAddr` function which not all libraries have. The default we override looks at X-Forwarded-For only

instrumentation/jaxrs2/src/main/java/brave/jaxrs2/TracingContainerFilter.java
```diff
@@ -0,0 +1,147 @@
...
+import javax.ws.rs.core.Context;
+import javax.ws.rs.core.Response;
+import javax.ws.rs.ext.Provider;
+import zipkin.Constants;
+
+import static javax.ws.rs.RuntimeType.SERVER;
+
+@Provider
+@Priority(0) // to make the span in scope visible to other filters
+@ConstrainedTo(SERVER)
```
jplock: should we add https://jersey.java.net/apidocs/2.12/jersey/javax/ws/rs/container/PreMatching.html on here as well so this provider happens before any routing decisions are made?

adriancole: sure. will do and add a comment to the effect. thanks for your help!

adriancole: hmm that breaks all the tests :P

adriancole:
implemented async servlet and servlet 2.5 support. only thing left is async http client, look into jaxrs annotation things then cleanups.

adriancole:
to test servlet 2.5 will be a minor pain (a maven invoker test, like we have in zipkin-server). I will definitely add this as otherwise it is punting to eyeballing and hoping it works.

instrumentation/servlet/src/main/java/brave/servlet/TracingFilter.java
```diff
@@ -0,0 +1,88 @@
...
+  final TraceContext.Extractor<HttpServletRequest> extractor;
+
+  TracingFilter(Builder builder) {
+    tracer = builder.tracing.tracer();
+    handler = new HttpServletHandler(builder.parser);
+    extractor = builder.tracing.propagation().extractor(HttpServletRequest::getHeader);
+  }
+
+  @Override
+  public void doFilter(ServletRequest request, ServletResponse response, FilterChain filterChain)
```
adriancole: @devinsba this servlet filter passes all tests sync and async. Will add an invoker test to make sure it also works on servlet 2.5 in practice (we have had people still using that, yes).

instrumentation/servlet/src/main/java/brave/servlet/TracingFilter.java
```diff
@@ -0,0 +1,89 @@
...
+  /** Creates a tracing filter with defaults. Use {@link #newBuilder(Tracing)} to customize. */
+  public static TracingFilter create(Tracing tracing) {
+    return newBuilder(tracing).build();
+  }
+
+  public static Builder newBuilder(Tracing tracing) {
+    return new Builder(tracing);
+  }
+
+  public static final class Builder {
```
adriancole: ps I looked at doing something like making a base builder class, but it made generics crazy.. it is more distracting to try and do that (or even autovalue) vs implement directly.

adriancole:
ok servlet 2.5 tests are in (via maven invoker). Glad I added them as I had a bug in my class detection stuff.

adriancole:
next thing I have is the apache async http client (mostly to show it can work).

I'm pretty happy with the api, but will likely polish it a little after stepping back and looking at each instrumentation closely.

instrumentation/servlet/src/main/java/brave/servlet/TracingFilter.java
```diff
@@ -0,0 +1,89 @@
...
+import javax.servlet.Filter;
+import javax.servlet.FilterChain;
+import javax.servlet.FilterConfig;
+import javax.servlet.ServletException;
+import javax.servlet.ServletRequest;
+import javax.servlet.ServletResponse;
+import javax.servlet.http.HttpServletRequest;
+import javax.servlet.http.HttpServletResponse;
+
+public final class TracingFilter implements Filter {
```
adriancole: @johnament FYI there will be some adjustments tomorrow (likely HttpTracing, which will be a component that includes Tracing and default http client and server tagging policy). However, most of this code will look the same. We won't be adding new Brave 3 style instrumentation moving forward. If you want feign or CDI, it would be based on the new stuff. Have a look!

adriancole:
ok added HttpTracing component, which I think dramatically simplifies construction of instrumentation. Most are less than 100lines all-in

brave/src/main/java/brave/http/HttpTracing.java
```diff
@@ -0,0 +1,43 @@
+package brave.http;
+
+import brave.Tracing;
+import com.google.auto.value.AutoValue;
+
+@AutoValue
+public abstract class HttpTracing {
```
adriancole: @llinder @jplock @hyleung @johnament @takezoe @devinsba so I think this is the latest thing to use in injection/configuraion. HttpTracing includes everything, including the normal things in the Tracing component (like the tracer and propagation), and the replacement for SpanNameProvider.

HttpTracing is nice because it should work nicely in any dependency injection system, like Spring, Guice or CDI, enclosing the non-instrumentation specific tracing configuration.

I haven't doc'd this class yet, but I would just note that the client and server parsers are where you do things like add tags based on servlet principal or http headers

adriancole:
added tests to ensure grpc is compatible with both v1.2 and v1.3

adriancole:
rewrote all the readmes.

waiting on this to merge https://github.com/openzipkin/zipkin/pull/1580

need to add http redirect tests, as I'm pretty sure some clients will have unexpected behavior

instrumentation/servlet/src/main/java/brave/servlet/TracingFilter.java
adriancole: ok thrash is mostly over. Only thing I'm thinking of tuning is how remote addresses are parsed.

adriancole:
renaming httpclient4 to httpclient since apache has made a new artifact httpclient5 (although not released anything on that in over a year)

adriancole:
ok httpasync client works.

TODO:
 * Add tests for http client redirects
 * Revisit exception parsing
 * Revisit remote address parsing

instrumentation/jaxrs2/src/test/java/brave/jaxrs2/ITTracingFeature_Client.java
```diff
@@ -0,0 +1,104 @@
...
+  }
+
+  @Override
+  @Test(expected = AssertionError.class) // doesn't yet close a span on exception
+  public void addsErrorTagOnTransportException() throws Exception {
+    super.addsErrorTagOnTransportException();
+  }
+
+  @ClientAsyncExecutor
+  static class TracingExecutorServiceProvider implements ExecutorServiceProvider {
```
adriancole: cc @jplock this works now..

adriancole:
redirect test indeed broke a couple of the clients... fixing now

instrumentation/httpasyncclient/src/main/java/brave/httpasyncclient/TracingHttpAsyncClientBuilder.java
```diff
@@ -0,0 +1,260 @@
...
+   *
+   * @see Constants#SERVER_ADDR
+   */
+  public TracingHttpAsyncClientBuilder remoteServiceName(String remoteServiceName) {
+    if (remoteServiceName == null) throw new NullPointerException("remoteServiceName == null");
+    this.remoteServiceName = remoteServiceName;
+    return this;
+  }
+
+  @Override public CloseableHttpAsyncClient build() {
```
adriancole: @shakuzen heh no comments yet.. but none needed, right? :p

adriancole:
ok epic work to get redirects operating with expected output (span per attempt) is in.

next up is polishing exception handling a remote address parsing. final lap

adriancole:
ok I have hidden everything not required to expose. tests are good, and I think the code is, too.

I have to document things tonight, but codewise, would appreciate any valiance in review, notably the customization model for HttpTracing. Here's a unit test of that.

For example, to override the naming policy for spans, you can do this, which will affect any http client:
```java
        .clientParser(new HttpClientParser() {
          // note you can take advantage of the actual request. ex req instanceof HttpServletRequest!
          @Override public <Req> String spanName(HttpAdapter<Req, ?> adapter, Req req) {
            return adapter.method(req).toLowerCase() + " " + adapter.path(req);
          }
        })
        .build();
```

Then when you are configuring a specific http client, you can assign the server name like so:

```java
    httpTracing = httpTracing.toBuilder().serverName("remote-service").build();
```

adriancole:
added a few client benchmarks.. I'll add the others including server-side later.

adriancole:
fyi here's an example run of the few client benchmarks (they are real requests, but against a stupid server)

```java
Benchmark                                                                       Mode  Cnt   Score   Error  Units
b.httpasyncclient.ApacheHttpAsyncClientBenchmarks.client_get                    avgt   15  61.399 ± 2.494  us/op
b.httpasyncclient.ApacheHttpAsyncClientBenchmarks.tracedButUnsampledClient_get  avgt   15  67.709 ± 4.623  us/op
b.httpasyncclient.ApacheHttpAsyncClientBenchmarks.tracedClient_get              avgt   15  69.701 ± 1.161  us/op
b.httpclient.ApacheHttpClientBenchmarks.client_get                              avgt   15  38.741 ± 1.196  us/op
b.httpclient.ApacheHttpClientBenchmarks.tracedButUnsampledClient_get            avgt   15  40.383 ± 0.768  us/op
b.httpclient.ApacheHttpClientBenchmarks.tracedClient_get                        avgt   15  45.378 ± 0.615  us/op
b.okhttp3.OkHttpClientBenchmarks.client_get                                     avgt   15  48.136 ± 1.143  us/op
b.okhttp3.OkHttpClientBenchmarks.tracedButUnsampledClient_get                   avgt   15  52.169 ± 2.923  us/op
b.okhttp3.OkHttpClientBenchmarks.tracedClient_get                               avgt   15  54.037 ± 1.181  us/op
```

adriancole:
ok all http client impls are now benchmarked. going to the http server ones next

adriancole:
just pushed benchmarks for all the http implementations client and server alike!

adriancole:
there's something up with the apachehc benchmark, as it crashes on socket problems immediately. All the others ran. Here's an example run

```
Benchmark                                                              Mode  Cnt    Score    Error  Units
b.httpasyncclient.ApacheHttpAsyncClientBenchmarks.client_get           avgt   15   72.363 ±  9.468  us/op
b.httpasyncclient.ApacheHttpAsyncClientBenchmarks.tracedClient_get     avgt   15   76.197 ±  2.698  us/op
b.httpasyncclient.ApacheHttpAsyncClientBenchmarks.unsampledClient_get  avgt   15   67.840 ±  0.797  us/op
b.jaxrs2.JaxRs2ClientBenchmarks.client_get                             avgt   15   90.393 ± 15.926  us/op
b.jaxrs2.JaxRs2ClientBenchmarks.tracedClient_get                       avgt   15  106.585 ± 18.322  us/op
b.jaxrs2.JaxRs2ClientBenchmarks.unsampledClient_get                    avgt   15   99.672 ± 13.117  us/op
b.jaxrs2.JaxRs2ServerBenchmarks.server_get                             avgt   15   87.435 ±  2.102  us/op
b.jaxrs2.JaxRs2ServerBenchmarks.tracedServer_get                       avgt   15   90.488 ±  1.378  us/op
b.jaxrs2.JaxRs2ServerBenchmarks.unsampledServer_get                    avgt   15   87.958 ±  1.392  us/op
b.okhttp3.OkHttpClientBenchmarks.client_get                            avgt   15   50.602 ±  0.936  us/op
b.okhttp3.OkHttpClientBenchmarks.tracedClient_get                      avgt   15   60.113 ±  2.676  us/op
b.okhttp3.OkHttpClientBenchmarks.unsampledClient_get                   avgt   15   55.833 ±  1.254  us/op
b.servlet.ServletBenchmarks.server_get                                 avgt   15   65.671 ±  0.932  us/op
b.servlet.ServletBenchmarks.tracedServer_get                           avgt   15   72.415 ±  1.698  us/op
b.servlet.ServletBenchmarks.unsampledServer_get                        avgt   15   66.801 ±  2.277  us/op
b.sparkjava.SparkBenchmarks.server_get                                 avgt   15   75.719 ±  1.139  us/op
b.sparkjava.SparkBenchmarks.tracedServer_get                           avgt   15   81.540 ±  1.892  us/op
b.sparkjava.SparkBenchmarks.unsampledServer_get                        avgt   15   77.176 ±  1.425  us/op
b.spring.web.RestTemplateBenchmarks.client_get                         avgt   15   67.928 ±  1.042  us/op
b.spring.web.RestTemplateBenchmarks.tracedClient_get                   avgt   15   77.949 ±  0.984  us/op
b.spring.web.RestTemplateBenchmarks.unsampledClient_get                avgt   15   74.508 ±  0.779  us/op
b.spring.webmvc.WebMvcBenchmarks.server_get                            avgt   15   97.270 ±  1.697  us/op
b.spring.webmvc.WebMvcBenchmarks.tracedServer_get                      avgt   15  100.218 ±  1.742  us/op
b.spring.webmvc.WebMvcBenchmarks.unsampledServer_get                   avgt   15  100.051 ±  3.035  us/op
```

adriancole:
the jaxrs client benchmarks look suspiciously bad. I'll check and see if @jplock's advice about the annotation is at fault, or something else.

adriancole:
actually even the untraced jaxrs-client perf is significantly worse than others. Maybe there's just some overhead that's implicit with the configuration involved.

instrumentation/p6spy/src/main/java/brave/p6spy/TracingJdbcEventListener.java
```diff
@@ -0,0 +1,88 @@
...
+
+  final String remoteServiceName;
+
+  TracingJdbcEventListener(String remoteServiceName) {
+    this.remoteServiceName = remoteServiceName;
+  }
+
+  @Override public void onBeforeAnyExecute(StatementInformation info) {
+    Tracer tracer = Tracing.currentTracer();
+    if (tracer == null) return;
```
adriancole: This method is special-cased variant of `Tracing.tracer() != null` Anyway the rationale was expanded on here https://github.com/openzipkin/brave/pull/379#discussion_r111049601

adriancole: main point is some libraries optimize for routine use of static wiring, brave doesn't. In doing so things are a lot simpler (and it discourages people from using edge-case config like currentTracer routinely)

instrumentation/servlet/README.md
```diff
@@ -0,0 +1,25 @@
...
+        .getServletContext()
+        .addFilter("TracingFilter", TracingFilter.create(tracing))
+        .addMappingForUrlPatterns(EnumSet.allOf(DispatcherType.class), true, "/*");
+  }
+
+  @Override
+  public void contextDestroyed(ServletContextEvent servletContextEvent) {
+  }
+}
+```
```
adriancole: thx will do

instrumentation/jaxrs2/src/main/java/brave/jaxrs2/TracingContainerFilter.java
adriancole: added javadoc on why, and we're free to figure out a different way!

```java
  /**
   * This implementation peeks to see if the request is async or not, which means {@link
   * PreMatching} cannot be used: pre-matching doesn't inject the resource info!
   */
```

instrumentation/jaxrs2/README.md
```diff
@@ -0,0 +1,108 @@
...
+The `TracingClientFilter` adds trace headers to outgoing requests.
+`TracingContainerFilter` extracts trace state from incoming requests.
+Both report to Zipkin how long each request takes, along with relevant
+tags like the http url.
+
+## Configuration
+
+### Dependency Injection
+Tracing always needs an instance of type `brave.http.HttpTracing`
+configured. Make sure this are in place before proceeding.
```
adriancole: done

instrumentation/servlet/README.md
adriancole: done

adriancole:
merging so that @jplock @llinder @devinsba and others can try it out for a few days via snapshots prior to release. As a reminder, the snapshot repository is http://oss.jfrog.org/artifactory/oss-snapshot-local

adriancole:
ps rewrote the PR description so it is more accessible

adriancole:
during integration testing with an example app, I noticed a glitch in how we handle scope. We should remove the span from scope *before* reporting the span, as otherwise it is confusing in logging case.

Heads up for a revision which tightens up the tests to make sure context is cleaned *before* reporting the span.

adriancole:
https://github.com/openzipkin/brave/pull/393

PR #356, opened 2017-02-23
Brave4 instrument 166

kmaci3k:
An attempt to implement requirements mentioned in https://github.com/openzipkin/brave/issues/166. 

Commits are mainly related to attaching context (i.e. for MDC) when some portion of code in done in context of a Span.

### DIFF
--- a/brave-apache-http-interceptors/src/main/java/brave/httpclient4/HttpClient4Instrumentation.java
+++ b/brave-apache-http-interceptors/src/main/java/brave/httpclient4/HttpClient4Instrumentation.java
@@ -1,0 +1,95 @@
+package brave.httpclient4;
+
+import brave.ClientHandler;
+import brave.Span;
+import brave.Tracer;
+import brave.parser.Parser;
+import brave.parser.TagsParser;
+import brave.propagation.Propagation;
+import brave.propagation.TraceContext;
+import java.io.IOException;
+import java.net.URI;
+import org.apache.http.HttpMessage;
+import org.apache.http.HttpRequestInterceptor;
+import org.apache.http.HttpResponseInterceptor;
+import org.apache.http.client.methods.HttpGet;
+import org.apache.http.client.protocol.HttpClientContext;
+import org.apache.http.impl.client.CloseableHttpClient;
+import org.apache.http.impl.client.HttpClients;
+import zipkin.TraceKeys;
+
+public final class HttpClient4Instrumentation {
+
+  public static class Config extends ClientHandler.Config<HttpClientContext, HttpClientContext> {
+
+    @Override protected Parser<HttpClientContext, String> spanNameParser() {
+      return c -> c.getRequest().getRequestLine().getMethod();
+    }
+
+    @Override protected Parser<HttpClientContext, zipkin.Endpoint> responseAddressParser() {
+      return new ServerAddressParser("");
+    }
+
+    @Override protected TagsParser<HttpClientContext> requestTagsParser() {
+      return (req, span) -> span.tag(TraceKeys.HTTP_URL,
+          req.getRequest().getRequestLine().getUri());
+    }
+
+    @Override protected TagsParser<HttpClientContext> responseTagsParser() {
+      return (res, span) -> {
+        int httpStatus = res.getResponse().getStatusLine().getStatusCode();
+        if (httpStatus < 200 || httpStatus > 299) {
+          span.tag(TraceKeys.HTTP_STATUS_CODE, String.valueOf(httpStatus));
+        }
+      };
+    }
+  }
+
+  // all this still will be set via builder
+  Tracer tracer = Tracer.newBuilder().build(); // add reporter etc
+  ClientHandler<HttpClientContext, HttpClientContext> clientHandler =
+      ClientHandler.create(new Config());
+  TraceContext.Injector<HttpMessage> injector =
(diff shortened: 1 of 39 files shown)


brave/src/main/java/brave/propagation/TraceContextHolder.java
```diff
@@ -0,0 +1,49 @@
+package brave.propagation;
+
+public class TraceContextHolder {
```
devinsba: My personal opinion is that we should avoid a class like this that uses statics to keep state. If there was a Factory in between the implementation and the users that would be more desirable to me

kmaci3k: I did it in this way because I needed a single object which would keep state for each thread. 

I'm wondering how I can achieve similar result using Factory. @devinsba, maybe you have something in mind ?

brave/src/test/java/brave/features/async/OneWaySpanTest.java
```diff
@@ -61,7 +61,7 @@
             .start().flush(); // start the server side and flush instead of processing a response
 
         flushedIncomingRequest.countDown();
-        // eventhough the client doesn't read the response, we return one
+        // even though the client doesn't read the response, we return one
```
adriancole: hehe

adriancole:
copying in @reta as he's had similar interest

adriancole:
Here are some notes from https://gitter.im/openzipkin/zipkin

## General idea

There are two lifetimes: Lifetime of the span and Lifetime of the context
* The span’s lifetime is start to finish
* The span has a lifetime within a context (attach/detach from a thread) 

https://docs.google.com/document/d/16mKbo6pWKHq30KvYYVSqo6c8uTOTkNLDSgjKwUs1PGs/edit#

## Implementation idea

```
try (Thing attachment = tracer.withCurrentSpan(span)) {
   // now span is in a thread local, MDC whatever
}
```
This PR is codebase for implementing the attaching part. One thing unlikely is that this would be a hook in the Span type. I think better to be in the Tracer itself, or most separate would be an object that only deals with in-memory context propagation.

The try/finally in this case is a good one for many reasons including that the implicit close is safe (no chance of accidentally bubbling exceptions when reporting etc). `Thing` above is most likely a sub-type of Closeable which doesn't throw.

## Takeaway

from @kmaci3k
> if I understand correctly I should do following changes:
Attachment of Span to ThreadLocal should be moved from Span to Tracer (or completely separate object). TraceContext should not have anything to do with attaching to/detaching from ThreadLocal.
tracer.withCurrentSpan(span) should not return Span but an object which just closes current Span, nothing more.

adriancole:
ps while there's no goal to exactly match, and we do want MDC plugin for sure, here's another project that has a somewhat similar api hook. you can play with it to see how it works (Even if there's only a no-op impl) https://github.com/google/instrumentation-java/blob/master/core/src/main/java/com/google/instrumentation/trace/ContextSpanHandler.java

brave-apache-http-interceptors/src/main/java/brave/httpclient4/HttpClient4Instrumentation.java
```diff
@@ -0,0 +1,95 @@
...
+   * Use this to explicitly control propagation.
+   *
+   * <p>For example, you'd create a new span for the request and response like so.
+   * <pre>{@code
+   *  Span clientSpan = spanFactory.start(parent);
+   * }</pre>
+   */
+  public HttpRequestInterceptor requestInterceptor(final TraceContext parent) {
+    return (request, context) -> {
+      Span span = parent == null ? tracer.newTrace() : tracer.newChild(parent);
```
reta: Could we simplify `tracer` API to support just something like that `Span span = tracer.newTrace(parent);` so the `tracer`  will make a decision about newTrace / newChild flow?

brave-grpc/src/main/java/brave/grpc/BraveGrpcClientInterceptor.java
```diff
@@ -0,0 +1,71 @@
...
+  public <ReqT, RespT> ClientCall<ReqT, RespT> interceptCall(
+      final MethodDescriptor<ReqT, RespT> method,
+      final CallOptions callOptions, final Channel next) {
+    return new ForwardingClientCall.SimpleForwardingClientCall<ReqT, RespT>(
+        next.newCall(method, callOptions)) {
+
+      @Override
+      public void start(Listener<RespT> responseListener, Metadata headers) {
+        // TODO: get current span
+        Span span = tracer.newTrace();
```
reta: Right, a bit relayed to related to previous comment, `tracer.newTrace()` could make a decision itself based on current span if it has to start a newTrace or newChild ... (TODO item)

brave-apache-http-interceptors/src/main/java/brave/httpclient4/HttpClient4Instrumentation.java
adriancole: newTrace would be a bad name, right? because it is not a new trace if there's an existing parent. maybe nextSpan?

reta: Agree, `nextSpan` or `startSpan`, would be better ones

adriancole:
here's a first draft of the way to plugin custom thread state handlers https://github.com/openzipkin/brave/pull/369

PR #358, opened 2017-02-28
Starts backfilling server integration tests

adriancole:
Before we re-implement instrumentation, we need to ensure we capture
existing functionality. This starts backfilling server instrumentation
tests, correcting some bugs found in the servlet filter along the way.

See #300


adriancole:
cc @lijunyong I'll do something similar for jaxrs shortly

brave-jersey2/src/test/java/com/github/kristofa/brave/jersey2/ITBraveTracingFeature_Server.java
```diff
@@ -0,0 +1,75 @@
...
+import org.junit.After;
+import org.junit.AssumptionViolatedException;
+import org.junit.Test;
+
+public class ITBraveTracingFeature_Server extends ITHttpServer {
+
+  @Override
+  @Test
+  public void reportsSpanOnTransportException() throws Exception {
+    // TODO: it seems the only way to process exceptions in a standard way is ExceptionMapper.
```
adriancole: @lijunyong can you raise a separate issue on the lack of "error" annotation in jaxrs-2 container filters. I've jotted down the reason why it currently doesn't work here.

adriancole:
follow-up pull request for jax-rs 2 on the way

PR #359, opened 2017-03-01
Instructions on how to close span on error with JAX-RS

adriancole:
@lijunyong noticed jax-rs spans don't close on error. This is
problematic as it can leak spans. It is also problematic as spans aren't
highlighted red in the UI.

Unfortunately, there's no hook in JAX-RS for unhandled exceptions. The
only way I can find is to ensure there's always an exception mapper.

Alternatively, we could use framework-specific classes. However, this
would require adding framework-specific modules, which don't yet exist.
If we did that, we'd likely want to do so as a part of #300

### DIFF
--- a/brave-jersey2/src/test/java/com/github/kristofa/brave/jersey2/ITBraveTracingFeature_Server.java
+++ b/brave-jersey2/src/test/java/com/github/kristofa/brave/jersey2/ITBraveTracingFeature_Server.java
@@ -7,7 +7,11 @@ import com.github.kristofa.brave.jaxrs2.BraveTracingFeature;
 import java.io.IOException;
 import javax.ws.rs.GET;
 import javax.ws.rs.Path;
+import javax.ws.rs.WebApplicationException;
+import javax.ws.rs.container.ContainerResponseFilter;
 import javax.ws.rs.core.Response;
+import javax.ws.rs.ext.ExceptionMapper;
+import javax.ws.rs.ext.Provider;
 import org.eclipse.jetty.servlet.ServletContextHandler;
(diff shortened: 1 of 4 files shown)


adriancole:
cc @llinder @devinsba @jplock for ideas

brave-jersey2/src/test/java/com/github/kristofa/brave/jersey2/ITBraveTracingFeature_Server.java
```diff
@@ -59,11 +52,33 @@ public Response disconnect() throws IOException {
     newServer(BraveTracingFeature.builder(brave).spanNameProvider(spanNameProvider).build());
   }
 
+  /**
+   * {@link ContainerResponseFilter} has no means to handle uncaught exceptions. Unless you provide
+   * a catch-all exception mapper, requests that result in unhandled exceptions will leak until they
+   * are eventually flushed.
+   */
+  @Provider
+  public static class ExceptionMapper implements ExtendedExceptionMapper<Exception> {
```
adriancole: we can move this to the jersey2 readme, but I don't think this is safe to add by default as it could interfere with user supplied exception mappers. My guess is that for jersey2, we'd eventually want to use framework-specific code, and this approach only as a workaround. @cschneider @schlosna thoughts?

jplock: Yea I wouldn't add this as a default as that would conflict with other registered mappers.

adriancole:
ok rebased with tests that show the instructions work

PR #367, opened 2017-03-15
Adds async tests for propagation and error handling

adriancole:
This adds async tests for thread-propagation and error handling. In
order to unpin tests from servlet 3.0, some redundant tests were
removed. This avoids creating different test modules based on servlet
version.


adriancole:
first commit is from #366

llinder:
It all looks like a good approach IMHO.

adriancole:
thanks @llinder I rebased this and fixed a broke test in jersey 2.x (wasn't pulling in the right dep for servlet 3.x support). will merge on green

PR #368, opened 2017-03-16
Add zipkin support for spark-framework

songxin1990:
Add zipkin support for spark-framework.


brave-spark-framework-filter/README.md
```diff
@@ -0,0 +1,56 @@
...
+public class BraveConfiguration {
+
+    /*@Bean 
+    Sender sender() {
+        return OkHttpSender.create("http://127.0.0.1:9411/api/v1/spans");
+    }*/
+
+    @Bean
+    public Reporter reporter() {
+        return new MyLogReporter();// I just report to log files,then use flume to gather.
```
adriancole: :) nice to know, but I don't think we want to steer people towards custom reporters when introducing spark.  how about looking at the other READMEs.. I think most of them focus on what's particular to the framework itself.

brave-spark-framework-filter/src/main/java/com/github/kristofa/brave/spark/BraveSparkRequestFilter.java
```diff
@@ -0,0 +1,67 @@
...
+            return new BraveSparkRequestFilter(this);
+        }
+    }
+
+    private final ServerRequestInterceptor requestInterceptor;
+    private final ServerSpanThreadBinder serverThreadBinder;
+    private final SpanNameProvider spanNameProvider;
+
+
+    BraveSparkRequestFilter(SpanNameProvider spanNameProvider, Brave brave) {
```
adriancole: you won't need a constructor like this as the type is new (old ctors like this were due to old code which predated builders)

brave-spark-framework-filter/pom.xml
```diff
@@ -0,0 +1,35 @@
...
+            <artifactId>brave-http</artifactId>
+            <version>${project.version}</version>
+        </dependency>
+        <dependency>
+            <groupId>com.sparkjava</groupId>
+            <artifactId>spark-core</artifactId>
+            <version>2.5</version>
+        </dependency>
+        <dependency>
+            <groupId>org.apache.httpcomponents</groupId>
```
adriancole: why is this dep here?

brave-spark-framework-filter/src/main/java/com/github/kristofa/brave/spark/SparkHttpServerResponse.java
```diff
@@ -0,0 +1,22 @@
+package com.github.kristofa.brave.spark;
+
+import com.github.kristofa.brave.http.HttpResponse;
+import spark.Response;
+
+/**
+ * Created by 00013708 on 2017/3/10.
+ */
+public class SparkHttpServerResponse implements HttpResponse {
```
adriancole: types like these don't need to be public (ps remove the automated javadoc please)

brave-spark-framework-filter/README.md
```diff
@@ -0,0 +1,56 @@
...
+        return brave;
+    }
+}
+
+```
+
+Third,use it in Spark
+
+```
+    ApplicationContext context = new ClassPathXmlApplicationContext("applicationContext.xml");
```
adriancole: is spring needed for spark?

brave-spark-framework-filter/src/test/java/com/github/kristofa/brave/spark/ITBraveSparkExample.java
```diff
@@ -0,0 +1,74 @@
...
+import java.io.BufferedReader;
+import java.io.IOException;
+import java.io.InputStreamReader;
+
+import static org.junit.Assert.assertEquals;
+
+/**
+ * Created by 00013708 on 2017/3/13.
+ */
+public class ITBraveSparkExample {
```
adriancole: in order to test all of the behaviors needed, you'd subclass ITHttpServer similar to how others are doing.

brave-spark-framework-filter/src/test/webapp/.gitignore
```diff
@@ -0,0 +1,4 @@
+# Ignore everything in this directory
+*
+# Except this file
+!.gitignore
```
adriancole: hmm I don't think we want to rely on a webapp directory. is this something spark is creating? can it be changed to be a part of the "target" directory instead?

songxin1990:
@adriancole  OK,I will take all your advice,fix these things AQAP.I am fresh as an open source contributor Thanks a lot,sincerely! : )

adriancole:
@songxin1990 that's a method that doesn't exist in all servlet versions.. can you check minimum servlet version of spark?

brave-spark-framework-filter/src/main/java/com/github/kristofa/brave/spark/BraveSparkExceptionHandler.java
```diff
@@ -0,0 +1,48 @@
...
+        this.brave = brave;
+        this.exceptionHandler = exceptionHandler;
+    }
+
+    @Override
+    public void handle(Exception ex, Request request, Response response) {
+        try {
+            try {
+                ServerTracer serverTracer = brave.serverTracer();
+                if (serverTracer != null) {
```
adriancole: why would serverTracer be null?

adriancole: did you mean to check if the exception is null?

brave-spark-framework-filter/src/test/java/com/github/kristofa/brave/spark/ITBraveSparkTest.java
```diff
@@ -0,0 +1,67 @@
...
+import org.junit.After;
+import org.junit.AssumptionViolatedException;
+import org.junit.Test;
+import spark.*;
+
+/**
+ * Created by 00013708 on 2017/3/17.
+ */
+public class ITBraveSparkTest extends ITHttpServer {
+    /*after filter executed before NotFoundRoute,there is no way to add 404 code after serverspan
```
adriancole: good note

brave-spark-framework-filter/src/test/java/com/github/kristofa/brave/spark/ITBraveSparkTest.java
```diff
@@ -0,0 +1,67 @@
...
+    @Override
+    @Test
+    public void addsStatusCodeWhenNotOk() throws Exception {
+        throw new AssumptionViolatedException("TODO: fix error reporting");
+    }
+
+    @Override
+    protected void init(Brave brave, SpanNameProvider spanNameProvider) throws Exception {
+        stop();
+        Spark.init();
```
adriancole: spark always uses static initialization? maybe there's a junit rule for spark?

brave-spark-framework-filter/src/main/java/com/github/kristofa/brave/spark/BraveSparkExceptionHandler.java
songxin1990: serverTracer can not be null,so it's unnecessary.

brave-spark-framework-filter/src/test/java/com/github/kristofa/brave/spark/ITBraveSparkTest.java
songxin1990: I reply all questions here.
1.Spark depends on jetty9.3.x，jetty9 with java-servlet-api version 3.1.
2.Spark always uses static initialization,anything like Spark.init(),Spark.get(),Spark.post() etc,
will init Spark only instance and once.

adriancole: ok

1) latest commit on brave moves the test runner to servlet 3.x (which means you can test async).

2) I'd double-check https://github.com/perwendel/spark/blob/master/src/test/java/spark/BooksIntegrationTest.java#L58 to ensure initialization is the same form, then..

songxin1990: 1.OK,I will have a try using test async.
2.Spark.awaitInitialization() is a method blocks wait spark initialization finished.I had used it ,but some problems happen.There should be a method for Spark.waitStop().
https://github.com/perwendel/spark/pull/730

adriancole:
>
> 2.Spark.awaitInitialization() is a method blocks wait spark initialization
> finished.I had used it ,but some problems happen.There should be a method
> for Spark.waitStop().
> perwendel/spark#730 <https://github.com/perwendel/spark/pull/730>
>
cool thanks for letting us know!

songxin1990:
Recently I prepare for a Kafka lecture,so a little busy.After Spark initialized or stopped,there is no way to get the variable 'initialized' in Spark.Service,nor there is a callback,so I can't use test async.
https://github.com/perwendel/spark/blob/master/src/main/java/spark/Service.java

I have looked for some other solutions,but can't.So,what can I do now?

songxin1990:
Sorry for Reopen this PR,not familiar.

adriancole:
For a something you cant test, override the method in your subclass and
throw AssumptionViolatedException.  This will ignore failures. Put a todo
comment of why it is being ignored.

songxin1990:
@adriancole  I have try my best,all Junit tests got passed.And  I add some more detailed comments on some Class and methods.

brave-spark-framework-filter/README.md
```diff
@@ -0,0 +1,39 @@
...
+
+Here is a use Example:
+
+Add a dependency:
+
+```
+<dependency>
+    <groupId>io.zipkin.brave</groupId>
+    <artifactId>brave-spark-framework-filter</artifactId>
+    <version>4.0.6</version>
```
adriancole: remove the version or use something that dynamically updates it.

brave-spark-framework-filter/README.md
```diff
@@ -0,0 +1,39 @@
...
+
+Use it in Spark:
+
+```
+    Sender sender = OkHttpSender.create("http://127.0.0.1:9411/api/v1/spans");
+    
+    Reporter reporter = new LoggingReporter();
+    //Reporter reporter = AsyncReporter.builder(sender()).build();
+    
+    Brave brave = new Brave.Builder("service-name").reporter(reporter).build();
```
adriancole: you can axe the logic here and above.. showing how to initialize brave is something covered elsewhere. also, this removes the extra dep (okhttp) and the commenting out of it (asyncreporter) which distract from the main issue.

brave-spark-framework-filter/README.md
```diff
@@ -0,0 +1,39 @@
...
+    
+    Spark.before(BraveSparkRequestFilter.create(brave));
```
adriancole: these parts don't look like code.. is this how it would look? wouldn't you set the `before` `exception` `after` stuff *before* using spark? If so, I'd make a comment like 

```java
// when setting up spark, add filters and handlers like so..
...

// now, a request like below will be traced
Spark.get("/foo", (req, res) -> "bar");
```

brave-spark-framework-filter/README.md
```diff
@@ -0,0 +1,39 @@
...
+    <groupId>io.zipkin.brave</groupId>
+    <artifactId>brave-spark-framework-filter</artifactId>
+    <version>4.0.6</version>
+</dependency>
+        
+```
+
+Use it in Spark:
+
+```
```
adriancole: note..  in code blocks you can add a formatting hint right after the triple back-tick ex. ```java

brave-spark-framework-filter/pom.xml
```diff
@@ -0,0 +1,41 @@
...
+            <artifactId>spark-core</artifactId>
+            <version>2.5.5</version>
+        </dependency>
+        <dependency>
+            <groupId>${project.groupId}</groupId>
+            <artifactId>brave-http-tests</artifactId>
+            <scope>test</scope>
+        </dependency>
+
+        <dependency>
```
adriancole: it doesn't appear this dependency is used. If it is needed, add an xml comment why?

brave-spark-framework-filter/src/main/java/com/github/kristofa/brave/spark/BraveSparkExceptionHandler.java
```diff
@@ -0,0 +1,57 @@
...
+import spark.ExceptionHandler;
+import spark.Request;
+import spark.Response;
+import zipkin.Constants;
+
+import static com.github.kristofa.brave.internal.Util.checkNotNull;
+
+/**
+ * All your ExceptionHandler implements must be wrapped by BraveSparkExceptionHandler,
+ * If not,when exception happens,server span can't be generated,because AfterFilters won't be executed.
```
adriancole: watch for spacing. Also, wrap types in `{@link ExceptionHandler}` or `{@linkplain ExceptionHandler}`

brave-spark-framework-filter/src/main/java/com/github/kristofa/brave/spark/BraveSparkExceptionHandler.java
```diff
@@ -0,0 +1,57 @@
...
+ * See https://github.com/perwendel/spark/issues/715 .
```
adriancole: linebreak then add `<p>` before the code example

brave-spark-framework-filter/src/main/java/com/github/kristofa/brave/spark/BraveSparkExceptionHandler.java
```diff
@@ -0,0 +1,57 @@
...
+ * {@code
+ * Spark.exception(Exception.class, BraveSparkExceptionHandler.create(brave, new ExceptionHandler() {
+        @Override
+        public void handle(Exception exception, Request request, Response response) {
+            //do something you want
+        }
+    }));
+ * }
+ */
+public class BraveSparkExceptionHandler implements ExceptionHandler {
```
adriancole: make all public types final or comment why they aren't

brave-spark-framework-filter/src/main/java/com/github/kristofa/brave/spark/BraveSparkExceptionHandler.java
```diff
@@ -0,0 +1,57 @@
...
+
+    private BraveSparkExceptionHandler(Brave brave, ExceptionHandler exceptionHandler) {
+        this.brave = brave;
+        this.exceptionHandler = exceptionHandler;
+    }
+
+    @Override
+    public void handle(Exception ex, Request request, Response response) {
+        try {
+            try {
```
adriancole: why double-try?

brave-spark-framework-filter/src/main/java/com/github/kristofa/brave/spark/BraveSparkExceptionHandler.java
```diff
@@ -0,0 +1,57 @@
...
+                ServerTracer serverTracer = brave.serverTracer();
+                serverTracer.submitBinaryAnnotation(Constants.ERROR, ex.getMessage());
```
adriancole: seems like you need to check null on ex? Also, the message might be null. check other code for how this is handled..

brave-spark-framework-filter/src/main/java/com/github/kristofa/brave/spark/BraveSparkRequestFilter.java
```diff
@@ -0,0 +1,69 @@
...
+import com.github.kristofa.brave.http.SpanNameProvider;
+import com.github.kristofa.brave.spark.internal.MaybeAddClientAddressFromRequest;
+import spark.Filter;
+import spark.Request;
+import spark.Response;
+
+import static com.github.kristofa.brave.internal.Util.checkNotNull;
+
+
+public class BraveSparkRequestFilter implements Filter {
```
adriancole: final

brave-spark-framework-filter/src/main/java/com/github/kristofa/brave/spark/BraveSparkRequestFilter.java
```diff
@@ -0,0 +1,69 @@
...
+    }
+
+    @Override
+    public void handle(Request request, Response response) throws Exception {
+        this.requestInterceptor.handle(new HttpServerRequestAdapter(new SparkHttpServerRequest(request), this.spanNameProvider));
+        if (maybeAddClientAddressFromRequest != null) {
+            maybeAddClientAddressFromRequest.accept(request.raw());
+        }
+        ServerSpan serverSpan =  serverThreadBinder.getCurrentServerSpan();
+        System.out.println("REQ serverSpan=>"+serverSpan.toString());
```
adriancole: remove all things like this

brave-spark-framework-filter/src/main/java/com/github/kristofa/brave/spark/BraveSparkRequestFilter.java
```diff
@@ -0,0 +1,69 @@
...
+        this.requestInterceptor = b.brave.serverRequestInterceptor();
+        this.serverThreadBinder = b.brave.serverSpanThreadBinder();
+        this.spanNameProvider = b.spanNameProvider;
+        maybeAddClientAddressFromRequest = MaybeAddClientAddressFromRequest.create(b.brave);
+    }
+
+    @Override
+    public void handle(Request request, Response response) throws Exception {
+        this.requestInterceptor.handle(new HttpServerRequestAdapter(new SparkHttpServerRequest(request), this.spanNameProvider));
+        if (maybeAddClientAddressFromRequest != null) {
```
adriancole: this won't ever be null

brave-spark-framework-filter/src/main/java/com/github/kristofa/brave/spark/BraveSparkResponseFilter.java
```diff
@@ -0,0 +1,60 @@
...
+import com.github.kristofa.brave.http.HttpServerResponseAdapter;
+import com.github.kristofa.brave.http.SpanNameProvider;
+import spark.Filter;
+import spark.Request;
+import spark.Response;
+
+import static com.github.kristofa.brave.internal.Util.checkNotNull;
+
+
+public class BraveSparkResponseFilter implements Filter {
```
adriancole: final

brave-spark-framework-filter/src/main/java/com/github/kristofa/brave/spark/BraveSparkResponseFilter.java
```diff
@@ -0,0 +1,60 @@
...
+    BraveSparkResponseFilter(Builder b) { // intentionally hidden
+        this.responseInterceptor = b.brave.serverResponseInterceptor();
+        this.serverThreadBinder = b.brave.serverSpanThreadBinder();
+        this.spanNameProvider = b.spanNameProvider;
+        this.serverTracer = b.brave.serverTracer();
+    }
+
+
+    @Override
+    public void handle(Request request, Response response) throws Exception {
```
adriancole: does this need to be in two types? maybe you can make a single type instead of BraveSparkRequestFilter  and BraveSparkResponseFilter and see which of the params are null? Seems like less code to maintain that way. ex it could be named BraveSparkFilter

brave-spark-framework-filter/src/test/java/com/github/kristofa/brave/spark/ITBraveSparkTest.java
```diff
@@ -0,0 +1,67 @@
...
+import spark.Spark;
+
+public class ITBraveSparkTest extends ITHttpServer {
+
+    /*after filter executed before NotFoundRoute,there is no way to add 404 code after serverspan
+    has been setServerSend,may be throw a not found exceeption and process it in ExceptionHandler*/
+    @Override
+    @Test
+    public void addsStatusCodeWhenNotOk() throws Exception {
+        throw new AssumptionViolatedException("TODO: fix error reporting");
```
adriancole: change TODO... to "Filters run before NotFoundRoute can be determined"

brave-spark-framework-filter/src/test/java/com/github/kristofa/brave/spark/ITBraveSparkTest.java
```diff
@@ -0,0 +1,67 @@
...
+    }
+    @Override
+    protected void init(Brave brave, SpanNameProvider spanNameProvider) throws Exception {
+        stop();
+        LocalTracer localTracer = brave.localTracer();
```
adriancole: remove the test code.

brave-spark-framework-filter/src/test/java/com/github/kristofa/brave/spark/ITBraveSparkTest.java
```diff
@@ -0,0 +1,67 @@
...
+        Spark.after(respFilter);
+        Spark.awaitInitialization();
+    }
+
+    @Override
+    protected String url(String path) {//default port 4567
+        return "http://localhost:4567" + path;
+    }
+
+    /*Spark stop asynchronously but share one class Instance,
```
adriancole: looks like you aren't using a code formatter. can you use the one in CONTRIBUTING.md please?

songxin1990:
OK,after work I will fix these things.

brave-spark-framework-filter/src/main/java/com/github/kristofa/brave/spark/BraveSparkExceptionHandler.java
songxin1990: yes,you are right.

brave-spark-framework-filter/README.md
songxin1990: It doesn't matter the order of before exception after.

brave-spark-framework-filter/src/main/java/com/github/kristofa/brave/spark/BraveSparkResponseFilter.java
songxin1990: Yes,I thought it before,it can be in one types.But I think it's Spark convention and if in one BraveSparkFilter,there can't be two Override handle method.
Request and response are not null.

brave-spark-framework-filter/src/test/java/com/github/kristofa/brave/spark/ITBraveSparkTest.java
songxin1990: not quite understand,which code is test code,LocalTracer localTracer = brave.localTracer()?

brave-spark-framework-filter/src/main/java/com/github/kristofa/brave/spark/BraveSparkResponseFilter.java
songxin1990: image

brave-spark-framework-filter/src/test/java/com/github/kristofa/brave/spark/ITBraveSparkTest.java
adriancole: this is an initialization method, not a unit test. So, I'm asking to remove the unnecessary logic from here as it isn't needed to initialize the trace instrumentation

brave-spark-framework-filter/src/main/java/com/github/kristofa/brave/spark/BraveSparkResponseFilter.java
songxin1990: Maybe, you like this?

adriancole:
First.. I noticed we should probably use afterAfter as that will close the
span even if there is an exception. That makes it optional for users to use
an exception mapper.

Maybe something like this (which will be similar for the brave 4 version
https://github.com/openzipkin/brave/pull/350)

BraveTracing tracing = BraveTracing.create(brave);
Spark.before(tracing.before()); // request side, which starts the span
Spark.afterAfter(tracing.afterAfter()); // response/exception side, which
closes the span

Optionally, you can have an exception mapper, who only adds the tag/binary
annotation "error" (but doesn't close the span as afterAfter will)

cc @perwendel wdyt?

songxin1990:
Yes,Spark.afterAfter() will be executed no matter what happens,very like **finally**.But  use afterAfter,the binary annotation will not be tagged,so will not show in Zipkin UI.

brave-spark-framework-filter/src/main/java/com/github/kristofa/brave/spark/BraveSparkExceptionHandler.java
songxin1990: Even if exception happens when Zipkin trace,the user's exceptionHandler will be executed

brave-spark-framework-filter/src/main/java/com/github/kristofa/brave/spark/BraveSparkExceptionHandler.java
songxin1990: ex can't be null,ex.getMessage() may be null.I am fixed it now.Maybe if ex.getMessage() is null,insert 
response.raw().getStatus() and response.body() into binaryAnnotation value?

adriancole:
right, just clarifying that if the response filter is afterAfter, then the exception mapper is only adding the binary annotation aka tag "error". In this case, its responsibility is easier, which is particularly helpful as the exception mapper component needs to be customized by the user. For example, even if they forget or misconfigure this, the span will still close if the response filter is afterAfter.

brave-spark-framework-filter/README.md
```diff
@@ -7,33 +7,13 @@ create and submit span with `sr`', `ss` annotations.
...
-    //Reporter reporter = AsyncReporter.builder(sender()).build();
-    
-    Brave brave = new Brave.Builder("service-name").reporter(reporter).build();
-    
-    Spark.before(BraveSparkRequestFilter.create(brave));
-    
+    Spark.before(BraveSparkFilter.create(brave).requestFilter());
     Spark.get("/foo", (req, res) -> "bar");
-    
+    //if you need add binaryAnnotation "error",use this.
```
adriancole: I'd still recommend always setting this (ex remove the clarification, or change it to "unless you add this, spans will not be marked as error).

Also, it is still sensible to have the configuration comments above the user code. It is more clear what is brave related and what is user related. In other words..

```java
// setup brave tracing...
Spark.before(braveTracing.before());
Spark.afterAfter(braveTracing.afterAfter());
Spark.exception(Exception.class, BraveSparkExceptionHandler.create(brave, new ExceptionHandlerImpl()));

// any routes you add are now traced, such as the below
 Spark.get("/foo", (req, res) -> "bar");
```

songxin1990: OK,I will fix it.

songxin1990:
Any more problems?

brave-spark-framework-filter/src/main/java/com/github/kristofa/brave/spark/BraveSparkExceptionHandler.java
```diff
@@ -0,0 +1,49 @@
...
+import spark.utils.StringUtils;
+import zipkin.Constants;
+
+import static com.github.kristofa.brave.internal.Util.checkNotNull;
+
+public final class BraveSparkExceptionHandler implements ExceptionHandler {
+
+    private final Brave brave;
+
+    private final ExceptionHandler exceptionHandler;
```
adriancole: rename to delegate (more common convention for forwarding types)

brave-spark-framework-filter/src/main/java/com/github/kristofa/brave/spark/BraveSparkExceptionHandler.java
```diff
@@ -0,0 +1,49 @@
...
+        this.exceptionHandler = exceptionHandler;
+    }
+
+    @Override
+    public void handle(Exception ex, Request request, Response response) {
+        try {
+            ServerTracer serverTracer = brave.serverTracer();
+            StringBuilder sb = new StringBuilder();
+            if (StringUtils.isNotEmpty(ex.getMessage())) {
+                sb.append(ex.getMessage()).append("\n");
```
adriancole: in other code, we have been only doing the exception message. probably a better idea to just follow that as it is also less code to test. People can always make fancier exception handlers

```java
      String message = throwable.getMessage();
      if (message == null) message = throwable.getClass().getSimpleName();
      span.tag(Constants.ERROR, message);
```

less code is better especially in new integrations

brave-spark-framework-filter/src/main/java/com/github/kristofa/brave/spark/BraveSparkFilter.java
```diff
@@ -0,0 +1,77 @@
+package com.github.kristofa.brave.spark;
+
+import com.github.kristofa.brave.*;
```
adriancole: no wildcard imports.. also new files are supposed to use 2 space indent.

easiest way is to use square codestyle and reformat all your code

https://github.com/square/java-code-styles

brave-spark-framework-filter/src/main/java/com/github/kristofa/brave/spark/BraveSparkFilter.java
```diff
@@ -0,0 +1,77 @@
...
+    BraveSparkFilter(BraveSparkFilter.Builder b) { // intentionally hidden
+        this.requestInterceptor = b.brave.serverRequestInterceptor();
+        this.responseInterceptor = b.brave.serverResponseInterceptor();
+        this.serverThreadBinder = b.brave.serverSpanThreadBinder();
+        this.spanNameProvider = b.spanNameProvider;
+        this.serverTracer = b.brave.serverTracer();
+        maybeAddClientAddressFromRequest = MaybeAddClientAddressFromRequest.create(b.brave);
+    }
+
+    public Filter requestFilter() {
```
adriancole: I still prefer naming these methods after what they are for. ex before and afterAfter. this makes them self-documenting. `responseFilter()` doesn't help someone know to use afterAfter or after

brave-spark-framework-filter/src/main/java/com/github/kristofa/brave/spark/BraveSparkExceptionHandler.java
```diff
@@ -0,0 +1,49 @@
...
+
+import static com.github.kristofa.brave.internal.Util.checkNotNull;
+
+public final class BraveSparkExceptionHandler implements ExceptionHandler {
+
+    private final Brave brave;
+
+    private final ExceptionHandler exceptionHandler;
+
+    public static BraveSparkExceptionHandler create(Brave brave, ExceptionHandler exceptionHandler) {
```
adriancole: I prefer hiding this type and exposing `BraveSparkFilter.exception(ExceptionHandler delegate)`

ps BraveSparkFilter name change in another comment..

brave-spark-framework-filter/src/main/java/com/github/kristofa/brave/spark/BraveSparkFilter.java
```diff
@@ -0,0 +1,77 @@
...
+import com.github.kristofa.brave.http.HttpServerResponseAdapter;
+import com.github.kristofa.brave.http.SpanNameProvider;
+import com.github.kristofa.brave.spark.internal.MaybeAddClientAddressFromRequest;
+import spark.Filter;
+import spark.Request;
+import spark.Response;
+
+import static com.github.kristofa.brave.internal.Util.checkNotNull;
+
+public final class BraveSparkFilter {
```
adriancole: I'd rename this type to `BraveTracing` as it flows nicer, also helps as not everyone knows brave is for tracing.. plus #350 is going for this convention.

Ex.

```java
BraveTracing tracing = BraveTracing.create(brave);
Spark.before(tracing.before());
Spark.exception(Exception.class, tracing.exception(new ExceptionHandlerImpl()));
Spark.afterAfter(tracing.afterAfter());
```

brave-spark-framework-filter/src/main/java/com/github/kristofa/brave/spark/internal/MaybeAddClientAddressFromRequest.java
```diff
@@ -0,0 +1,36 @@
...
+
+import javax.servlet.http.HttpServletRequest;
+
+/**
+ * Parses the {@link Constants#CLIENT_ADDR client address}, by looking at "X-Forwarded-For", then
+ * the remote address of the request. This performs no DNS lookups.
+ *
+ * <p>This is a hack as {@link ServerResponseInterceptor} doesn't yet support client addresses.
+ */
+public final class MaybeAddClientAddressFromRequest
```
adriancole: add a dependency on web-servlet-filter so you don't have to copy/paste this and also its tests

brave-spark-framework-filter/src/test/java/com/github/kristofa/brave/spark/ITBraveSparkTest.java
```diff
@@ -0,0 +1,66 @@
...
+    @Override
+    @Test
+    public void addsStatusCodeWhenNotOk() throws Exception {
+        throw new AssumptionViolatedException("Filters run before NotFoundRoute can be determined");
+    }
+    @Override
+    protected void init(Brave brave, SpanNameProvider spanNameProvider) throws Exception {
+        stop();
+        Spark.before(BraveSparkFilter.builder(brave).spanNameProvider(spanNameProvider).build().requestFilter());
+        Spark.get("/foo", (req, res) -> "bar");
```
adriancole: please move the route definitions below the brave setup code, as eventhough it is valid, it is confusing to read a mix of concerns.

brave-spark-framework-filter/src/test/java/com/github/kristofa/brave/spark/ITBraveSparkTest.java
```diff
@@ -0,0 +1,66 @@
...
+import spark.ExceptionHandler;
+import spark.Request;
+import spark.Response;
+import spark.Spark;
+
+public class ITBraveSparkTest extends ITHttpServer {
+
+    /**
+     * after filter executed before NotFoundRoute,there is no way to add 404 code after serverspan
+     * has been setServerSend,may be throw a not found exceeption and process it in ExceptionHandler
```
adriancole: nit: watch for punctuation and spelling errors like exceeption and lack of space after comma

songxin1990:
Thank you very much for spending your much time And I learn a lot  from you.
I will fix these things after work ASAP.

brave-spark-framework-filter/src/main/java/com/github/kristofa/brave/spark/BraveSparkExceptionHandler.java
songxin1990: The same thought as you,but I thing change brave-spark-framework-filter  to sparkjava better.

brave-spark-framework-filter/src/main/java/com/github/kristofa/brave/spark/BraveSparkFilter.java
songxin1990: Maybe change BraveSparkFilter  to BraveTracingSparkJava?

songxin1990:
@adriancole  I have finished.

brave-sparkjava/README.md
```diff
@@ -0,0 +1,22 @@
+# brave-spark-framework-filter
```
adriancole: nit rename to sparkjava

brave-sparkjava/src/main/java/com/github/kristofa/brave/sparkjava/BraveTracingSparkJava.java
```diff
@@ -0,0 +1,99 @@
...
+import com.github.kristofa.brave.servlet.internal.MaybeAddClientAddressFromRequest;
+import spark.ExceptionHandler;
+import spark.Filter;
+import spark.Request;
+import spark.Response;
+import zipkin.Constants;
+
+import static com.github.kristofa.brave.internal.Util.checkNotNull;
+
+public final class BraveTracingSparkJava {
```
adriancole: I feel the SparkJava suffix is implied. The intent with the `BraveTracing` name was that in sparkjava stack traces or debug, you'd see "brave and tracing" there.

SparkJava is redundant as all the types returned are defined in sparkjava.

If we adopt a more verbose naming convention, then it will make other things a bit longer, too. Ex in okhttp as opposed to BraveTracingInterceptor, it would be BraveTracingOkHttpInterceptor (eventhough it would only be usable in okhttp due to the type being declared in okhttp)

I know naming conventions are hard, but hoping you might agree that dropping the SparkJava suffix doesn't hurt and also makes for a more consistent and less verbose naming strategy for us to apply to other code.

songxin1990: Yes,your advise is better.I will change BraveTracingSparkJava  to BraveTracing. As user add dependence ,they will know BraveTracing is used for traceing in SparkJava.

songxin1990:
@adriancole  I have finished.

songxin1990:
If you need any help just leave me a message or use @ mention me. 
Maybe,I should  view the issues list,see what I can do.

adriancole:
thanks for the help! yeah next step will be porting it to brave 4

adriancole:
oops!  we forgot to link this to the parent pom.xml so it wasn't released. will fix in a sec

adriancole:
https://github.com/openzipkin/brave/pull/376

songxin1990:
OK，ASAP.

adriancole:
ok it is out on brave 4.1.1 now

Issue #370, opened 2017-03-28
brave-grpc could not work  when update Grpc to 1.2.0


Mylog:
an @ExperimentalApi  method had been removed in  io.grpc.ServerCall ,
So compile error occured at   BraveGrpcServerInterceptor   line 134,
change to 
  SocketAddress socketAddress = (SocketAddress)this.call.getAttributes().get(Key.of("remote-addr"));
would fix the bug.

adriancole:
Ack. Care to raise a PR?

adriancole:
actually the suggested fix won't work because grpc keys are compared by identity (so making a copy wouldn't match).

adriancole:
there's actually two problems here.. one is that "attributes" has been renamed to "getAttributes" the next is where did the remote address attribute go to?

Safest thing at the moment is to remove the client address functionality as it is a locks up the implementation to <grpc 1.2, which is a bad thing..

adriancole:
ok dropping support for client address is the best thing for now. Let's revisit later as I'm aware grpc are adding native tracing support anyway

#371

adriancole:
Brave 4.3 is being released now and has a new package for GRPC. we can update this to detect things about client address, but keep in mind GRPC 1.3 has a new tracing library

PR #371, opened 2017-04-02
Enables an upgrade path to gRPC 1.2 by removing references to ServerCall

adriancole:
Until people move off gRPC 1.0.x, it may not be easy to re-introduce support
for client address.

See #370


adriancole:
If we want to revisit this, what we'd need to do is something like okhttp's Platform and reflectively use a different approach to support grpc <1.2 and 1.2+ Intentionally deferring that complexity.

PR #376, opened 2017-04-10
Links sparkjava to build so that it is published

adriancole:
sparkjava wasn't added to the root pom, so it never built, nor was
published.

This also uses standard formatting and fixes the naming convention on
the integration test so it isn't run twice.


adriancole:
will release brave 4.1.1 so that what we announced is actually available.

PR #377, opened 2017-04-11
Makes circleci use maven wrapper and forces retrolambda to fork

adriancole:
Before, circleci's test phase didn't use the maven wrapper which was
inconsistent with other phases.

Also, retrolambda wasn't forking and this caused some spurious issues
attempting to backport lambdas.


adriancole:
note the retrolambda fail only happened in circleci (not osx)

PR #379, opened 2017-04-11
Adds Tracer.current(): last resort when you can't otherwise get a tracer

### LINKED ISSUE #206: Add service locator for Brave

adriancole:
This adds `Tracer.current()` which particularly paves the way for trace
instrumentation on resources that initialize prior to the tracer, like
JDBC drivers.

static method `Tracer.current()` represents the last initialized Tracer
instance method `Tracer.close()` deregisters a tracer if it is current

This was based on an implementation in #210 by @ewhauser
The close hook hopefully satisfies @schlosna deregistration request

This isn't a separate type or pluggable in attempts to keep the behavior
easy to understand.

Closes #210
Fixes #206

### DIFF
--- a/brave/src/main/java/brave/Tracer.java
+++ b/brave/src/main/java/brave/Tracer.java
@@ -44,7 +44,35 @@ import zipkin.reporter.Sender;
  * @see Span
  * @see Propagation
  */
-public final class Tracer {
+public final class Tracer implements Closeable {
+  // volatile for visibility on get. writes guarded by Tracer.class
+  static volatile Tracer current = null;
+
+  /**
+   * Returns the most recently created tracer iff it hasn't been closed. null otherwise.
+   *
+   * <p>This object should not be cached.
+   */
+  @Nullable public static Tracer current() {
+    return current;
+  }
+
+  private void maybeSetCurrent() {
+    if (current != null) return;
+    synchronized (Tracer.class) {
+      if (current == null) current = this;
+    }
+  }
+
+  /** Ensures this tracer can be garbage collected, by making it not {@link #current()} */
+  @Override public void close() {
+    if (current != this) return;
+    // don't blindly set most recent to null as there could be a race
+    synchronized (Tracer.class) {
+      if (current == this) current = null;
+    }
+  }
+
   public static Builder newBuilder() {
     return new Builder();
   }
@@ -170,6 +198,7 @@ public final class Tracer {
     this.sampler = builder.sampler;
     this.currentTraceContext = builder.currentTraceContext;
     this.traceId128Bit = builder.traceId128Bit;
+    maybeSetCurrent();
   }
 
   /** Used internally by operations such as {@link Span#finish()}, exposed for convenience. */
--- a/brave/src/main/java/brave/propagation/CurrentTraceContext.java
+++ b/brave/src/main/java/brave/propagation/CurrentTraceContext.java
@@ -34,9 +34,12 @@ public abstract class CurrentTraceContext {
     @Override void close();
   }
 
-  /** Default implementation which is backed by an inheritable thread local */
+  /** Default implementation which is backed by a static inheritable thread local */
   public static final class Default extends CurrentTraceContext {
-    final InheritableThreadLocal<TraceContext> local = new InheritableThreadLocal<>();
+    // static as we want one context per thread, not one context per thread-instance.
+    // if this is not static, patterns that coordinate via statics (like Tracer.current()) will break.
+    // static ThreadLocal was also used in Brave 3's ThreadLocalServerClientAndLocalSpanState
+    static final InheritableThreadLocal<TraceContext> local = new InheritableThreadLocal<>();
 
     @Override public TraceContext get() {
       return local.get();
--- a/brave/README.md
+++ b/brave/README.md
@@ -28,6 +28,11 @@ tracer = Tracer.newBuilder()
               .localServiceName("my-service")
               .reporter(reporter)
               .build();
+
+// When all tracing tasks are complete, close the tracer and reporter
+// This might be a shutdown hook for some users
+tracer.close();
+reporter.close();
 ```
 
 ## Tracing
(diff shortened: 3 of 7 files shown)


adriancole:
@devinsba @semyonslepov @llinder looking for a review as this helps me complete the re-instrumentation and frees hands for fun things like elasticsearch optimizations!

brave/src/main/java/brave/propagation/CurrentTraceContext.java
```diff
@@ -34,9 +34,12 @@
...
   }
 
-  /** Default implementation which is backed by an inheritable thread local */
+  /** Default implementation which is backed by a static inheritable thread local */
   public static final class Default extends CurrentTraceContext {
-    final InheritableThreadLocal<TraceContext> local = new InheritableThreadLocal<>();
+    // static as we want one context per thread, not one context per thread-instance.
+    // if this is not static, patterns that coordinate via statics (like Tracer.current()) will break.
+    // static ThreadLocal was also used in Brave 3's ThreadLocalServerClientAndLocalSpanState
+    static final InheritableThreadLocal<TraceContext> local = new InheritableThreadLocal<>();
```
adriancole: without this (which was a port miss from brave 3), Tracer.current() would result in broken tracing

brave/src/main/java/brave/Tracer.java
```diff
@@ -44,7 +44,35 @@
  * @see Span
  * @see Propagation
  */
-public final class Tracer {
+public final class Tracer implements Closeable {
+  // volatile for visibility on get. writes guarded by Tracer.class
+  static volatile Tracer current = null;
+
+  /**
+   * Returns the most recent tracer that hasn't been closed.
```
devinsba: Maybe its better as:

`Returns the most recently created tracer iff it hasn't been closed. null otherwise

adriancole: sgtm

brave/src/main/java/brave/Tracer.java
```diff
@@ -44,7 +44,35 @@
...
+   *
+   * <p>This object should not be cached.
+   */
+  @Nullable public static Tracer current() {
```
llinder: Is it expected that instrumentation that uses this would just perform a delayed retry if `Tracer.current()` was null?  Its probably overkill but maybe some kind of callback approach would be better better suited as apposed to null.

adriancole: Here's what the code looks like for p6spy (and mysql which are the only instrumentation so far I have seen need this)

```java
@Override public void onBeforeAnyExecute(StatementInformation info) {
    Tracer tracer = Tracer.current();
    if (tracer == null) return;

    Span span = tracer.nextSpan();
    ...
```

One option is to create a noop tracer, or some composite that allows none. This affects the codebase more for what it a special case (statically initialized trace instrumentation). While not objective, I think it also encourages a pattern we should discourage, which is to rely on `Tracer.current()` for routine work. If we did, we'd need to rethink what that would return, or make a delegate type.

Ex with a special lazy initialized tracer (which implies defining what empty means)
```java
   // note tracer currently has luxury of being a final type.
   // This implies we make something that doles out span ids even in noop case!
   // Doing so could break propagation when a tracer is eventually initialized.
   final Tracer currentTracer = Tracer.current(); // can be static or instance

@Override public void onBeforeAnyExecute(StatementInformation info) {
    Span span = tracer.nextSpan();
    ...
```

If we did a callback approach, I think it would be a bit more complex, especially from a GC perspective, as we'd have a registration (ex. MySQL would register with an intermediary for tracer update callbacks). I think the callsite of the instrumentation would be slightly less cluttered, but the machinery would be substantial and perhaps maintenance causing.

```java
// static or instance, but needs to be volatile since callback can happen on any thread.
volatile Tracer currentTracer; 

@Override public void onCurrentTracer(Tracer currentTracer) {
   this.currentTracer = currentTracer;
}

@Override public void onBeforeAnyExecute(StatementInformation info) {
    if (currentTracer == null) return;

    Span span = tracer.nextSpan();
    ...
```

With the proposed, we can make our instrumentation indirect access to Tracer.current() anyway, availing users to manage how they see fit. Users could use something framework specific that might be a better fit (ex static injection).

```java
/** Override to use a mechanism besides Tracer.current(), such as static injection */
@Nullable protected Tracer currentTracer() {
  return Tracer.current();
}

@Override public void onBeforeAnyExecute(StatementInformation info) {
    Tracer tracer = currentTracer();
    if (tracer == null) return;

    Span span = tracer.nextSpan();
    ...
```

wdyt?

llinder: All very good points which illustrator the simplicity of your current approach.  Given that current use cases are already calling `Tracer.current()` for each execution and that this pattern is discourage, I think it makes a sense to keep it as is.  Thanks for the thorough reply :)

adriancole: NP!

adriancole:
ps it is understood that some frameworks may support alternative static
injection, or more sophisticated scoping. As use of `Tracer.current()`
is often coupled with service loader or `Class.newInstance()`, such
extensibility would require statically available machinery similar to [google's](https://github.com/google/instrumentation-java/blob/19060b8e067ec191a9b177ac35c60f3dc18c1262/core/src/main/java/com/google/instrumentation/common/Provider.java).

Since that is complicated and can be added later, I'm opting to punt until demand warrants it.

adriancole:
fyi here's a working example of this in use: https://github.com/openzipkin/brave/pull/350/files#diff-2ec9bf3b5f2be4e730bcc3e3ccfcac5f

PR #381, opened 2017-04-20
Adds brave.Tracing an object graph that holds things you need

adriancole:
Before, we added some helpers like `Tracer.clock()` because when
instrumenting, you sometimes need more than just the Tracer. Recently,
we needed to decide how to expose the propagation format.. should we add
another convenience method?

Taking experience from Zipkin (which took its experience from Dagger and
Spring), another way is to expose an object graph.. a component that
provides types often used together.

This adds `brave.Tracing` which can be passed around or implemented
differently as needed. Instrumentation would consume this to get common
types they need. For example, `JaxRs2Tracing` would accept an instance
of tracing, so that it has what it needs, a tracer and a propagation
handler.

Instances of this can be wired up in Spring (Dagger, Guice or CDI),
making instrumentation wiring far simpler than chaining builders.

Thanks to @bogdandrutu and @llinder for discussions leading to this.

### DIFF
--- a/brave/src/main/java/brave/Tracing.java
+++ b/brave/src/main/java/brave/Tracing.java
@@ -1,0 +1,246 @@
+package brave;
+
+import brave.internal.Internal;
+import brave.internal.Nullable;
+import brave.internal.Platform;
+import brave.propagation.CurrentTraceContext;
+import brave.propagation.Propagation;
+import brave.propagation.TraceContext;
+import brave.sampler.Sampler;
+import java.io.Closeable;
+import zipkin.Endpoint;
+import zipkin.reporter.AsyncReporter;
+import zipkin.reporter.Reporter;
+import zipkin.reporter.Sender;
+
+/**
+ * This provides utilities needed for trace instrumentation. For example, a {@link Tracer}.
+ *
+ * <p>Instances built via {@link #newBuilder()} are registered automatically such that statically
+ * configured instrumentation like JDBC drivers can use {@link #current()}.
+ *
+ * <p>This type can be extended so that the object graph can be built differently or overridden, for
+ * example via spring or when mocking.
+ */
+public abstract class Tracing implements Closeable {
+
+  public static Builder newBuilder() {
+    return new Builder();
+  }
+
+  /** All tracing commands start with a {@link Span}. Use a tracer to create spans. */
+  abstract public Tracer tracer();
+
+  /**
+   * When a trace leaves the process, it needs to be propagated, usually via headers. This utility
+   * is used to inject or extract a trace context from remote requests.
+   */
+  // Implementations should override and cache this as a field.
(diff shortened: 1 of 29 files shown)


adriancole:
cc @devinsba

brave-core/src/main/java/com/github/kristofa/brave/TracerAdapter.java
```diff
@@ -1,6 +1,7 @@
 package com.github.kristofa.brave;
 
 import brave.Tracer;
+import brave.Tracing;
```
devinsba: I'm guessing you backed out some changes in this file. Probably don't need the import

brave/src/main/java/brave/Tracing.java
```diff
@@ -0,0 +1,215 @@
...
+
+  /** Ensures this component can be garbage collected, by making it not {@link #current()} */
+  @Override abstract public void close();
+
+  public static final class Builder {
+    String localServiceName;
+    Endpoint localEndpoint;
+    Reporter<zipkin.Span> reporter;
+    Clock clock;
+    Sampler sampler = Sampler.ALWAYS_SAMPLE;
```
devinsba: I'm not sure about defaulting to always, though I've started doing this in our projects so :man_shrugging:

brave/src/main/java/brave/Tracing.java
```diff
@@ -0,0 +1,215 @@
...
+      maybeSetCurrent();
+    }
+
+    @Override public Tracer tracer() {
+      return tracer;
+    }
+
+    @Override public Propagation.Factory propagationFactory() {
+      return propagationFactory;
+    }
```
devinsba: Is this new too? I don't remember this from before

brave/src/main/java/brave/Tracing.java
adriancole: I think we have done this mainly a preference for first starters.. I think finagle defaults to 0.1, which is good but first starters tend to feel things are broke (fwiw this isn't a change I know it is hard to see that)

brave/src/main/java/brave/Tracing.java
adriancole: this is the only really new thing. everything else was pushing things from one place to another. should have called that out actually :)

brave-core/src/main/java/com/github/kristofa/brave/TracerAdapter.java
adriancole: good catch

adriancole:
will merge as soon as I rebase the other PR over top of it (unless I hear screaming)

adriancole:
added `Tracing.currentTracer()` as a shortcut to checking null `Tracing.current().tracer()` as especially mysql does not need any other components like propagation.

PR #384, opened 2017-04-23
Adds Tracer.nextSpan(Extractor, Carrier) to simplify instrumentation

adriancole:


adriancole:
This was noticed by @reta and I'd like an opinion from @llinder @devinsba and/or @felixbarny (though all are welcome)

here's an example of this in use https://github.com/openzipkin/brave/pull/350/commits/7a6f36cf2425993c3462c80c671690e841d08513

brave/src/main/java/brave/Tracer.java
```diff
@@ -261,6 +263,17 @@ public SpanInScope withSpanInScope(Span span) {
     return currentContext != null ? toSpan(currentContext) : null;
   }
 
+  /**
+   * Conditionally joins a span, or starts a new trace, depending on if a trace context was
+   * extracted from the carrier (usually an incoming request).
+   */
+  @Nullable public <C> Span nextSpan(Extractor<C> extractor, C carrier) {
```
adriancole: using nextSpan though not sure if that's an awesome name. can't think of a better one.

brave/src/main/java/brave/Tracer.java
```diff
@@ -261,6 +263,17 @@ public SpanInScope withSpanInScope(Span span) {
     return currentContext != null ? toSpan(currentContext) : null;
   }
 
+  /**
+   * Conditionally joins a span, or starts a new trace, depending on if a trace context was
+   * extracted from the carrier (usually an incoming request).
+   */
+  public <C> Span nextSpan(Extractor<C> extractor, C carrier) {
```
adriancole: we could also consider naming extractSpan?

llinder: Nice simplification!  I'm on the fence about the name though.  `nextSpan` still feels better since its creating a span from the extracted context.  `extractSpan` feels like its materializing an existing span rather than creating a new one.

adriancole:
other thoughts.. another key here is that in the future we might fork a
child (ex if not doing client+server spans, rather single-host ones, we'd
likely implement that policy inside the same function)

resumeTrace - because we are attempting to resume a trace that started
somewhere else
nextSpanFromIncoming - writing this eventhough I think the name is horrible
spanFromRequest - not always a request, could be a message, but mostly true

leaning towards nextSpan still..

adriancole:
merging, though we can change this anytime between now and next release

PR #389, opened 2017-05-02
Adds MDCCurrentTraceContext for log correlation

adriancole:
This adds trace and span IDs to the SLF4J Mapped Diagnostic Context (MDC)
so that you can search or aggregate logs accordingly.

To enable this, configure `brave.Tracing` with `MDCCurrentTraceContext`
like so:

```java
tracing = Tracing.newBuilder()
    .currentTraceContext(MDCCurrentTraceContext.create())
    ...
    .build();
```

Then, in your log configuration, you can use `traceId` and or `spanId`.

Here's an example logback configuration:

```xml
<pattern>%d [%X{traceId}/%X{spanId}] [%thread] %-5level %logger{36} - %msg%n</pattern>
```

When a trace is in progress, it would log statements like this:
```
2017-05-02 23:36:04,789 [fcd015bf6f8b05ba/fcd015bf6f8b05ba] [main] INFO  c.a.FooController - I got here!
```

Users could then copy/paste the trace ID into the zipkin UI, or use log
correlation to further debug a problem.

Fixes #150
See #369

PR #392, opened 2017-05-06
Adds Log4J 2 ThreadContext integration

adriancole:
This adds trace and span IDs to the Log4J 2 Thread Context so that you
can search or aggregate logs accordingly.

To enable this, configure `brave.Tracing` with `ThreadContextCurrentTraceContext`
like so:

```java
    .currentTraceContext(ThreadContextCurrentTraceContext.create())
    ...
    .build();
```

See #389

### DIFF
--- a/context/log4j2/README.md
+++ b/context/log4j2/README.md
@@ -1,0 +1,29 @@
+# brave-context-log4j2
+This adds trace and span IDs to the Log4J 2 Thread Context so that you
+can search or aggregate logs accordingly.
+
+To enable this, configure `brave.Tracing` with `ThreadContextCurrentTraceContext`
+like so:
+
+```java
+tracing = Tracing.newBuilder()
+    .currentTraceContext(ThreadContextCurrentTraceContext.create())
+    ...
+    .build();
+```
+
+Then, in your log configuration, you can use `traceId` and or `spanId`.
+
+Here's an example log4j2.properties pattern:
+
+```xml
+appender.console.layout.pattern = %d{ABSOLUTE} [%X{traceId}/%X{spanId}] %-5p [%t] %C{2} (%F:%L) - %m%n
+```
+
+When a trace is in progress, it would log statements like this:
+```
+11:01:29,799 [e2ffceb485bdfb1d/e2ffceb485bdfb1d] INFO  [main] c.a.FooController (FooController.java:30) - I got here!
+```
+
+Users could then copy/paste the trace ID into the zipkin UI, or use log
+correlation to further debug a problem.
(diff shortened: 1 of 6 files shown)


adriancole:
cc @kuroneko25

adriancole:
merging so I can use this in our example app. won't release for a couple days in case there's feedback

kuroneko25:
Do you have an example of how to use this with gRPC? I'm using `brave-grpc` specifically.

adriancole:
>
> Do you have an example of how to use this with gRPC? I'm using brave-grpc
> specifically.
>
example of using log4j2 with grpc? or example of having brave use grpc's
Context?

kuroneko25:
Sorry I didn't realize there was a big refactoring recently to grpc instrumentation and I've switched to using 
```
    <dependency>
      <artifactId>brave-instrumentation-grpc</artifactId>
      <version>4.3.0</version>
    </dependency>
```

Although it doesn't seem like there is anything in `ThreadContext` when I tried to access it from within an API implementation. Here is how I set it up:

```
...
      GrpcTracing tracing = GrpcTracing.create(
          Tracing.newBuilder().localServiceName(serviceName)
              .reporter(Reporter.NOOP)
              .sampler(Sampler.ALWAYS_SAMPLE)
              .currentTraceContext(ThreadContextCurrentTraceContext.create())
              .build()
      serverBuilder.addService(ServerInterceptors.intercept(
          serviceDefinition,
          tracing.newServerInterceptor()
      );
     serverBuilder.build();
...
```

In the API implementation I'm doing something like this
```
...
  @Override
  public void sayHello(HelloRequest request,
      StreamObserver<HelloResponse> responseObserver) {
    System.out.println(ThreadContext.get("traceId"));
    responseObserver.onNext(responseForRequest(request));
    responseObserver.onCompleted();
  }
...
```

`ThreadContext.get("traceId")` is null in this case. I've stepped through the debugger and it looks like `ThreadContextCurrentTraceContext` is populating `TraceContext` with the correct trace and span. Maybe this is due to the internal threading model of gRPC? I noticed this at the end of the page https://logging.apache.org/log4j/2.x/manual/thread-context.html. Perhaps it's related?

```
With the ThreadContext logging statements can be tagged so log entries that were related in some way can be linked via these tags. The limitation is that this only works for logging done on the same application thread (or child threads when configured).

Some applications have a thread model that delegates work to other threads, and in such models, tagging attributes that are put into a thread-local map in one thread are not visible in the other threads and logging done in the other threads will not show these attributes.

Log4j 2.7 adds a flexible mechanism to tag logging statements with context data coming from other sources than the ThreadContext. See the manual page on extending Log4j for details.
```

adriancole:
can you bump a new issue specific to grpc? I can get on this.

kuroneko25:
#425

PR #393, opened 2017-05-07
Ensures current span is always closed prior to finishing

adriancole:
This changes logic in http handlers such that a span is always closed
prior to being finished. This has two advantages: one is that we can
test that instrumentation do not leak span contexts. Another is that
log context synchronization will not add trace IDs of a closed span.


adriancole:
note I tried to pay for this change by removing duplicate logic in tests

brave/src/main/java/brave/http/HttpClientHandler.java
```diff
@@ -12,35 +13,25 @@ public HttpClientHandler(HttpAdapter<Req, Resp> adapter, HttpClientParser parser
     this.parser = parser;
   }
 
-  public Req handleSend(Req request, Span span) {
-    if (span.isNoop()) return request;
+  public void handleSend(Req request, Span span) {
```
adriancole: these return values weren't used, or when used encouraged people to use a span in scope.

adriancole:
boatload of test backfills next

Issue #396, opened 2017-05-08
Mitigate impact of double-instrumentation


adriancole:
*Note* this issue is uncommon and can be solved in configuration. However, it may suggest we should make certain application layer instrumentation have the option to not create client or server spans (local instead).

Due to layering, it is possible to accidentally double-trace clients and servers.  For example, you could double-instrument netty and also servlet or servlet and also webmvc. Each of these do not expect to have a parent, so they blindly read headers ignoring any prior state. This results in two traces for the same incoming request. One for each time the trace is extracted.

On the other side, you could double-instrument httpclient and also resttemplate or okhttp and also jaxrs. In this case, 2 RPC spans would be created and the trace identifiers would be injected twice.

Here's an example of servlet + webmvc redundantly traced and apachehc and rest template redundantly traced.

![Uploading Screen Shot 2017-05-08 at 1.44.42 PM.png…]()
```json
[
  [
    {
      "traceId": "6e67065cc0eb64d3",
      "id": "6e67065cc0eb64d3",
      "name": "get",
      "timestamp": 1494221947722032,
      "duration": 884120,
      "annotations": [
        {
          "timestamp": 1494221947722032,
          "value": "sr",
          "endpoint": {
            "serviceName": "brave-webmvc-example",
            "ipv4": "192.168.1.10"
          }
        },
        {
          "timestamp": 1494221948606152,
          "value": "ss",
          "endpoint": {
            "serviceName": "brave-webmvc-example",
            "ipv4": "192.168.1.10"
          }
        }
      ],
      "binaryAnnotations": [
        {
          "key": "ca",
          "value": true,
          "endpoint": {
            "serviceName": "",
            "ipv4": "127.0.0.1",
            "port": 58308
          }
        },
        {
          "key": "http.path",
          "value": "/a",
          "endpoint": {
            "serviceName": "brave-webmvc-example",
            "ipv4": "192.168.1.10"
          }
        }
      ]
    },
    {
      "traceId": "6e67065cc0eb64d3",
      "id": "e28ad196c1ce889f",
      "name": "get",
      "parentId": "6e67065cc0eb64d3",
      "timestamp": 1494221947802502,
      "duration": 785297,
      "annotations": [
        {
          "timestamp": 1494221947802502,
          "value": "cs",
          "endpoint": {
            "serviceName": "brave-webmvc-example",
            "ipv4": "192.168.1.10"
          }
        },
        {
          "timestamp": 1494221948587799,
          "value": "cr",
[...]

Issue #395, opened 2017-05-07
Http Sampler


adriancole:
Now that we have a generic way to get the http path from any instrumented server, we should be able to configure a parameterized sampler.

For example, this could provide features similar to [amazon x-ray](http://docs.aws.amazon.com/xray/latest/devguide/xray-sdk-java-configuration.html#xray-sdk-java-configuration-sampling)

```json
{
  "version": 1,
  "rules": [
    {
      "description": "Player moves.",
      "service_name": "*",
      "http_method": "*",
      "url_path": "/api/move/*",
      "fixed_target": 0,
      "rate": 0.05
    }
  ],
  "default": {
    "fixed_target": 1,
    "rate": 0.1
  }
}
```

Implementation would be a sampler that is attached to the `brave.http.HttpTracing` component. It may need to be a factory in nature. We may have to revise `Tracer.nextSpan` to allow passing of a sampler.

Implementation of a path based sampler would use code like this:

```java
  final HttpAdapter<Req, ?> adapter;
  final HttpAdapter<Req, ?> req;

  public boolean boolean isSampled(long traceId) {
    String path = adapter.path(req);
    // make a decision
  }
```

I know we do custom sampling in zipkin itself, so having something like this would make that not custom code. I'm sure others would be interested in this, too, but worth asking.

adriancole:
incidentally, Tracer.nextSpan hasn't yet been released, so we can revise it as necessary.

adriancole:
Here's a sketch I'm pretty sure will work

```java
/**
 * Decides whether to start a new trace based on http request properties such as path.
 *
 * <p>Ex. Here's a sampler that only traces api requests
 * <pre>{@code
 * httpTracingBuilder.serverSampler((adapter, request) -> {
 *   if (adapter.path(req).startsWith("/api")) {
 *     return true;
 *   } else {
 *     return false;
 *   }
 * });
 * }</pre>
 */
interface HttpSampler {
  HttpSampler NULL = new HttpSampler() {
    @Override public <Req> Boolean trySample(HttpAdapter<Req, ?> adapter, Req request) {
      return null;
    }

    @Override public String toString() {
      return "NullHttpSampler";
    }
  };

  /** Returns an overriding sampling decision for a new trace or null to punt to default. */
  @Nullable <Req> Boolean trySample(HttpAdapter<Req, ?> adapter, Req request);
}
```

adriancole:
[...]

PR #397, opened 2017-05-08
Refines Http client and server handlers and documents

adriancole:
During integration of the new apis, I noticed some glitches around any
future in sampling (#395) and double-instrumentation (#396). The main
problem was the location of decisions around trace context injection
and extraction, as well when a span was created. The solution to this
problem is to centralize such logic, which has an advantage of making
instrumentation easier to write as well.

Here's how the interfaces now work:

The first step in developing http client instrumentation is implementing
a `HttpClientAdapter` for your native library. This ensures users can
portably control tags using `HttpClientParser`.

Next, you'll need to indicate how to insert trace IDs into the outgoing
request. Often, this is as simple as `Request::setHeader`.

With these two items, you now have the most important parts needed to
trace your server library. You'll likely initialize the following in a
constructor like so:
```java
MyTracingFilter(HttpTracing httpTracing) {
  tracer = httpTracing.tracing().tracer();
  handler = HttpClientHandler.create(httpTracing, new MyHttpClientAdapter());
  extractor = httpTracing.tracing().propagation().injector(Request::setHeader);
}
```

Synchronous interception is the most straight forward instrumentation.
You generally need to...
1. Start the span and add trace headers to the request
2. Put the span in scope so things like log integration works
3. Invoke the request
4. Catch any errors
5. Complete the span

```java
Span span = handler.handleSend(injector, request); // 1.
Throwable error = null;
try (Tracer.SpanInScope ws = tracer.withSpanInScope(span)) { // 2.
  response = invoke(request); // 3.
  error = e; // 4.
  throw e;
} finally {
  handler.handleReceive(response, error, span); // 5.
}
```

The first step in developing http server instrumentation is implementing
a `HttpServerAdapter` for your native library. This ensures users can
portably control tags using `HttpServerParser`. See [HttpServletAdapter](./servlet/src/main/java/brave/servlet/HttpServletAdapter.java)
as an example (you may even be able to use it!).

Next, you'll need to indicate how to extract trace IDs from the incoming
request. Often, this is as simple as `Request::getHeader`.

With these two items, you now have the most important parts needed to
trace your server library. You'll likely initialize the following in a
constructor like so:
```java
MyTracingInterceptor(HttpTracing httpTracing) {
  tracer = httpTracing.tracing().tracer();
  handler = HttpServerHandler.create(httpTracing, new MyHttpServerAdapter());
  extractor = httpTracing.tracing().propagation().extractor(Request::getHeader);
}
```

Synchronous interception is the most straight forward instrumentation.
You generally need to...
1. Extract any trace IDs from headers and start the span
2. Put the span in scope so things like log integration works
3. Invoke the request
4. Catch any errors
5. Complete the span

```java
Span span = handler.handleReceive(extractor, request); // 1.
Throwable error = null;
try (Tracer.SpanInScope ws = tracer.withSpanInScope(span)) { // 2.
  response = invoke(request); // 3.
  error = e; // 4.
  throw e;
} finally {
  handler.handleSend(response, error, span); // 5.
}
```


adriancole:
cc @llinder @devinsba @reta @jplock 

note: this is about how to write new instrumentation.. future work should be able to proceed without breaking apis. This could include making "application spans" when things like Jaxrs are layered over instrumented libraries like netty or apache hc  (#396). This could also include making a generic http-request based sampling strategy (#395)

adriancole:
merging so folks can test it with real things (we don't publish snapshots on pull requests)

http://oss.jfrog.org/artifactory/oss-snapshot-local is where they will be

https://github.com/openzipkin/brave-webmvc-example already uses snapshots at the moment

any feedback post-merge very welcome!

PR #398, opened 2017-05-09
Adds the ability to customize http errors, defaulting to status 400+

adriancole:
This adds HttpClientParser and HttpServerParser.error, which by default
adds an error tag when the status is >=400 or there's an exception.

This policy is similar to ratpack, but overridable.

### DIFF
--- a/brave/src/main/java/brave/http/HttpAdapter.java
+++ b/brave/src/main/java/brave/http/HttpAdapter.java
@@ -45,6 +45,17 @@ public abstract class HttpAdapter<Req, Resp> {
    */
   @Nullable public abstract Integer statusCode(Resp response);
 
+  String parseError(@Nullable Resp response, @Nullable Throwable error) {
+    if (error != null) {
+      String message = error.getMessage();
+      return message != null ? message : error.getClass().getSimpleName();
+    }
+    Integer httpStatus = statusCode(response);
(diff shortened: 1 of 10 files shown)


adriancole:
cc @hyleung

hyleung:
@adriancole thx! I think we've got those cases covered on the Ratpack side, too - except looks like we do add an extra `TraceKeys.HTTP_STATUS_CODE` tag, and the error tag has slightly different data in it...

https://github.com/hyleung/ratpack-zipkin/blob/master/src/main/java/ratpack/zipkin/internal/DefaultServerTracingHandler.java#L48-L62

I may have asked this before, but is there something I can look at that describes the semantics around the various tags - often I'm not sure when to use which tags and what they are meant to represent.

adriancole:
hi there.

yeah there is a description, but we don't have something to render into
markdown
https://github.com/openzipkin/zipkin-api/blob/master/thrift/zipkinCore.thrift

ps I don't know how to finish this, but here's a spike of converting to
code on this pull request
https://github.com/hyleung/ratpack-zipkin/pull/30

hyleung:
@adriancole thanks! I'll take a look at that PR - hoping to get back to that stuff (now that I can get a snapshot to work against), I can take some of that stuff and incorporate it.

adriancole:
@hyleung I just updated this. I think this policy is probably fine for ratpack, but you can also change it now.

PR #400, opened 2017-05-09
Proves autowiring works on spring interceptors

adriancole:
Cleans up auto-wired configuration and shows it works


adriancole:
thx @sirtyro for noticing cc @devinsba

PR #401, opened 2017-05-09
Ensures legacy apps written in Spring 3 can use new instrumentation

adriancole:
This reverts the default spring version to 3 to make sure we don't
accidentally use new apis.


adriancole:
@sirtyro @devinsba fyi soon I will update https://github.com/openzipkin/brave-webmvc-example to prove deployment works with spring 3+ XML+servlet 2.5, spring 4 + servlet3, and finally spring boot.

I think this should be the most sincere way of showing all work.

PR #403, opened 2017-05-09
WIP archive old instrumentation

adriancole:
This moves the old instrumentation to a different directory, so as to
not confuse new users. They will still be built and published at least
until Brave 5, maybe longer. This also adds a new README


adriancole:
moved general tracking to https://github.com/openzipkin/brave/issues/410

PR #408, opened 2017-05-10
Adds HttpSampler for parameterized sampling of http requests

### LINKED ISSUE #395: Http Sampler

adriancole:
The prior sampling policy was to use the default (trace ID) sampler for
server and client requests.  For example, if there's a incoming request
that has no trace IDs in its headers, the sampler indicated by
`Tracing.Builder.sampler` decides whether or not to start a new trace.
Once a trace is in progress, it is used for any outgoing http client
requests.

On the other hand, you may have http client requests that didn't originate
from a server. For example, you may be bootstrapping your application,
and that makes an http call to a system service. The default policy will
start a trace for any http call, even ones that didn't come from a server
request.

This change makes it possible to make sampling decisions based on the
http request, in a way that applies to all http instrumentation uniformly.

Here's an example which only starts traces for requests
originating from the  `/api` endpoint. It chooses 100% of such requests.

```java
    .serverSampler(new HttpSampler() {
       @Override public <Req> Boolean trySample(HttpAdapter<Req, ?> adapter, Req request) {
         return adapter.path(request).startsWith("/api");
       }
     })
    .clientSampler(HttpSampler.NEVER_SAMPLE)
    .build();
```

Fixes #395

### DIFF
--- a/instrumentation/README.md
+++ b/instrumentation/README.md
@@ -22,14 +22,15 @@ httpTracing = HttpTracing.newBuilder(tracing).serverName("github").build();
 okhttp = TracingCallFactory.create(httpTracing, new OkHttpClient());
 ```
 
-### Http tracing
+# Http tracing
 Most instrumentation are based on http communication. For this reason,
 we have specialized handlers for http clients and servers. All of these
 are configured with `HttpTracing`.
 
-The `HttpTracing` class holds a reference to a tracing component and also
-includes instructions on what to put into http spans.
+The `HttpTracing` class holds a reference to a tracing component,
+instructions on what to put into http spans, and sampling policy.
 
+## Tagging policy
 By default, the following is added for both http clients and servers:
 * Span.name as the http method in lowercase: ex "get"
 * Tags/binary annotations:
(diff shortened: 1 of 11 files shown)


adriancole:
cc @llinder @sirtyro @garyd203 @llinder @hyleung @takezoe

adriancole:
note I just threw up the api. I will implement it today

adriancole:
added implementation, tests and also a client+server example

instrumentation/README.md
```diff
@@ -63,6 +64,36 @@ apache = TracingHttpClientBuilder.create(httpTracing.clientOf("s3")).build();
...
+```java
+httpTracing = httpTracing.toBuilder()
+    .serverSampler(new HttpSampler() {
+       @Override public <Req> Boolean trySample(HttpAdapter<Req, ?> adapter, Req request) {
+         return adapter.path(request).startsWith("/api");
+       }
+     })
+    .clientSampler(HttpSampler.NEVER_SAMPLE)
+    .build();
+```
```
devinsba: An excluded paths example might be good here as well

adriancole: sure I'll change the readme

adriancole: done

llinder:
Late to the party but looks great.  Can't wait to start using this!

PR #413, opened 2017-05-11
Implements letClear semantics via setting null in scope

adriancole:
In edge cases, you may need to clear the current span temporarily. For
example, launching a task that should not be associated with the current
request. To do this, simply pass null to `withSpanInScope`.

```java
try (SpanInScope cleared = tracer.withSpanInScope(null)) {
  startBackgroundThread();
}
```

### DIFF
--- a/brave/README.md
+++ b/brave/README.md
@@ -293,6 +293,16 @@ try (SpanInScope ws = tracer.withSpanInScope(span)) {
 }
 ```
 
+In edge cases, you may need to clear the current span temporarily. For
+example, launching a task that should not be associated with the current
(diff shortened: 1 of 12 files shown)


adriancole:
Besides launching tasks which should be in new traces...

One use case for this is zipkin self-tracing. We don't want to amplify writes by creating a tracing loop. By clearing the span we can do logic based on whether Tracer.currentSpan returns null or not.

Another is if you make a logging reporter. If you write a logging reporter, you probably don't want to add the trace ID into your MDC as it would be confusing (since the log message itself is a span).

PR #414, opened 2017-05-12
Adds Cassandra client and server integration

adriancole:
This contains tracing instrumentation for [Cassandra](https://github.com/apache/cassandra/blob/trunk/src/java/org/apache/cassandra/tracing/Tracing.java) and the [DataStax Java Driver](https://github.com/datastax/java-driver).

`brave.cassandra.Tracing` extracts trace state from the custom payload
of incoming requests. How long each request takes, each suboperation,
and relevant tags like the session ID are reported to Zipkin.

`brave.cassandra.driver.TracingSession` tracks the client-side of cassandra and
adds trace context to the custom payload of outgoing requests. If
server integration is in place, cassandra will contribute data to these 
RPC spans.

### DIFF
--- a/instrumentation/cassandra-driver/src/main/java/brave/cassandra/driver/TracingSession.java
+++ b/instrumentation/cassandra-driver/src/main/java/brave/cassandra/driver/TracingSession.java
@@ -30,22 +30,22 @@ import static zipkin.internal.Util.checkNotNull;
 
 public final class TracingSession extends AbstractSession {
   public static Session create(Tracing tracing, Session delegate) {
-    return new TracingSession(CassandraDriverTracing.create(tracing), delegate);
+    return new TracingSession(CassandraClientTracing.create(tracing), delegate);
   }
 
-  public static Session create(CassandraDriverTracing cassandraTracing, Session delegate) {
+  public static Session create(CassandraClientTracing cassandraTracing, Session delegate) {
     return new TracingSession(cassandraTracing, delegate);
   }
 
   final Tracer tracer;
-  final CassandraDriverSampler sampler;
-  final CassandraDriverParser parser;
+  final CassandraClientSampler sampler;
+  final CassandraClientParser parser;
   final String remoteServiceName;
   final TraceContext.Injector<Map<String, ByteBuffer>> injector;
   final ProtocolVersion version;
   final Session delegate;
 
-  TracingSession(CassandraDriverTracing cassandraTracing, Session target) {
+  TracingSession(CassandraClientTracing cassandraTracing, Session target) {
     checkNotNull(cassandraTracing, "cassandraTracing");
     this.delegate = checkNotNull(target, "delegate");
     tracer = cassandraTracing.tracing().tracer();
--- a/instrumentation/cassandra-driver/src/main/java/brave/cassandra/driver/TracingSession.java
+++ b/instrumentation/cassandra-driver/src/main/java/brave/cassandra/driver/TracingSession.java
@@ -1,0 +1,167 @@
+package brave.cassandra.driver;
+
+import brave.Span;
+import brave.Tracer;
+import brave.Tracing;
+import brave.propagation.SamplingFlags;
+import brave.propagation.TraceContext;
+import com.datastax.driver.core.AbstractSession;
+import com.datastax.driver.core.CloseFuture;
+import com.datastax.driver.core.Cluster;
+import com.datastax.driver.core.PreparedStatement;
+import com.datastax.driver.core.ProtocolVersion;
+import com.datastax.driver.core.RegularStatement;
+import com.datastax.driver.core.ResultSet;
+import com.datastax.driver.core.ResultSetFuture;
+import com.datastax.driver.core.Session;
+import com.datastax.driver.core.SimpleStatement;
+import com.datastax.driver.core.Statement;
+import com.google.common.util.concurrent.FutureCallback;
+import com.google.common.util.concurrent.Futures;
+import com.google.common.util.concurrent.ListenableFuture;
+import java.net.InetSocketAddress;
+import java.nio.ByteBuffer;
+import java.util.LinkedHashMap;
+import java.util.Map;
+import zipkin.Endpoint;
+
+import static brave.Span.Kind.CLIENT;
+import static zipkin.internal.Util.checkNotNull;
+
+public final class TracingSession extends AbstractSession {
+  public static Session create(Tracing tracing, Session delegate) {
+    return new TracingSession(CassandraDriverTracing.create(tracing), delegate);
+  }
+
+  public static Session create(CassandraDriverTracing cassandraTracing, Session delegate) {
+    return new TracingSession(cassandraTracing, delegate);
+  }
+
+  final Tracer tracer;
+  final CassandraDriverSampler sampler;
+  final CassandraDriverParser parser;
+  final String remoteServiceName;
+  final TraceContext.Injector<Map<String, ByteBuffer>> injector;
+  final ProtocolVersion version;
+  final Session delegate;
+
+  TracingSession(CassandraDriverTracing cassandraTracing, Session target) {
+    checkNotNull(cassandraTracing, "cassandraTracing");
+    this.delegate = checkNotNull(target, "delegate");
+    tracer = cassandraTracing.tracing().tracer();
+    sampler = cassandraTracing.sampler();
+    parser = cassandraTracing.parser();
+    String remoteServiceName = cassandraTracing.remoteServiceName();
+    this.remoteServiceName = remoteServiceName != null
+        ? remoteServiceName
+        : target.getCluster().getClusterName();
+    injector = cassandraTracing.tracing().propagation().injector((carrier, key, v) -> {
+      int length = v.length(); // all values are ascii
+      byte[] buf = new byte[length];
+      for (int i = 0; i < length; i++) {
+        buf[i] = (byte) v.charAt(i);
+      }
+      carrier.put(key, ByteBuffer.wrap(buf));
+    });
+    version = delegate.getCluster().getConfiguration().getProtocolOptions().getProtocolVersion();
+  }
+
+  @Override public ResultSetFuture executeAsync(Statement statement) {
+    Span span = nextSpan(statement);
+    if (!span.isNoop()) parser.request(statement, span.kind(CLIENT));
+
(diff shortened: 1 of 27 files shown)


instrumentation/cassandra/README.md
```diff
@@ -0,0 +1,52 @@
...
+Clients must enable tracing. If they want to continue an existing trace,
+they must add propagation fields to the custom payload of a statement.
+
+```java
+// minimally, you need to prepare a statement and enable tracing
+preparedStatement.enableTracing();
+
+// By default, B3 style is used, so instrumented clients do something like this
+Map<String, ByteBuffer> payload = new LinkedHashMap<>();
+payload.set("X-B3-TraceId", byteBuffer("463ac35c9f6413ad"));
```
adriancole: note that brave allows you to change the propagation format. I intentionally decided to not go with the finagle binary encoding as most people don't understand what that is. B3 is reasonably well documented, so a better default IMHO.

adriancole:
PS I will raise another pull request that depends on this for datastax client shortly (harvesting code from zipkin and some buffing)

adriancole:
Ok added the client side to this pull request. I have some test cases to fix, and need to put in basic tagging policy. Should be ready in the next day or two.

adriancole:
I got all the normal tests working. So, folks can test this (I'll also change zipkin to use it).

Next steps are to put in some basic tag policy logic, and add edge-case tests we don't usually cover, like write timeouts.

adriancole:
added tag policy hooks to the client side (the only code zipkin needs at the moment)

still TODO: edge case tests and server-side policy

adriancole:
buffed the `CassandraDriverParser`, so it looks prettier and accepts Tagger not Span

adriancole:
I'm going to rename CassandraDriverX to CassandraClientX in a separate commit. @llinder etc let me know if you disagree with this choice.

adriancole:
OK I think this is good for a first cut. I'll integrate in zipkin and if it works have no more changes planned for the initial impl. We can always add more functionality later (and likely will need to)

adriancole:
updated to use the SpanCustomizer type

adriancole:
I'm not going to merge this until some folks savvy with cassandra have a chance to play around with it.

adriancole:
ps considering the scope of this includes 3 modules, it might make more sense to bump this to a separate repository brave-cassandra. Any thoughts?

devinsba:
:+1: to new repo if there will be a bunch of churn on that. Not sure what the cadence is for the cassandra project

llinder:
Separate repository sounds good to me.

llinder:
Looks like a solid start.  I will give this a go in one of our services in pre-production as soon as I can.  Just want to finish up the Ratpack work first.

instrumentation/cassandra-driver/src/main/java/brave/cassandra/driver/TracingSession.java
```diff
@@ -0,0 +1,167 @@
...
+    });
+    version = delegate.getCluster().getConfiguration().getProtocolOptions().getProtocolVersion();
+  }
+
+  @Override public ResultSetFuture executeAsync(Statement statement) {
+    Span span = nextSpan(statement);
+    if (!span.isNoop()) parser.request(statement, span.kind(CLIENT));
+
+    // o.a.c.tracing.Tracing.newSession must use the same propagation format
+    if (version.compareTo(ProtocolVersion.V4) >= 0) {
```
michaelsembwever: does it break the request if outgoingPayload is added to it against a server that doesn't support it (ie < V4) ?

instrumentation/cassandra-driver/src/main/java/brave/cassandra/driver/TracingSession.java
```diff
@@ -0,0 +1,167 @@
...
+    ResultSetFuture result;
+    try {
+      result = delegate.executeAsync(statement);
+    } catch (RuntimeException | Error e) {
+      if (span.isNoop()) throw e;
+      parser.error(e, span);
+      span.finish();
+      throw e;
+    }
+    Futures.addCallback(result, new FutureCallback<ResultSet>() {
```
michaelsembwever: have we any idea on the impact adding all the callbacks will cause.
EG in a client that processes 10k+ requests per second this could cause trouble, and a suitable load-shedding mechanism be appropriate.

instrumentation/cassandra/src/main/java/brave/cassandra/CassandraTraceKeys.java
```diff
@@ -0,0 +1,11 @@
+package brave.cassandra;
+
+public final class CassandraTraceKeys {
```
michaelsembwever: it's interesting that you use `final class` here, as appropriate, yet not on the other class.
i would not presume that the other classes are designed for being subclassed (which are in fact broadening the API being published beyond what you're probably intending it to be.)

adriancole:
>
> +public final class CassandraTraceKeys {
>
> it's interesting that you use final class here, as appropriate, yet not
> on the other class.
> i would not presume that the other classes are designed for being
> subclassed (which are in fact broadening the API being published beyond
> what you're probably intending it to be.)
>
This is just a habit for constant files, i wouldn't read too much into
that. If someone were subclassing, they could I guess make a subclass of
TraceKeys.. wasn't thinking about that.

The other class is subclassable mainly as it seems very likely to need it.
I mentioned some notes in the type doc about configuration maybe being
something not awesome with system properties. Another thing is how
subordinate work is recorded.. right now, for simplicity and also to keep
traces from getting too "bottom heavy", I put annotations instead of child
spans. Subclassing would allow someone to override that, and since it is
likely they'll need to override the construction of the type anyway (bc it
is literally new'ed) seemed to fit.

I suspect this code will flex a bit before final, but there were my
thoughts going into it..

instrumentation/cassandra-driver/src/main/java/brave/cassandra/driver/TracingSession.java
adriancole: heh actually looking at the code, this is a bit silly.. I am checking span.isNoop for each callback. Will buff this out and move the conditional in front. I don't think we'd want to sample 10K spans/second, so this should take care of that concern.

Anyway, you highlight something which is that we don't yet have benchmarks on this for normal vs instrumented case.

instrumentation/cassandra-driver/src/main/java/brave/cassandra/driver/TracingSession.java
adriancole: you'd know better than me :) To test this though, I suppose we could add an invoker test that uses an older version of cassandra..

adriancole:
Tracing in Cassandra is actively maintained, ex there are some notes about
things that work in 3.11. I suspect it will actually be the testing and
focus concerns that make this more sensible as a separate repo beyond just
the several modules reason.

For example, we will likely need a matrix build at some point albeit via
maven invoker or docker. This is easier as a separate repo.

On focus, there is a lot of activity in brave unrelated to cassandra.
Another repo could keep the spam level down a bit.

instrumentation/cassandra/src/main/java/brave/cassandra/Tracing.java
```diff
@@ -0,0 +1,149 @@
...
+    return contextOrFlags.context() != null
+        ? tracer.joinSpan(contextOrFlags.context())
+        : tracer.newTrace(contextOrFlags.samplingFlags());
+  }
+
+  @Override protected final void stopSessionImpl() {
+    ZipkinTraceState state = (ZipkinTraceState) get();
+    if (state != null) state.incoming.finish();
+  }
+
```
michaelsembwever: how come you didn't also override `public void doneWithLocalSession(TraceState)` ?

adriancole: possibly an oversight?

instrumentation/cassandra/src/main/java/brave/cassandra/Tracing.java
```diff
@@ -0,0 +1,149 @@
...
+  protected void parseRequest(TraceState state, String request, Map<String, String> parameters,
+      SpanCustomizer customizer) {
+    customizer.name(parseSpanName(state, request));
+    customizer.tag(CassandraTraceKeys.CASSANDRA_REQUEST, request);
+    customizer.tag(CassandraTraceKeys.CASSANDRA_SESSION_ID, state.sessionId.toString());
+  }
+
+  @Override protected final TraceState newTraceState(InetAddress coordinator, UUID sessionId,
+      TraceType traceType) {
+    throw new AssertionError();
```
michaelsembwever: why is this an AssertionError?

adriancole:
I've moved all this code here for future updates. feel free to adjust anything! https://github.com/openzipkin/brave-cassandra

instrumentation/cassandra/src/main/java/brave/cassandra/Tracing.java
michaelsembwever: Or `public TraceState initializeFromMessage(final MessageIn<?> message)` ?
Or `public Map<String, byte[]> getTraceHeaders()`

(by not implementing outbound tracing, you won't be getting the tracing across the nodes *within* the cluster)

instrumentation/cassandra/src/main/java/brave/cassandra/Tracing.java
adriancole: because this method shouldn't be called because we override what calls it. That said, if the upstream code drifts that could cease to be the case.

instrumentation/cassandra/src/main/java/brave/cassandra/Tracing.java
adriancole: honestly I only implemented barebones to get things off the ground. Notably there's a lot of moving parts just in basics, so I was hoping to get something together that people can contribute to to add more features.

adriancole:
added https://github.com/openzipkin/brave-cassandra/issues/1 to not lose
this point

adriancole:
ok! the brave-cassandra repo is in business. For example, you can grab the snapshot cassandra jar here http://oss.jfrog.org/oss-snapshot-local/io/zipkin/brave/cassandra/brave-instrumentation-cassandra/0.0.1-SNAPSHOT/

Thanks for the review folks, and please direct all future things here, por favor! https://github.com/openzipkin/brave-cassandra

PR #416, opened 2017-05-13
Adds brave.Tagger, a safer type to expose for simple customization

adriancole:
Before, we passed `brave.Span` to callbacks that only add tags. This has
a number of problems, including that it is harder to test and tempts
users to interfere with the Span (like calling finish).

This changes our callbacks to use a simpler type `brave.Tagger`, which
existed in an early version of Brave 4, and recommends exposing this
instead of `brave.Span` for user callbacks like tag customization.

### DIFF
--- a/brave/src/main/java/brave/http/HttpParser.java
+++ b/brave/src/main/java/brave/http/HttpParser.java
@@ -1,6 +1,6 @@
 package brave.http;
 
-import brave.Span;
+import brave.Tagger;
 import brave.internal.Nullable;
 import zipkin.Constants;
 import zipkin.TraceKeys;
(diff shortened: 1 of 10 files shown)


adriancole:
cc @llinder @devinsba @reta I think this is a pretty important thing to lock-down before our "formal" release of Brave 4.3

brave/src/main/java/brave/http/HttpParser.java
```diff
@@ -1,6 +1,6 @@
 package brave.http;
 
-import brave.Span;
+import brave.Tagger;
```
adriancole: notice.. now the http instrumentation customizers no longer reference span at all.

instrumentation/jaxrs2/src/main/java/brave/jaxrs2/TracingContainerFilter.java
```diff
@@ -69,11 +67,6 @@ public void filter(ContainerRequestContext request, ContainerResponseContext res
     } else {
       return; // unknown state
     }
-
-    Response.StatusType statusInfo = response.getStatusInfo();
```
adriancole: this was redundant code I noticed

devinsba:
Sorry I didn't get to this in time, I do have an API level comment though. I think it would probably be nice to have the ability to add annotations (logs) also in addition to tags/BinaryAnnotations

adriancole:
>
> Sorry I didn't get to this in time, I do have an API level comment though.
> I think it would probably be nice to have the ability to add annotations
> (logs) also in addition to tags/BinaryAnnotations
>
This interface is passed for callbacks to add tags, though. For example
request and response tags which are done at the same time as "cs" or "sr"
are sent. It would be confusing to have a method called requestTags and add
annotations during that, right?

I'd expect that we'd want a different interface for general purpose
callbacks, if you were thinking about what to use for people setting things
in scope that weren't normal http parsing.

like SpanCustomizer which extends Tagger, allows annotations and maybe even
exposes the trace context?

adriancole:
On the other hand, i am all for simplifying things.. eventhough these
particular hooks say tags on them, we can rename them. If we add .name to
whatever we call the span customizer, we can remove a span naming hook and
just call it customizeRequest(..., req, customizer)

So,
abstract class SpanCustomizer {
  abstract Customizer name(String name);

  abstract Customizer annotate(String value);
  abstract Customizer tag(String key, String value);
}

While parsing has no use case for annotations, this could be safer to
expose to general purpose callbacks like user interceptors even via request
attributes. Or, you could make an injection handler for it.

This cant be implemented as a lambda but whatev :)

@devinsba wdyt?

adriancole:
Actually this is awesome because we can do things like this..

new SpanCustomizerLimiter()
  .nameLength(255)
  .tagValueLength(255)
  .limit(userProvidedCustomizer);

Very nice separation of concerns and this would also prevent things like I
did like accidentally add mysql tag that was 15MiB

adriancole:
https://github.com/openzipkin/brave/pull/417

PR #417, opened 2017-05-14
Replaces Tagger with SpanCustomizer for user callbacks

adriancole:
Before, we introduced `brave.Tagger` to limit the scope when adding tags
to spans. This widens the scope a bit to handle all routine changes made
by users to in-flight spans:

* setting the span name
* adding tags
* adding annotations

This type is called `SpanCustomizer`, and only exposes the above rather
than lifecycle hooks such as `start()`. This has the following benefits:

* Users do not need to learn the entire span type, or be tempted to finish one!
* Users can prevent accidental huge annotations or tags by decorating the `SpanCustomizer`

Thanks to @devinsba for the idea. Now's the best time to change this as
we haven't released any of the affected code formally.


adriancole:
@reta I think this is something of interest to you, too (in CXF)

brave/src/main/java/brave/SpanCustomizer.java
```diff
@@ -0,0 +1,40 @@
...
+import zipkin.TraceKeys;
+
+/**
+ * Simple interface users can customize a span with. For example, this can add custom tags useful
+ * in looking up spans.
+ *
+ * <p>This type is safer to expose directly to users than {@link Span}, as it has no hooks that
+ * can affect the span lifecycle.
+ */
+public abstract class SpanCustomizer {
```
reta: Apologies if there are some conventions behind, but going with simple interface seems like more lightweight solution, what do you think?

```
public interface SpanCustomizer {
}
```

reta:
@adriancole Definitely, thanks for heads up. Personally, I like it more than previously introduced `Tagger`, more options are exposed through the API, thanks!

brave/src/main/java/brave/SpanCustomizer.java
reta: The `Customizer` part of the name looks a little bit off here (as for me). Functionally wise, we enrich / decorate `Span` with annotations or tags. May be `SpanDecorator` or `SpanEnricher`? Just a suggestion ...

adriancole:
>
> Apologies if there are some conventions behind, but going with simple interface seems like more lightweight solution, what do you think?
>
> problem with interface is that this is exposed to users, so we can never
add methods to it without breaking compilation. I'm ok with that, just
letting you know.

reta:
@adriancole The `default` methods won't fit?

adriancole:
>
> @adriancole <https://github.com/adriancole> The default methods won't fit?
>
not for pre java 8 code (we still get requests for servlet 2.5!)

that said, I think we know the span model pretty well and can resist adding
methods :P

adriancole:
switched to interface and added a comment reminding us to not break it :)

incidentally most of our types like this are interfaces

adriancole:
hah whoops will fix the typo! I meant "until Java 8 is required"

Issue #425, opened 2017-05-16
Tracing information in log4j2 with gRPC


kuroneko25:
Short summary of what I want to accomplish:
We are building gRPC services and want to automatically include traceId and spanId in every log entry we emit from the service so we can correlate them. The logging library we use is Log4j2.

Brave libraries we use:
```
<dependency>
  <groupId>io.zipkin.brave</groupId>
  <artifactId>brave-instrumentation-grpc</artifactId>
  <version>4.3.0</version>
</dependency>
<dependency>
  <groupId>io.zipkin.brave</groupId>
  <artifactId>brave-context-log4j2</artifactId>
  <version>4.3.0</version>
</dependency>
```

Here is how I set up the gRPC server to use Brave:
```
...
GrpcTracing tracing = GrpcTracing.create(
    Tracing.newBuilder().localServiceName(serviceName)
        .reporter(Reporter.NOOP)
        .sampler(Sampler.ALWAYS_SAMPLE)
        .currentTraceContext(ThreadContextCurrentTraceContext.create())
        .build()
serverBuilder.addService(ServerInterceptors.intercept(
    serviceDefinition,
    tracing.newServerInterceptor()
);
serverBuilder.build();
...
```

We have a custom log4j layout class that spits out JSON. The relevant piece of code is something like this:
```
...
private JSONObject getTrace() {
  JSONObject json = new JSONObject();
  addValue(json, "traceId", ThreadContext.get(TRACE_ID));
  addValue(json, "spanId", ThreadContext.get(SPAN_ID));
  return json;
}
...
```

However `ThreadContext.get("traceId")` seems to be null always. I've stepped through the debugger and it looks like `ThreadContextCurrentTraceContext` is populating `TraceContext` with the correct trace and span in the server interceptor. Maybe this is due to the internal threading model of gRPC? Since the thread that runs the interceptor is different from the thread that runs the API code. They did mention something along these lines at the bottom of this page https://logging.apache.org/log4j/2.x/manual/thread-context.html. What should be the right way to achieve this? Should I implement my own `CurrentTraceContext` that puts tracing information into a gRPC context instead?

adriancole:
I raised a PR about this. If that doesn't work OOB, it might be a case where...

* we need to slightly redo how the log4j2 integration works like [this](https://logging.apache.org/log4j/2.x/manual/extending.html#Custom_ContextDataInjector)
or
* we are simply missing wrapping something in our grpc impl, and can take it from there

[...]

PR #426, opened 2017-05-16
Adds GrpcCurrentTraceContext

adriancole:
See #425

### DIFF
--- a/context/grpc/src/main/java/brave/context/grpc/GrpcCurrentTraceContext.java
+++ b/context/grpc/src/main/java/brave/context/grpc/GrpcCurrentTraceContext.java
@@ -1,0 +1,26 @@
+package brave.context.grpc;
+
+import brave.propagation.CurrentTraceContext;
+import brave.propagation.TraceContext;
+import io.grpc.Context;
+
+/** Manages scope with a {@link Context} as opposed to a thread local. */
+public final class GrpcCurrentTraceContext extends CurrentTraceContext {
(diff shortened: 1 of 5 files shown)


context/grpc/src/main/java/brave/context/grpc/GrpcCurrentTraceContext.java
```diff
@@ -0,0 +1,26 @@
+package brave.context.grpc;
+
+import brave.propagation.CurrentTraceContext;
+import brave.propagation.TraceContext;
+import io.grpc.Context;
+
+/** Manages scope with a {@link Context} as opposed to a thread local. */
+public final class GrpcCurrentTraceContext extends CurrentTraceContext {
```
adriancole: @kuroneko25 can you copy/paste this file into your project, then wrap it with the log4j2 context and see if it works for you?

ex. `ThreadContextCurrentTraceContext.create(GrpcCurrentTraceContext.create())`

kuroneko25: I wrapped the log4j2 context and that didn't work. But just to make sure I'm accessing it correctly `ThreadContext.get("traceId")` is the expected usage?

adriancole:
@kuroneko25 so this had nothing to do with the bug, but I can merge if you are interested in using it instead of thread locals for other reasons. Eventually, we'd likely add something like this in order to join traces with instrumentation-java. Anyway, let me know.

adriancole:
closing this as it was unrelated to the problem. If someone wants this we can re-open

PR #427, opened 2017-05-17
Eagerly provisions span so it is visible to httpclient interceptors

adriancole:
Thanks @TilinC for reporting


adriancole:
@kuroneko25 I will make a test like this for gRPC next.

adriancole:
@devinsba if this change makes sense to you, feel free to merge and cut a minor while away. Otherwise, we'll sync my morning

PR #429, opened 2017-05-17
Updates release timeout to 30m in response to latent maven central syncs

adriancole:
Last two releases include a lot more jars, and have timed out syncing to
Maven Central. The builds actually worked, as did the maven central
sync. However, the build failure causes alarm. This raises the timeout
to 30m until we no longer have timeouts (likely when we stop publishing
archived jars)

### DIFF
--- a/circle.yml
+++ b/circle.yml
@@ -54,8 +54,10 @@ deployment:
   publish-stable:
     owner: openzipkin
     tag: /\d+\.\d+\.\d+/
+    # triples the timeout to 30 minutes as maven central sync takes a lot longer than 10m
     commands:
-      - ./build-support/publish-stable.sh
+      - ./build-support/publish-stable.sh:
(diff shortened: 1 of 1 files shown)


adriancole:
cc @openzipkin/devops-tooling @fedj fyi This was due to build https://circleci.com/gh/openzipkin/brave/284 and a recent one that was the same (fail on maven central sync)

adriancole:
pps https://circleci.com/docs/1.0/configuration/#modifiers

eirslett:
The nineties called and wanted its dial-up modem back :fax:

devinsba:
Unfortunately this did not fix the issue. I think the maven plugin will need to be updated to fix this. I'll file an issue there

adriancole:
no delays lately, though the bomb is still here

PR #436, opened 2017-06-08
Ensures gRPC implementations can see the current span

### LINKED ISSUE #425: Tracing information in log4j2 with gRPC

adriancole:
Before, we handled user interceptors, but not the server implementation.

Fixes #425

### DIFF
--- a/instrumentation/grpc/src/main/java/brave/grpc/TracingServerInterceptor.java
+++ b/instrumentation/grpc/src/main/java/brave/grpc/TracingServerInterceptor.java
@@ -6,8 +6,8 @@ import brave.Tracing;
 import brave.propagation.Propagation;
 import brave.propagation.TraceContext;
 import brave.propagation.TraceContextOrSamplingFlags;
-import io.grpc.ForwardingServerCall;
 import io.grpc.ForwardingServerCall.SimpleForwardingServerCall;
+import io.grpc.ForwardingServerCallListener.SimpleForwardingServerCallListener;
 import io.grpc.Metadata;
 import io.grpc.ServerCall;
 import io.grpc.ServerCallHandler;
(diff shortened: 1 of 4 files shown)


adriancole:
going out as brave 4.3.4

Issue #303, opened 2016-12-16
Allow disabling tracing for certain resources with brave-jaxrs2


robinst:
Hi! We started using this library and it's really nice so far. We're using `brave-jaxrs2` to configure tracing on Jersey resources.

We have some resources where we want to disable tracing (a healthcheck endpoint that gets called frequently by Amazon's ELB). To do that, I've written a `DynamicBraveTracingFeature` that works like this:

```java
public class DynamicBraveTracingFeature implements DynamicFeature {

    private final Brave brave;
    private final SpanNameProvider spanNameProvider;

    public DynamicBraveTracingFeature(Brave brave, SpanNameProvider spanNameProvider) {
        this.brave = brave;
        this.spanNameProvider = spanNameProvider;
    }

    @Override
    public void configure(ResourceInfo resourceInfo, FeatureContext context) {
        if (resourceInfo.getResourceClass().isAnnotationPresent(DisableTracing.class) ||
                resourceInfo.getResourceMethod().isAnnotationPresent(DisableTracing.class)) {
            return;
        }
        context.register(BraveContainerRequestFilter.builder(brave).spanNameProvider(spanNameProvider).build());
        context.register(BraveContainerResponseFilter.create(brave));
    }
}
```

There's also an annotation called `DisableTracing` (omitted for brevity). What that allows is to disable tracing for a specific resource on the class or method level like this:

```java
@Path("/healthcheck")
@DisableTracing
public class HealthcheckResource {
    // ...
}
```

I'm wondering if that is something that you'd like to see us upstream. If yes, I can add a builder and documentation and submit a PR. What do you think?

adriancole:
definitely sensible to filter out resources. Unless you are expecting other dynamic features, I might name this feature FilteringBraveFeature or similar

Before you invest too much in a specific annotation. I'd look at how others are doing pattern exclusion for http tracing. If you need breadcrumbs let me know.

robinst:
> Before you invest too much in a specific annotation.

Not sure what you mean, it would be a new annotation living in the `brave-jaxrs2` artifact.

> I'd look at how others are doing pattern exclusion for http tracing. If you need breadcrumbs let me know.

Others? In Jersey? Breadcrumbs please?

adriancole:
> Before you invest too much in a specific annotation.
>
> Not sure what you mean, it would be a new annotation living in the brave-jaxrs2 artifact.

Sorry, what I meant to say is that something like this would bind
[...]

PR #437, opened 2017-06-08
Adds HttpRuleSampler for assigning sample rates to http endpoints

### LINKED ISSUE #432: Custom per url http sampler feature request
After an HTTP sampler that will allow the specification of different sampling rates for particular path and method combinations which will then fall through to the default rate set in brave.Tracing

Idea being you can specify rules like:
- GET /api/person/ has a sample rate of 0.1
- POST /api/person has a sample rate of 0.5
- Otherwise, fall through to brave.Tracing sampler

Regex on the path would be a nice to have, would settle for an absolute value.

adriancole:
This allows you to declare rules based on http patterns. These decide
which sample rate to apply.

Ex. Here's a sampler that traces 80% requests to /foo and 10% of POST
requests to /bar. Other requests will use a global rate provided by the
tracing component.

```java
httpTracingBuilder.serverSampler(HttpRuleSampler.newBuilder()
  .addRule(null, "/foo", 0.8f)
  .addRule("POST", "/bar", 0.1f)
  .build());
```

Fixes #432

### DIFF
--- a/brave/src/main/java/brave/sampler/ParameterizedSampler.java
+++ b/brave/src/main/java/brave/sampler/ParameterizedSampler.java
@@ -1,0 +1,58 @@
+package brave.sampler;
+
+import brave.propagation.SamplingFlags;
+import java.util.List;
+import javax.annotation.Nullable;
+
+/**
+ * This is an implementation of how to decide whether to trace a request using ordered rules. For
+ * example, you could write rules to look at an HTTP method and path, or a RabbitMQ routing key and
+ * queue name.
+ *
+ * <p>This looks at runtime parameters to see if they {@link Rule#matches(Object) match} a rule. If
+ * all calls to a java method should have the same sample rate, consider {@link DeclarativeSampler}
+ * instead.
+ *
+ * @param <P> The type that encloses parameters associated with a sample rate. For example, this
+ * could be a pair of http and method..
+ */
+public final class ParameterizedSampler<P> {
+  public static final <P> ParameterizedSampler<P> create(List<? extends Rule<P>> rules) {
+    if (rules == null) throw new NullPointerException("rules == null");
+    return new ParameterizedSampler<>(rules);
+  }
+
+  public static abstract class Rule<P> {
+    final Sampler sampler;
+
+    /**
+     * @param rate percentage of requests to start traces for. 1.0 is 100%
+     */
+    protected Rule(float rate) {
+      sampler = CountingSampler.create(rate);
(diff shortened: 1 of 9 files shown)


adriancole:
Note: polish still needed (ex markdown docs) cc @takezoe @hyleung @jplock @reta

reta:
Very useful feature, @adriancole

jplock:
If I had more time, I'd like to figure out a way to do the annotation based samplers as I think that would be easier to work with for Jersey users (similar to https://github.com/openzipkin/brave/issues/303).  I think this implementation is really useful for starters though.  I can expose this within the dropwizard configuration in https://github.com/smoketurner/dropwizard-zipkin

adriancole:
only thing pending here besides test fix is see if it can be abstracted to work with non-http ex messaging, since there could be similar rules for topic and message headers.

adriancole:
extracted ParameterizedSampler underneath for use in non-http use cases.

@sirtyro can you have a look? if looks good in principle, I'll clean up things fix tests etc.

brave/src/main/java/brave/sampler/ParameterizedSampler.java
```diff
@@ -0,0 +1,58 @@
...
+ * queue name.
+ *
+ * <p>This looks at runtime parameters to see if they {@link Rule#matches(Object) match} a rule. If
+ * all calls to a java method should have the same sample rate, consider {@link DeclarativeSampler}
+ * instead.
+ *
+ * @param <P> The type that encloses parameters associated with a sample rate. For example, this
+ * could be a pair of http and method..
+ */
+public final class ParameterizedSampler<P, R extends ParameterizedSampler.Rule<P>> {
```
adriancole: I'll probably quiet down the generics here if I can to only the parameter type

adriancole: hushed

adriancole:
fixed unrelated bug in apachehc parsing

llinder:
This looks really good.  I like that its not strictly tied to http, this will be helpful for several non http use cases for us.

adriancole:
great I'll polish it up for merge

sirtyro:
Looks good to me also, cheers for this!

adriancole:
cool. I have a couple tests and a readme left. then will merge (likely in an hour or two if lucky)

adriancole:
ready to merge (and release) on green

PR #438, opened 2017-06-08
Makes maven fail on error during integration tests

adriancole:
Exceptions are not assertions, so materialize as errors. Errors don't
cause the integration-test phase to fail, while it will for verify.


adriancole:
@wdittmer-mp this bud's for you

PR #439, opened 2017-06-08
Adds DeclarativeSampler and JAX-RS ContainerAdapter

### LINKED ISSUE #433: Parameterized non http sampler feature request
Looking for a way to have custom sampling rates set for traces that are started by a scheduled task, custom annotation or managed operation (JMX).

A possible solution would be that when we call tracer.nextSpan() we provide an opaque object which is used as the key to lookup (or create if it doesn't exist) an individual counting sampler and the desired rate.

We have our own annotation that we use to specify custom tracing start points in our code. So we would then be able to do something like @TraceThisMethod(sampleRate = 0.12) and pass in the instance of the annotation and the sampling rate from it.

Under the hook you would do something like a conditional lookup on the object.
samplerCache.get(target, target -> CountingSampler.create(rate));

You wouldn't need to fall back to Tracing.sampler() as it would create one instead (making the call to nextSpan with the parameters means you have short circuited the default sampler).

adriancole:
This adds functionality needed to make sampling decisions based on 
annotations present on a Java method (as opposed to runtime parameters).

Specifically, this adds `DeclarativeSampler` which parses a description
of a method into a sample rate (or null to defer the decision).

It also adds `ContainerSampler`, which uses `DeclarativeSampler` to 
allow decisions based on the JAX-RS `ResourceInfo` type.

Finally, this adds a feature test showing how arbitrary tracing might 
look like using the `DeclarativeSampler` with AspectJ.

Fixes #433

### DIFF
--- a/instrumentation/http-tests/src/main/java/brave/http/ITHttpServer.java
+++ b/instrumentation/http-tests/src/main/java/brave/http/ITHttpServer.java
@@ -3,6 +3,7 @@ package brave.http;
 import brave.SpanCustomizer;
 import brave.internal.HexCodec;
 import brave.sampler.Sampler;
+import java.io.IOException;
 import okhttp3.OkHttpClient;
 import okhttp3.Request;
 import okhttp3.Response;
@@ -38,16 +39,12 @@ public abstract class ITHttpServer extends ITHttp {
     final String parentId = traceId;
     final String spanId = "48485a3953bb6124";
 
-    Request request = new Request.Builder().url(url(path))
+    get(new Request.Builder().url(url(path))
         .header("X-B3-TraceId", traceId)
         .header("X-B3-ParentSpanId", parentId)
         .header("X-B3-SpanId", spanId)
         .header("X-B3-Sampled", "1")
-        .build();
-
-    try (Response response = client.newCall(request).execute()) {
-      assertThat(response.isSuccessful()).isTrue();
-    }
+        .build());
 
     assertThat(spans).allSatisfy(s -> {
       assertThat(HexCodec.toLowerHex(s.traceId)).isEqualTo(traceId);
@@ -61,12 +58,7 @@ public abstract class ITHttpServer extends ITHttp {
     httpTracing = HttpTracing.create(tracingBuilder(Sampler.NEVER_SAMPLE).build());
     init();
 
-    String path = "/foo";
-
-    Request request = new Request.Builder().url(url(path)).build();
-    try (Response response = client.newCall(request).execute()) {
-      assertThat(response.isSuccessful()).isTrue();
-    }
+    get("/foo");
 
     assertThat(spans)
         .isEmpty();
@@ -77,15 +69,7 @@ public abstract class ITHttpServer extends ITHttp {
    */
   @Test
   public void async() throws Exception {
-    String path = "/async";
-
-    Request request = new Request.Builder().url(url(path)).build();
-    try (Response response = client.newCall(request).execute()) {
-      if (response.code() == 404) throw new AssumptionViolatedException(path + " not supported");
-      assertThat(response.isSuccessful()).isTrue();
-    } catch (AssumptionViolatedException e) {
-      throw e;
-    }
+    get("/async");
 
     assertThat(spans).hasSize(1);
   }
@@ -98,16 +82,7 @@ public abstract class ITHttpServer extends ITHttp {
    */
   @Test
   public void createsChildSpan() throws Exception {
-    String path = "/child";
-
-    Request request = new Request.Builder().url(url(path)).build();
-    try (Response response = client.newCall(request).execute()) {
-      if (response.code() == 404) throw new AssumptionViolatedException(path + " not supported");
-    } catch (AssumptionViolatedException e) {
-      throw e;
-    } catch (Exception e) {
-      // ok, but the span should include an error!
-    }
+    get("/child");
 
     assertThat(spans).hasSize(2);
 
@@ -122,12 +97,7 @@ public abstract class ITHttpServer extends ITHttp {
 
   @Test
   public void reportsClientAddress() throws Exception {
-    String path = "/foo";
-
-    Request request = new Request.Builder().url(url(path)).build();
-    try (Response response = client.newCall(request).execute()) {
-      assertThat(response.isSuccessful()).isTrue();
-    }
+    get("/foo");
 
     assertThat(spans)
         .flatExtracting(s -> s.binaryAnnotations)
@@ -137,14 +107,9 @@ public abstract class ITHttpServer extends ITHttp {
 
   @Test
   public void reportsClientAddress_XForwardedFor() throws Exception {
-    String path = "/foo";
-
-    Request request = new Request.Builder().url(url(path))
+    get(new Request.Builder().url(url("/foo"))
         .header("X-Forwarded-For", "1.2.3.4")
-        .build();
-    try (Response response = client.newCall(request).execute()) {
-      assertThat(response.isSuccessful()).isTrue();
-    }
+        .build());
 
     assertThat(spans)
         .flatExtracting(s -> s.binaryAnnotations)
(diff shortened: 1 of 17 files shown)


adriancole:
cc @sirtyro @jplock @devinsba this is a draft design, using AOP

adriancole:
@reta you may also be interested in this (plumbing when annotations drive traces)

reta:
@adriancole Definitely, perfect fit for JAX-RS (f.e. annotating API endpoints, using `DynamicFeature`, ...)

sirtyro:
@adriancole, looks good. Only thought so far is clean up? How do avoid the keyToSamplers getting too massive? Perhaps a WeakKeyConcurrentHashMap could help?

adriancole:
>
> @adriancole <https://github.com/adriancole>, looks good. Only thought so
> far is clean up? How do avoid the keyToSamplers getting too massive?
> Perhaps a WeakKeyConcurrentHashMap could help?
>
actually high cardinality would be a problem. For example, a key shouldn't
be something with a lot of parameters in it, rather a classifier. I was
thinking about this last night actually. maybe we can have a different map
for the negative lookup which is bounded.

adriancole:
next steps on this after feedback here and a direct chat w/@sirtyro

This is a bit confusing juxtaposed with the HttpRuleSampler. Also, this only shows a trivial case, where there's a key that's used for all arguments (ex hard-set global rate per method). To push this a little more is needed because things like Messaging and JAX-RS would need to look at something else.

So, I'm thinking this.
step 1: lookup or create a sampler (the input to this might be a template, method, annotation or a PJP in aspectJ)
step 2.a: if that sampler is not parameterized, just get the next decision
step 2.b: if that sampler is parameterized, apply input to get the next decision

We currently can't do step 2.b with this impl.

Also, we should deal with the key for the sampler. In ideal case, this is bounded as no runtime parameters are keys. For example, an annotation instance is a good key as they will correspond to compiled symbols. A template is a good key as there are a small amount of these. A path isn't a good key as it is unbounded. In practice, the Keyed thing may not work at all for parameterized (2.b) sampling, but it is worth a shot to try and understand it.

Finally, this is tediously complex without real use cases to drive the design. For that reason, I won't blather on more until there's something realistic to reference (like a mock JAX-RS template sampler)

adriancole:
renamed to annotationsampler and added a lot of javadoc.

need to think slightly more about applicability to JAX-RS (since we also have HttpSampler there)

adriancole:
sorry about the name flip-flopping.. I just renamed to `DeclarativeSampler` which emphasizes that this uses parameters marked up declaratively (vs runtime). I made the type parameter `<M>`, which was to facilitate usage in JAX-RS (in anticipation of using for `ResourceInfo` which you can get access to in a dynamic feature etc). I've still to implement an http sampler using something besides an annotation.

adriancole:
@reta @jplock @sirtyro mind taking a look? I think this is ready but want some feedback if you can. I've re-written the description. Note this complements but does not replace the HttpRuleSampler. The `HttpRuleSampler` only works on runtime parameters such as the current http path. Code here is based on static information such as annotations on the java method.

instrumentation/jaxrs2/src/main/java/brave/jaxrs2/TracingContainerFilter.java
```diff
@@ -46,7 +46,9 @@
   @Context ResourceInfo resourceInfo;
 
   @Override public void filter(ContainerRequestContext request) {
+    if (resourceInfo != null) request.setProperty(ResourceInfo.class.getName(), resourceInfo);
```
adriancole: @llinder you might want to look at this including plumbing as it might be an option for ratpack

instrumentation/jaxrs2/src/test/java/brave/jaxrs2/ITTracingFeature_Container.java
```diff
@@ -102,18 +112,72 @@ public Response toResponse(Exception e) {
...
+    @GET
+    @Path("bar")
+    public Response bar() {
+      return Response.status(200).build();
+    }
+  }
 
+  @Test
+  public void declarative_sampling() throws Exception {
+    httpTracing = ContainerSampler.decorate(httpTracing, new RateForMethod<ResourceInfo>() {
```
adriancole: ex. here's an example of weaving in a custom trace annotation into JAX-RS cc @reta @jplock

instrumentation/http-tests/src/main/java/brave/http/ITHttpServer.java
```diff
@@ -265,16 +227,4 @@ private void addsErrorTagOnException(String path) {
...
-    String uri = "/foo?z=2&yAA=1";
-
-    Request request = new Request.Builder().url(url(uri)).build();
-    try (Response response = client.newCall(request).execute()) {
-      assertThat(response.isSuccessful()).isTrue();
-    }
-
-    assertReportedTagsInclude(TraceKeys.HTTP_PATH, "/foo");
-  }
 }
```
adriancole: refactored this file to pay for some of the lines of code, although it didn't really pay for many of them :P

jplock:
@adriancole should we include a new `@Traced` annotation in this PR so downstream dependencies can use it?

adriancole:
>
> @adriancole <https://github.com/adriancole> should we include a new
> @Traced annotation in this PR so downstream dependencies can use it?
>
interesting idea. I was thinking to somehow not introduce a dependency on
user-code, but then again being able to opt-in is probably better than
having everyone make their own annotations. OTOH, we'd have to decide what
properties to assign.

ex. sampleRate spanName maybe other things maybe nothing?

I know @sirtyro has an annotation they have been using (possibly multiple
annotations)
Also sleuth has a number of them
https://github.com/spring-cloud/spring-cloud-sleuth/tree/0f29735c1112e8586103a274b1d8fcbdc8a7d189/spring-cloud-sleuth-core/src/main/java/org/springframework/cloud/sleuth/annotation

Does anything resonate with you?

jplock:
I definitely want to keep it simple.  From my narrow view of supporting dropwizard-zipkin which is only dealing with Jersey, a single `@Traced(rate=0.1, enabled=true)` annotation would suit my needs I think.  Otherwise I'll add that in the dropwizard bundle and expose that to my users.

sirtyro:
We've taken it a step further. Our @Trace annotation has got a propagation property so users can specify when a new child span should start. Annotation is an interesting beast, perhaps a different PR?

adriancole:
@jplock @reta  here's a thought. I think it would be best to move the JAX-RS container sampler to the test tree. For example, if you are only using annotations, it can be done simpler. Plus, users don't have to see it if using via dropwizard anyway. This lowers the surface area of the change. I'd leave ContainerAdapter exposed because this helps isolate how the resource method magic gets stitched in. After looking at how this is used in dropwizard and CXF, if we should top-level ContainerSampler, we can always do that.

reta:
@adriancole I think it sounds reasonable and makes sense to me as well, keeping `ContainerAdapter` exposed and moving `ContainerSampler` to test tree.

instrumentation/jaxrs2/src/test/java/brave/jaxrs2/features/sampling/Traced.java
```diff
@@ -0,0 +1,37 @@
...
+import brave.http.HttpSampler;
+import brave.jaxrs2.ContainerAdapter;
+import brave.sampler.DeclarativeSampler;
+import java.lang.annotation.Retention;
+import java.lang.annotation.RetentionPolicy;
+import java.lang.reflect.Method;
+import javax.annotation.Nullable;
+
+@Retention(RetentionPolicy.RUNTIME)
+public @interface Traced {
```
adriancole: @reta @jplock here's the test implementation of annotation-based sampling. It isn't much code, but I placed it here (as well tests for the impl). Hopefully, this will guide usage!

adriancole:
4.4.0-SNAPSHOT from snapshot repo oss.jfrog.org/artifactory/oss-snapshot-local will have this shortly

instrumentation/jaxrs2/src/test/java/brave/jaxrs2/features/sampling/Traced.java
reta: Thanks, @adriancole, the annotation is useful baseline, could be interpreted by frameworks / apps differently

PR #440, opened 2017-06-11
Add zipkin support for Netty http handler

songxin1990:
Add zipkin support for Netty http handler.
But there is one problem:
inherited httpTracing block the Netty httpserver start,so the test can't pass.
For details,see ITNettyTracing.java.


adriancole:
hmm mind raising this from a branch? when raising a PR from master, it is easy to get things cluttered up. Right now, there's some accidental changes to spark-java which aren't related to this.

songxin1990:
@adriancole  so what should i do? create a new feature branch to do something specific?maybe I should figure it out myself.

Take a look at the code especially the ITNettyTracing.java. If I pass the inherited httpTracing,the netty httpserver can't start up successfully.

adriancole:
yes create a new feature branch, and make sure it only has new code in it
related to netty (in a single commit). this will save us time later when we
try to get your code working as github would let me or others edit that
branch.

songxin1990:
Ok,thanks.I will do it tomorrow.

Issue #442, opened 2017-06-14
Add abandon method to Span


sirtyro:
We've run into an interesting scenario whereby we need to do span.abandon() as we have our own custom instrumentation around async messaging. So we start a span before we find out that its a duplicate. We don't want to finish or flush it as we could get up to four messages like this, which is a lot of noise. We could add a tag and then extend the reporter to simply not report spans that have a duplicate tag but it feels a bit hacky

Could we get a span.abandon() implementation?

adriancole:
cc @felixbarny I think you have a use case for late-dropping spans when the app is a monolith (ex in this case it is more an after-the-fact sampling decision)

adriancole:
So, this is very much like a local, after-the-fact sampling decision raised earlier in stagemonitor https://github.com/stagemonitor/stagemonitor/issues/268

`span.abandon()` means dump what you have and finish, but runs a resurrection risk.

`span.unsample()` means don't report this span (and also don't propagate it), but not necessarily finish.

`span.unsample()` could return a noopspan with the sampling flag unset in the trace context. In impl, it would write a field to MutableSpan used in the recorder so it remembers to not report it (even on GC).

sirtyro:
I could accept that it is an after-the-fact sampling decision.

`span.unsample()` returning a noop span is confusing but i otherwise agree with what you said, i like that it would not be reported on GC either.

adriancole:
note: I chose `unsample()` for option 2 as opposed to `sample(boolean)` as data is thrown out or not recorded at all when the span starts unsampled. late unsampling is viable, but late sampling would create junk.

sirtyro:
I prefer unsample() as it feels more terminal, which it should. And if it returns a noop span that then helps to stop propagation?

adriancole:
@sirtyro another choice could be to name it `span.setNoop()` which is effectively the same and might make sense as to why the result would be a noop span.

sirtyro:
Could be confusing then as to why there is no undo, ie setReal() ? I'm ok with unsample. That is clearer than setNoop to me. But ultimately, I don't mind too much.

At the end of the day, the javadoc should make it clear why things are happening.

adriancole:
in-process you can propagate the noop span to prevent expensive tags from being recorded or prevent new traces from being made. If you happen to be in a terminal part of code, you wouldn't propagate it anyway.

sirtyro:
Yep cool, concur with that.

Happy with unsample or setNoop

[...]

PR #443, opened 2017-06-14
Add abandon method to span

sirtyro:
implementation of #442

### DIFF
--- a/brave/src/test/java/brave/RealSpanTest.java
+++ b/brave/src/test/java/brave/RealSpanTest.java
@@ -61,6 +61,13 @@ public class RealSpanTest {
         .containsExactly(3L);
   }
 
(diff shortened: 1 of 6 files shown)


brave/src/test/java/brave/RealSpanTest.java
```diff
@@ -61,6 +61,13 @@
         .containsExactly(3L);
   }
 
+  @Test public void abandon() {
+    span.start();
+    span.abandon();
+
+    assertThat(spans).hasSize(0);
```
adriancole: nit `.isEmpty()`

adriancole:
merging as-is and will follow-up in brave-opentracing to help with the late sampling flag thing

PR #445, opened 2017-06-16
Adds Tracing.setNoop to disable tracing at runtime

### LINKED ISSUE #435: Kill switch feature
It would be good to have a switch within brave that can be used to stop anything and everything related to tracing.

For our adoption, we need to provide some risk mitigation that if we're having any issues we can easily and safely stop tracing. We have a good way to signalling to our artefacts that they need to turn off brave. What we don't have is a good way to actually turn brave off.

At the moment, we're wrapping all our samplers and reporter with delegates that we can switch off, which will simply stop delegating. 

It would also mean that we can easily turn off tracing during a build, rather than what we do at the moment which is to have a test jar present with a marker class and our configuration has conditional beans meaning you get a no-op reporter for builds.

adriancole:
If you are in a situation where you need to turn off tracing at runtime,
invoke `Tracing.setNoop(true)`. This will turn any new spans into "noop"
spans, and drop any data until `Tracing.setNoop(false)` is invoked.

Fixes #435

### DIFF
--- a/brave/README.md
+++ b/brave/README.md
@@ -178,6 +178,7 @@ next = tracer.newSpan(oneWayReceive.context()).name("step2").start();
 There's a working example of a one-way span [here](src/test/java/brave/features/async/OneWaySpanTest.java).
 
 ## Sampling
+
 Sampling may be employed to reduce the data collected and reported out
 of process. When a span isn't sampled, it adds no overhead (noop).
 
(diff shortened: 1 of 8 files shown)


adriancole:
This is pending review

sirtyro:
Looks good. I like that it circumvents sampling too!

adriancole:
thx for having a look

PR #447, opened 2017-06-22
Adds an end user friendly span customiser

sirtyro:
As per discussions with @adriancole, we don't want to expose the need to deal with tracer.currentSpan and its possibility of being null to end users. This adds a class that nicely wraps that up, which can be made into a spring bean that end users can wire in.

@adriancole, a few things:
- not sure about whether we should just have a reference to Tracer or whether we should always do Tracing.tracer().etc ?
- You mentioned returning a Noop span if there isn't one current, however you cannot create one without a TraceContext so it seemed simpler to just not do the operation if there isn't one and then return customiser rather than the real span.

### DIFF
--- a/brave/src/main/java/brave/CurrentSpanCustomizer.java
+++ b/brave/src/main/java/brave/CurrentSpanCustomizer.java
@@ -3,7 +3,7 @@ package brave;
 /**
  * Provides a mechanism for end users to be able to customise the current span.
  *
- * Handles the case of there being no current span in scope.
+ * <p>Handles the case of there being no current span in scope.
  */
 public final class CurrentSpanCustomizer implements SpanCustomizer {
 
--- a/brave/src/main/java/brave/CurrentSpanCustomizer.java
+++ b/brave/src/main/java/brave/CurrentSpanCustomizer.java
@@ -1,0 +1,56 @@
+package brave;
+
+/**
+ * Provides a mechanism for end users to be able to customise the current span.
+ *
+ * Handles the case of there being no current span in scope.
+ */
+public class CurrentSpanCustomizer implements SpanCustomizer {
+
+  private Tracing tracing;
+
+  /** Creates a span customizer that will affect the current span in scope if present */
+  public static CurrentSpanCustomizer create(Tracing tracing) {
+    return new CurrentSpanCustomizer(tracing);
+  }
+
+  CurrentSpanCustomizer(Tracing tracing) {
+    this.tracing = tracing;
+  }
+
+  /** {@inheritDoc} */
+  @Override public SpanCustomizer name(String name) {
+    Span currentSpan = tracing.tracer().currentSpan();
+    if (currentSpan != null) {
+      currentSpan.name(name);
+    }
+    return this;
+  }
+
+  /** {@inheritDoc} */
+  @Override public SpanCustomizer tag(String key, String value) {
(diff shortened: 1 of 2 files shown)


brave/src/main/java/brave/CurrentSpanCustomizer.java
```diff
@@ -0,0 +1,56 @@
+package brave;
+
+/**
+ * Provides a mechanism for end users to be able to customise the current span.
+ *
+ * Handles the case of there being no current span in scope.
+ */
+public class CurrentSpanCustomizer implements SpanCustomizer {
+
+  private Tracing tracing;
```
adriancole: I'd make this ref final and to Tracer (so that each call doesn't need to do tracing.tracer)

brave/src/main/java/brave/CurrentSpanCustomizer.java
```diff
@@ -0,0 +1,56 @@
+package brave;
+
+/**
+ * Provides a mechanism for end users to be able to customise the current span.
+ *
+ * Handles the case of there being no current span in scope.
+ */
+public class CurrentSpanCustomizer implements SpanCustomizer {
```
adriancole: nit: final

brave/src/main/java/brave/CurrentSpanCustomizer.java
```diff
@@ -0,0 +1,56 @@
+package brave;
+
+/**
+ * Provides a mechanism for end users to be able to customise the current span.
+ *
+ * Handles the case of there being no current span in scope.
```
adriancole: add a `<p>` in front of the H or javadoc won't do what you want.

sirtyro:
Re: static final SpanCustomizer NOOP = new SpanCustomizer() ...

Its probably more work to return the NOOP than the current approach so I'll leave it as is?

adriancole:
yeah agreed NOOP would be more work in hindsight.

brave/src/main/java/brave/CurrentSpanCustomizer.java
```diff
@@ -0,0 +1,56 @@
+package brave;
+
+/**
+ * Provides a mechanism for end users to be able to customise the current span.
+ *
+ * <p>Handles the case of there being no current span in scope.
+ */
+public final class CurrentSpanCustomizer implements SpanCustomizer {
```
adriancole: @reta @jplock @llinder @devinsba FYI, this is a way to prevent users from having or needing a reference to the tracer. For example, provide `SpanCustomizer` via this type. Then if users employ `SpanCustomizer`, they can contribute to a current span w/o using any advanced types.

adriancole: ex in the rare case users want to customize a span, they can inject it

```java
@Inject SpanCustomizer span; // this won't be null, if provided by CurrentSpanCustomizer

 void userCode() {
   span.annotate("tx.started");
   ...
 }
```

reta: Thanks @adriancole! Nice way to deal with spans without bringing the whole tracer instance, the similar approach we are using in JAX-RS using contextual injection.

adriancole:
added docs https://github.com/openzipkin/brave/commit/a3f5f0aeb518b6d7fd024ae02826e09f09a9ccd0

PR #452, opened 2017-07-03
updated the JSR305 to servicemix bundle

kameshsampath:
Updated the brave JSR305 with service mix JSR305 3.0.2_1 bundle to make the `javax.annotation;version="[1.1,2.0)"` instead of `javax.annotation;version="[3.0.0,4.0.0)"` to make it deployable in OSGi containers that already export javax.annotation.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -332,9 +332,9 @@
        This is added as a root-level dependency to prevent javadoc from failing.
        -->
     <dependency>
-      <groupId>com.google.code.findbugs</groupId>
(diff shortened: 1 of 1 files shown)


cschneider:
I tested the PR and it works fine for me. So I recommend to merge.

adriancole:
Thanks for the help from you both!

kameshsampath:
thanks @cschneider, my pleasure @adriancole

PR #455, opened 2017-07-14
Adds ThreadLocalSpan to show how to deal with callbacks w/o attributes

adriancole:
Sometimes you have to instrument a library where there's no attribute
namespace shared across request and response. For this scenario, you can
use `ThreadLocalSpan` to temporarily store the span between callbacks.

Here's an example:
```java
class MyFilter extends Filter {
  final ThreadLocalSpan threadLocalSpan;

  public void onStart(Request request) {
    // Allocates a span and places it in scope so that code between here and onFinish can see it
    Span span = threadLocalSpan.next();
    if (span == null || span.isNoop()) return; // skip below logic on noop

    // Assume you have code to start the span and add relevant tags...
  }

  public void onFinish(Response response, Attributes attributes) {
    // as long as we are on the same thread, we can read the span started above
    Span span = threadLocalSpan.remove();
    if (span == null || span.isNoop()) return; // skip below logic on noop

    // Assume you have code to complete the span
  }
}
```

### DIFF
--- a/brave/src/main/java/brave/propagation/ThreadLocalSpan.java
+++ b/brave/src/main/java/brave/propagation/ThreadLocalSpan.java
@@ -1,0 +1,98 @@
+package brave.propagation;
+
+import brave.Span;
+import brave.Tracer;
+import javax.annotation.Nullable;
+
+/**
+ * This type allows you to place a span in scope in one method and access it in another without
+ * using an explicit request parameter.
+ *
+ * <p>Many libraries expose a callback model as opposed to an interceptor one. When creating new
+ * instrumentation, you may find places where you need to place a span in scope in one callback
+ * (like `onStart()`) and end the scope in another callback (like `onFinish()`).
+ *
+ * <p>Provided the library guarantees these run on the same thread, you can simply propagate the
+ * result of {@link Tracer#withSpanInScope(Span)} from the starting callback to the closing one.
+ * This is typically done with a request-scoped attribute.
+ *
+ * Here's an example:
+ * <pre>{@code
+ * class MyFilter extends Filter {
+ *   public void onStart(Request request, Attributes attributes) {
+ *     // Assume you have code to start the span and add relevant tags...
+ *
+ *     // We now set the span in scope so that any code between here and
+ *     // the end of the request can see it with Tracer.currentSpan()
+ *     SpanInScope spanInScope = tracer.withSpanInScope(span);
+ *
+ *     // We don't want to leak the scope, so we place it somewhere we can
+ *     // lookup later
+ *     attributes.put(SpanInScope.class, spanInScope);
+ *   }
+ *
+ *   public void onFinish(Response response, Attributes attributes) {
+ *     // as long as we are on the same thread, we can read the span started above
+ *     Span span = tracer.currentSpan();
+ *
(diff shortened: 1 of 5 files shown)


adriancole:
cc @sirtyro @jakerobb I think both of you have asked similar things.. If this looks helpful, let me know and I'll write tests for it.

cc @llinder @devinsba

jakerobb:
Somehow I completely overlooked this. Seems like a good idea to me!

brave/src/main/java/brave/propagation/ThreadLocalSpan.java
```diff
@@ -0,0 +1,98 @@
...
+
+  final Tracer tracer;
+  final ThreadLocal<Tracer.SpanInScope> currentSpanInScope = new ThreadLocal<>();
+
+  ThreadLocalSpan(Tracer tracer) {
+    this.tracer = tracer;
+  }
+
+  public void set(Span span) {
+    currentSpanInScope.set(tracer.withSpanInScope(span));
```
jakerobb: I'm fairly sure this won't play nice with BraveActiveSpanScope in https://github.com/openzipkin/brave-opentracing/pull/43, which also calls tracer.withSpanInScope. Then you end up with the same span in the stack twice, where the second one thinks that the first one (in fact the same span) is its parent, which leads to the strange parent-child relationships that threw me off for so long.

adriancole: I don't quite understand where having the same span placed in scope twice can cause a problem, as long as whoever puts something in scope also closes it. Should maybe I put some unit tests to show the state isn't corrupted when a span is double-scoped? Maybe we aren't testing this and should I guess?

adriancole: I think what will happen is BraveActiveSpanScope will have no idea about this type :) To test, you could in BraveActiveSpanScope's test put a span in scope opentracing doesn't know about, and close that scope (maybe interleaved between ActiveSpan doing the same).

Sometime soon, I'll make a base test class for CurrentTraceContext and put a test in for double-scoping to make sure it doesn't die.

jakerobb: Ooh, sorry, I didn't explain the problem very well. And, in fact, I think it turns out that there's no problem _here_. 

I'm going to submit a separate issue.

adriancole:
holding off on this until it is more popular or a week passes. If it doesn't get rule-of-three by then I'll drop it.

sirtyro:
Sorry about the lag, I consider this highly valuable. We had to implement this ourselves and couldn't understand why it wasn't in brave, so yes please!

jonathan-lo:
:+1: 

Didn't notice this earlier!

adriancole:
>
> Sorry about the lag, I consider this highly valuable. We had to implement
> this ourselves and couldn't understand why it wasn't in brave, so yes
> please!
>
the rule of three :) if you notice something missing, ask. features are
added late to ensure maintenance is worthwhile

adriancole:
OK revised the design a little. By provisioning a span up-front, some null checks and indentation are avoided. Let me know what you think.

PR #458, opened 2017-07-20
P6spy log with jdbc parameter values

jorgheymans:
allow to collect jdbc statements including parameter values during p6spy tracing.


jorgheymans:
There are tests failing in brave-okhttp, seems not related to my PR.

jorgheymans:
yep like it as well, i'll amend the PR thanks for the fast feedback

adriancole:
Good stuff. Thanks for the help!

PR #459, opened 2017-07-24
Avoiding NPEs in `p6spy` edge-cases

suls:
In rare cases, `p6spy` is unable to intercept the executed SQL properly.
This will then lead to an NPE in the
`TracingJdbcEventListener#onBeforeAnyExecute` method.
This change is adding a `null` check and will use `query` as the SQL
instead.

note: this is related to https://github.com/p6spy/p6spy/pull/385

cc @kristofa @adriancole

### DIFF
--- a/instrumentation/p6spy/src/main/java/brave/p6spy/TracingJdbcEventListener.java
+++ b/instrumentation/p6spy/src/main/java/brave/p6spy/TracingJdbcEventListener.java
@@ -13,6 +13,10 @@ import zipkin.Endpoint;
 import zipkin.TraceKeys;
 
 final class TracingJdbcEventListener extends SimpleJdbcEventListener {
+  /**
+   * Use this instead of the actual sql if p6spy wasn't able to intercept it.
+   */
+  static final String QUERY_PLACEHOLDER = "query ";
 
   final String remoteServiceName;
   final boolean includeParameterValues;
@@ -30,6 +34,7 @@ final class TracingJdbcEventListener extends SimpleJdbcEventListener {
     // regardless of noop or not, set it in scope so that custom contexts can see it (like slf4j)
     if (!span.isNoop()) {
       String sql = includeParameterValues ? info.getSqlWithValues() : info.getSql();
+      if (sql == null) sql = QUERY_PLACEHOLDER;
       span.kind(Span.Kind.CLIENT).name(sql.substring(0, sql.indexOf(' ')));
       span.tag(TraceKeys.SQL_QUERY, sql);
       parseServerAddress(info.getConnectionInformation().getConnection(), span);
--- a/instrumentation/p6spy/src/test/java/brave/p6spy/TracingJdbcEventListenerTest.java
+++ b/instrumentation/p6spy/src/test/java/brave/p6spy/TracingJdbcEventListenerTest.java
@@ -1,5 +1,6 @@
 package brave.p6spy;
 
+import static org.assertj.core.api.Assertions.assertThat;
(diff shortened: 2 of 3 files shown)


instrumentation/p6spy/src/test/java/brave/p6spy/TracingJdbcEventListenerTest.java
```diff
@@ -66,4 +78,22 @@
...
+    when(statementInformation.getConnectionInformation()).thenReturn(ci);
+    when(ci.getConnection()).thenReturn(connection);
+    when(connection.getMetaData()).thenReturn(metaData);
+    when(metaData.getURL()).thenReturn(url);
+
+    TracingJdbcEventListener listener = new TracingJdbcEventListener("", false);
+    listener.onBeforeAnyExecute(statementInformation);
+    listener.onAfterAnyExecute(statementInformation,1,null);
+
+    assertThat(spans.getFirst().name, is(equalTo(TracingJdbcEventListener.QUERY_PLACEHOLDER.trim())));
```
adriancole: nit: prefer assertj since we use it everywhere. assertThat(...).isEqualTo(

Also, best to use the string constant vs having people look it up :) a touch of anti DRY in tests can make it easier to understand.

instrumentation/p6spy/src/main/java/brave/p6spy/TracingJdbcEventListener.java
```diff
@@ -30,6 +34,7 @@
     // regardless of noop or not, set it in scope so that custom contexts can see it (like slf4j)
     if (!span.isNoop()) {
       String sql = includeParameterValues ? info.getSqlWithValues() : info.getSql();
+      if (sql == null) sql = QUERY_PLACEHOLDER;
       span.kind(Span.Kind.CLIENT).name(sql.substring(0, sql.indexOf(' ')));
```
adriancole: if you use tertiary op here, you don't need to play with extra span.

Ex. scrub the constant and 

```java
// use "query" if p6spy wasn't able to intercept the SQL.
.name(sql == null ? "query" : sql.substring(0, sql.indexOf(' ')));
```

suls: Could do, but we also need it the line below: `span.tag(TraceKeys.SQL_QUERY, sql);`

instrumentation/p6spy/src/test/java/brave/p6spy/TracingJdbcEventListenerTest.java
suls: done

adriancole:
maybe..

if (sql == null) sql = "query"; // mention why

then as tag value flip what we're checking against..

sql.indexOf(' ') ? sql.substring(0, sql.indexOf(' ')) : sql

main thing is trying to rid a trailing space as a tag

adriancole:
Actually we shouldnt tag sql.query unless it is sql. So choices are to abandon or not start a span with no sql, or start with a fake name like "query" but dont add a tag. Make sense?

adriancole:
Or add sql.query tag as empty string (which is weird but less than invalid sql)

adriancole:
so, the question here is..

Is it better to span.abandon() the vague span which has no data its query? If not, we should probably not add the sql tag as it isn't SQL.

Which sounds better? Since this is NPE causing we should probably fix it one way or another soon.

adriancole:
I closed this with the same code, except not starting a span if there is null SQL. This is something we can revisit if people show interest later

PR #461, opened 2017-07-25
Add Kafka instrumentation for SPSC

ImFlog:
Relates to https://github.com/openzipkin/brave/issues/211

For now SPSC.
Next steps : Kafka Streams and SPMC.


instrumentation/kafka/src/main/java/brave/kafka/RecordTracing.java
```diff
@@ -0,0 +1,55 @@
...
+import brave.propagation.TraceContext;
+import org.apache.kafka.clients.consumer.ConsumerRecord;
+import org.apache.kafka.common.header.Header;
+import org.apache.kafka.common.header.Headers;
+import zipkin.internal.Util;
+
+/**
+ * Util class which creates a child span according to the headers.
+ */
+public class RecordTracing {
```
adriancole: final

instrumentation/kafka/src/main/java/brave/kafka/RecordTracing.java
```diff
@@ -0,0 +1,55 @@
...
+
+    /**
+     * Continues the trace extracted from the headers or creates a new one if the extract fails.
+     * Call this method while consumming your kafka records.
+     */
+    public Span nexSpanFromRecord(ConsumerRecord record) {
+        Span newChild;
+        TraceContext context = extractor.extract(record.headers()).context();
+        if (context != null) {
+            newChild = tracing.tracer().newChild(context);
```
adriancole: nit: just early return

instrumentation/kafka/src/main/java/brave/kafka/RecordTracing.java
```diff
@@ -0,0 +1,55 @@
...
+                    if (header == null) return null;
+                    return new String(header.value(), Util.UTF_8);
+                });
+    }
+
+    /**
+     * Continues the trace extracted from the headers or creates a new one if the extract fails.
+     * Call this method while consumming your kafka records.
+     */
+    public Span nexSpanFromRecord(ConsumerRecord record) {
```
adriancole: I'd make this method simpler and more similar to http, just nextSpan (like HttpClientHandler.nextSpan)

instrumentation/kafka/src/main/java/brave/kafka/TracingConsumer.java
```diff
@@ -0,0 +1,181 @@
...
+
+import java.util.Collection;
+import java.util.List;
+import java.util.Map;
+import java.util.Set;
+import java.util.concurrent.TimeUnit;
+import java.util.regex.Pattern;
+
+
+public class TracingConsumer<K, V> implements Consumer<K, V> {
```
adriancole: make this package private and instead make a decorator similar to how some of the other libraries work.
This can help centralize config I think.

ex. 
```java
traced = kafkaTracing.consumer(consumer);
```

For example, KafkaTracing may have additional configuration, and this could be done in a builder vs ctor overloads here. Since we add no new methods, hiding the type by returning Consumer (and making ctor package private) is safest start!

adriancole: however this is exposed (via a decorating factory method or like this) needs type-level javadoc w/example

instrumentation/kafka/src/main/java/brave/kafka/TracingProducer.java
```diff
@@ -0,0 +1,118 @@
...
+import org.apache.kafka.common.errors.ProducerFencedException;
+import org.apache.kafka.common.header.Headers;
+import zipkin.internal.Util;
+
+import java.util.List;
+import java.util.Map;
+import java.util.concurrent.Future;
+import java.util.concurrent.TimeUnit;
+
+public class TracingProducer<K, V> implements Producer<K, V> {
```
adriancole: similar comment about how to expose this (look at GrpcTracing for example)

instrumentation/kafka/src/main/java/brave/kafka/TracingProducer.java
```diff
@@ -0,0 +1,118 @@
...
+
+    public static final String KAFKA_TOPIC_TAG = "kafka.topic";
```
adriancole: if we're doing this, extract to another file I think..

instrumentation/kafka/src/main/java/brave/kafka/TracingProducer.java
```diff
@@ -0,0 +1,118 @@
...
+    }
+
+    /**
+     * We wrap the send method to add tracing.
+     */
+    @Override
+    public Future<RecordMetadata> send(ProducerRecord<K, V> record, Callback callback) {
+        Span span = tracing.tracer().nextSpan();
+        span.tag(KAFKA_TOPIC_TAG, record.topic());
+        span.tag(KAFKA_KEY_TAG, record.key().toString());
```
adriancole: key might be null? also maybe just start with topic? or get input from others before doing key.

instrumentation/kafka/src/main/java/brave/kafka/RecordTracing.java
```diff
@@ -0,0 +1,55 @@
...
+            // Should we throw an exception ? Create a new span ?
+            newChild = tracing.tracer().nextSpan();
+        }
+        return newChild;
+    }
+
+    /**
+     * Close the producer async span extracted from the headers.
+     */
+    void finishProducerSpan(ConsumerRecord record) {
```
adriancole: I use maybe prefix on things that might not do anything.

instrumentation/kafka/src/test/java/brave/kafka/ITTracingKafka.java
```diff
@@ -0,0 +1,161 @@
...
+import org.apache.kafka.common.TopicPartition;
+import org.apache.kafka.common.utils.Time;
+import org.junit.After;
+import org.junit.Before;
+import org.junit.Test;
+import zipkin.Span;
+
+public class ITTracingKafka {
+
+    private static final String LOCALHOST = "127.0.0.1";
```
adriancole: nit I tend to use just normal instance fields in junit tests as it is less distracting to read.

instrumentation/kafka/src/test/java/brave/kafka/ITTracingKafka.java
```diff
@@ -0,0 +1,161 @@
...
+            .isEqualTo(Long.toHexString(producerSpans.getFirst().traceId));
+
+        RecordTracing recordTracing = new RecordTracing(consumerTracing);
+        for (ConsumerRecord<String, String> record : records) {
+            brave.Span span = recordTracing.nexSpanFromRecord(record);
+            assertThat(span.context().parentId()).isEqualTo(producerSpans.getLast().traceId);
+        }
+    }
+
+    private void initKafkaServer() throws IOException {
```
adriancole: I'd pull this stuff out to another file. Look at zipkin-reporter and zipkin.. I think we have a file somewhere that can be a template for approach.

instrumentation/kafka/src/test/java/brave/kafka/RecordTracingTest.java
```diff
@@ -0,0 +1,101 @@
...
+    private static final String SAMPLED = "1";
+
+    private static final String TEST_TOPIC = "myTopic";
+    private static final String TEST_KEY = "foo";
+    private static final String TEST_VALUE = "bar";
+
+    private RecordTracing recordTracing;
+    private ConsumerRecord<String, String> fakeRecord;
+
+    private ConcurrentLinkedDeque<zipkin.Span> spans = new ConcurrentLinkedDeque<>();
```
adriancole: unless concurrent, don't use concurrent. it is distracting as I will start thinking there is concurrency going on :)

instrumentation/pom.xml
```diff
@@ -31,6 +31,7 @@
     <module>sparkjava</module>
     <module>spring-web</module>
     <module>spring-webmvc</module>
+    <module>kafka</module>
```
adriancole: probably name this like the jar kafka-clients

instrumentation/kafka/src/test/java/brave/kafka/ITTracingKafka.java
```diff
@@ -0,0 +1,161 @@
...
+    public void produce_and_consume_kafka_message() throws Exception {
+        TracingProducer<String, String> tracingProducer = createTracingProducer();
+        TracingConsumer<String, String> tracingConsumer = createTracingConsumer();
+
+        Future<RecordMetadata> send = tracingProducer.send(new ProducerRecord<>(TEST_TOPIC, TEST_KEY, TEST_VALUE));
+        // Block for synchronous send
+        send.get();
+
+        ConsumerRecords<String, String> records = tracingConsumer.poll(1000);
+        // Oddly we need to poll twice to retrieve content
```
dgrabows: I think that poll needs to be called in a loop that exits once a record is returned or some safeguard timeout is reached. Calling poll is not guaranteed to return records when called, even if there are messages on the broker that could be consumed. See http://grokbase.com/p/kafka/users/155mqpwf3n/consumer-poll-returns-no-records-unless-called-more-than-once-why for more details. That post is a couple years old, but I'm pretty sure the advice is still accurate.

A test case I've written that deals with this is https://github.com/openzipkin/zipkin/blob/master/zipkin-collector/kafka10/src/test/java/zipkin/collector/kafka10/KafkaCollectorTest.java#L135. In that test case, `receivedSpans` is a blocking queue and the Kafka consumer polling loop is buried in the KafkaCollector, but the end result should be the same.

instrumentation/kafka/README.md
```diff
@@ -0,0 +1,26 @@
+# Brave Kafka instrumentation
+
+Add decorators for Kafka producer and consumer to enable tracing for Single Producer/Single Consumer (for now).
+The producer propagates the B3 headers in the records headers.
+The consumer is responsible to close the producer span.
+
+To use the producer simply wrap it like this : 
+```java
+Producer<String, String> stringProducer = new KafkaProducer<>(settings);
+TracingProducer<String, String> tracingProducer = new TracingProducer<>(tracing, mockProducer);
```
dgrabows: Is `mockProducer` supposed to be `stringProducer`?

instrumentation/kafka/src/test/java/brave/kafka/ITTracingKafka.java
```diff
@@ -0,0 +1,161 @@
...
+    private void setupKafka(String zkAddress) throws IOException {
+        Properties brokerProps = new Properties();
+        brokerProps.setProperty("zookeeper.connect", zkAddress);
+        brokerProps.setProperty("broker.id", "0");
+        brokerProps.setProperty("log.dirs", Files.createTempDirectory("kafka-").toAbsolutePath().toString());
+        brokerProps.setProperty("listeners", "PLAINTEXT://" + LOCALHOST + ":" + KAFKA_PORT);
+
+        brokerProps.setProperty("replica.socket.timeout.ms", "1000");
+        brokerProps.setProperty("controller.socket.timeout.ms", "1000");
+        brokerProps.setProperty("offsets.topic.replication.factor", "1");
```
dgrabows: I don't think it is contributing to the intermittent test failure, but you could reduce churn on broker startup by also reducing the number of partitions for the consumer offsets topic from the default of 50 to 1.

`brokerProps.setProperty("offsets.topic.num.partitions", "1");`

instrumentation/kafka/README.md
```diff
@@ -0,0 +1,26 @@
+# Brave Kafka instrumentation
```
devinsba: Might be good to make a note on the readme that this is only compatible with the recent versions of Kafka

instrumentation/kafka/README.md
ImFlog: Could be any kind of producer. You made me realize that I wrongly copy/pasted this code for the documentation.
Will change to ```realProducer``` and remove the types.

instrumentation/kafka/pom.xml
```diff
@@ -0,0 +1,79 @@
+<?xml version="1.0"?>
+<project xmlns="http://maven.apache.org/POM/4.0.0" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 http://maven.apache.org/xsd/maven-4.0.0.xsd">
+    <parent>
+        <groupId>io.zipkin.brave</groupId>
+        <artifactId>brave-instrumentation-parent</artifactId>
+        <version>4.4.1-SNAPSHOT</version>
```
adriancole: 4.5.2-SNAPSHOT :)

instrumentation/kafka/pom.xml
```diff
@@ -0,0 +1,79 @@
...
+    <properties>
+        <main.basedir>${project.basedir}/../..</main.basedir>
+        <kafka.version>0.11.0.0</kafka.version>
+    </properties>
+
+    <dependencies>
+        <dependency>
+            <groupId>org.apache.kafka</groupId>
+            <artifactId>kafka-clients</artifactId>
+            <version>${kafka.version}</version>
```
adriancole: this should be a provided scope, so that we don't interfere with versions

instrumentation/kafka/src/main/java/brave/kafka/TracingProducer.java
ImFlog: Starting with topic only.

instrumentation/kafka/pom.xml
```diff
@@ -0,0 +1,82 @@
+<?xml version="1.0"?>
+<project xmlns="http://maven.apache.org/POM/4.0.0"
+    xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
+    xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 http://maven.apache.org/xsd/maven-4.0.0.xsd">
+  <parent>
+    <groupId>io.zipkin.brave</groupId>
+    <artifactId>brave-instrumentation-parent</artifactId>
+    <version>4.5.2-SNAPSHOT</version>
```
adriancole: 4.5.3-SNAPSHOT :P

instrumentation/kafka/pom.xml
```diff
@@ -0,0 +1,82 @@
...
+      <exclusions>
+        <exclusion>
+          <groupId>org.slf4j</groupId>
+          <artifactId>slf4j-log4j12</artifactId>
+        </exclusion>
+      </exclusions>
+    </dependency>
+  </dependencies>
+
+  <build>
```
adriancole: is this used?

instrumentation/kafka/src/main/java/brave/kafka/TracingConsumer.java
```diff
@@ -0,0 +1,187 @@
...
+  public TracingConsumer(Tracing tracing, Consumer<K, V> consumer) {
+    this.recordTracing = new RecordTracing(tracing);
+    this.wrappedConsumer = consumer;
+  }
+
+  @Override
+  public ConsumerRecords<K, V> poll(long timeout) {
+    ConsumerRecords<K, V> records = wrappedConsumer.poll(timeout);
+    for (ConsumerRecord<K, V> record : records) {
+      recordTracing.finishProducerSpan(record);
```
adriancole: this will need to change as we aren't sharing span IDs anymore

ImFlog:
Closing this in favor of https://github.com/openzipkin/brave/pull/467, will keep the TODOs from this one.

PR #465, opened 2017-08-04
Updates to Zipkin 1.29

adriancole:
This works in support of Messaging spans, though docs and examples will
be follow-up work.


adriancole:
cc @ImFlog

PR #470, opened 2017-08-16
Allows non-inheriting trace context

adriancole:
Eventhough convenience may mean an inheriting thread local is a good
choice, there are problems with this approach, such as the inability to
cleanup data. This encourages use of non-inheriting thread local w/o
changing the default.

fixes https://github.com/openzipkin/brave-opentracing/issues/48


adriancole:
@gnoremac @schlosna do you think it is fair to recommend against using inheritable thread locals (leaving the default inheritable since we used to)?

adriancole:
simplified the javadoc a bit, notably ridding the discussion about static references as the underlying problem is different tracer instances.

PR #471, opened 2017-08-17
Clears static references to Tracing after each test

adriancole:
A side-effect of `Tracing.newBuilder` is that it registers itself if no
other instance has been. This can create side-effects on tests that rely
on the current tracer, such as MySQL. This clears all tests that use
`Tracing.newBuilder` so they can't taint others.


adriancole:
the motivation for this, btw was #470.

Innocently changing from inheriting to non-inheriting thread local broke a test, but only in Maven. I thought it was my code, or something weird about thread local inheritance, but in reality it was just a non-clean test setup. This change may seem overkill, but I want to drive home how statics hurt.

PR #475, opened 2017-08-24
Adds pointers for unit testing

adriancole:
cc @llinder @devinsba @songxin1990 @imflog @sirtyro @anuraaga @takezoe

### DIFF
--- a/brave/src/main/java/brave/propagation/StrictCurrentTraceContext.java
+++ b/brave/src/main/java/brave/propagation/StrictCurrentTraceContext.java
@@ -1,9 +1,13 @@
-package brave.internal;
+package brave.propagation;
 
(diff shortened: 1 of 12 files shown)


brave/src/main/java/brave/propagation/StrictCurrentTraceContext.java
```diff
@@ -1,9 +1,13 @@
-package brave.internal;
-
-import brave.propagation.CurrentTraceContext;
-import brave.propagation.TraceContext;
-
-/** Useful when developing instrumentation as state is enforced more strictly */
+package brave.propagation;
```
adriancole: made this not internal as that makes it safe for other libraries to use.

PR #479, opened 2017-09-01
update p6spy version, important because https://github.com/p6spy/p6sp…

jorgheymans:
…y/pull/386 made that there was no way to exclude logging of binary content

### DIFF
--- a/instrumentation/p6spy/pom.xml
+++ b/instrumentation/p6spy/pom.xml
@@ -12,7 +12,7 @@
 
   <properties>
     <main.basedir>${project.basedir}/../..</main.basedir>
-    <p6spy.version>3.0.0</p6spy.version>
+    <p6spy.version>3.2.0</p6spy.version>
     <derby.version>10.13.1.1</derby.version>
   </properties>
 


jorgheymans:
welp sorry for the long PR title :-/

adriancole:
no code change needed etc right?

adriancole:
I think in some places we accidentally didn't put provided scope, which means we can override user-supplied versions... perhaps we should fix this?

jorgheymans:
I don't know the right approach to compile/provided scope for libraries. It's a conveniency / accidentally overriding user space trade-off ..

adriancole:
>
> I don't know the right approach to compile/provided scope for libraries.
> It's a conveniency / accidentally overriding user space trade-off ..
>
yeah on this side we've interfered before, which is why we set to provided.
can revisit, but probably still better to be consistent.

jorgheymans:
tend to agree for very common libs like guava / commons-* / jackson etc .. but very likely p6spy is not present in a project that was not already tracing so in this case it's probably ok to more lean towards convenience.

PR #480, opened 2017-09-01
Fixes places where we didn't use provided scope

adriancole:
Provided helps us not interfere with user-defined versions

### DIFF
--- a/instrumentation/p6spy/pom.xml
+++ b/instrumentation/p6spy/pom.xml
@@ -22,6 +22,7 @@
       <groupId>p6spy</groupId>
       <artifactId>p6spy</artifactId>
       <version>${p6spy.version}</version>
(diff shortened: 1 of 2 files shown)


adriancole:
cc @jorgheymans

jorgheymans:
:+1:

adriancole:
about your comment on p6spy.. 
> tend to agree for very common libs like guava / commons-* / jackson etc .. but very likely p6spy is not present in a project that was not already tracing so in this case it's probably ok to more lean towards convenience.

it actually makes sense. I can take out provided and put rationale instead (ex unlikeliness that users are already using it w/o tracing, so leave as compile scope) cc @felixbarny

felixbarny:
I also chose to use compile scope for p6spy in stagemonitor.

adriancole:
Cool will note and continue. Thx for the feedback!

PR #484, opened 2017-09-04
Adds propagation base test classes

adriancole:
This allows us to see implementation-specific behaviour more explicitly.
It also allows us to see wrinkles in apis as they apply to different
implementations (for example inability to inject a black context w/
Injector).


adriancole:
some typos left and want to add base tests for injector/extractor Otherwise a good start I think. Lemme know what you think

cc @jcchavezs @llinder @takezoe @sirtyro

adriancole:
injector extractor tests will be separate as otherwise gets too big

PR #488, opened 2017-09-11
[P6SPY] Provide a way to set the remote service name per …

hypnoce:
…jdbc connection url.

### DIFF
--- a/instrumentation/p6spy/src/main/java/brave/p6spy/TracingJdbcEventListener.java
+++ b/instrumentation/p6spy/src/main/java/brave/p6spy/TracingJdbcEventListener.java
@@ -8,11 +8,16 @@ import com.p6spy.engine.event.SimpleJdbcEventListener;
 import java.net.URI;
 import java.sql.Connection;
 import java.sql.SQLException;
+import java.util.regex.Matcher;
+import java.util.regex.Pattern;
 import zipkin.Constants;
 import zipkin.Endpoint;
 import zipkin.TraceKeys;
 
 final class TracingJdbcEventListener extends SimpleJdbcEventListener {
+
+  private final static Pattern URL_SERVICE_NAME_FINDER =
+      Pattern.compile("zipkinServiceName=(\\w*)");
 
   final String remoteServiceName;
   final boolean includeParameterValues;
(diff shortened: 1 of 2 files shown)


hypnoce:
Fixes # 487

instrumentation/p6spy/src/main/java/brave/p6spy/TracingJdbcEventListener.java
```diff
@@ -69,9 +74,17 @@
   void parseServerAddress(Connection connection, Span span) {
     try {
       URI url = URI.create(connection.getMetaData().getURL().substring(5)); // strip "jdbc:"
+      String actualServiceName = remoteServiceName;
```
adriancole: maybe rename the field to defaultRemoteServiceName

instrumentation/p6spy/src/main/java/brave/p6spy/TracingJdbcEventListener.java
```diff
@@ -69,9 +74,17 @@
...
+      final Matcher matcher = URL_SERVICE_NAME_FINDER.matcher(url.toString());
```
adriancole: one other thing.. in our mysql interceptor, it is this:

```java
      String remoteServiceName = connection.getProperties().getProperty("zipkinServiceName");
```

is there a way to avoid manual parsing?

hypnoce: I will check that.

hypnoce: Actually `getProperties` is part of the internal MySQL API and not jdbc.

adriancole:
final note is to use "zipkinServiceName" as the JDBC url property so that it matches the same for mysql (better than having 2 slightly different conventions)

hypnoce:
Btw, if there is any documentation to update else then the p6spy README, let me know.

hypnoce:
> LG. if you don't mind, please use https://github.com/square/java-code-styles to reformat
Done

adriancole:
great job!

PR #491, opened 2017-09-12
Avoids actual changes done by the Reporter when it is noop.

jcchavezs:
This PR adds a check before executing actual changes in the recorder when it is a noop.

Ping @adriancole

### DIFF
--- a/brave/src/main/java/brave/internal/recorder/Recorder.java
+++ b/brave/src/main/java/brave/internal/recorder/Recorder.java
@@ -40,29 +40,34 @@ public final class Recorder {
 
   /** @see brave.Span#start(long) */
   public void start(TraceContext context, long timestamp) {
+    if (noop.get()) return;
     spanMap.getOrCreate(context).start(timestamp);
   }
 
   /** @see brave.Span#name(String) */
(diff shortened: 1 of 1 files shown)


brave/src/main/java/brave/internal/recorder/Recorder.java
```diff
@@ -71,12 +76,14 @@ public void tag(TraceContext context, String key, String value) {
...
   /** @see brave.Span#remoteEndpoint(Endpoint) */
   public void remoteEndpoint(TraceContext context, Endpoint remoteEndpoint) {
+    if (noop.get()) return;
     if (remoteEndpoint == null) throw new NullPointerException("remoteEndpoint == null");
     spanMap.getOrCreate(context).remoteEndpoint(remoteEndpoint);
   }
 
   /** @see Span#finish() */
   public void finish(TraceContext context, long finishTimestamp) {
+    if (noop.get()) return;
```
adriancole: for these down, you don't want to early return as they remove data from the map (we want that data removed even in noop in case a span started before the tracing component was set to noop status)

adriancole:
thanks and congrats on first java PR evar!

PR #493, opened 2017-09-13
prevent npe when onAfter is getting called out of band

jorgheymans:
... also bumped version of p6spy to 3.3.0


adriancole:
This logic in centralized in #455 though was waiting on another to say worth it before merge.

for this change, only thing I might suggest is only read the thread local once:

```java
Tracer.SpanInScope scope = currentSpanInScope.get();
if (scope != null) {
  scope.close();
  currentSpanInScope.remove();
}
```

adriancole:
thanks again!

PR #503, opened 2017-09-22
Removes hard-dependency on zipkin v1 reporter

adriancole:
This adds detection such that Zipkin v1 types are not invoked at
runtime. However, there are some limitations to this.

For example, reflective configuration (like spring XML beans) needs no
Zipkin v1 dependency at all. However, compiled configuration (like Guice
or Spring AutoConfig) need an optional dependency on
`io.zipkin.java:zipkin` to affect the local or remote endpoint. This is
because these methods are overloaded and to compile an overloaded
method, you need to access all the types.

For example, if you don't have `io.zipkin.java:zipkin` in your classpath,
when you compile a call to `Tracing.Builder.localEndpoint`, you will
get an error like this from the compiler:

```
TracingConfiguration.java:[48,9] cannot access zipkin.Endpoint
  class file for zipkin.Endpoint not found
```

However, once compiled, if you are using `zipkin2.Endpoint`, you
don't actually need the v1 types at runtime.

In Brave 5, we should remove the zipkin v1 types from the api,
eliminating this problem. Until then, `io.zipkin.java:zipkin` is left
as a required dependency.


adriancole:
Ps I took the optional dep route. Another way is we keep compile deps (which mean users dont need to specify a v1 optional dep at compile time). This might be better as people can always exclude if they want to recover the couple hundred KiB of v1 jars.

cc @jorgheymans @llinder @devinsba  preferences?

adriancole:
ok found a sweet spot. this removes the hard dependency on zipkin reporter v1, but can't avoid the one on zipkin v1 due to method overloading. I think this is the best we can do for now (luckily zipkin artifacts even with old types share the same version)

PR #506, opened 2017-10-02
Removes expectation a setter can clear a key

adriancole:
I made a mistake documenting Propagation.Setter as being able to clear
a key. This is not guaranteed to work, for example you can't clear
headers made with URLConnection, only replace or add to them.

Instead, this documents that in reusable carriers, you must clear keys
on your own. If carriers are single-shot, there's no need to clear
anything first.


adriancole:
cc @jcchavezs @llinder I think I mentioned to you two that trying to set null was in hindsight a mistake. this is the clarification around it.

PR #507, opened 2017-10-03
Clarifies close instructions

adriancole:
The README on the core library doesn't say why close is needed. This
fixes that and other places where instructions were too loose.

Spring automatically detects closeable things, so removed explicitness
there.

### DIFF
--- a/spring-beans/README.md
+++ b/spring-beans/README.md
@@ -11,8 +11,7 @@ Bean Factories exist for the following types:
 
 Here are some example beans using the factories in this module:
 ```xml
-  <bean id="sender" class="zipkin.reporter2.okhttp3.OkHttpSender" factory-method="create"
-      destroy-method="close">
+  <bean id="sender" class="zipkin.reporter2.okhttp3.OkHttpSender" factory-method="create">
     <constructor-arg type="String" value="http://localhost:9411/api/v2/spans"/>
   </bean>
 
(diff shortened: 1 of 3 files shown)


adriancole:
cc @jorgheymans

spring-beans/README.md
```diff
@@ -11,8 +11,7 @@ Bean Factories exist for the following types:
 
 Here are some example beans using the factories in this module:
 ```xml
-  <bean id="sender" class="zipkin.reporter2.okhttp3.OkHttpSender" factory-method="create"
-      destroy-method="close">
+  <bean id="sender" class="zipkin.reporter2.okhttp3.OkHttpSender" factory-method="create">
```
adriancole: @shakuzen in spring, anything that's closeable automatically is closed anyway right? This includes factory beans, yeah? https://github.com/openzipkin/brave/blob/master/spring-beans/src/main/java/brave/spring/beans/TracingFactoryBean.java#L35

shakuzen: If it is a singleton Bean managed by Spring, it should be. Here are some notes on the auto-detection of the destroy method: https://docs.spring.io/spring-framework/docs/current/javadoc-api/org/springframework/context/annotation/Bean.html#destroyMethod--

I can't say I've ever done anything with FactoryBeans directly, but I would expect it to work the same.

adriancole: heh added a unit test. it works indeed!

PR #511, opened 2017-10-04
 document how to use the factory beans in javaconfig

jorgheymans:


spring-beans/README.md
```diff
@@ -34,3 +34,36 @@ Here are some example beans using the factories in this module:
     <property name="tracing" ref="tracing"/>
   </bean>
 ```
+The factory beans can also be used in java configuration:
```
adriancole: why wouldn't you just use normal beans?

jorgheymans: the normal beans will not be closed() automatically by Spring unless i misunderstood, to be honest i did not know about the factory beans until you mentioned them here https://github.com/openzipkin/brave/pull/507.

jorgheymans:
actually i was wrong, as documented here https://docs.spring.io/spring-framework/docs/current/javadoc-api/org/springframework/context/annotation/Bean.html#destroyMethod-- 
Spring will already call close() automatically if the method is present, i should have read in more detail.

> As a convenience to the user, the container will attempt to infer a destroy method against an object returned from the @Bean method. For example, given an @Bean method returning an Apache Commons DBCP BasicDataSource, the container will notice the close() method available on that object and automatically register it as the destroyMethod. **This 'destroy method inference' is currently limited to detecting only public, no-arg methods named 'close' or 'shutdown'.** The method may be declared at any level of the inheritance hierarchy and will be detected regardless of the return type of the @Bean method (i.e., detection occurs reflectively against the bean instance itself at creation time).

adriancole:
Sorry I may have diverted too much focus to the factory bean. Thx anyway!

PR #512, opened 2017-10-05
Allows extraction of just a trace ID (not span ID) from an incoming request

adriancole:
`Extractor.extract` returns a union type `TraceContextOrSamplingFlags` which
contains one of:
* `TraceContext` if trace and span IDs were present.
* `TraceIdContext` if a trace ID was present, but not span IDs.
* `SamplingFlags` if no identifiers were present

`TraceIdContext` is new to this change and allows you to work with pre-ordained trace IDs, such as are present in `x-amzn-trace-id` header when a request originates from an ELB.

### DIFF
--- a/instrumentation/kafka-clients/src/main/java/brave/kafka/clients/KafkaTracing.java
+++ b/instrumentation/kafka-clients/src/main/java/brave/kafka/clients/KafkaTracing.java
@@ -45,11 +45,13 @@ public final class KafkaTracing {
    * available.
    */
   public Span joinSpan(ConsumerRecord record) {
-    TraceContextOrSamplingFlags contextOrSamplingFlags = extractor.extract(record.headers());
-    if (contextOrSamplingFlags.context() != null) {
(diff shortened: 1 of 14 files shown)


adriancole:
Backfilled tests and documentation cc @jcchavezs

instrumentation/kafka-clients/src/main/java/brave/kafka/clients/KafkaTracing.java
```diff
@@ -1,64 +1,66 @@
...
-public final class KafkaTracing {
-
-  public static KafkaTracing create(Tracing tracing) {
-    return new KafkaTracing(tracing);
-  }
-
-  private final Tracing tracing;
-  private final TraceContext.Extractor<Headers> extractor;
-
-  KafkaTracing(Tracing tracing) { // hidden constructor
```
adriancole: this didn't match because the file had windows linefeeds which I corrected

jcchavezs:
:+1:

PR #514, opened 2017-10-13
Adds `TraceContext.extra` and supporting command `Tracer.nextSpan`

### LINKED ISSUE #390: More interoperable propagation via extra stuff in TraceContext
Right now, you cannot use wider trace context propagation formats than B3 because there's no way to carry that data from the incoming request to any outgoing requests. For example, AWS X-Ray includes a timestamp in their trace ID format. If we lost this, we break traces.

Also, many folks have asked in various ways to carry things like request ids so that they can be accessed in custom-code or tagged in log files.

I'm resisting the term baggage as change proposed are data structure only, and "baggage" itself tends to have more implications on api. I'm intentionally punting any api changes.

Here are some use cases that could be supported by adding `List<Object> TraceContext.extra()` and `List<Object> SamplingFlags.extra()` for when there is extra stuff, but not a trace yet. I think this is the only change needed to support this, as everything else is existing apis.

Making a custom propagation handler for Amazon X-Ray
```java
@Override public void inject(TraceContext traceContext, C carrier) {
  for (Object thing : traceContext.extra()) {
    if (thing instanceof XRayTraceContext) {
      inject(traceContext, (XRayTraceContext) thing, C carrier)
      break;
    }
  }
}

// dumbed down, but for example
void inject(TraceContext traceContext, XRayTraceContext xrayTraceContext, C carrier) {
  setter.put(carrier, "X-Amzn-Trace-Id", xrayTraceContext.root() + ";Parent=" + HexCodec.toLowerHex(traceContext.spanId()) + ";Sampled=1");
}
```

Adding type-safe data to a root span
```java
// let's pretend there's a standard "debug-id" that many use regardless of trace format!
for (Object thing : span.context().extra()) {
  if (thing instanceof DebugTraceId) {
    span.tag("debug-id", thing.toString());
    break;
  }
}
```

Making custom propagated stuff visible *without* requiring people to use brave's api
```java
// Assume Sony defines a header format for pre-defined tags, and a propagation handler
// User code wants to use PlayStationContext.currentGame() which is controlled by them
// The below facilitates propagation without pinning anyone to Brave's api. 
// This means they can switch impl to any other means and no problemo. 
// IOTW zero, like zero lock-in. No lock in to B3 header format.
// No lock in to a specific version of a tracer api. This is a choice not a requirement.

class PlayStationAwareCurrentTraceContext extends CurrentTraceContext {
  --snip--

  @Override public Scope newScope(TraceContext currentSpan) {
    Scope scope = delegate.newScope(currentSpan);
    Object current = findPlayStationContext(currentSpan.extra());
    if (current == null) return scope;
    final Object previous = PlayStationContext.save();
    PlayStationContext.set(current);
    return () -> {
      scope.close();
      PlayStationContext.set(previous);
    };
  }
}
```

Details..

For example, I can easily create a decorating propagation system. Ex new `CustomPropagationFactory(Propagation.B3)` where the custom one parses wider trace ids such as Amazon X-Ray. However, there's no place to store it, so when the CurrentTraceContext code executes, it would neither see nor be able to propagate that data. The impact of this is that we cannot continue a trace in such a system without breaking it.

This is a similar problem to what @nbmorgan brought up about customizing context to include application-specific keys. For example, if one wanted to make pony-currentUser, they would need to add interceptors before Brave's, duplicate the effort of propagating it in-process, finally duplicate the effort of attaching it to the outgoing requests. This could also be solved by the same tool, ex CustomPropagationFactory, if there was a place to store this custom data.

We may at some point be able to leverage or implement a [baggage](https://github.com/JonathanMace/tracingplane) handler, but minimally, we need to be able to extra "extra" stuff into a read-only object, then allow injection of that stuff on the other side.

The api would need to be on TraceContext itself, and it is likely safest to use an Object as opposed to a map to describe that custom "stuff". That's because it may take the form of a data structure or binary content, which is not a Map. For example baggage buffers, or type-safe user data.

This discussion is further elaborated here, though I think the salient issue for Brave is being able to interop with other trace context formats. The fact that this also enables a limited use case of arbitrary propagation is a bonus  https://docs.google.com/document/d/1rq-oCR8gNOi2-G0GFhQJZgSZCmKkxL8jbMZCHhyj268/edit

adriancole:
This adds the ability to carry extra data from one side of the process
to the other. It is available as `TraceContext.extra`, intentionally
opaque as it is a `Propagation` implementation detail.

For example, a propagation implementation could push extra fields
alongside the trace context, and optionally inject it into outgoing
requests.
```java
ExtraFieldPropagation.newFactory(Propagation.Factory.B3, "user-id");
```

Such a plugin could expose or integrate with other code to make the
extra data visible, given a trace context.

Using the example above, the following is possible using
`Tracing.currentTraceContext()`:
```java
userId = ExtraFieldPropagation.get("user-id");
```

To support this, and to ensure this "extra" data isn't lost, a new
operation is added `Tracer.nextSpan()` which takes the result of
`TraceContext.Extractor.extract()`. Along the way, existing code has
been ported to use it (incidentally fixing bugs not caught before).

A future change can integrate this "extra" with `AWSPropagation` or
expose `ExtraFieldPropagation` if desirable.

Fixes #390


brave/src/main/java/brave/Tracer.java
```diff
@@ -192,10 +198,42 @@ public final Span joinSpan(TraceContext context) {
...
 
   /**
-   * Like {@link #newChild(TraceContext)}, where the trace ID is present, but not a span ID.
+   * Explicitly creates a child within an existing. The result will be have its parent ID set to the
+   * input's span ID.
+   *
+   * <p>To implicitly create a new trace, or a span within an existing one, use {@link
+   * #nextSpan()}.
    */
-  public Span newTrace(TraceIdContext traceIdContext) {
```
adriancole: this api wasn't released so safe to trade

instrumentation/kafka-clients/src/main/java/brave/kafka/clients/KafkaTracing.java
```diff
@@ -43,16 +43,28 @@ public static KafkaTracing create(Tracing tracing) {
   /**
    * Retrieve the span extracted from the record headers. Creates a root span if the context is not
    * available.
+   *
+   * @deprecated this results in appending to a span already complete. Please use {@link
+   * #nextSpan(ConsumerRecord)}
    */
+  @Deprecated
   public Span joinSpan(ConsumerRecord record) {
```
adriancole: @ImFlog found a bug.. we shouldn't encourage this (better remove or repurpose) as this will lead to writing to a finished span. I also noticed we could probably make a single span for the batch consumption (which could lead to significantly less load) cc @jonathan-lo

instrumentation/grpc/src/test/java/brave/grpc/ITTracingServerInterceptor.java
```diff
@@ -114,6 +114,7 @@ public void start(Listener<RespT> responseListener, Metadata headers) {
             headers.put(Key.of("X-B3-TraceId", ASCII_STRING_MARSHALLER), traceId);
             headers.put(Key.of("X-B3-ParentSpanId", ASCII_STRING_MARSHALLER), parentId);
             headers.put(Key.of("X-B3-SpanId", ASCII_STRING_MARSHALLER), spanId);
+            headers.put(Key.of("X-B3-Sampled", ASCII_STRING_MARSHALLER), "1");
```
adriancole: bug! "join" tests used to pass without this

adriancole:
here's a working example of extra propagation https://github.com/openzipkin/brave/pull/515

instrumentation/kafka-clients/src/main/java/brave/kafka/clients/KafkaTracing.java
ImFlog: Aren't we heading to the first problem i.e. how do we know when we start the record treatment ?

adriancole: so in messaging span, we have something defined like this:
[producer span], [consumer span], [processor span]

I think we (me) accidentally missed reviewing this impl, that "joinSpan" is too late. It would do something like this..
[producer span], [consumer span{end}       {late data}]

So the first problem is to not expose something that will always mutate a completed span when a trace is in progress, nextSpan accomplishes this by creating a separate span for this activity.

The other thing I noticed was that the "consumer span" could be no problem if used for the bulk task, as it would be the single parent of each processor span.
Ex  after we fix the joinSpan -> nextSpan, for a 2 message batch we have
```
[producer span] -> [consumer span] -> [processor span]
                   [consumer span] -> [processor span]
```
Seems we could make this more efficient by removing the message key from the consumer span (processor could add that), and then re-use one consumer span for the batch, like this.

```
[producer span] -> [consumer span] -> [processor span]
                                  \-> [processor span]
```

If there are problems with this, I currently don't see them. Can you reply if there are (and reply generally as you are key in this decision!)

adriancole:
I will remove the change to kafka, and put it in a different pull request

instrumentation/kafka-clients/src/main/java/brave/kafka/clients/KafkaTracing.java
adriancole: moved discussion here

https://github.com/openzipkin/brave/pull/518

adriancole:
one problem I noticed is this doesn't support "extra stuff" encoded in prefix headers. stand-by

adriancole:
prefix headers are too complicated and we've not been asked to support them. Punting

PR #515, opened 2017-10-13
Adds ExtraFieldPropagation

adriancole:
Sometimes you need to propagate extra fields, such as a request ID or an alternate trace context.
For example, if you are in a Cloud Foundry environment, you might want to pass the request ID:

```java
// when you initialize the builder, define the extra field you want to propagate
tracingBuilder.propagationFactory(
  ExtraFieldPropagation.newFactory(B3Propagation.FACTORY, "x-vcap-request-id")
);

// later, you can tag that request ID or use it in log correlation
requestId = ExtraFieldPropagation.current("x-vcap-request-id");
```

You may also need to propagate a trace context you aren't using. For example, you may be in an
Amazon Web Services environment, but not reporting data to X-Ray. To ensure X-Ray can co-exist
correctly, pass-through its tracing header like so.

```java
tracingBuilder.propagationFactory(
  ExtraFieldPropagation.newFactory(B3Propagation.FACTORY, "x-amzn-trace-id")
);
```

See #390

### DIFF
--- a/instrumentation/http-tests/src/main/java/brave/http/ITHttp.java
+++ b/instrumentation/http-tests/src/main/java/brave/http/ITHttp.java
@@ -2,6 +2,8 @@ package brave.http;
 
 import brave.Tracing;
 import brave.propagation.CurrentTraceContext;
+import brave.propagation.ExtraFieldPropagation;
+import brave.propagation.Propagation;
 import brave.propagation.StrictCurrentTraceContext;
 import brave.propagation.TraceContext;
 import brave.sampler.Sampler;
@@ -41,6 +43,7 @@ public abstract class ITHttp {
           }
           spans.add(s);
         })
+        .propagationFactory(ExtraFieldPropagation.newFactory(Propagation.Factory.B3, "user-id"))
         .currentTraceContext(currentTraceContext)
         .sampler(sampler);
   }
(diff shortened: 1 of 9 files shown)


instrumentation/http-tests/src/main/java/brave/http/ITHttp.java
```diff
@@ -41,6 +43,7 @@
           }
           spans.add(s);
         })
+        .propagationFactory(ExtraFieldPropagation.newFactory(Propagation.Factory.B3, "user-id"))
```
adriancole: so this means do B3 and also if you see "user-id" pass it along and onto outgoing headers

instrumentation/http-tests/src/main/java/brave/http/ITServlet25Container.java
```diff
@@ -72,11 +73,12 @@ public void doFilter(ServletRequest request, ServletResponse response, FilterCha
   @Test public void currentSpanVisibleToOtherFilters() throws Exception {
     String path = "/foo";
 
-    Request request = new Request.Builder().url(url(path)).build();
+    Request request = new Request.Builder().url(url(path))
```
adriancole: example test showing this works

jcchavezs:
LGTM

jcchavezs:
:+1:

ajishashah:
Can i set extra field from my controller ?  and get it downstream?

adriancole:
please use gitter for q/a as opposed to adding questions to old issues.
adding questions spams people watching this repository.

if looking for help with sleuth please ask here
https://gitter.im/spring-cloud/spring-cloud-sleuth

PR #518, opened 2017-10-14
Deprecates KafkaTracing.joinSpan for nextSpan

adriancole:
`KafkaTracing.joinSpan` will always mutate a completed span when a trace
is in progress. `KafkaTracing.nextSpan` accomplishes this by creating a
separate span for this activity.

The problem with `KafkaTracing.joinSpan` is that it returns a span that
is finished. Any tags or annotations will be delayed until Brave flushes
them:

```
[producer span], [consumer span{end} {late data}]
```

`KafkaTracing.nextSpan` instead returns a child which can be started and
stopped without tainting its parent:

```
[producer span], [consumer span], [processor span]
```

### DIFF
--- a/instrumentation/kafka-clients/src/main/java/brave/kafka/clients/TracingConsumer.java
+++ b/instrumentation/kafka-clients/src/main/java/brave/kafka/clients/TracingConsumer.java
@@ -6,6 +6,7 @@ import brave.propagation.TraceContext.Extractor;
 import brave.propagation.TraceContext.Injector;
 import brave.propagation.TraceContextOrSamplingFlags;
 import java.util.Collection;
+import java.util.LinkedHashMap;
 import java.util.List;
 import java.util.Map;
 import java.util.Set;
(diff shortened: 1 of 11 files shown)


adriancole:
The other thing I noticed was that the "consumer span" could be no problem if used for the bulk task, as it would be the single parent of each processor span.
Ex  after we fix the joinSpan -> nextSpan, for a 2 message batch we have
```
[producer span] -> [consumer span] -> [processor span]
                   [consumer span] -> [processor span]
```
Seems we could make this more efficient by removing the message key from the consumer span (processor could add that), and then re-use one consumer span for the batch, like this.

```
[producer span] -> [consumer span] -> [processor span]
                                  \-> [processor span]
```

If there are problems with this, I currently don't see them. cc @ImFlog @jonathan-lo

ImFlog:
I don't see a problem either. LGTM

adriancole:
@ImFlog cool. I am going to go ahead and implement the batch consumer span thing, too.

adriancole:
PS we can't use one span for the batch of records, unless there's no propagated in any of the records! I'm refactoring on account of this glitch

brave/src/main/java/brave/Tracer.java
```diff
@@ -224,16 +227,45 @@ public Span newChild(TraceContext parent) {
    * extracted span IDs. This means the extracted context (if any) is the parent of the span
    * returned.
    *
+   * <p><em>Note:</em> If a context could be extracted from the input, that trace is resumed, not
```
adriancole: cc @jcchavezs subtle, but important. When you are calling something like kafka.poll, you might already be in a trace. Currently, the zipkin model does not have support for multiple parents (or links as defined in census). The impact of this is that when you can extract a span from headers, it won't be allocated to the current trace (as the extracted one is in a different trace).

In many cases, there is no impact as tools like message processors are often entrypoints. However, it is interesting to keep in mind in the future. In census, you can establish a trace+span link from one trace to another, though I've not seen it used in practice. One day, we might revisit this, but not today!

instrumentation/kafka-clients/src/main/java/brave/kafka/clients/TracingConsumer.java
```diff
@@ -43,38 +43,48 @@
...
-      if (record.key() != null) {
-        span.tag(KafkaTags.KAFKA_KEY_TAG, record.key().toString());
-      }
-      span.tag(KafkaTags.KAFKA_TOPIC_TAG, record.topic());
+  static void addTagsAndStart(Span consumerSpanForTopicPartition, TopicPartition partition) {
+    if (!consumerSpanForTopicPartition.isNoop()) {
+      consumerSpanForTopicPartition
+          .kind(Span.Kind.CONSUMER)
+          .tag(KafkaTags.KAFKA_TOPIC_TAG, partition.topic())
+          .tag(KafkaTags.KAFKA_PARTITION_TAG, String.valueOf(partition.partition()))
```
ImFlog: No more KEY_TAG ?
The partition is just an ID, It could be useful but the key still makes sense IMHO.

adriancole:
No more KEY_TAG ?

The partition is just an ID, It could be useful but the key still makes
sense IMHO

Good questions. First, topic and parition are always good tags on consumer
span.
* they serialize naturally
* you can search topic=foo and partition=1

That said, fine to leave out the partition if you dont think people will
want to refine by it or know which partition the consumer was assigned to
at the time..

Things on key..

first I noticed the key isnt great without looking at other things
* it is nullable (read NPE though we can guard on that)
* people encode junk into it sometimes, for example base64 trace
identifiers (previously used this!)
* more about last, but it is binary.. though we could look at serializer
and add if ascii marshaller

Whichever spans we add the key, we have to bear this in mind. Ex
* check null
* check marshaller is ascii

Say we add the key, we have 3 choices, and in a complete trace remember
that having the key in any one of these makes it searchable!

* producer span
  * good because it is a single message
  * if you add here you can still search when consumer isnt traced.

* consumer span
  * complicated because it is a bulk task
    * to reuse a consumer span for NNN messages means you don't add a key
(record cardinality)
    * for example concatenating keys doesnt work with search and could
become huge
  * we could defensively check if only one record, but that is erratic
  * we can definitely add key when we had an inbound context, but if we
do.. that probably implies the key was already added by the producer!

* processor span (kafka.nextSpan(record))
  * safe as single record
  * adding key here makes it searchable even when the producer isnt traced.


So.. my advice is to defensively (check null ascii) default to adding the
key on producer span always and processor span only when there is no
incoming context. Skip adding on the consumer phase as it is complicated
due to bulk and if either producer or nextSpan is called you can still
query. I can remove the partition until someone asks for it.

I will work in above and later we can add a dynamic policy. Sound good?

Lemme know any more feedback! Thanks for thinking this through.

adriancole:
ok updated and did a big README update too

ImFlog:
LGTM

PR #531, opened 2017-10-31
add null checks before putting values to MDC

skastenholz:
Since JBoss Logging 3 does not support putting null values to MDC and it is not required by the SLF4J API I added simple null checks. As a result Brave SLF4J context can be used in conjuction with JBoss Logging 3.


adriancole:
I think in other code we are conditionally calling "remove" should we be doing that here (instead of setting null?)

skastenholz:
Thank you for your remark. I incorporated your request. Also brave.context.slf4j.MDCCurrentTraceContextTest runs successfully now.

adriancole:
thanks!

PR #533, opened 2017-11-05
Adds TracingCachingHttpClientBuilder for httpclient-cache

### LINKED ISSUE #431: Instrumenting Apache HttpClients built with CachingHttpClientBuilder
Since 4.3 its not possible to instrument Apache HttpClients built with the `CachingHttpClientBuilder` since the only way to created instrumented clients is using the `TracingHttpClientBuilder`.

It would be good if there was a way to apply instrumentation as interceptors that could be applied to any builder rather than requiring the use of the tracing builder.

adriancole:
This refactors types so that we can introduce identical functionality
from TracingHttpClientBuilder for TracingCachingHttpClientBuilder.

Fixes #431


adriancole:
sorry @sgoo fell off radar!

PR #539, opened 2017-11-23
Fixes some state bugs in our servlet filter thanks Nic Munroe

adriancole:
@nicmunroe pointed out some problems in our filter and how to test for
them https://github.com/Nike-Inc/wingtips/pull/49#issuecomment-336529316


nicmunroe:
Looks good!

PR #540, opened 2017-11-25
Makes Tracer's toString include info helpful in troubleshooting

adriancole:
Problems can occur when state isn't cleaned up. This adds minimal info
to determine if there's leaked state or not. This should help debugging
problems including circleci.

### DIFF
--- a/archive/brave-core/src/test/java/brave/interop/TracerAdapterTest.java
+++ b/archive/brave-core/src/test/java/brave/interop/TracerAdapterTest.java
@@ -13,6 +13,7 @@ import org.junit.After;
 import org.junit.Test;
 import zipkin.Constants;
 import zipkin2.Annotation;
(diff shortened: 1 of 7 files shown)


adriancole:
Ex we get test failures like this pretty routinely. hoping the message after this change is more helpful cc @shakuzen

```
Your build ran 769 tests with 1 failure
startWithCurrentLocalSpanAndFinishWithTracer - brave.interop.TracerAdapterTest
lessjava.lang.AssertionError: 

Expecting actual not to be empty
	at brave.interop.TracerAdapterTest.checkLocalSpanReportedToZipkin(TracerAdapterTest.java:136)
	at brave.interop.TracerAdapterTest.startWithCurrentLocalSpanAndFinishWithTracer(TracerAdapterTest.java:58)
```

PR #543, opened 2017-11-26
Adds more output to debug flakey test

adriancole:
builds on #540 as the test is flakey but can't quite tell why yet


archive/brave-core/src/test/java/brave/interop/TracerAdapterTest.java
```diff
@@ -40,7 +44,11 @@
       .tracer();
   Brave brave3 = TracerAdapter.newBrave(brave4);
 
-  @After public void close(){
+  @Before public void clearBrave3State() {
+    ThreadLocalServerClientAndLocalSpanState.clear();
```
adriancole: wild guess is this was the source of flakes. In the brave3 model you could leak contexts because there was no try/finally mechanism.

Issue #516, opened 2017-10-13
the currentTimeMicroseconds of DefaultClock is slightly inaccurate


cgfork:
The currentTimeMicroseconds of DefaultClock or Platform is not equal to the System.currentTimeMillis(). The delta > 1000us. System.currentTimeMillis() is the system time.  System.nanoTime() is the time of running Java Virtual Machine's high-resolution time. When the system time is changed (by NTP),  System.nanoTime() is not changed. So the currentTimeMicroseconds is inaccurate.  I think it is necessary to reset the start time when the system time changes.

```java
classs DefaultClock implements Clock {
        final Ticker ticker;

        DefaultClock() {
            ticker = new Ticker();
        }

        @Override
        public long currentTimeMicroseconds() {
            return ticker.nowMicroMillis();
        }
    }

class Ticker {

        private final AtomicReference<Marker> ref;

        Ticker() {
            ref = new AtomicReference<>(new Marker());
        }

        public long nowMicroMillis() {
            for (int i = 0; i < 3; i++) {
                Marker marker = ref.get();
                Marker currentMarker = new Marker();

                long microMillis = currentMarker.currentMicroMills(marker);
                if (Math.abs(currentMarker.startTs - microMillis) < 1000) {
                    return microMillis;
                }

                if (ref.compareAndSet(marker, currentMarker)) {
                    return (System.nanoTime() - currentMarker.startTick) / 1000L + currentMarker.startTs;
                }
            }

            return new Marker().startTs;
        }
    }

class Marker {
        final long startTs;
        final long startTick;

        Marker() {
            this.startTs = System.currentTimeMillis() * 1000;
            this.startTick = System.nanoTime();
        }

        long currentMicroMills(Marker old) {
            return (startTick - old.startTick) / 1000L + old.startTs;
        }
    }
```

adriancole:
I think @llinder noticed this too at some point...

adriancole:
note that in java 9 we can get an accurate clock by default, avoiding this (though there's still risk of the clock changing between start and finish):

```java
    return Tracing.newBuilder()
        .clock(() -> {
          Instant instant = java.time.Clock.systemUTC().instant();
          return SECONDS.toMicros(instant.getEpochSecond()) + NANOSECONDS.toMicros(instant.getNano());
        })
```

adriancole:
cc also @shakuzen @jorgheymans

llinder:
[...]

PR #547, opened 2017-11-29
Read system time once per trace, and use tick offsets thereafter

adriancole:
This moves logic such that clock readings occur once per trace. In doing
so, we get clock consistency throughout the trace, without risk of 
becoming out-of-sync with NTP updates.

An expiring approach was considered for sharing of timestamps outside the scope of a trace. This intended to reduce the overhead of reading current time, yet allowing NTP updates to be visible. However, benchmarks showed the infrastructure needed to expire a timestamp under contention was more costly than reading time. Hence, this only locks time per-trace.

See #516

cc @jorgheymans @llinder


jorgheymans:
LGTM, with this i longer have cs,sr,ss,cr out of order in the UI :+1:

jorgheymans:
:+1:

adriancole:
cutting this as 4.11

PR #550, opened 2017-12-07
Makes end-to-end benchmarks more realistic

adriancole:
Chat with @wu-sheng made me realize our end-to-end benchmarks were not
that realistic, on account of using a noop reporter. Most people use
`AsyncReporter` and eventhough that code lives outside Brave, it is a
very important factor in latency which should be included.

For example, here's with 1KiB message size (to avoid GC thrash in JMH)
```
Benchmark                                           Mode  Cnt    Score    Error  Units
EndToEndBenchmarks.server_get                       avgt   15  159.013 ±  9.592  us/op
EndToEndBenchmarks.traced128Server_get              avgt   15  223.974 ± 14.517  us/op
EndToEndBenchmarks.traced128Server_get_resumeTrace  avgt   15  226.344 ±  9.706  us/op
EndToEndBenchmarks.tracedServer_get                 avgt   15  220.218 ±  6.028  us/op
EndToEndBenchmarks.tracedServer_get_resumeTrace     avgt   15  218.240 ±  2.942  us/op
EndToEndBenchmarks.tracedawsServer_get              avgt   15  218.895 ±  6.939  us/op
EndToEndBenchmarks.tracedawsServer_get_resumeTrace  avgt   15  219.882 ± 12.913  us/op
EndToEndBenchmarks.unsampledServer_get              avgt   15  166.991 ±  2.342  us/op
```

Overhead without the async reporter is usually up to 30us. The above
shows significant opportunity to reduce overhead there. FWIW, it is
mainly due to encoding spans on the calling thread. There are other
ways to accomplish the goal of memory-bounding now that the data
structures are simpler (in zipkin v2).

### DIFF
--- a/instrumentation/benchmarks/src/main/java/brave/EndToEndBenchmarks.java
+++ b/instrumentation/benchmarks/src/main/java/brave/EndToEndBenchmarks.java
@@ -27,6 +27,7 @@ import org.openjdk.jmh.runner.Runner;
 import org.openjdk.jmh.runner.RunnerException;
 import org.openjdk.jmh.runner.options.Options;
 import org.openjdk.jmh.runner.options.OptionsBuilder;
+import zipkin2.reporter.AsyncReporter;
 import zipkin2.reporter.Reporter;
 
 import static javax.servlet.DispatcherType.REQUEST;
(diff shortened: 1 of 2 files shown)


adriancole:
https://github.com/openzipkin/zipkin-reporter-java/issues/81 for the follow-up work

PR #554, opened 2017-12-13
Corrects missing status code in servlet 2.5 environments

adriancole:
Before, we couldn't detect the status code unless we wrapped it first.
This made layered tools on servlet api not work unless they could affect
the servlet response. For example, Spring MVC's interceptor cannot.

Benchmarking was used to show that caching the reflective lookup saves
an order of magnitude of overhead than doing so otherwise.

### DIFF
--- a/instrumentation/servlet/src/main/java/brave/servlet/ServletRuntime.java
+++ b/instrumentation/servlet/src/main/java/brave/servlet/ServletRuntime.java
@@ -4,6 +4,10 @@ import brave.Span;
 import brave.http.HttpServerHandler;
 import brave.internal.Nullable;
 import java.io.IOException;
+import java.lang.reflect.Method;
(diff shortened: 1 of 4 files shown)


instrumentation/servlet/src/main/java/brave/servlet/ServletRuntime.java
```diff
@@ -4,13 +4,18 @@
 import brave.http.HttpServerHandler;
 import brave.internal.Nullable;
 import java.io.IOException;
+import java.lang.reflect.Method;
+import java.util.LinkedHashMap;
```
adriancole: FYI in case someone is curious.. I tested perf against identityhashmap and a custom size-bounded map. no difference in throughput. chose LinkedHashMap as it almost always the best choice anyway

PR #558, opened 2017-12-20
Adds zipkin2 support to Brave 3 api and removes internal package sharing

adriancole:
This removes the need to share internal packages between the brave 3 and
4 apis and zipkin. In doing so, it unlocks some OSGi problems.

This also makes it easier to transition to zipkin2 by allowing reporter2
as a Brave.Builder.spanReporter arg.

See https://github.com/apache/camel/pull/2151


adriancole:
verified with camel

PR #559, opened 2017-12-20
Removes global osgi setup for more precise config, removes more internal

adriancole:
See #588 which didn't quite work


archive/brave-core/bnd.bnd
```diff
@@ -1,3 +1,9 @@
+Import-Package: \
+  !zipkin,\
```
adriancole: this is wrong bc brave-core still has a hard dep on zipkin package

PR #572, opened 2018-01-08
WIP Spring Messaging impl

codefromthecrypt:
This is inspired by Spring Cloud Sleuth, but uses the messaging spans
from the v2 model. There's a certain amount of work left, just raising
early.

### DIFF
--- a/instrumentation/spring-messaging/README.md
+++ b/instrumentation/spring-messaging/README.md
@@ -1,0 +1,20 @@
+# brave-instrumentation-spring-messaging
+This module contains a tracing interceptor for [Spring Messaging](https://docs.spring.io/spring-boot/docs/current/reference/html/boot-features-messaging.html).
+`TracingChannelInterceptor` completes a producer span per message, propagating it via headers. When
+consuming, a child span is created from trace in headers if present.
+
+## Configuration
+
+Tracing always needs a bean of type `Tracing` configured. Make sure
+it is in place before proceeding. Here's an example in [XML](../../spring-beans/README.md).
+
+Then, wire `TracingChannelInterceptor` and add it via `InterceptableChannel.addInterceptor`, or
+something that does the same thing like below:
+
+```java
+@Bean
+@GlobalChannelInterceptor // handy to apply to all channels if you depend on spring-integration-core
+public ChannelInterceptor tracingChannelInterceptor(Tracing tracing) {
+  return TracingChannelInterceptor.create(tracing);
+}
+```
--- a/instrumentation/spring-messaging/pom.xml
+++ b/instrumentation/spring-messaging/pom.xml
@@ -1,0 +1,72 @@
+<?xml version="1.0"?>
+<project xmlns="http://maven.apache.org/POM/4.0.0"
+    xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
+    xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 http://maven.apache.org/xsd/maven-4.0.0.xsd">
+  <parent>
+    <groupId>io.zipkin.brave</groupId>
+    <artifactId>brave-instrumentation-parent</artifactId>
+    <version>5.6.2-SNAPSHOT</version>
+  </parent>
+  <modelVersion>4.0.0</modelVersion>
+
+  <artifactId>brave-instrumentation-spring-messaging</artifactId>
+  <name>Brave Instrumentation: Spring Messaging</name>
+
+  <properties>
+    <main.basedir>${project.basedir}/../..</main.basedir>
+    <main.java.version>1.6</main.java.version>
+    <main.signature.artifact>java16</main.signature.artifact>
+    <spring-boot.version>1.5.19.RELEASE</spring-boot.version>
+  </properties>
+
+  <dependencies>
+    <dependency>
+      <groupId>org.springframework</groupId>
+      <artifactId>spring-messaging</artifactId>
+      <version>${spring4.version}</version>
+      <scope>provided</scope>
+    </dependency>
+    <dependency>
+      <groupId>org.springframework</groupId>
+      <artifactId>spring-core</artifactId>
+      <version>${spring4.version}</version>
+      <scope>provided</scope>
+    </dependency>
+
+    <dependency>
+      <groupId>org.springframework.integration</groupId>
+      <artifactId>spring-integration-core</artifactId>
+      <!-- look at the version corresponding to spring boot used or else test classpath will be sad -->
+      <!-- http://search.maven.org/remotecontent?filepath=org/springframework/boot/spring-boot-dependencies/1.5.19.RELEASE/spring-boot-dependencies-1.5.19.RELEASE.pom -->
+      <version>4.3.19.RELEASE</version>
+      <scope>provided</scope>
+    </dependency>
+
+    <dependency>
+      <groupId>org.springframework</groupId>
+      <artifactId>spring-context</artifactId>
+      <version>${spring4.version}</version>
+      <scope>test</scope>
+    </dependency>
+
+    <dependency>
+      <groupId>${project.groupId}</groupId>
+      <artifactId>brave-tests</artifactId>
+      <scope>test</scope>
+    </dependency>
+
+    <dependency>
+      <groupId>org.springframework.boot</groupId>
+      <artifactId>spring-boot-starter-test</artifactId>
+      <version>${spring-boot.version}</version>
+      <scope>test</scope>
+    </dependency>
+
+    <dependency>
+      <groupId>org.springframework</groupId>
+      <artifactId>spring-test</artifactId>
+      <version>${spring4.version}</version>
+      <scope>test</scope>
+    </dependency>
+  </dependencies>
+</project>
--- a/instrumentation/spring-messaging/src/main/java/brave/spring/messaging/TracingChannelInterceptor.java
+++ b/instrumentation/spring-messaging/src/main/java/brave/spring/messaging/TracingChannelInterceptor.java
@@ -1,0 +1,157 @@
+package brave.spring.messaging;
+
+import brave.Span;
+import brave.SpanCustomizer;
+import brave.Tracing;
+import brave.propagation.ThreadLocalSpan;
+import brave.propagation.TraceContext;
+import brave.propagation.TraceContextOrSamplingFlags;
+import org.springframework.beans.factory.annotation.Autowired;
+import org.springframework.messaging.Message;
+import org.springframework.messaging.MessageChannel;
+import org.springframework.messaging.MessageHandler;
+import org.springframework.messaging.support.ChannelInterceptorAdapter;
+import org.springframework.messaging.support.ExecutorChannelInterceptor;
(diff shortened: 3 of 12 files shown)


adriancole:
cc @jonathan-lo and @marcingrzejszczak

instrumentation/spring-messaging/pom.xml
```diff
@@ -0,0 +1,49 @@
...
+      <scope>provided</scope>
+    </dependency>
+
+    <dependency>
+      <groupId>${project.groupId}</groupId>
+      <artifactId>brave-tests</artifactId>
+      <scope>test</scope>
+    </dependency>
+
+    <dependency>
```
adriancole: note: not used yet

marcingrzejszczak:
BTW - https://github.com/spring-cloud/spring-cloud-sleuth/issues/821

jonathan-lo:
looking good. is end goal for sleuth to use this instrumentation point?

adriancole:
thanks, yeah this will (ideally) obviate the same code in sleuth, except
using the style we are using here for kafka. Plus being this will be
consistent across spring and spring boot variants.

Sleuth will focus on configuring this per "Oh Thank Heaven for 7-Eleven"
https://github.com/spring-cloud/spring-cloud-sleuth/issues/711

adriancole:
pretty much done here. just will make some executor based tests and also make a integration one that uses normal kafka on one side cc @shakuzen @ImFlog

adriancole:
@lchayoun can you please provide a test case relating to https://github.com/spring-cloud/spring-cloud-sleuth/issues/821? meanwhile I will try to guess

instrumentation/spring-messaging/src/main/java/brave/spring/messaging/TracingChannelInterceptor.java
```diff
@@ -0,0 +1,113 @@
...
+  @Override
+  public void afterSendCompletion(Message<?> message, MessageChannel channel, boolean sent,
+      Exception ex) {
+    finishSpan(Span.Kind.PRODUCER, ex);
+  }
+
+  @Override public Message<?> postReceive(Message<?> message, MessageChannel channel) {
+    MessageHeaderAccessor headers = MessageHeaderAccessor.getMutableAccessor(message);
+    startConsumerSpan(headers);
+    return new GenericMessage<>(message.getPayload(), headers.getMessageHeaders());
```
artembilan: I think this really fixes the use-case for the `QueueChannel`  from Spring Integration.
Although I would also use `setLeaveMutable(true)` to allow downstream flow to reuse our *mutable* message.

I guess the concern from the `DirectChannel` perspective cannot be fixed here and we even might not need to worry about it. See https://github.com/spring-cloud/spring-cloud-sleuth/issues/821.

The problem there that a `MessageHandler` is called directly from the `send()` call stack. So, `MessageHandler.handle()` is in between `preSend()` and `afterSendCompletion()`, therefore it is still the part of the `PRODUCER` span. If that is really fine from the tracing perspective and there is no reason to fine-grain it for the `CONSUMER` span around this `MH.handle()` just because we still in the same thread and, yeah, in that case the producer span will be closed after that internal consumer span, so, we might just need to document that is it an expected behavior do not have separate span for the `MH` when we use `DirectChannel`. 

On the other hand I see the situation like: `DirectChannel -> RequestReplyMessageHandler -> DirectChannel` and so on. In this case we have several `preSend()` and `afterSendCompletion()` in the same thread. Therefore we start several producer spans, where the next one will be as a child of the previous. That will looks as a regular callstack. Having this assumption I would say that it might be good idea to really  fine-grain it for that child consumer span as well...

WDYT?

lchayoun:
@adriancole The test case is to use the `Sink` interface, currently it is not working with "SubscribableChannel"

instrumentation/spring-messaging/src/main/java/brave/spring/messaging/TracingChannelInterceptor.java
adriancole: right now, there are separate thread locals for producer and consumer, but if indeed there are several layers of nesting in-process, this would break. This is one of the tricky parts which relies on callbacks to be unwound the same way they were wound (ex a after for every pre). Stack semantics could work if this is the case

instrumentation/spring-messaging/README.md
```diff
@@ -0,0 +1,20 @@
+# brave-instrumentation-spring-messaging
+This module contains a tracing interceptor for [Spring Messaging](https://spring.io/guides/gs/messaging-rabbitmq/).
```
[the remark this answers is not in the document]
adriancole: link is indeed imprecise good catch. best link suggestion?

instrumentation/spring-messaging/src/main/java/brave/spring/messaging/TracingChannelInterceptor.java
artembilan: Well, it doesn't because as I said: the `MH` in `DirectChannel` is called in between `preSend` and `afterSendComplete`:
```
message = interceptors.preSend(message, this, interceptorStack);
...
sent = this.doSend(message, timeout);
...
interceptors.postSend(message, this, sent);
interceptors.afterSendCompletion(message, this, sent, null, interceptorStack);
```

As you see there is on `beforeHandle` and `afterMessageHandled` because such a `MessageChannel` is not an `Executor`-aware. That's how we don't have there `consumer` span and that is what was an original request in that Spring Cloud Stream issue. 

My idea on the matter to check the channel type and perform consumer span withing `preSend/afterSendCompletion` call stack. However that looks for me like an overlap the producer span and consumer anyway...

Another thought is like a `TracingRequestHandlerAdvice` implementation to apply exactly to the `MessageHandler.handleRequestMessage()` call. This way with the scenario like: `dc -> mh -> dc -> mh` we may definitely have properly stacked producer and consumer spans even if everything is still in a single thread.

Thoughts ?

instrumentation/spring-messaging/README.md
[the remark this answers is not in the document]
artembilan: Maybe this: https://docs.spring.io/spring-boot/docs/current/reference/html/boot-features-messaging.html? Not sure that we should always rely on the Getting Started Guides...

instrumentation/spring-messaging/src/main/java/brave/spring/messaging/TracingChannelInterceptor.java
adriancole: I don't quite see the alignment problem as yet, if stacks are in use even if there's a handle between the send.

I think all it needs is a test case to make me understand. Any chance you can give a code snippet for a dumb `dc -> mh -> dc -> mh` that I can ensure passes? You can link me a similar test in spring-cloud-stream if easier

instrumentation/spring-messaging/README.md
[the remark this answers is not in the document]
adriancole: works for me :) it will be mildly imprecise because some will use this without spring boot, but anyway good link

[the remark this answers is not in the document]
artembilan: Yeah, indeed, not everyone is with Boot, but at least this link summaries what we have about Messaging in Spring.

instrumentation/spring-messaging/src/main/java/brave/spring/messaging/TracingChannelInterceptor.java
artembilan: Ok. I will try to come up with something tomorrow, but bear in mind there is reason to overhead it with Spring Cloud Stream, because the problem is exactly on the `MessageChannel` level. So, simple unit test should be enough.

adriancole:
Thx agreed unit tests should suffice, just something higher level helps
with later bug squashing. All help appreciated!

adriancole:
Started porting some of sleuth's integration tests and switched the scoping mechanism to stack based

adriancole:
For those following along.. this creates chained spans just like in kafka. For example, the producer span starts and stops, then the consumer span, then any processor (if there is one). In this case the subscribed message handler is the processor.

Importantly, header-based propagation is still used when polling consumers are, as there's no thread relationship between poll and a later (if ever) processor. This is why there's a utility function to resume a span from a message.

https://github.com/openzipkin/brave/tree/master/instrumentation/kafka-clients#whats-happening

In the near future, we'll likely make an abstraction, if for no other reason but to centralize docs on how messaging semantics apply to java code

adriancole:
Here's some insights so far:

Typically we think of a producer consumer model as producer starts and stops, then consumer starts and stops (queue). However, it seems spring-messaging has synchronous and asynchronous dispatch (DirectChannel, ExecutorChannel, PublishSubscribeChannel) which are different in nature to things like QueueChannel which is more similar to our messaging span.

For example, the dispatch-oriented ones can use normal or executor-instrumented thread state propagation. The Queue one(s) will need the "inject/extract" message propagation. In either of these styles it isn't clear at this level of abstraction how you'd know if there was a remote downstream or not, eventhough you can tell to a degree if your handler is local (ex it will be in DirectChannel)

impact: I think we should probably cherry-pick a few channel types for testing like DirectChannel QueueChannel etc, but not expect all of them to be messaging (like queue based modeling). Moreover, we need to integrate a real backend like rabbit or kafka and see how this plays out in a remote scenario.

@dsyer also feel free to clarify me.. or pass some hints

adriancole:
There's a small tension as we don't know if tasks are queue in nature or not until we know types or observe behavior. For example, sometimes preSend is used for message handler while other times it is beforeHandle. Sometimes (preSend, afterSendCompletion) encloses (postReceive, afterReceiveCompletion) and sometimes they are happens after (like when queuing).

I think the code will be easier to follow if we focus on modeling the work at the abstraction of spring-messaging, and decouple that from pretending we can know if that work is remote or not. For example, we can create spans etc, but not mark them as queueing (producer/consumer) or remote until such time we know they are.

In general, we know more as we move up in layers. spring-messaging, we can't know if something is remote or not, or if we can, it is guessing. We need more, as for example, even to model like a queue requires us to know if we are a QueueChannel or not (that class is in spring-integration). binders like kafka are even further up the stack in spring-cloud-stream

ACTIONS:

* add a lot of tests for the known types provided in spring-messaging, so that we can pin-down the span modeling behavior and as many gotchas as possible (like preHandle not being called for whatever reason)

* In spring-messaging types do defensive header extraction, but don't do header injection. Create a utility which can inject headers onto a message when it is definitely remote. This could be used in another interceptor or some conditional hook by upper layer code.

* make a separate package (or platform detection) for spring-integration as we at least need to know if QueueChannel is in use. That one should use Producer/Consumer

* get some integrated example with stream, possibly with kafka, to figure out how we can tell which parts of a flow are remote, using any of the spring-messaging, -integration or -cloud-stream types.

@marcingrzejszczak  is attacking this in sleuth and we're collaborating both sides to try to help make a permanent solution to the modeling concerns above. Might take some time, more advice and practice before solid.

adriancole:
rebased this.. need to re-look at sleuth as a lot may have changed since last look

devinsba:
Oh I didn't realize this was here, I think I may have done redundant work when I built: https://github.com/spring-cloud/spring-cloud-sleuth/pull/1218

PR #574, opened 2018-01-11
Adds extraction mechanism to ThreadLocalSpan and supports stacking

adriancole:
This adds utilities needed for spring-messaging and similar patterns
where a thread local is used to bridge two callbacks.


adriancole:
will backfill unit tests before release. this class somehow never had any

PR #576, opened 2018-01-12
Remove unneeded @Autowired annotation

geoand:
The TracingHandlerInterceptor is not annotated with one of the Spring
stereotype annotations and is instead created explicitly using Java
Config or XML config, therefore there is no need have the constructor
annotated with @Autowired

### DIFF
--- a/instrumentation/spring-webmvc/src/main/java/brave/spring/webmvc/TracingHandlerInterceptor.java
+++ b/instrumentation/spring-webmvc/src/main/java/brave/spring/webmvc/TracingHandlerInterceptor.java
@@ -40,7 +40,7 @@ public final class TracingHandlerInterceptor implements HandlerInterceptor {
   final HttpServerHandler<HttpServletRequest, HttpServletResponse> handler;
   final TraceContext.Extractor<HttpServletRequest> extractor;
 
-  @Autowired TracingHandlerInterceptor(HttpTracing httpTracing) { // internal
+  TracingHandlerInterceptor(HttpTracing httpTracing) { // internal
     tracer = httpTracing.tracing().tracer();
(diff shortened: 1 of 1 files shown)


adriancole:
Curious.. is this causing a problem?

In recent spring it is possible to Import a type that has an autowired
annotation (even if examples dont do that).

Is preventing autowiring buying us something?

geoand:
@adriancole Nope, no problems at all, or any explicit gain for this :). 

It's just that (if I am not seriously mistaken) it's unnecessary. Spring won't try to autowire the constructor since the bean has already been created via Java Config or XML config.

adriancole:
@geoand so to put this another way, if we keep `Autowired` we should tell people to choose one of scan the brave package or manually wire, but not both?

geoand:
Unless I am missing something obvious, scanning of the brave package won't pickup this class, because it's not annotated with a stereotype annotation like `@Component`.
If my understanding is correct, only manual wiring works in this case, and that doesn't change by the presence or absence of `@Autowired` on the constructor

adriancole:
I think the following would work in spring 4+ but not spring 3. Mildly less
annoying than making bean methods, but again maybe not idiomatic or not
helpful enough to support. wdyt?

// NOTE: while bean configuration via @Import works with Spring 4, it does
not with Spring 3

@Configuration @Import(TracingHandlerInterceptor.class)
static class InstrumentationConfiguration {

geoand:
That would work great!
It looks good to me :)

adriancole:
closing for now. if we re-open this, likely we'd do the same for anything spring related. appreciate the dialogue!

geoand:
Sounds good! 
Thanks for taking the time to discuss!

PR #577, opened 2018-01-16
Adds CorrelationFields in support of in and between process propagation

adriancole:
## Overview
This change aims to support correlation fields which may be request-scoped (and propagated downstream), or locally scoped (propagated in process, but not to the next). Example use cases include request-scoped fields like current game or country code, or local fields, such as the current operation name. Where cross-process propagation is desired, the ideal encoding is the [w3c Correlation Context](https://github.com/w3c/distributed-tracing/tree/master/correlation_context).

## Background on existing work in Brave
The first propagation mechanics of brave included propagation of the [TraceContext](https://github.com/openzipkin/brave/blob/master/brave/src/main/java/brave/propagation/TraceContext.java) in and to the next process. This object's duty is to place spans in the correct position in a potentially distributed operation (via parent IDs). Secondary duties include consistent sampling decisions, and correlation with other contexts such as logging. The data is mostly the same as the [trace-context](https://github.com/w3c/distributed-tracing) header being defined in w3c.

Brave has two roles in propagation: making data visible in-process, and injecting and extracting data out of process.

The [CurrentTraceContext](https://github.com/openzipkin/brave/blob/master/brave/src/main/java/brave/propagation/CurrentTraceContext.java) plugin synchronizes a `TraceContext` struct with implicitly visible state, such as a thread local or a logging context. Users can then call `Tracer.getCurrentSpan` for example, or read the logging context variable `traceId`. This in-process propagation is critical for correlation between systems, as often libraries have no other means to coordinate.

The [Propagation](https://github.com/openzipkin/brave/blob/master/brave/src/main/java/brave/propagation/Propagation.java) plugin connects traces across message boundaries by injecting or extracting the trace context from headers. For example, an incoming trace from another process is extracted from headers, the tracer creates a child, and later instrumentation use the above `CurrentTraceContext` to synchronize the current trace ID with whatever contexts that need to see it.

Notably due to demand created by OpenTracing and Sleuth, we have had requests to push arbitrary data along with the trace-context (implicitly via the above apis described). We've also had a number of non-arbitrary requests to [propagate extra data](https://github.com/openzipkin/brave/issues/390). For example, domain-specific data like the current playstation game, or extra properties needed for Amazon X-Ray.

Towards that end, we added an extension to our trace context object called "extra". [Extra](https://github.com/openzipkin/brave/blob/master/brave/src/main/java/brave/propagation/TraceContext.java#L107) is a list of opaque data plugins can use to ensure their propagation needs are met. 

We also added a default implementation that handles the common use-case of sending a mix of system and user fields. Here's an example:

Setup your tracing instance with allowed fields:
```java
tracingBuilder.propagationFactory(
  ExtraFieldPropagation.newFactoryBuilder(B3Propagation.FACTORY)
                       .addField("x-vcap-request-id")
                       .addPrefixedFields("baggage-", Arrays.asList("country-code", "user-id"))
                       .build()
);
```

Later, you can call below to affect the country code of the current trace context
```java
ExtraFieldPropagation.set("country-code", "FO");
String countryCode = ExtraFieldPropagation.get("country-code");
```
 
## Why a change is needed
The current code is ok, but limited in implementation which makes it difficult to integrate with metrics contexts and scopes smaller than trace.

Using `ExtraFieldPropagation`, correlation with other fields involve the following trade-offs:
* In `ExtraFieldPropagation` (and OpenTracing) propagating something in-process also means adding overhead out of process
* `ExtraFieldPropagation` is harder to manage systematically as it uses a separate header per field
* `ExtraFieldPropagation` coupling with propagation fields implies a whitelist approach
* Users would want to wrap `ExtraFieldPropagation` as it isn't a great user-level api

For example, if you propagate the span name, to export changes to it for log correlation, you'll need to customize the log correlation plugins and also end up with another header on the wire that isn't useful on the other side.

On the other hand, with a `CorrelationFields` type, we can decouple the usage of propagated tags, their policy and their encoding scheme (defaulting to the correlation-context header). Like `ExtraFieldPropagation`, this could be added to B3, allowing a clean transition. In fact, `ExtraFieldPropagation` could be an implementation. At the end of the day, we end up with flexibly scoped propagation code with coherent usage patterns for context synchronization.

## Scoping constraints
An understood limitation is that since this api exists in the scope of Brave (a tracing api), Correlation fields cannot be added before a trace context has been created. For example, an integration with metrics context could imply ordering metrics after tracing, eager provision of an unstarted trace context, or an integration to copy currently scoped metrics tags into the a newly provisioned `CorrelationFields` object.

## See also
A lot of this work is biased towards the emerging [w3c Correlation Context](https://github.com/w3c/distributed-tracing/tree/master/correlation_context) specification.

This also has feature overlap (and integration possibilities) with Baggage (both [the original](http://brownsys.github.io/tracing-framework/docs/tutorials/baggage.html) and [OpenTracing](http://opentracing.io/documentation/pages/api/cross-process-tracing.html)) and [OpenCensus Tags](https://github.com/census-instrumentation/opencensus-specs/blob/master/tags/TagContext.md). Note that feature overlap isn't 100% between any of these.

### DIFF
--- a/brave/README.md
+++ b/brave/README.md
@@ -893,3 +893,68 @@ Propagating a trace context instead of a span is a right fit for several reasons
 ### Public namespace
 Brave 4's public namespace is more defensive that the past, using a package
 accessor design from [OkHttp](https://github.com/square/okhttp).
+
+### CorrelationFields Api
+
+The CorrelationFields functionality is similar to OpenTracing (OT) Baggage, and
+dissimilar to Census tags. This is a legacy preference, allowing Sleuth users
+(which has a baggage api similar to OT) to transition to Brave. This design
+implies string based correlation fields are stored in a span or a trace context
+for read-back, whereas census tags are immutable scoped variables. OT, Census
+and Brave all decouple thread scoping, header injection and extraction from the
+apis that mutate propagated fields.
+
+Here's an example of using OpenTracing baggage:
+```java
+// mutate the span's baggage adding the method
+span.setBaggageItem("method", fullMethodName);
+
+// set the span in scope
+try (Scope scope = tracer.scopeManager().activate(span, false)) {
+  // If downstream code call tracer.activeSpan(), they'll see the above span.
+  // It can then read span.getBaggageItem("method")
+  //               or span.context().baggageItems()
+}
+// Repeating the same block might have different results as downstream code
+// or other threads can overwrite the key name "method".
+```
+
+Here's an example of using Census tagging:
+```java
+// Tags are encouraged to be constants
+static final TagKey RPC_METHOD = TagKey.create("method");
+...
+
+// make a copy of the parent tag context, with the method tag
+TagContext withMethod = tagger.toBuilder(parentCtx)
+                              .put(RPC_METHOD, TagValue.create(fullMethodName)).build();
+
+// set the tag context in scope
+try (Scope scope = tagger.withTagContext(withMethod)) {
+  // If downstream code call tagger.getCurrentTagContext(), they'll see the above tags.
+  // However, as of v0.10, TagContext.getIterator() is hidden, so only census impls can read them!
+}
+// Repeating the same block has the exact same result as tag contexts are immutable.
+```
+
+The OT-centric design has a few implications, notably shared mutable state:
+
+* String keys and values are easy to use, but don't imply up-front validation.
+  * Census tags are types, encouraging prevalidation and use as constants.
+* Sharing mutable data across threads can lead to inconsistent reads of fields.
+  * For example, child threads affect the same state their parent reads.
+  * Immutable tag contexts eliminate this
+* A tracer api is more complex than a tagger api, and has an ordering concern.
+  * To contribute tags, you need to have tracing start before metrics.
+  * library compatibility is a higher liability with larger apis
+
+The following are some areas where `CorrelationFields` mitigates some of this:
+
+* `CorrelationFields` is a separate api, limiting the api surface area
+  * `CurrentCorrelationFields` can obviate explicitly exposing tracer code.
+* `CorrelationFields.NOOP` is default
+  * Users who don't use or want correlation fields are burdened less.
+* `CorrelationFields` is addressable via the `TraceContext`.
+  * `TraceContext` is a lighter object than `Span`, with no heavy references.
+* Brave's api allows pre-allocation of `TraceContext` instances.
+  * This means you can attach correlation fields prior to starting a span.
--- a/brave/src/main/java/brave/internal/correlation/CorrelationFields.java
+++ b/brave/src/main/java/brave/internal/correlation/CorrelationFields.java
@@ -1,0 +1,127 @@
+package brave.internal.correlation;
+
+import brave.internal.Nullable;
+import brave.propagation.Propagation;
+import java.util.Map;
+
+/**
+ * Associates key/value pairs with the current trace context. Correlation fields are stored as a
+ * part of the trace context, adding overhead. For this reason, {@link #isNoop() noop mode} is
+ * default: correlation fields are opt-in.
+ *
+ * <p>Here's an example of instrumentation publishing a user ID field for downstream code to read
+ * with {@link CurrentCorrelationFields}.
+ * <pre>{@code
+ * // mutate the trace context's correlation fields
+ * span.context().correlationFields().set("user-id", currentUserId);
+ *
+ * // set the context in scope
+ * try (Scope scope = tracing.currentTraceContext().newScope(traceContext)) {
+ *   // downstream can read fields with CurrentCorrelationFields.get("user-id")
+ *   //                              or CurrentCorrelationFields.forEach((k, v) ->{});
+ * }
+ * // Repeating the same block might have different results as downstream code
+ * // or other threads can overwrite the key name "method"
+ * }</pre>
+ *
+ * <h3>Data constraints</h3>
+ *
+ * <p>This implementation is strict with regards to key and value policy. Currently limited to 255
+ * printable characters respectively. That said, there's no requirement for fields to end up on the
+ * wire with the same names. For example, in <a href="https://github.com/w3c/distributed-tracing/tree/master/correlation_context">Correlation-Context</a>,
+ * they are embedded in a single header, joined on semi-colon.
+ *
+ * <h3>Relationship to Span tags</h3>
+ *
+ * <p>{@link brave.Span#tag(String, String) tags} are different than correlation fields in the
+ * following ways:
+ * <pre>
+ * <ul>
+ *   <li>Tags are sent out-of-band where correlation fields are in-band, usually embedded in request headers</li>
+ *   <li>As correlation fields are used to synchronize other contexts, they support readback</li>
+ * </ul>
+ * </pre>
+ *
+ * Put another way, a metrics or logging context can readily access correlation fields once they are
+ * added.
+ *
+ * <h3>Relationship to Propagation injection and extraction</h3>
+ * {@link Propagation} injection and extraction is intentionally decoupled from correlation fields.
+ * The reason for this is we don't know what the encoding format would be, and if it correlation
+ * fields will be in a separate propagation field or not. For example, if the propagation format is
+ * Amazon's, correlation fields coexist in the same header field as trace identifiers. B3 has no
+ * correlation field format. However, you can add <a href="https://github.com/w3c/distributed-tracing/tree/master/correlation_context">Correlation-Context</a>
+ * to B3 to accomplish both.
+ *
+ * <h3>Application notes</h3>
+ *
+ * <p>Where possible, instrument nearest to the original request. For example, ordering before
+ * metrics can allow any extracted fields to be visible as a metrics dimension or logging context key.
+ *
+ * <p>Do not use correlation fields as ad-hoc storage, as they will add weight to the request and
+ * expose them to arbitrary access. A password would be a very bad example of a correlation field.
+ */
+public abstract class CorrelationFields implements Cloneable {
+
+  /** When true, no fields are recorded. Use this flag to avoid performing expensive computation. */
+  public abstract boolean isNoop();
+
+  // cannot depend on java 8 apis
+  public interface Consumer {
+    void accept(String name, String value);
+  }
+
+  public abstract void set(String name, String value);
+
+  /** Hidden as used for internal merging */
+  public abstract void setAll(CorrelationFields other);
+
+  /** Retrieves a correlation key by name or returns null if not found. */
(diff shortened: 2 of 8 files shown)


adriancole:
cc @felixbarny @wu-sheng @marcingrzejszczak @devinsba @llinder @jcchavezs @basvanbeek @shakuzen @nicmunroe @jkschneider (and any others) please check me on commentary etc.

Actual implementation will follow shortly as homework (the hard part) is done

brave/src/main/java/brave/propagation/CorrelationFields.java
```diff
@@ -0,0 +1,196 @@
...
+
+    @Override public void forEach(Consumer consumer) {
+    }
+
+    @Override public String toString() {
+      return "NoopCorrelationFields{}";
+    }
+  };
+
+  static final class Real extends CorrelationFields {
```
adriancole: PS I am starting with LHM eventhough it has more memory cost than array-offset indexing.

brave/src/main/java/brave/propagation/CorrelationFields.java
```diff
@@ -0,0 +1,196 @@
...
+  public interface Consumer {
+    void accept(String name, String value);
+  }
+
+  public abstract void put(String name, String value);
+
+  /**
+   * Retrieves a correlation key by name or returns null if not found.
+   *
+   * @throws IllegalArgumentException if the input name is not valid
```
adriancole: this is out-of-date. throwing is bad as it can break users (who are likely to ignore this). Current impl just silently drops on problems.

brave/src/main/java/brave/propagation/CorrelationFields.java
```diff
@@ -0,0 +1,196 @@
...
+      for (int i = 0; i < size; i++) {
+        char c = input.charAt(i);
+        if (c >= ' ' && c <= '~') continue;
+        return log(type + " " + input + " contains an unprintable character");
+      }
+      return true;
+    }
+
+    private static boolean log(String message) {
+      return false;
```
felixbarny: TODO add logger?

brave/src/main/java/brave/propagation/CorrelationFields.java
```diff
@@ -0,0 +1,196 @@
...
+ *   context is shared regardless of spans. This can affect performance of code that creates a lot
+ *   of spans</li>
+ *   <li>Span-scoped data implies sharing of data must happen after a tracer api. This means metrics
+ *   can't reliably collaborate with this data unless metrics perform after span allocation.</li>
+ *   <li>Tracer apis are more complex than tagging apis, and increases the compatibility complexity
+ *   verses a standalone tagging api.</li>
+ * </ul>
+ * </pre>
+ *
+ * <p>The following are some areas where the Brave can mitigate some of these concerns.
```
felixbarny: > the Brave

pun intended?

adriancole: I wish

adriancole:
moved the huge rationale section to where the other rationale are in the brave README. Added `CurrentCorrelationFields` which makes this DI friendly

brave/README.md
```diff
@@ -772,3 +772,68 @@ Propagating a trace context instead of a span is a right fit for several reasons
...
+// Repeating the same block has the exact same result as tag contexts are immutable.
+```
+
+The OT-centric design has a few implications, notably shared mutable state:
+
+* String keys and values are easy to use, but don't imply up-front validation.
+  * Census tags are types, encouraging prevalidation and use as constants.
+* Sharing mutable data across threads can lead to inconsistent reads of fields.
+  * For example, child threads affect the same state their parent reads.
+  * Immutable tag contexts eliminate this
```
ivantopo: It might be worth mentioning that immutable tags eliminate this, at the cost of having to manage scopes for the TagContext.

brave/README.md
```diff
@@ -772,3 +772,68 @@ Propagating a trace context instead of a span is a right fit for several reasons
...
+* A tracer api is more complex than a tagger api, and has an ordering concern.
+  * To contribute tags, you need to have tracing start before metrics.
```
ivantopo: I'm not entirely sure I get the point of bringing up metrics here. Ideally only a subset of span tags (the ones ensured to be low cardinality + `http.template`) and none of these seem to be related to correlation fields. What's the link I'm missing here?

aldex32:
Hi guys,

is it possible to have the possibility to set a custom header by using `ExtraFieldPropagation`?

brave/README.md
adriancole: good point.. author died in the middle of this point i think

brave/README.md
adriancole: yeah wording was a bit crap. (tags are the census name for correlation fields).

The jist is that if the correlation api is separate, it can scope fields before or even if tracing doesn't exist. One use case is to use the same correlation field in metrics and tracing (and further down the line, such as country-code)

To have metrics re-use correlation fields from tracing, for example, it would likely need to start after tracing started (or at the same time). Similar problem exists in logging. If you want logs to contain a correlation field, that field should be scoped as early as possible. Otherwise, you see request handling without and MDC value until sometime later when tracing kicks in.

The idea is that a tagging api could be made at the head of a request, and be simpler or at least less dangerously, ensuring that whether it is tracing metrics or logging (or something else), they all read exactly the same values, with no ordering interdependency apart from tagger.

What I'm calling out here is that if we assign that responsibility to brave (for convenience because parsing, inject extract propagation etc is already handled), there is at least an interdependency problem, which is most easily sorted if trace starts before other subsystems on a request, or at least as early as possible.

Now.. if I did clear up anything (mildly possible), could you help me revise it into a shorter bullet or two? :)

adriancole:
@aldex32 yeah I'd expect ExtraFieldPropagation to use this feature for mutability

adriancole:
@aldex32 I'm going to address the mutability of `ExtraFieldPropagation` so that there's less time pressure on this PR. I have some discomfort on local changes, particularly around extracting a context which are not a concern with `ExtraFieldPropagation`

adriancole:
@aldex32 Actually, there's a limitation in the current Propagation design where it can only affect traces that originated remotely. That impacts this, too.. will raise something independently

adriancole:
https://github.com/openzipkin/brave/issues/579 needs to be explored prior to finishing this design. Will work on that

adriancole:
Had some chats with folks like @ivantopo and I feel I want to think this through a bit more. I generally don't like how this feels like it needs to hook together (want to see if there's another way to avoid special hooks in the tracer or not).

Meanwhile I've rejigged things in support of prefixed keys (and mutation) of the existing extra field propagation code. I'm fairly happy with this, as the logic is completely optional to those not pushing extra fields. Going to cut that below before picking this up again. https://github.com/openzipkin/brave/pull/581

adriancole:
I've pushed some [interim improvements to extra field propagation](https://github.com/openzipkin/brave/tree/master/brave#prefixed-fields) in brave 4.9.3 to buy time here. Will revisit shortly

adriancole:
did a huge revamp on the description. updates around the code will follow next

shakuzen:
Just sharing our use case that I'm hoping this can solve: we have an additional ID that we would like to propagate throughout the trace and add to all services' logs for log correlation based on that ID.

adriancole:
>
> Just sharing our use case that I'm hoping this can solve: we have an
> additional ID that we would like to propagate throughout the trace and add
> to all services' logs for log correlation based on that ID.
>
Is that ID constant for all downstream requests? Also, does that need to be
sent in a specific header, or does it matter as long as in-process
correlation occurs?

shakuzen:
> Is that ID constant for all downstream requests?

The service that starts the trace (or possibly an uninstrumented service before the trace starts) should discern and set the ID. It shouldn't be changed.

> Also, does that need to be sent in a specific header, or does it matter as long as in-process
correlation occurs?

Things aren't quite set in stone yet, but correlation is the primary goal. As alluded to above, the service initially setting this ID may be uninstrumented and therefore a well-known header is probably the best way to receive this in an instrumented app, was my line of thinking. If everything were instrumented, I guess it wouldn't matter how it is propagated.

adriancole:
@SergeyKanzhelev ps if you like please review the PR description eventhough the code has drifted a bit This is in response to notes in the census issue that came out of the same discussion https://github.com/census-instrumentation/opencensus-specs/issues/45

adriancole:
feedback from @mpetazzoni included that if this feature didn't require a header-per-field encoding (whitelist), it would be easier to use

> In the context of a broader code base where I define the tracer in one place (and then available via DI), it's not really convenient to define that whitelist of fields. Really depends on the service, use-case, etc.

adriancole:
I've rejigged the code so that it is compilable now, but it isn't right. I'll try to peck at it over a flight

adriancole:
here are some design thoughts:

An implementation of correlation context could be "accumulate and publish later". For example, data is placed into the correlation fields object, but only made visible when the span is re-scoped.

There are generally 3 scopes of data which one might correlate with:

process scoped: ex current amazon instance ID
request scoped: ex origin user id, country code
span scoped: ex span name

To address the scoping of data, we could consider options such as a flag per field, indicating whether it is to be published downstream or not, or if it is to expire with the span. Note that in the Correlation-Context spec, there's been discussion of a ttl flag (interpreted remotely), which for example the value 1 means drop at the next hop.

adriancole:
deprioritizing this as during the tracing workshop, there are especially security concerns about adding user data to outgoing requests by default. Basically, what some call "baggage" is not only dangerous to the size of production requests, but also a potentially large security risk (imagine accidentally propagating data to a third party)

We should be careful when implementing it to make 100% optional and something that potentially re-validates per hop whether to send the correlation keys out or not. 

In the mean time, users can use the whitelist approach with ExtraFieldPropagation. We may still want to continue with in-process propagation efforts, but at any rate be very careful with remote propagation (cc @garyd203 @sirtyro)

https://docs.google.com/document/d/1YJ0oriH62Ixk3W37z31X0SeuXpgXpXBetiz8EWxG6NU/edit#

adriancole:
still waiting as prefer to see what census decides before continuing this: census-instrumentation/opencensus-specs#200

adriancole:
Most notably, there is an api for filtering proposed. This is something that's currently possible in brave anyway as the "getter' concept can internally implement filtering for any key (trace id or a tag) OTOH has discussed in this PR, the fields are likely to be concatenated into a single header.. propagation decision logic makes sense.

Also, while this we are interested in what census does, Kamon has already rolled out an impressive feature (hats off @ivantopo @dpsoft) https://github.com/kamon-io/Kamon/pull/552

What we do here will consider all angles in other words (cc also long time friend @nicmunroe)

adriancole:
fixed by #1130

Issue #592, opened 2018-01-25
manually publish brave 4.13.4 and 4.14.0


adriancole:
CircleCI has been failing incessantly (usually by socket disconnect or the vm dying), but should fail less when we stop publishing archive #567. Unfortunately, I can't get CircleCI to pass a deploy (even after 6 attempts), so publishing manually

adriancole:
push intended version tag
```bash
$ ./mvnw versions:set -DnewVersion=4.13.4  -DgenerateBackupPoms=false
# edit /pom.xml and change tag to 4.13.4
$ git commit -am"prepare release 4.13.4"
$ ./mvnw clean install -DskipTests
$ git tag 4.13.4
$ git push origin 4.13.4
```

make next version and push master
```bash
$ ./mvnw versions:set -DnewVersion=4.14.0-SNAPSHOT  -DgenerateBackupPoms=false
# edit /pom.xml and change tag to HEAD
$ git commit -am"prepare for next development iteration"
$ ./mvnw clean install -DskipTests
$ git push origin master
```

adriancole:
build automation kicked off the 4.13.4 deploy, which initially died, but re-kicked. If the following also dies, I'll do the deploy manually as well https://circleci.com/gh/openzipkin/brave/916

as mentioned in other topics to @openzipkin/devops-tooling  I think it is time to remove circleci from the critical path as I am constantly kicking builds and when it gets to release is paralyzing. If I was on holiday I'm sure no-one would want to do a release in this state, so best to only use circle for very small work if at all

PR #593, opened 2018-01-25
Uses travis to deploy instead of CircleCI

adriancole:
This reverts some work in 82c1232ea8cc2c1569813d5240e222e883e588ff such that failures or quotas in CircleCI can't prevent us from releasing brave

See #592


adriancole:
added some test deflaking logic here #594

adriancole:
both travis checkers passed. pushing again to ensure they do twice in a row

PR #597, opened 2018-01-29
Adds vertx-web module

### LINKED ISSUE #307: brave-vertx module
Requests to trace vertx have come up more often in gitter https://gitter.im/openzipkin/zipkin

There is an implementation here https://github.com/emmanuelidi/vertx-zipkin, and we've also gotten requests for other versions of vertx

Does it make sense to make a multi-module repo openzipkin/brave-vertx ? If ok with @emmanuelidi, we could actually move his repo there and reorganize it to support multiple versions and the publication process we have. Alternatively, we could start again, yet another option is to just push requests to emmanuelidi/vertx-zipkin

Thoughts? @mkargar @lintuxlin @emmanuelidi @fzakaria @hosuaby @cescoffier @vietj

adriancole:
This adds support for vertx-web's Routing Context. This was developed
while looking at prior work in the following repositories:
* https://github.com/emmanuelidi/vertx-zipkin/tree/master/src/main/java/io/vertx/ext/zipkin/web
* https://github.com/opentracing-contrib/java-vertx-web

Notably, the latter had notes about rerouting, which were helpful as we
don't yet have a generic http test for forwarding/re-routing.

Fixes #307

### DIFF
--- a/instrumentation/netty-codec-http/README.md
+++ b/instrumentation/netty-codec-http/README.md
@@ -2,9 +2,6 @@
 
 This module contains a tracing decorators for [Netty's Http Codec](https://github.com/netty/netty/tree/4.1/codec-http) 4.x.
 
-This module contains JAX-RS 2.x compatible tracing filters and a feature
-to automatically configure them.
-
 `NettyHttpTracing.serverHandler()` extracts trace state from incoming requests,
 and reports to Zipkin how long each take, along with relevant tags like the
 http url.
--- a/instrumentation/pom.xml
+++ b/instrumentation/pom.xml
@@ -34,6 +34,7 @@
     <module>spring-webmvc</module>
     <module>kafka-clients</module>
     <module>netty-codec-http</module>
+    <module>vertx-web</module>
   </modules>
 
   <!-- ${project.groupId}:brave version is set in the root pom.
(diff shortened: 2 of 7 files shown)


adriancole:
Here are some impl notes.

As mentioned a couple times, there was a discussion on opentracing's PR about reroute which was helpful. I'm not sure who started it, but @pavolloffay summarized some parts of it:
https://github.com/opentracing-contrib/java-vertx-web/pull/2#issuecomment-312860001

In a related note Brave instrumentation typically do not hold the RPC span open until the complete body is read (eventhough such an idea was piloted as a child span by @yschimke). By following our normal behaviour, we avoid a lot of the dodgy implications about what happens if a connection reset occurs etc. In the future, we may make ending a span based on body read vs status quo (headers sent), but for now all good. 

https://docs.google.com/document/d/1yIo_zZvMyAS8J7bpCz85h-FISwU0Jr-Zu-POIvp7B9E/edit

Another note was that by using our currentcontext api, I was able to get all of our standard tests passing. I am aware that there are places we can break propagation as eluded to by @emmanuelidi and @vietj. Hopefully we can make a failing test which we can follow-up with later on whatever the best issue is.

For now, I think this code has the basics in place and even if is not great passes all integration tests

PR #600, opened 2018-01-31
Prevents java.lang.NoClassDefFoundError when searching for platform

### LINKED ISSUE #599: Get NoClassDefFoundError of ThreadLocalRandom  when using jre6
when run in jre1.6.0_33，I got a exception:
Exception in thread "main" java.lang.NoClassDefFoundError: java/util/concurrent/ThreadLocalRandom
	at brave.internal.Platform.findPlatform(Platform.java:104)
	at brave.internal.Platform.<clinit>(Platform.java:27)
	at brave.Tracing$Builder.build(Tracing.java:271)

adriancole:
Recently, we added `Platform.nextTraceIdHigh()`, which compiled against
a JDK 7 type `ThreadLocalRandom` due to being indirectly referenced in
a utility method. This changes the signature such that it can't imply
this anymore.

This also removes AutoValue from platform as it adds unnecessary classes
to the jar.

Fixes #599

### DIFF
--- a/brave/src/main/java/brave/internal/Platform.java
+++ b/brave/src/main/java/brave/internal/Platform.java
@@ -3,7 +3,6 @@ package brave.internal;
 import brave.Clock;
 import brave.Tracer;
 import brave.Tracing;
-import com.google.auto.value.AutoValue;
 import java.net.InetAddress;
 import java.net.NetworkInterface;
 import java.security.SecureRandom;
@@ -110,7 +109,7 @@ public abstract class Platform {
     if (jre7 != null) return jre7;
 
     // compatible with JRE 6
-    return Jre6.build(zipkinV1Present);
+    return new Jre6(zipkinV1Present);
   }
 
   /**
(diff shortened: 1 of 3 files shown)


brave/src/main/java/brave/internal/Platform.java
```diff
@@ -175,63 +172,81 @@ static Jre9 buildIfSupported(boolean zipkinV1Present) {
...
-    @IgnoreJRERequirement
-    @Override public long randomLong() {
-      return java.util.concurrent.ThreadLocalRandom.current().nextLong();
+    Jre9(boolean zipkinV1Present) {
+      super(zipkinV1Present);
     }
 
-    @IgnoreJRERequirement
-    @Override public long nextTraceIdHigh() {
-      return nextTraceIdHigh(java.util.concurrent.ThreadLocalRandom.current());
```
adriancole: using a static method like this indirectly caused the code to compile against the TLR type, subverting the attempt to defer usage until runtime. I fixed the signature so that it couldn't do this anymore.

adriancole:
try 4.14.2

PR #601, opened 2018-01-31
Removes temporary measure to migrate http middleware to zipkin v2

adriancole:
http client and server middleware incubated with zipkin v1 apis. To
give people time to update, we deprecated hooks to parse remote
endpoints using v1 types. All but two projects on github have migrated
off those types, obviating the special handling.

Note: this doesn't remove any common-course support for v1 endpoint,
such as users adding it to a span or to a tracer, eventhough those are
deprecated, too. This only removes from middleware plugins to simplify
their build scripts, removing a quite usualy source of an optional
dependency.

I think we can merge and cut this once the below are migrated:
https://github.com/xjdr/xio/issues/174 cc @pdex
https://github.com/dreamming/dmz-inward-/issues/1 cc @yinming

### DIFF
--- a/brave/src/main/java/brave/internal/Platform.java
+++ b/brave/src/main/java/brave/internal/Platform.java
@@ -27,9 +27,6 @@ public abstract class Platform {
 
   volatile Endpoint endpoint;
 
-  /** Ensure we don't raise a {@linkplain ClassNotFoundException} calling deprecated methods */
-  public abstract boolean zipkinV1Present();
-
   public Reporter<zipkin2.Span> reporter() {
     return LoggingReporter.INSTANCE;
   }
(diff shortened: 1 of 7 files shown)


adriancole:
cc @llinder @marcingrzejszczak @devinsba @shakuzen @takezoe @hyleung y'all aren't affected, but might end up with neater builds as a result

adriancole:
ps to clean this up further would be to remove the method `Span.remoteEndpoint(zipkin.Endpoint)` as this is holding non-http instrumentation in such a way that they have to compile with old classes present even if not used.

We could remove Span.remoteEndpoint(zipkin.Endpoint)` without removing the more routinely accessed, and therefore more risky to remove, methods on Tracing.Builder

adriancole:
ps my other suggestion is much more breaking, so leaving that for another day

PR #602, opened 2018-01-31
Implements parsing of http route and uses it by default for span name

adriancole:
"http.route" is an expression such as "/items/:itemId" representing an application endpoint. This adds support for parsing this, employing it as the default span name policy when present. A route based span name looks like "/users/{userId}", "not_found" or "redirected".  More below:

## Http Route
The http route is an expression such as `/items/:itemId` representing an
application endpoint. `HttpAdapter.route()` parses this from a response,
returning the route that matched the request, empty if no route matched,
or null if routes aren't supported. This value is either used to create
a tag "http.route" or as an input to a span naming function.

### Http route cardinality
The http route groups similar requests together, so results in limited
cardinality, often a better choice for a span name vs the http method.

For example, the route `/users/{userId}`, matches `/users/25f4c31d` and
`/users/e3c553be`. If a span name function used the http path instead,
it could DOS-style attack vector on your span name index, as it would
grow unbounded vs `/users/{userId}`. Even if different frameworks use
different formats, such as `/users/[0-9a-f]+` or `/users/:userId`, the
cardinality is still fixed with regards to request count.

The http route can be "" (empty) on redirect or not-found. If you use
http route for metrics, coerce empty to constants like "redirected" or
"not_found" with the http status. Knowing the difference between not
found and redirected can be a simple intrusion detection signal. The
default span name policy uses constants when a route isn't known for
reasons including sharing the span name as a metrics correlation field.

### Supporting HttpAdapter.route(response)

Although the route is associated with the request, not the response,
it is parsed from the response object. The reason is that many server
implementations process the request before they can identify the route.

Instrumentation authors implement support via extending HttpAdapter
accordingly. There are a few patterns which might help.

#### Callback with non-final type
When a framework uses an callback model, you are in control of the type
being parsed. If the response type isn't final, simply subclass it with
the route data.

For example, if Spring MVC, it would be this:
```java
    // check for a wrapper type which holds the template
    handler = HttpServerHandler.create(httpTracing, new HttpServletAdapter() {
      @Override public String template(HttpServletResponse response) {
        return response instanceof HttpServletResponseWithTemplate
            ? ((HttpServletResponseWithTemplate) response).route : null;
      }

      @Override public String toString() {
        return "WebMVCAdapter{}";
      }
    });

--snip--

// when parsing the response, scope the route from the request

    Object template = request.getAttribute(BEST_MATCHING_PATTERN_ATTRIBUTE);
    if (route != null) {
      response = new HttpServletResponseWithTemplate(response, route.toString());
    }
    handler.handleSend(response, ex, span);
```

#### Callback with final type
If the response type is final, you may be able to make a copy and stash
the route as a synthetic header. Since this is a copy of the response,
there's no chance a user will receive this header in a real response.

Here's an example for Play, where the header "brave-http-template" holds
the route temporarily until the parser can read it.
```scala
    result.onComplete {
      case Failure(t) => handler.handleSend(null, t, span)
      case Success(r) => {
        // add a synthetic header if there was a routing path set
        var resp = template.map(t => r.withHeaders("brave-http-template" -> t)).getOrElse(r)
        handler.handleSend(resp, null, span)
      }
    }
--snip--
    override def template(response: Result): String =
      response.header.headers.apply("brave-http-template")
```

#### Common mistakes

For grouping to work, we want routes that are effectively the same, to
in fact be the same. Here are a couple things on that.

* Always start with a leading slash
  * This allows you to differentiate the root path from empty (no route)
  * This prevents accidental partitioning like `users/:userId` from `/users/:userId`
* Take care not to duplicate slashes
  * When joining nested paths, avoid writing templates like `/nested//users/:userId`
  * The `ITHttpServer` test will catch some of this

Fixes https://github.com/openzipkin/zipkin/issues/1874

### DIFF
--- a/instrumentation/http/README.md
+++ b/instrumentation/http/README.md
@@ -9,13 +9,16 @@ instructions on what to put into http spans, and sampling policy.
 
 ## Span data policy
 By default, the following are added to both http client and server spans:
-* Span.name as the http method in lowercase: ex "get"
+* Span.name is the http method in lowercase: ex "get" or a route described below
 * Tags/binary annotations:
   * "http.method", eg "GET"
   * "http.path", which does not include query parameters.
   * "http.status_code" when the status us not success.
   * "error", when there is an exception or status is >=400
 * Remote IP and port information
+
+A route based span name looks like "/users/{userId}", "not_found" or
+"redirected". There's a longer section on Http Route later in this doc.
 
 Naming and tags are configurable in a library-agnostic way. For example,
 the same `HttpTracing` component configures OkHttp or Apache HttpClient
@@ -29,6 +32,7 @@ httpTracing = httpTracing.toBuilder()
     .clientParser(new HttpClientParser() {
       @Override
       public <Req> void request(HttpAdapter<Req, ?> adapter, Req req, SpanCustomizer customizer) {
+        customizer.name(spanName(adapter, req)); // default span name
         customizer.tag(TraceKeys.HTTP_URL, adapter.url(req)); // the whole url, not just the path
       }
     })
@@ -45,13 +49,20 @@ Ex:
 ```java
 overrideSpanName = new HttpClientParser() {
   @Override public <Req> String spanName(HttpAdapter<Req, ?> adapter, Req req) {
-    return adapter.method(req).toLowerCase() + " " + adapter.path(req);
+    // If using JAX-RS, maybe we want to use the resource method
+    if (adapter instanceof ContainerAdapter) {
+      Method method = ((ContainerAdapter) adapter).resourceMethod(req);
+      return method.getName().toLowerCase();
+    }
+    // If not using framework-specific knowledge, we can use http
+    // attributes or go with the default.
+    return super.spanName(adapter, req);
   }
 };
 ```
 
 Note that span name can be overwritten any time, for example, when
-parsing the response.
+parsing the response, which is the case when route-based names are used.
 
 ## Sampling Policy
 The default sampling policy is to use the default (trace ID) sampler for
@@ -87,6 +98,31 @@ httpTracingBuilder.serverSampler(HttpRuleSampler.newBuilder()
   .addRule("POST", "/bar", 0.1f)
   .build());
 ```
+
+## Http Route
+The http route is an expression such as `/items/:itemId` representing an
+application endpoint. `HttpAdapter.route()` parses this from a response,
+returning the route that matched the request, empty if no route matched,
+or null if routes aren't supported. This value is either used to create
+a tag "http.route" or as an input to a span naming function.
+
+### Http route cardinality
+The http route groups similar requests together, so results in limited
+cardinality, often a better choice for a span name vs the http method.
+
+For example, the route `/users/{userId}`, matches `/users/25f4c31d` and
+`/users/e3c553be`. If a span name function used the http path instead,
+it could DOS-style attack vector on your span name index, as it would
+grow unbounded vs `/users/{userId}`. Even if different frameworks use
+different formats, such as `/users/[0-9a-f]+` or `/users/:userId`, the
+cardinality is still fixed with regards to request count.
+
+The http route can be "" (empty) on redirect or not-found. If you use
+http route for metrics, coerce empty to constants like "redirected" or
+"not_found" with the http status. Knowing the difference between not
+found and redirected can be a simple intrusion detection signal. The
+default span name policy uses constants when a route isn't known for
+reasons including sharing the span name as a metrics correlation field.
 
 # Developing new instrumentation
 
@@ -181,3 +217,75 @@ try (Tracer.SpanInScope ws = tracer.withSpanInScope(span)) { // 2.
   handler.handleSend(response, error, span); // 5.
 }
 ```
+
+### Supporting HttpAdapter.route(response)
+
+Although the route is associated with the request, not the response,
+it is parsed from the response object. The reason is that many server
+implementations process the request before they can identify the route.
+
+Instrumentation authors implement support via extending HttpAdapter
+accordingly. There are a few patterns which might help.
+
+#### Callback with non-final type
+When a framework uses an callback model, you are in control of the type
+being parsed. If the response type isn't final, simply subclass it with
+the route data.
+
+For example, if Spring MVC, it would be this:
+```java
+    // check for a wrapper type which holds the template
+    handler = HttpServerHandler.create(httpTracing, new HttpServletAdapter() {
+      @Override public String template(HttpServletResponse response) {
+        return response instanceof HttpServletResponseWithTemplate
+            ? ((HttpServletResponseWithTemplate) response).route : null;
+      }
+
+      @Override public String toString() {
+        return "WebMVCAdapter{}";
+      }
+    });
+
+--snip--
+
+// when parsing the response, scope the route from the request
+
+    Object template = request.getAttribute(BEST_MATCHING_PATTERN_ATTRIBUTE);
+    if (route != null) {
+      response = new HttpServletResponseWithTemplate(response, route.toString());
+    }
+    handler.handleSend(response, ex, span);
+```
+
+#### Callback with final type
+If the response type is final, you may be able to make a copy and stash
+the route as a synthetic header. Since this is a copy of the response,
+there's no chance a user will receive this header in a real response.
+
+Here's an example for Play, where the header "brave-http-template" holds
+the route temporarily until the parser can read it.
+```scala
+    result.onComplete {
+      case Failure(t) => handler.handleSend(null, t, span)
+      case Success(r) => {
+        // add a synthetic header if there was a routing path set
+        var resp = template.map(t => r.withHeaders("brave-http-template" -> t)).getOrElse(r)
+        handler.handleSend(resp, null, span)
+      }
+    }
+--snip--
+    override def template(response: Result): String =
+      response.header.headers.apply("brave-http-template")
+```
+
+#### Common mistakes
+
+For grouping to work, we want routes that are effectively the same, to
+in fact be the same. Here are a couple things on that.
+
+* Always start with a leading slash
+  * This allows you to differentiate the root path from empty (no route)
+  * This prevents accidental partitioning like `users/:userId` from `/users/:userId`
+* Take care not to duplicate slashes
+  * When joining nested paths, avoid writing templates like `/nested//users/:userId`
+  * The `ITHttpServer` test will catch some of this
--- a/instrumentation/jersey-server/src/main/java/brave/jersey/server/TracingApplicationEventListener.java
+++ b/instrumentation/jersey-server/src/main/java/brave/jersey/server/TracingApplicationEventListener.java
@@ -7,15 +7,19 @@ import brave.http.HttpServerHandler;
 import brave.http.HttpTracing;
 import brave.propagation.Propagation.Getter;
 import brave.propagation.TraceContext;
+import java.util.List;
 import javax.inject.Inject;
 import javax.ws.rs.ext.Provider;
 import org.glassfish.jersey.server.ContainerRequest;
 import org.glassfish.jersey.server.ContainerResponse;
+import org.glassfish.jersey.server.ExtendedUriInfo;
 import org.glassfish.jersey.server.ManagedAsync;
+import org.glassfish.jersey.server.internal.routing.RoutingContext;
 import org.glassfish.jersey.server.monitoring.ApplicationEvent;
 import org.glassfish.jersey.server.monitoring.ApplicationEventListener;
 import org.glassfish.jersey.server.monitoring.RequestEvent;
 import org.glassfish.jersey.server.monitoring.RequestEventListener;
+import org.glassfish.jersey.uri.UriTemplate;
 
 @Provider
 public final class TracingApplicationEventListener implements ApplicationEventListener {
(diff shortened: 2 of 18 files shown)


adriancole:
@jplock do you think it makes sense to add jersey2 support to brave (which then your dropwizard project could use instead of normal jaxrs?) If so, we could get the http template as it isn't available at the jax-rs level ex https://github.com/micrometer-metrics/micrometer/blob/9ae63906cc78d44c444dbb43340b5b287d6b645e/micrometer-jersey2/src/main/java/io/micrometer/jersey2/server/DefaultJerseyTagsProvider.java

jplock:
Yea we might have more control over things at the jersey level than just jaxrs

adriancole:
so revamped this based on some experience doing this with vert.x and spring web mvc. TL;DR: seems better to process the template in the response phase even if it is about the request.

here's why:
* request parsing should happen reliably as otherwise if you have an
error, you don't have good context for it. Deferring request parsing
can interfere
* if you have something at request time, but need it at response,
propagating just a string representing the template is relatively safe
(vs holding onto the routing object)
* metrics from spans can only happen after response anyway, so they don't care
* it is less confusing to have one place to parse than two (request/response)

So.. yeah thinking it is more than reasonable for the template
property to be parsed from the response.

adriancole:
next steps: verify this with play, ratpack and jersey2 (otherwise the api might be crap but don't know how yet)

instrumentation/spring-webmvc/src/main/java/brave/spring/webmvc/TracingHandlerInterceptor.java
```diff
@@ -68,6 +82,16 @@ public void afterCompletion(HttpServletRequest request, HttpServletResponse resp
     Span span = tracer.currentSpan();
     if (span == null) return;
     ((SpanInScope) request.getAttribute(SpanInScope.class.getName())).close();
-    handler.handleSend(response, ex, span);
+    Object template = request.getAttribute(BEST_MATCHING_PATTERN_ATTRIBUTE);
+    if (template == null) { // skip thread-local overhead if there's no attribute
```
adriancole: ps I know this stuff is mostly copy/paste between here and vert.x (except here the attribute is an object not a string). I'm letting that pressure build before generalizing code

adriancole:
Here's an example setting the span name to the http template (when using spring MVC on an endpoint with a path variable)

screen shot 2018-02-12 at 6 06 31 pm

instrumentation/spring-webmvc/src/main/java/brave/spring/webmvc/TracingHandlerInterceptor.java
```diff
@@ -68,6 +81,19 @@ public void afterCompletion(HttpServletRequest request, HttpServletResponse resp
...
     if (span == null) return;
     ((SpanInScope) request.getAttribute(SpanInScope.class.getName())).close();
+    Object template = request.getAttribute(BEST_MATCHING_PATTERN_ATTRIBUTE);
+    if (template != null) {
+      response = new HttpServletResponseWithTemplate(response, template.toString());
+    }
     handler.handleSend(response, ex, span);
   }
+
+  static class HttpServletResponseWithTemplate extends HttpServletResponseWrapper {
```
adriancole: cc @llinder @hyleung When an http library provides a wrapper, we get a *big* benefit, which is the ability to stash things through response-shaped holes. You can look at the note on vert.x for the opposite (not saying vert.x is bad, just that wrappers are safer for us to use)

cc @cescoffier

instrumentation/vertx-web/src/main/java/brave/vertx/web/TracingRoutingContextHandler.java
```diff
@@ -23,6 +23,17 @@
...
+ * <h3>Why use a thread local for the http template when parsing {@linkplain HttpServerResponse}?</h3>
+ * <p>When parsing the response, we use a thread local to make the current route's path visible.
+ * This is an alternative to wrapping {@linkplain HttpServerResponse} or declaring a custom type.
+ * We don't wrap {@linkplain HttpServerResponse}, because this would lock the instrumentation to the
+ * signatures currently present on it (for example, if a method is added, we'd have to recompile).
+ * If a wrapper is eventually provided by vertx, we could use that, but it didn't exist at the time.
+ * We could also define a custom composite type like ResponseWithTemplate. However, this would
+ * interfere with people using "instanceof" in http samplers or parsers: they'd have to depend on a
+ * brave type. The least impact means was to use a thread local, as eventhough this costs a little,
+ * it prevents revision lock or routine use of custom types.
```
adriancole: note to self "prevents version lock"

instrumentation/spring-webmvc/src/main/java/brave/spring/webmvc/TracingHandlerInterceptor.java
hyleung: I don't think Ratpack has wrappers out-of-the-box, but iirc we had to _introduce_ wrapper types to work around Ratpack's types (e.g. we don't really get a "Request", but a `RequestSpec` - so we have to wrap it in something that is more request-like).

adriancole: very confident this api is fine for ratpack https://github.com/hyleung/ratpack-zipkin/commit/bb498012771bc11ae32a1be1c6fcf7c6bb2b3ab9

adriancole:
status update: verified this integrates cleanly with ratpack https://github.com/hyleung/ratpack-zipkin/commit/bb498012771bc11ae32a1be1c6fcf7c6bb2b3ab9

next step: check out play and jersey2

adriancole:
synthetic header approach should work with play (even if I can't get it to work right now.. this is more about me and scala+play vs this change)

```scala
    result.onComplete {
      case Failure(t) => handler.handleSend(null, t, span)
      case Success(r) => {
        // add a synthetic header if there was a routing path set
        var resp = template.map(t => r.withHeaders("brave-http-template" -> t)).getOrElse(r)
        handler.handleSend(resp, null, span)
      }
    }
--snip--
    override def template(response: Result): String =
      response.header.headers.apply("brave-http-template")
```

adriancole:
Updated description based on various inputs. cc @rakyll @jkschneider

PS besides notes above (ex need to complete jersey2 prior to merge), I'd like to make a default span namer (which handles redirect etc). However, this requires knowing the difference between no matched route, and the framework doesn't support http routing. Needs some thought and advice welcome.

cc also @llinder @hyleung

adriancole:
added an issue about jersey support https://github.com/openzipkin/brave/issues/611

jkschneider:
@adriancole Careful about instrumenting WebMVC requests with a `HandlerInterceptor` because they are never called when a servlet filter intervenes earlier (e.g. spring-security blocks access to a resource).

Also, in the case of WebMVC, the code that sets `BEST_PATTERN_MATCHING_ATTRIBUTE` does not normalize away extra '/' in the URI. That is the purpose of this [nonsense](https://github.com/micrometer-metrics/micrometer/blob/master/micrometer-spring-legacy/src/main/java/io/micrometer/spring/web/servlet/WebMvcTags.java#L98). I'd suggest this is something to test for in other implementations as well.

Sorry if this is a repeat of something you've heard before :)

adriancole:
Very good notes, jon. One thing interesting in Brave is that it is a
collection of libraries.. we can apply good defaults in Boot (via sleuth)
or DW, but on their own, we can only advise. I think it is worth updating
the WebMVC notes with the hazards as they arent in the readme right now.

We have an old issue #396 which I think plays in here. We could have a
scenerio where instrumentation can "contribute to" but not own the task of
an RPC call. For example, you could imagine undertow owning the latency and
http tags, but webmvc contributing a route name (to the same span, where it
works). This hasnt been well thought through but I suspect we will end up
somewhere nearby.

adriancole:
@jkschneider opened 
https://github.com/openzipkin/brave/issues/612 about the advice on impact of ordering. This should be commodity knowledge as it is indeed very very important

adriancole:
working on a slash deduper..

adriancole:
slash thing wasn't needed afterall. I figured out what jersey is doing

instrumentation/jersey-server/src/main/java/brave/jersey/server/TracingApplicationEventListener.java
```diff
@@ -75,6 +78,34 @@ public RequestEventListener onRequest(RequestEvent requestEvent) {
       return request.getHeaderString(name);
     }
 
+    /**
+     * This returns the matched template as defined by a base URL and path expressions.
+     *
+     * <p>Matched templates are pairs of (resource path, method path). This code skips redundant
+     * slashes from either source caused by Path("/") or implicitly by Path("").
+     */
+    @Override public String route(ContainerResponse response) {
```
adriancole: @jkschneider fyi. feel free to steal this.. managed to get to the bottom of why there were double-slashing. jersey actually normalizes "" to "/" btw. This is a lot more efficient in benches

adriancole: added a commit.. maybe wait until merged :)

adriancole:
did some special casing of route parsing as it can be expensive (started about 3 ops/microsecond). Now, pretty good.

```java
Benchmark                                                            Mode  Cnt   Score   Error   Units
TracingApplicationEventListenerAdapterBenchmarks.parseRoute         thrpt   15  80.678 ± 0.923  ops/us
TracingApplicationEventListenerAdapterBenchmarks.parseRoute_nested  thrpt   15  20.599 ± 0.238  ops/us
```

instrumentation/http/README.md
```diff
@@ -85,6 +88,29 @@ httpTracingBuilder.serverSampler(HttpRuleSampler.newBuilder()
...
+## Http Route
+The http route is an expression such as `/items/:itemId` representing an
+application endpoint. `HttpAdapter.route()` parses this from a response,
+returning the route that matched the request, empty if no route matched,
+or null if routes aren't supported. This value is either used to create
+a tag "http.route" or as an input to a span naming function.
+
+### Http route cardinality
+The http route groups similar requests together, so results in limited
+cardinality, often better choice for a span name vs the http method.
```
jkschneider: often _a_ better choice...

instrumentation/http/README.md
```diff
@@ -85,6 +88,29 @@ httpTracingBuilder.serverSampler(HttpRuleSampler.newBuilder()
...
+
+For example, the route `/users/{userId}`, matches `/users/25f4c31d` and
+`/users/e3c553be`. If a span name function used the http path instead,
+it could DOS-style attack vector on your span name index, as it would
```
jkschneider: it could _result in a_ DOS-style...

instrumentation/http/README.md
```diff
@@ -85,6 +88,29 @@ httpTracingBuilder.serverSampler(HttpRuleSampler.newBuilder()
...
+grow unbounded vs `/users/{userId}`. Even if different frameworks use
+different formats, such as `/users/[0-9a-f]+` or `/users/:userId`, the
+cardinality is still fixed with regards to request count.
+
+The http route can can be empty on redirect or not-found. If you are
```
jkschneider: can ~~can~~

instrumentation/http/README.md
```diff
@@ -85,6 +88,29 @@ httpTracingBuilder.serverSampler(HttpRuleSampler.newBuilder()
...
+using the route for metrics, coerce empty to constants like "redirected"
+or "not_found" using the http status. For example, the default span name
```
jkschneider: ~~For example, t~~The default...

instrumentation/http/README.md
```diff
@@ -177,4 +203,76 @@ try (Tracer.SpanInScope ws = tracer.withSpanInScope(span)) { // 2.
 } finally {
   handler.handleSend(response, error, span); // 5.
 }
-```
\ No newline at end of file
+```
+
+### Supporting HttpAdapter.route(response)
+
+Eventhough the route is associated with the request, not the response,
```
jkschneider: Event though

instrumentation/http/README.md
```diff
@@ -177,4 +203,76 @@ try (Tracer.SpanInScope ws = tracer.withSpanInScope(span)) { // 2.
...
+it is parsed from the response object. The reason is that many server
+implementations process the request before they can identify the route.
+
+Instrumentation authors implement support via extending HttpAdapter
```
jkschneider: extending `HttpAdapter`

instrumentation/http/README.md
```diff
@@ -177,4 +203,76 @@ try (Tracer.SpanInScope ws = tracer.withSpanInScope(span)) { // 2.
...
+accordingly. There are a few patterns which might help.
+
+#### Callback with non-final type
+When a framework uses an callback model, you are in control of the type
```
jkschneider: uses a~~n~~ callback

instrumentation/http/README.md
```diff
@@ -177,4 +203,76 @@ try (Tracer.SpanInScope ws = tracer.withSpanInScope(span)) { // 2.
...
+being parsed. If the response type isn't final, simply subclass it with
+the route data.
+
+For example, if spring MVC, it would be this:
```
jkschneider: in ~~s~~Spring MVC

Issue #508, opened 2017-10-03
instrumenting RabbitMQ via spring-amqp


jonathan-lo:
There's been some interest at my work with instrumenting rabbit interactions that are made using spring-amqp. @adriancole mentioned that there's been previous requests for this instrumentation channel e.g. openzipkin/zipkin#1428.

I can have a look at this if there's sufficient interest. Since spring-amqp has some nice hooks into the consume / produce lifecycle, this looks feasible.

jonathan-lo:
The plan of attack I'm thinking is:
producer via RabbitTemplate - use the MessagePostProcessor hook to inject trace headers / annotate
consumer via RabbitListener - use SimpleRabbitListenerContainerFactory's adviceChain which allows us to specify an aspect to extract the trace headers and handle appropriately

adriancole:
These types all are in the same maven dependency, right? If so
instrumentation-spring-amq sounds right. If any amqp works junit driven
intergration test might be easier as rabbit isnt java. Even if not is ok as
we can add the rabbitmq service to ci

jonathan-lo:
Hmm you're right, whilst RabbitTemplate and RabbitListener is from spring-rabbit, the extension points are from the spring-amqp maven dep, so instrumentation-spring-amqp sounds reasonable.

Will have a play around with the integration tests, at the moment it seems the easiest path would be to add rabbitmq service to ci.

liangxp:
I also want to trace rabbitMQ Message

jonathan-lo:
should we close this now that @adriancole is working on a generic spring-messaging instrumentation?

adriancole:
Well depends. The spring-messaging abstraction doesnt know if a queue is
remote, the topic name etc. It is a bit harder to get right.. Do you use
spring messaging to access spring amqp?

jonathan-lo:
I see. My spike is using spring amqp's abstractions. So the value add here could be digging the exchange / routing key out of the MessageProperties

adriancole:
your spike looks good. only question I have really is if there is another way apart from aop to handle the consumer side. I'm guessing instrumenting with spring-amqp (or spring-rabbit) could have an advantage of decoupling the incompatible rabbit 4.x from 5.x series as well https://github.com/openzipkin/brave/compare/master...jonathan-lo:master

cc @shakuzen

skastenholz:
[...]

PR #603, opened 2018-02-04
spring amqp instrumentation

jonathan-lo:
As per #508, here is a first cut of a spring-amqp based instrumentation for brave.

(btw thanks @skastenholz who saved me from doing some cleanup with my spike!)


instrumentation/spring-amqp/README.md
```diff
@@ -0,0 +1,23 @@
+# brave-instrumentation-spring-amqp
+
+## Tracing for Spring Rabbit
```
adriancole: Maybe retitle to..

Tracing for Spring AMQP (RabbitMQ)

instrumentation/spring-amqp/README.md
```diff
@@ -0,0 +1,23 @@
...
+`TracingMessagePostProcessor` adds trace headers to outgoing
+rabbit messages. It then reports to Zipkin how long each request takes.
+
+To configure tracing for rabbit message producers, 
+ensure `Tracing` is in the Spring context, then create `TracingMessagePostProcessor` bean. 
+Wire this bean as a BeforePublishPostProcessor on the RabbitTemplate that you wish to trace. 
+
+
+### Message Consumer
+Tracing is supported for spring-rabbit @RabbitListener based services.
```
adriancole: backtick `@RabbitListener` or github thinks it is a person

instrumentation/spring-amqp/pom.xml
```diff
@@ -0,0 +1,40 @@
...
+
+	<dependencies>
+		<dependency>
+			<groupId>org.springframework.amqp</groupId>
+			<artifactId>spring-amqp</artifactId>
+			<version>${spring-amqp.version}</version>
+			<scope>provided</scope>
+		</dependency>
+		<dependency>
+			<groupId>org.springframework</groupId>
```
adriancole: add a comment why aop is needed (usually we don't depend on other libraries that aren't direct dependencies)

instrumentation/spring-amqp/README.md
```diff
@@ -0,0 +1,23 @@
+# brave-instrumentation-spring-amqp
+
+## Tracing for Spring Rabbit
+This module provides instrumentation for spring-amqp based services. Currently, the only broker supported by spring-amqp is RabbitMQ.
+
+## Configuration
```
adriancole: Since this uses a non-default dependency on spring-aop, might want to mention this dependency is required (provided won't do this)

pom.xml
```diff
@@ -61,6 +61,7 @@
     <!-- Ensure older versions of spring still work -->
     <spring4.version>4.3.13.RELEASE</spring4.version>
     <spring.version>3.2.18.RELEASE</spring.version>
+    <spring-amqp.version>1.7.5.RELEASE</spring-amqp.version>
```
adriancole: until used in multiple modules, or pinned to a specific spring version, keep defined as a property in the instrumentation module itself. The reason spring is listed here is because multiple dependents on core spring

instrumentation/spring-amqp/src/main/java/brave/spring/amqp/TracingRabbitListenerAdvice.java
```diff
@@ -0,0 +1,47 @@
...
+
+  @Override public Object invoke(MethodInvocation methodInvocation) throws Throwable {
+
+    final Message message = (Message) methodInvocation.getArguments()[1];
+    final TraceContextOrSamplingFlags extracted = extractor.extract(message.getMessageProperties());
+    final Span span = tracer.nextSpan(extracted).kind(Span.Kind.CONSUMER).start();
+    try (SpanInScope ws = tracer.withSpanInScope(span)) {
+      return methodInvocation.proceed();
+    } catch (Throwable t) {
+      span.tag("error", t.getMessage());
```
adriancole: t.getMessage can be null. If you look at httpparser you'll see a familiar dance on this

instrumentation/spring-amqp/src/main/java/brave/spring/amqp/TracingRabbitListenerAdvice.java
```diff
@@ -0,0 +1,47 @@
...
+
+import brave.Span;
+import brave.Tracer;
+import brave.Tracer.SpanInScope;
+import brave.Tracing;
+import brave.propagation.Propagation;
+import brave.propagation.TraceContext.Extractor;
+import brave.propagation.TraceContextOrSamplingFlags;
+
+public class TracingRabbitListenerAdvice implements MethodInterceptor {
```
adriancole: Maybe add some javadoc on why AOP (I know it was discussed before, but the code should know about it)

instrumentation/spring-amqp/src/main/java/brave/spring/amqp/TracingRabbitListenerAdvice.java
```diff
@@ -0,0 +1,47 @@
...
+import org.springframework.amqp.core.MessageProperties;
+
+import brave.Span;
+import brave.Tracer;
+import brave.Tracer.SpanInScope;
+import brave.Tracing;
+import brave.propagation.Propagation;
+import brave.propagation.TraceContext.Extractor;
+import brave.propagation.TraceContextOrSamplingFlags;
+
```
adriancole: Add some javadoc about how to configure this, ex what it applies to

jonathan-lo:
I'll have a look at the integration test next.

@adriancole not sure if the semantics of this is a little quirky as compared to the zipkin span messaging spec? Here's an example of SPMC case:

main

And here's the span information - noting that the span includes not only consuming the message from the broker but also the time taken to process the business logic in the app:

consumer-span

jonathan-lo:
Project to replicate above found at https://github.com/jonathan-lo/rabbit-playground.git

adriancole:
heh looks like random service ids.. that's probably not good :)

jonathan-lo:
hah, that's not the point though!

re-reading the doco and the kafka instrumentation. I'm not sure the current impl has the consumer span concepts totally correct. Taking SPSC as an example, are the following the spans to be recorded?

1. span representing some code does some logic, and sends a message
2. span representing msg produced to broker
3. span representing msg consumed from broker
4. span representing the application logic that acts on the message

jonathan-lo:
separate-spans

The alternative doesn't look too fantastic....

adriancole:
@jonathan-lo what do you mean by "The alternative doesn't look too fantastic...." ?

adriancole:
maybe add a span name so that we can tell what these are doing (even if temporarily or via osx preview "Tools, annotate")

jonathan-lo:
haha sorry, forgot that you can't click into screenshots :) I've updated the above with span names. This is a simple SPMC scenario where there is a http interaction (span 1) that produces a rabbit message (2). Span 3 is the consumer span for the first consumer, and span 4 is the one that 'does stuff' with the message. Span 5 and 6 are the same thing for a second consumer.

instrumentation/spring-amqp/src/main/test/brave/spring/amqp/ITSpringAmqpTracing.java
```diff
@@ -0,0 +1,222 @@
...
+import org.springframework.context.annotation.AnnotationConfigApplicationContext;
+import org.springframework.context.annotation.Bean;
+import org.springframework.context.annotation.Configuration;
+import zipkin2.Span;
+import zipkin2.reporter.Reporter;
+
+import static org.assertj.core.api.Assertions.assertThat;
+import static org.assertj.core.groups.Tuple.tuple;
+
+public class ITSpringAmqpTracing {
```
adriancole: the tests are skipped as they are not under the source root "src/test/java"

instrumentation/spring-amqp/pom.xml
```diff
@@ -0,0 +1,61 @@
+<?xml version="1.0"?>
+<project xmlns="http://maven.apache.org/POM/4.0.0" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
+	xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 http://maven.apache.org/xsd/maven-4.0.0.xsd">
+	<parent>
+		<groupId>io.zipkin.brave</groupId>
+		<artifactId>brave-instrumentation-parent</artifactId>
+		<version>4.14.4-SNAPSHOT</version>
```
adriancole: rebase over master to get latest version, though I don't think anything you are doing is affected by recent versions

instrumentation/spring-amqp/src/main/test/brave/spring/amqp/ITSpringAmqpTracing.java
```diff
@@ -0,0 +1,222 @@
...
+    public TracingRabbitConsumer tracingRabbitConsumer(CountDownLatch countDownLatch) {
+      return new TracingRabbitConsumer(countDownLatch);
+    }
+
+    @Bean CountDownLatch messageReceivedLatch() {
+      return new CountDownLatch(1);
+    }
+  }
+
+  private static class TracingRabbitProducer {
```
adriancole: nit: in your test class avoid the prefix Tracing unless it is an instrumented type. That's because our normal instrumentation types are prefixed Tracing. I was looking to see if this class was in main. Maybe instead use just "RabbitProducer" or "HelloWorldProducer"

instrumentation/spring-amqp/src/main/java/brave/spring/amqp/TracingRabbitListenerAdvice.java
```diff
@@ -0,0 +1,62 @@
...
+import brave.propagation.TraceContext.Extractor;
+import brave.propagation.TraceContextOrSamplingFlags;
+import org.springframework.amqp.rabbit.listener.SimpleMessageListenerContainer;
+
+/**
+ * TracingRabbitListenerAdvice is an AOP advice to be used with the {@link
+ * SimpleMessageListenerContainer#setAdviceChain(Advice...) SimpleMessageListenerContainer} in order
+ * to add tracing functionality to a spring-amqp managed message consumer. While a majority of
+ * brave's instrumentation points is implemented using a delegating wrapper approach, this extension
+ * point was ideal as it covers both programmatic and @RabbitListener annotation driven amqp consumers.
```
adriancole: `@RabbitListener` won't do what you want in javadoc. If you want it to look like that I think you need to wrap it in `{@code @RabbitListener}`

instrumentation/spring-amqp/src/main/java/brave/spring/amqp/TracingRabbitListenerAdvice.java
```diff
@@ -0,0 +1,62 @@
...
+      };
+  private final Extractor<MessageProperties> extractor;
+  private final Tracer tracer;
+
+  public TracingRabbitListenerAdvice(Tracing tracing) {
+    this.extractor = tracing.propagation().extractor(GETTER);
+    this.tracer = tracing.tracer();
+  }
+
+  @Override public Object invoke(MethodInvocation methodInvocation) throws Throwable {
```
adriancole: so this will apply to any method invocation, right? Feels like the methods that are consumer in nature (polling a message) would be traced differently than ones that are processor in nature (ex RabbitListeners)

I'll pull the code to try and figure out if this is the case or not

jonathan-lo: Right, so this is only meant to be used with the SimpleMessageListenerContainer (or DirectMessageListenerContainer in 2.0+), and would wrap a span around the *processing* - this is shown in the first screenshot that I'd attached earlier.

If we were to also have a separate span for the actual message consumption (e.g. the 2nd screenshot) that is possible with some changes. The processing would then become a SERVER span

jonathan-lo: Having said that I'd need to dig a bit further into the spring-amqp source as well

adriancole:
we should rename to brave-instrumentation-spring-rabbit as that's the entrypoint dep

adriancole:
interestingly, we haven't really discussed a hybrid modeling, but if the
consumption and processing is single step, then yeah it is pretty much
feeling like a server span. Not sure if we have dependency graph code to
cover this "interpretation"

adriancole:
opened openzipkin/zipkin#1931 about modeling. it is an interesting challenge

adriancole:
OK thought a lot about this, and have the following advice for wrapping this up.

* rename this to spring-rabbit (including package names etc)
* Introduce a type SpringRabbitTracing which is initialized like `KafkaTracing`. This will be the decorator
* change to a object factory approach, as this will let us change things later. This is described below
* Make all types except SpringRabbitTracing package private (allows us to change code later)
* In the advice do create the consumer span, but immediately end it and start a new one for the processor. I'd remove the trace headers from the message once you've extracted them.

So, instead of this..
```java
    @Bean
    public TracingRabbitListenerAdvice tracingRabbitListenerAdvice(Tracing tracing) {
      return new TracingRabbitListenerAdvice(tracing);
    }

    @Bean
    public SimpleRabbitListenerContainerFactory rabbitListenerContainerFactory(
        ConnectionFactory connectionFactory, Advice tracingRabbitListenerAdvice) {
      SimpleRabbitListenerContainerFactory factory = new SimpleRabbitListenerContainerFactory();
      factory.setConnectionFactory(connectionFactory);
      factory.setAdviceChain(tracingRabbitListenerAdvice);
      return factory;
    }
```

This..
```java
    @Bean
    public SpringRabbitTracing springRabbitTracing(Tracing tracing) {
      return SpringRabbitTracing.create(tracing);
    }

    @Bean
    public SimpleRabbitListenerContainerFactory rabbitListenerContainerFactory(
        SpringRabbitTracing springRabbitTracing,
        ConnectionFactory connectionFactory) {
      return springRabbitTracing.newSimpleRabbitListenerContainerFactory(connectionFactory);
    }
```

Internally, `newSimpleRabbitListenerContainerFactory` does the setup dance and can change independently. Same thing for `newRabbitTemplate()`

Sound good?

jonathan-lo:
Good tips. Having better encapsulation allows us to change the implementation later down the track easier.

I had a scary thought that by doing so would mean that my *real* use case - spring integration - wouldn't work. Thankfully this isn't the case :relieved:

adriancole:
after rebase, please add this to our bomb instrumentation/bom/pom.xml

that way people can more easily force a version

jonathan-lo:
oh btw i'll add received exchange / routing key / queue info as tags in the next changeset

instrumentation/spring-rabbit/README.md
```diff
@@ -0,0 +1,23 @@
+# brave-instrumentation-spring-rabbit
+
+## Tracing for Spring AMQP (RabbitMQ)
```
adriancole: Let's just remove the AMQP part and call it Spring Rabbit here and in the description

instrumentation/spring-rabbit/README.md
```diff
@@ -0,0 +1,23 @@
...
+
+### Message Producer
+This module contains a tracing interceptor for [Spring RabbitTemplate](https://spring.io/TODO/).
+`TracingMessagePostProcessor` adds trace headers to outgoing rabbit messages. 
+It then reports to Zipkin how long each request takes.
+
+To configure tracing for rabbit message producers, 
+ensure `Tracing` is in the Spring context, then create `TracingMessagePostProcessor` bean. 
+Wire this bean as a BeforePublishPostProcessor on the RabbitTemplate that you wish to trace. 
+
```
adriancole: add javaconfig example here and in the consume section based on what's in the IT

instrumentation/spring-rabbit/pom.xml
```diff
@@ -0,0 +1,61 @@
...
+
+	<properties>
+		<main.basedir>${project.basedir}/../..</main.basedir>
+		<main.java.version>1.6</main.java.version>
+		<main.signature.artifact>java16</main.signature.artifact>
+		<spring-amqp.version>1.7.5.RELEASE</spring-amqp.version>
+	</properties>
+
+	<dependencies>
+		<dependency>
```
adriancole: maybe remove the amqp dep as it is transitive from spring-rabbit, then change the variable name to spring-rabbit.version

instrumentation/spring-rabbit/pom.xml
```diff
@@ -0,0 +1,61 @@
...
+			<version>${spring-amqp.version}</version>
+			<scope>compile</scope>
+		</dependency>
+
+		<!-- spring-aop is required for the consumer side instrumentation point -->
+		<dependency>
+			<groupId>org.springframework</groupId>
+			<artifactId>spring-aop</artifactId>
+			<version>${spring4.version}</version>
+			<scope>compile</scope>
```
adriancole: this probably should be provided scope else you'll override the version a user chooses. same thing for spring-rabbit's dep

instrumentation/spring-rabbit/src/main/java/brave/spring/rabbit/SpringRabbitTracing.java
```diff
@@ -0,0 +1,50 @@
...
+
+  private SpringRabbitTracing(Tracing tracing) {
+    this.tracing = tracing;
+    this.tracingMessagePostProcessor = new TracingMessagePostProcessor(tracing);
+    this.tracingRabbitListenerAdvice = new TracingRabbitListenerAdvice(tracing);
+  }
+
+  /**
+   * Creates an instrumented rabbit template.
+   * @param connectionFactory
```
adriancole: kill the param and return tags here and the next method

instrumentation/spring-rabbit/src/main/java/brave/spring/rabbit/TracingRabbitListenerAdvice.java
```diff
@@ -0,0 +1,105 @@
...
+   */
+  @Override public Object invoke(MethodInvocation methodInvocation) throws Throwable {
+    final Message message = (Message) methodInvocation.getArguments()[1];
+    final TraceContextOrSamplingFlags extracted = extractTraceContextAndRemoveHeaders(message);
+
+    final Span consumerSpan = tracer.nextSpan(extracted).kind(CONSUMER).start();
+    tagReceivedMessageProperties(consumerSpan, message.getMessageProperties());
+    consumerSpan.finish();
+
+    final Span serverSpan = tracer.newChild(consumerSpan.context()).kind(SERVER).start();
```
adriancole: I know this was confusing due to earlier discussion, but server isn't a child of consumer, rather it is either/or. Just remove kind here is good (rename serverSpan to listenerSpan)

instrumentation/spring-rabbit/src/main/java/brave/spring/rabbit/TracingRabbitListenerAdvice.java
```diff
@@ -0,0 +1,105 @@
...
+    tagReceivedMessageProperties(serverSpan, message.getMessageProperties());
```
adriancole: only tag on the consumer span (you can still search no problem) here, you probably should just set a span name based on the invocation's method name

jonathan-lo:
the span name idea is proving a little tricky - our aop advice is around a ContainerDelegate's method call and doesn't have access to the actual listener / processor

adriancole:
maybe use the name "on-message" for now?

jonathan-lo:
yeah thats good enough for now. also the name of the consumer span should be the queue that we consumed from imo

adriancole:
Usually the name of a span doesnt have variables in it rather the operation
name.

adriancole:
anyway tentatively, suppose this is fine to have the queue name as the
span name. I suspect it will be limited cardinality and we can revisit

adriancole:
interestingly we haven't named the consumer spans in kafka! in OT they call them "send" and "receive" sometimes with the channel name as a space delimted suffix

adriancole:
Thanks very much!

PR #605, opened 2018-02-08
Fixes bug where we didn't merge extra fields from implicit context

adriancole:
While unusual, and likely a leak, it is possible to have a span in scope
when extracting a remote context. This fixes the code to merge the extra
fields in the current span with extracted fields and backfills tests.

Thanks to @aldex32 and @aukevanleeuwen for hunting this down

### DIFF
--- a/brave/src/main/java/brave/propagation/ExtraFieldPropagation.java
+++ b/brave/src/main/java/brave/propagation/ExtraFieldPropagation.java
@@ -167,7 +167,7 @@ public final class ExtraFieldPropagation<K> implements Propagation<K> {
    *
    * <p>Prefer {@link #set(TraceContext, String, String)} if you have a reference to a span.
    */
-   public static void set(String name, String value) {
+  public static void set(String name, String value) {
     TraceContext context = currentTraceContext();
     if (context != null) set(context, name, value);
   }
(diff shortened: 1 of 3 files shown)


brave/src/main/java/brave/propagation/ExtraFieldPropagation.java
```diff
@@ -252,32 +252,55 @@ public static void set(TraceContext context, String name, String value) {
...
+      // If there's an implicit context while extracting only extra fields, we will have two extras!
+      List<Object> extras = result.extra();
+      int thisExtraIndex = -1, parentExtraIndex = -1;
+      for (int i = 0, length = extras.size(); i < length; i++) {
+        if (extras.get(i) instanceof Extra) {
+          Extra extra = (Extra) result.extra().get(i);
+          if (!Arrays.equals(extra.fieldNames, fieldNames)) {
+            throw new IllegalStateException(
+                String.format("Mixed name configuration unsupported: found %s, expected %s",
+                    Arrays.asList(extra.fieldNames), Arrays.asList(fieldNames))
```
aukevanleeuwen: I find it rather hard to wrap my head around, so I don't have a lot of comments other than you can use `Arrays.toString(...)` instead of create a complete new list here. But yeah, it's in the exception flow and not in the critical path or something... :-)

adriancole: good hint!

adriancole:
@llinder or @devinsba this is waiting for zipkin-reporter-java 2.3.2 to sync to central (should happen soon, but I'll be on the plane).

When that's in, can you merge this and cut release-4.15.0? Note that brave uses travis for release.

cc also @jcchavezs

adriancole:
maven just synced I'll cut the release!

aldex32:
@adriancole thanks for the quick fix!

adriancole:
>
> @adriancole <https://github.com/adriancole> thanks for the quick fix!
>
so you are all good now? no more 3GB of trace contexts?

aldex32:
@adriancole it's not any more in {{PROD}}, now it needs to pass through all environments again. But I will let you know!
Thanks again!

aldex32:
@adriancole now this fix is running on our `PROD` environment for one month, and till now seems everything good.

PR #606, opened 2018-02-11
Deflakes async tests by switching to concurrent blocking queue

adriancole:
This removes some Thread.sleep and other guards from tests by changing
assertions to use a blocking queue. It also improves some of the state
checking, particularly around leaked trace contexts. This uncovered two
where async instrumentation (vertx and netty) redundantly scoped a span.

### DIFF
--- a/instrumentation/http-tests/src/main/java/brave/http/ITHttp.java
+++ b/instrumentation/http-tests/src/main/java/brave/http/ITHttp.java
@@ -1,33 +1,117 @@
 package brave.http;
 
 import brave.Tracing;
+import brave.internal.HexCodec;
 import brave.propagation.B3Propagation;
 import brave.propagation.CurrentTraceContext;
 import brave.propagation.ExtraFieldPropagation;
 import brave.propagation.StrictCurrentTraceContext;
 import brave.propagation.TraceContext;
 import brave.sampler.Sampler;
-import java.util.Map;
-import java.util.concurrent.ConcurrentLinkedDeque;
+import java.util.concurrent.BlockingQueue;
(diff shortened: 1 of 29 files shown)


adriancole:
fyi @marcingrzejszczak @llinder @devinsba @shakuzen @hyleung

Took a flight or two, but this gets rid of all the async "fudge" sleeps I could find. This is indirectly for the upcoming "http.template" code, as that is built on the maintenance work on this PR.

instrumentation/http-tests/src/main/java/brave/http/ITHttp.java
```diff
@@ -1,33 +1,117 @@
...
 import okhttp3.mockwebserver.MockWebServer;
 import org.junit.After;
 import org.junit.Rule;
-import org.junit.rules.ExpectedException;
+import zipkin2.Annotation;
 import zipkin2.Span;
 
 import static org.assertj.core.api.Assertions.assertThat;
 
+/**
```
adriancole: interesting, albeit long-winded notes here

instrumentation/http-tests/src/main/java/brave/http/ITHttp.java
```diff
@@ -37,22 +121,22 @@
         .spanReporter(s -> {
           // make sure the context was cleared prior to finish.. no leaks!
           TraceContext current = httpTracing.tracing().currentTraceContext().get();
+          boolean contextLeak = false;
           if (current != null) {
-            assertThat(current.spanId())
```
adriancole: this was completely busted check. the new code is correct and found two bugs

PR #613, opened 2018-02-19
Registers jaxrs only when a client or server

adriancole:
cc @jplock @cschneider I'm pretty sure we should have been doing this


jplock:
Looks good to me

PR #615, opened 2018-02-22
Http status is usually known, prefer using a primitive

adriancole:
There are only 2 instrumentations who don't always know the http status.
This changes to avoid boxing where possible, which can reduce memory
pressure. The status 0 is used when unknown.


instrumentation/http/src/main/java/brave/http/HttpParser.java
```diff
@@ -40,11 +40,14 @@
   // If this were not an abstraction, we'd use separate hooks for response and error.
   public <Resp> void response(HttpAdapter<?, Resp> adapter, @Nullable Resp res,
       @Nullable Throwable error, SpanCustomizer customizer) {
-    Integer httpStatus = res != null ? adapter.statusCode(res) : null;
-    if (httpStatus != null && (httpStatus < 200 || httpStatus > 299)) {
-      customizer.tag("http.status_code", String.valueOf(httpStatus));
+    int statusCode = 0;
```
adriancole: this starts to structure this code in a way that is used for the next PR about http route

PR #617, opened 2018-02-22
Exposes Jax-RS client filter public, so it can be used independently

adriancole:
We shouldn't use brave-instrumentation-jersey-server with the
TracingFeature as it will duplicate server instrumentation. This allows
the client filter to be re-used independently, such as in DropWizard.


adriancole:
cc @jplock

adriancole:
templating works nicely with dropwizard

screen shot 2018-02-22 at 5 13 06 pm

screen shot 2018-02-22 at 5 10 44 pm

PR #622, opened 2018-02-25
Fix result type of HttpServerHandler.create()

takezoe:
Fix result type of `HttpServerHandler.create()` to expose type parameters.


takezoe:
@adriancole I've just found it. :-) Can I push a fix for `HttpClientHandler` to this pull request as well? or should I create an another pull request?

adriancole:
Yeap same pull request is fine. Thanks!

adriancole:
going out in 4.16.2 (as will help cleanup play tracing, so worth pushing out!)

Thanks for the help!

PR #623, opened 2018-02-25
Adds Dubbo instrumentation 新年快乐

### LINKED ISSUE #504: Add support for dubbo
hello，i can not find duubo in instrumentation，can you help me ， thanks
https://github.com/alibaba/dubbo

adriancole:
This is a Tracing filter for RPC providers and consumers in [Dubbo](http://dubbo.io/books/dubbo-dev-book-en/impls/filter.html)

When used on a consumer, `TracingFilter` adds trace state as attachments
to outgoing requests. When a provider, it extracts trace state from
incoming requests. In either case, the filter reports to Zipkin how long
each request took, along with any error information.

Note: A Dubbo Provider is a server, and a Dubbo Consumer is a client in
Zipkin terminology.

screen shot 2018-02-25 at 3 12 09 pm

## Configuration

The filter "tracing" requires an extension of type `Tracing` configured.
Most typically, this is provided by Spring, so have this in place before
proceeding. Here's an example in [XML](https://github.com/openzipkin/brave/tree/master/spring-beans/README.md).

Then, configure the filter named "tracing" in your producer or consumer.

Here's an example of using Spring 2.5+ XML
```xml
<!-- a provider -->
<dubbo:service filter="tracing" interface="com.alibaba.dubbo.demo.DemoService" ref="demoService"/>

<!-- or a consumer -->
<dubbo:reference filter="tracing" id="demoService" check="false" interface="com.alibaba.dubbo.demo.DemoService"/>
```

谢谢 @blacklau @sdcuike and @jessyZu who wrote dubbo filters for Brave in the past

Fixes #504

### DIFF
--- a/instrumentation/README.md
+++ b/instrumentation/README.md
@@ -5,6 +5,7 @@ example, the directory "servlet" includes the artifact "brave-instrumentation-se
 
 Here's a brief overview of what's packaged here:
 
+* [dubbo-rpc](dubbo-rpc/README.md) - Tracing filter for RPC providers and consumers in [Dubbo](http://dubbo.io/books/dubbo-user-book-en/)
 * [grpc](grpc/README.md) - Tracing client and server interceptors for [grpc](github.com/grpc/grpc-java)
 * [httpasyncclient](httpasyncclient/README.md) - Tracing decorator for [Apache HttpClient](https://hc.apache.org/httpcomponents-asyncclient-dev/) 4.0+
 * [httpclient](httpclient/README.md) - Tracing decorator for [Apache HttpClient](http://hc.apache.org/httpcomponents-client-4.4.x/index.html) 4.3+
--- a/instrumentation/dubbo-rpc/README.md
+++ b/instrumentation/dubbo-rpc/README.md
@@ -1,4 +1,4 @@
-# brave-instrumentation-dubbo-grpc
+# brave-instrumentation-dubbo-rpc
 This is a tracing filter for RPC providers and consumers in [Dubbo](http://dubbo.io/books/dubbo-dev-book-en/impls/filter.html)
 
 When used on a consumer, `TracingFilter` adds trace state as attachments
--- a/instrumentation/dubbo-rpc/README.md
+++ b/instrumentation/dubbo-rpc/README.md
@@ -1,0 +1,28 @@
+# brave-instrumentation-dubbo-grpc
+This is a tracing filter for RPC providers and consumers in [Dubbo](http://dubbo.io/books/dubbo-dev-book-en/impls/filter.html)
+
+When used on a consumer, `TracingFilter` adds trace state as attachments
+to outgoing requests. When a provider, it extracts trace state from
+incoming requests. In either case, the filter reports to Zipkin how long
+each request took, along with any error information.
+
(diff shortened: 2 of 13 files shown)


adriancole:
cc @chickenlj

adriancole:
by the way, the screen shot was from the normal dubbo demo. I changed its config to say "tracing" in the filter list, added the dependency and spring XML for the Tracing component.

I suspect it would be similarly easy with https://github.com/alibaba/dubbo-spring-boot-starter and sleuth, but not tried, yet.

aderm:
good

chickenlj:
Hi,  adriancole 

This is really a good news for dubbo users.

But as a new one to `openzipkin`, it really takes some time for me to integrate `TracingFilter` into a dubbo demo project. 

I think it would be better if we can put the configuration of `[tracing](https://github.com/openzipkin/brave/tree/master/spring-beans)` in more prominent position so users can easily notice that.

And also, for users who choose not to use spring, i think we should also point out the way of how to use inject `Tracer`.

What do you think?

Best Regards

adriancole:
>
> But as a new one to openzipkin, it really takes some time for me to
> integrate TracingFilter into a dubbo demo project.
>
thanks for spending time on this. very valuable to have feedback especially
as not released yet.

I think it would be better if we can put the configuration of [tracing](
> https://github.com/openzipkin/brave/tree/master/spring-beans) in more
> prominent position so users can easily notice that.
>
Fo you mean this should be easier to notice in general, or easier to notice
in the dubbo README or something else? What should change in your opinion?

> And also, for users who choose not to use spring, i think we should also
> point out the way of how to use inject Tracer.
>
Do you mean like show how to make a
manual com.alibaba.dubbo.common.extension.ExtensionFactory for tracing? or
is there a different extension approach you have in mind? manual one won't
be too hard I think.

adriancole:
opened issue about api configuration as this task might be complex as whatever we doc should mention somehow to shutdown the reporter https://github.com/openzipkin/brave/issues/632

adriancole:
fyi I have a working example, but it isn't great due to issues noted in this issue. Please follow there for next work https://github.com/openzipkin/brave/issues/632#issuecomment-370355437

chickenlj:
Sorry for late reply.

> Fo you mean this should be easier to notice in general, or easier to notice
> in the dubbo README or something else? What should change in your opinion?

I mean `easier to notice in the dubbo-rpc README`, can be just a link. Just a suggestion, maybe configuration of  [`tracing`]( https://github.com/openzipkin/brave/tree/master/spring-beans) is a common sense for users familiar with openzipkin. 

> Do you mean like show how to make a
> manual com.alibaba.dubbo.common.extension.ExtensionFactory for tracing? or
> is there a different extension approach you have in mind? manual one won't
> be too hard I think.

Yes, a ExtensionFactory can be one way for creating `tracing` instance. And i am also thinking of users to  create a tracing instance manually and inject into ExtensionLoader as an extension.

chickenlj:
This is what i want, sorry i failed to notice this part in [README](https://github.com/openzipkin/brave/tree/master/instrumentation/dubbo-rpc)

> The filter "tracing" requires an extension of type Tracing configured. Most typically, this is provided by Spring, so have this in place before proceeding. Here's an example in XML.

PR #624, opened 2018-02-27
Adds instrumentation bom for pinning a version

### LINKED ISSUE #619: brave bom
A brave bom would allow applications to easily force-align all versions of the brave instrumentation libraries used.
```
<dependency>
        <artifactId>brave-bom</artifactId>
        <version>${brave.version}</version>
        <type>pom</type>
        <scope>import</scope>
      </dependency>
```

This could be especially useful for applications using brave via sleuth-starter, since it does not define all instrumentation libraries. Also, brave is probably going to crank out new versions more often than sleuth, so having to update just the bom version would be an easy way to track them.

@adriancole @devinsba unsure if we need different boms for brave instrumentation and zipkin reporters?

adriancole:
Thanks @jorgheymans for the start on this.

Please check my advice is valid cc @shakuzen @devinsba @marcingrzejszczak 

Fixes #619


instrumentation/README.md
```diff
@@ -31,6 +31,44 @@ Here are other tools we provide for configuring or testing instrumentation:
...
 
+### Version alignments
+When using multiple instrumentation packages, you'll want to align
+versions in one place. This allows you to more safely upgrade, with less
+worry about conflicts.
+
+You can use our Maven instrumentation BOM (Bill of Materials) for this:
+
+Ex. in your dependencies section, import the BOM like this:
+```
```
shakuzen: The pickiest of nits: add `xml` for prettier code highlighting.

adriancole:
>
> The pickiest of nits: add xml for prettier code highlighting.
>
good catch anyway!

jorgheymans:
Good stuff, is  the reporter bom also needed still?

adriancole:
Not sure really. We can wait until someone asks for that and then do it in
zipkin-reporter.

PR #627, opened 2018-03-01
Guards against malformed trace extraction without allocating exceptions

### LINKED ISSUE #625: don't die or do expensive things on malformed input
We should add tests that malformed B3 headers don't throw, or if they do we catch it. This was identified in a related discussion about Jaeger's B3 codec.

Here, I think we should make a special hex parser that skips out on malformed (to differentiate it from null). This can log the failure at debug level as opposed to initializing an exception, throwing then catching that.

we should also benchmark malformed propagation to ensure we don't accidentally create overhead or miss parts. This should include extra field propagation and AWS

related issue https://github.com/jaegertracing/jaeger-client-java/issues/354

adriancole:
This makes sure when we parse bad data, we can log, but don't do things
expensive in the mean time. This does so by inlining parsing.

Fixes #625

### DIFF
--- a/brave-tests/src/test/java/brave/propagation/B3PropagationTest.java
+++ b/brave-tests/src/test/java/brave/propagation/B3PropagationTest.java
@@ -48,4 +48,27 @@ public class B3PropagationTest extends PropagationTest<String> {
     assertThat(result)
         .isEqualTo(SamplingFlags.NOT_SAMPLED);
   }
+
+  @Test public void extractTraceContext_malformed() {
+    MapEntry mapEntry = new MapEntry();
+    map.put("X-B3-TraceId", "463ac35c9f6413ad48485a3953bb6124"); // ok
+    map.put("X-B3-SpanId",  "48485a3953bb6124"); // ok
+    map.put("X-B3-ParentSpanId", "-"); // not ok
+
+    SamplingFlags result = propagation().extractor(mapEntry).extract(map).samplingFlags();
+
+    assertThat(result)
+        .isEqualTo(SamplingFlags.EMPTY);
+  }
+
+  @Test public void extractTraceContext_malformed_sampled() {
+    MapEntry mapEntry = new MapEntry();
+    map.put("X-B3-TraceId", "-"); // not ok
+    map.put("X-B3-Sampled", "1"); // ok
+
+    SamplingFlags result = propagation().extractor(mapEntry).extract(map).samplingFlags();
+
+    assertThat(result)
+        .isEqualTo(SamplingFlags.EMPTY);
+  }
 }
--- a/brave/src/main/java/brave/propagation/B3Propagation.java
+++ b/brave/src/main/java/brave/propagation/B3Propagation.java
@@ -4,7 +4,6 @@ import java.util.Arrays;
 import java.util.Collections;
 import java.util.List;
 
-import static brave.internal.HexCodec.lowerHexToUnsignedLong;
 import static brave.internal.HexCodec.toLowerHex;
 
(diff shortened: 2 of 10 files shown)


adriancole:
laptop says this can fail B3 73+ times per microsecond, so probably ok

brave/src/main/java/brave/propagation/B3Propagation.java
```diff
@@ -115,35 +114,29 @@
...
-      result.spanId(lowerHexToUnsignedLong(spanId));
-      String parentSpanIdString = getter.get(carrier, propagation.parentSpanIdKey);
-      if (parentSpanIdString != null) result.parentId(lowerHexToUnsignedLong(parentSpanIdString));
-      return TraceContextOrSamplingFlags.create(result.build());
+      // Try to parse the span and parent ID into the context
+      if (result.parseSpanId(getter, carrier, propagation.spanIdKey)
+          && result.parseParentId(getter, carrier, propagation.parentSpanIdKey)) {
+        return TraceContextOrSamplingFlags.create(result.sampled(sampledV).debug(debug).build());
+      }
+      return TraceContextOrSamplingFlags.EMPTY; // trace context is malformed so return empty
```
jcchavezs: So we don't even use the sample information if the parse of span/trace ID fail?

adriancole:
If the trace context is malformed yeah I don't think there is value
trusting the sampling decision either. It is simpler this way and also
hopefully will help people find bugs.

adriancole:
Keep in mind a well formed request is still good. Ex sampled without trace
id is ok. Only if a trace id is malformed should we kick out.

brave-tests/src/test/java/brave/propagation/B3PropagationTest.java
```diff
@@ -48,4 +48,15 @@
     assertThat(result)
         .isEqualTo(SamplingFlags.NOT_SAMPLED);
   }
+
+  @Test public void extractTraceContext_malformed() {
+    MapEntry mapEntry = new MapEntry();
+    map.put("x-b3-traceid", "1"); // ok
```
jcchavezs: What about declaring more examples here on what happens by using a data provider? that would be more expressive in terms on which cases this will/won't fail. What do you think?

adriancole:
backfilled tests including that it doesn't log by default

adriancole:
>
> What about declaring more examples here on what happens by using a data
> provider? that would be more expressive in terms on which cases this
> will/won't fail. What do you think?
>
there's a fair amount of redundancy in the trace context tests vs here, so
was just testing lightly

>

adriancole:
circleci is busted due to broken config on the bom which is fixed externally #628

adriancole:
so benchmark results for extraction in normal case are almost exactly the same (within error range)

benchmark results for extracting a bad context are much different.


Scenario where there's a malformed parent ID, but everything else is OK (note: we toss the context based on malformed)

catch exception approach
```java
Benchmark                                    Mode  Cnt   Score   Error   Units
PropagationBenchmarks.extract_b3            thrpt   15  10.564 ± 0.469  ops/us
PropagationBenchmarks.extract_b3_malformed  thrpt   15   0.570 ± 0.027  ops/us
```

current code approach
```java
Benchmark                                    Mode  Cnt   Score   Error   Units
PropagationBenchmarks.extract_b3            thrpt   15  11.283 ± 0.424  ops/us
PropagationBenchmarks.extract_b3_malformed  thrpt   15  12.410 ± 0.380  ops/us
```

llinder:
Nice test cases :)

PR #628, opened 2018-03-01
WIP on BOM

marcingrzejszczak:
- added a separate pom with licenses etc. (central will require this info from any module) called `parent-pom` (didn't have an idea for a better name)
- the `brave-instrumentation` BOM doesn't use the `parent`, but the  `parent-pom`, that's why it doesn't inherit any concrete library versions such as spring etc.

IMO what's missing is a `brave-all-bom` that contains all brave projects with their versions

### DIFF
--- a/brave-bom/pom.xml
+++ b/brave-bom/pom.xml
@@ -1,18 +1,18 @@
 <project xmlns="http://maven.apache.org/POM/4.0.0"
-    xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
-    xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 http://maven.apache.org/xsd/maven-4.0.0.xsd">
+         xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
+         xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 http://maven.apache.org/xsd/maven-4.0.0.xsd">
   <modelVersion>4.0.0</modelVersion>
 
   <parent>
     <groupId>io.zipkin.brave</groupId>
     <artifactId>brave-pom-parent</artifactId>
     <version>4.16.3-SNAPSHOT</version>
-    <relativePath>../../brave-parent</relativePath>
+    <relativePath>../brave-parent</relativePath>
   </parent>
 
-  <artifactId>brave-instrumentation-bom</artifactId>
-  <name>Brave Instrumentation BOM</name>
-  <description>Bill Of Materials POM for all Brave instrumentation artifacts</description>
+  <artifactId>brave-bom</artifactId>
+  <name>Brave BOM</name>
+  <description>Bill Of Materials POM for all Brave artifacts</description>
   <packaging>pom</packaging>
 
   <properties>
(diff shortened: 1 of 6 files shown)


marcingrzejszczak:
Corresponding Sleuth branch - https://github.com/spring-cloud/spring-cloud-sleuth/tree/issues_%23872_brave_bom

adriancole:
makes sense.. maybe we should just have one bom with everything? doesn't matter what (as long as it is for external consumption eg not benchmarks)

cc @shakuzen @jorgheymans

jorgheymans:
not sure about the brave-all, how would you assign it a version number for example ?  This can get confusing to explain : "in order to depend on brave 4.16 with zipkin 2.4.5 you need to have brave-all 1.2.4" .

adriancole:
I mean things under io.zipkin.brave in maven central (which share a common
version)

think brave jar, slf4j context and a few instrumentation

jorgheymans:
I thought that was the purpose of brave-bom? 

At least from my viewpoint brave = core + instrumentation + reporter because that is what I need to get things working hence it would be nice to have them all in one go. Reporters are used without brave iiuc, that's why they do not follow the same release cycle. 

But from a user perspective it's confusing having to keep track of three version numbers (zipkins, brave, reporters) but I'm digressing perhaps.

marcingrzejszczak:
Yeah, as a user I'd like to just import the Brave BOM and only pick the dependencies. `core + instrumentation + reporter` makes sense.

marcingrzejszczak:
Let's do whatever suits the maintainers. From the user POV I can add 2 boms if necessary - I don't care ;)

adriancole:
If we do one bom per subdir you will have to import...

One for instrumentation
One for slf4j (and later grpc context)
One for amazon propagation (later also for trace-context)


Also
One for zipkin reporter

People using spring legacy also need to track spring beans


To me it doesnt make sense to even users to make so much work. That is what
I am trying to figure out. Gut feel is that in hind sight will be a mess.

Imagine having to explain to a user that to update grpc you need to import
and override 3 boms (just for brave!)

Lets go with one bom. We can still refactor later to make many boms but we
cant undo the many boms once we publish them.

I think we should follow same in reporter, too.

Open to ideas just dont want to dig a hole.. make sense?

marcingrzejszczak:
Makes sense to me

adriancole:
I dont have a lot of experience but the only other thing i can think of is
an externally managed brave bom. Ex one that could potentially also set
versions for other things like brave cassandra.

Downside is this would be manually updated like a release train thing (vs
brave bom here which automatically happens). Not sure we are ready for
release train..

marcingrzejszczak:
Maybe now it looks better?

adriancole:
Typo in slf4j but yeah ;) maybe missing some other things.. i can help
audit.

Again we can always later refactor brave-bom to include other boms.

Question before fix and merge.. should we do reporter first and have
brave-bom import that?

adriancole:
thanks, Marcin!

adriancole:
ps we can't really take action on the zipkin-reporter q until a bom exists, so going to do that now

jorgheymans:
> Question before fix and merge.. should we do reporter first and have brave-bom import that?

if reporter is used standalone you could have a reporter-bom and let brave-bom import that, I don't know about the audience that will actually use this reporter-bom outside of brave. 

You might also consider creating a repository just for the brave-bom then and move it there, because its release cycle will no longer be exclusively tied to brave-* , see https://github.com/FasterXML/jackson-bom

adriancole:
>
> if reporter is used standalone you could have a reporter-bom and let
> brave-bom import that, I don't know about the audience that will actually
> use this reporter-bom outside of brave.
>
download stats suggest reporter is more often downloaded than brave by a
significant amount (of IPs). It was made for applications that write zipkin
(including wingtips etc). Also folks scraping logs use the reporter
sometime.

> You might also consider creating a repository just for the brave-bom then
> and move it there, because its release cycle will no longer be exclusively
> tied to brave-* , see https://github.com/FasterXML/jackson-bom
>
"because its release cycle will no longer be exclusively tied to brave-*"
what do you mean by that? Right now it isn't. Are you saying that if we
import reporter we should then move brave-bom to a separate repo?

jorgheymans:
> what do you mean by that? Right now it isn't. Are you saying that if we import reporter we should then move brave-bom to a separate repo?

my bad, it's not tied to the brave releasy cycle I made a wrong assumption sorry. No need to move it to separate repo either, this can always be done later if really needed.

PR #629, opened 2018-03-02
Adds base test fixture for WebMVC based tests

adriancole:
This will allow us to backfill tests that are in sleuth which require
different types of configuration classes.


adriancole:
fixed by https://github.com/openzipkin/brave/pull/643

PR #631, opened 2018-03-03
Adds Java 9 module names with manifest entries

adriancole:
This is the easy way out of getting module names stable and mapped to
our entrypoint packages. I've only done this for the highly re-used
jars.

See square/okhttp#3743
thx @yschimke

### DIFF
--- a/brave-tests/pom.xml
+++ b/brave-tests/pom.xml
@@ -34,6 +34,16 @@
   <build>
     <plugins>
       <plugin>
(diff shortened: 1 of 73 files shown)


adriancole:
verified with `(find . -name \*4.17.1-SNAPSHOT.jar -exec /Library/Java/JavaVirtualMachines/jdk-9.jdk/Contents/Home/bin/jar -d -f {} \;)|less`

PR #633, opened 2018-03-03
Adds SpringRabbitTracing.Builder.remoteServiceName

adriancole:
`SpringRabbitTracing.Builder.remoteServiceName` allows you to set the
name the broker will have in the service graph. Usually you'll want a
service-scoped name like a cluster or similar.

This also adjusts the span names to always be operation names like other
instrumentation.


adriancole:
@jonathan-lo I took liberties here, knowing we'll end up needing parsers like we have in grpc, cassandra etc at some point. This policy is most like other code (including OT).

PR #634, opened 2018-03-03
Adds missing names for Kafka spans

adriancole:
Somehow we forgot to add span names based on the operations before!


adriancole:
cc @ImFlog @jonathan-lo whoops

instrumentation/kafka-clients/src/main/java/brave/kafka/clients/TracingProducer.java
```diff
@@ -69,7 +69,7 @@
       if (record.key() instanceof String && !"".equals(record.key())) {
         span.tag(KafkaTags.KAFKA_KEY_TAG, record.key().toString());
       }
-      span.tag(KafkaTags.KAFKA_TOPIC_TAG, record.topic()).kind(Kind.PRODUCER).start();
+      span.tag(KafkaTags.KAFKA_TOPIC_TAG, record.topic()).name("send").kind(Kind.PRODUCER).start();
```
adriancole: these names are simply based on the java method, which is the safest choice imho

PR #635, opened 2018-03-04
Makes it possible for tools layered on servlet to contribute tags and http.route

adriancole:
Layers over servlet, like web mvc have important data to add to a span, but shouldn't create new spans or finish spans started at the servlet layer. Such an approach keeps latency coherent and also helps make sure we have less chance of lifecycle bugs (like double-reporting)

At the very least, these "tagging interceptors or filters" set the servlet request attribute "http.route" with the best matching route as well any tags like the controller method. Since it is only adding tags, it doesn't need reference to a tracer, rather "span customizer".

This implementation looks at the request attribute "http.route" and when there,
exposes it to the response parsing phase. This should make instrumentation based
on servlet much more straight-forward to write. Tagging policy at this layer is
framework-specific.

See https://github.com/spring-cloud/spring-cloud-sleuth/pull/880

### DIFF
--- a/instrumentation/http-tests/src/main/java/brave/test/http/ITHttp.java
+++ b/instrumentation/http-tests/src/main/java/brave/test/http/ITHttp.java
@@ -12,7 +12,6 @@ import brave.sampler.Sampler;
 import java.util.concurrent.BlockingQueue;
 import java.util.concurrent.LinkedBlockingQueue;
 import java.util.concurrent.TimeUnit;
-import okhttp3.mockwebserver.MockWebServer;
 import org.junit.After;
 import org.junit.Rule;
 import org.junit.rules.TestRule;
@@ -82,8 +81,6 @@ import static org.assertj.core.api.Assertions.assertThat;
 public abstract class ITHttp {
   public static final String EXTRA_KEY = "user-id";
   static final String CONTEXT_LEAK = "context.leak";
-
-  @Rule public MockWebServer server = new MockWebServer();
 
   /**
    * When testing servers or asynchronous clients, spans are reported on a worker thread. In order
--- a/instrumentation/http-tests/src/main/java/brave/test/http/ITHttpClient.java
+++ b/instrumentation/http-tests/src/main/java/brave/test/http/ITHttpClient.java
@@ -13,10 +13,12 @@ import brave.propagation.SamplingFlags;
 import brave.sampler.Sampler;
 import java.util.Arrays;
 import okhttp3.mockwebserver.MockResponse;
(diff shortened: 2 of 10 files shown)


adriancole:
cc @marcingrzejszczak @jkschneider I think this will be handy as it teaches people to share knowledge about the http.route when they have it. This is useful regardless of if people are using brave, imho.

adriancole:
fyi this now starts integrating as I was hoping it would. There's a test fail, but should be obvious the intent

adriancole:
still working on this.. want to make sure the filters layer nicely i.e. solely rely on servlet

adriancole:
OK, this formalizes the name SpanCustomizingXXX for filters or interceptors that layer over (specifically brave.servlet.TracingFilter) only adding controller-specific data and/or the http route.

I have some more builders, tests to look into, and readmes to fix but once this is in, I think we can generically recommend any framework that layers over servlet to integrate like this.

cc @jplock not sure if dropwizard layers over servlet, or if it is independent.. lemme know

jplock:
@adriancole yes I believe it does, eventually using Jetty for the actual serving (see http://www.dropwizard.io/1.2.2/docs/manual/core.html#how-it-s-glued-together)

adriancole:
closer still. want to make a spring-specific servlet filter wrapper to reduce the pain of people switching off doing tracing via MVC handlers

adriancole:
smoked out some problems with servlet 2.5 integration (missing methods etc). The spring-webmvc README is up to date and accurate. Need to work on the other two

adriancole:
fyi this is generally deprecating JAX-RS container filters as we know they are crap (along with web mvc).

adriancole:
ok everything works now. readmes are up to date.. will backfill tests and split into a few commits

adriancole:
I messed up test fixtures yesterday. fixing them and cleaning up

adriancole:
pulled out https://github.com/openzipkin/brave/pull/642 which this will layer over (reducing the code count)

adriancole:
pulled out https://github.com/openzipkin/brave/pull/643 for the spring webmvc layer

adriancole:
this lowers access needed for customization https://github.com/openzipkin/brave/pull/644

adriancole:
pulled jersey into a separate pull request https://github.com/openzipkin/brave/pull/645

adriancole:
pulled out https://github.com/openzipkin/brave/pull/646 for jaxrs

adriancole:
added instructions for sparkjava https://github.com/openzipkin/brave/pull/649

adriancole:
all merged at this point

PR #636, opened 2018-03-06
Added an option to wrap current spring rabbit compontents

marcingrzejszczak:
without this change you can only explicitly create a RabbitTemplate or a SimpleRabbitListenerContainerFactory. In case of frameworks like Sleuth that want to wrap existing instances that's not possible.

with this change in a hacky (RabbitTemplate) and non-hacky (SRLCF) way we're allowing to instrument existing objects. For RT we're hacking into a field of before post processors and injecting there the traced instrumentation. For SRLCF we're adding the traced advice to an existing factory

### DIFF
--- a/instrumentation/spring-rabbit/src/main/java/brave/spring/rabbit/SpringRabbitTracing.java
+++ b/instrumentation/spring-rabbit/src/main/java/brave/spring/rabbit/SpringRabbitTracing.java
@@ -1,6 +1,14 @@
 package brave.spring.rabbit;
 
+import java.lang.reflect.Field;
+import java.util.ArrayList;
+import java.util.Arrays;
+import java.util.Collection;
+import java.util.List;
+
 import brave.Tracing;
+import org.aopalliance.aop.Advice;
+import org.springframework.amqp.core.MessagePostProcessor;
 import org.springframework.amqp.rabbit.config.SimpleRabbitListenerContainerFactory;
 import org.springframework.amqp.rabbit.connection.ConnectionFactory;
 import org.springframework.amqp.rabbit.core.RabbitTemplate;
@@ -60,6 +68,36 @@ public final class SpringRabbitTracing {
     return rabbitTemplate;
   }
 
+  /** Instruments an existing rabbit template. */
+  public RabbitTemplate fromRabbitTemplate(RabbitTemplate rabbitTemplate) {
+    try {
+      MessagePostProcessor[] processorsArray = currentProcessorsWithTracing(
+              rabbitTemplate);
+      rabbitTemplate.setBeforePublishPostProcessors(processorsArray);
+      return rabbitTemplate;
+    }
+    catch (NoSuchFieldException | IllegalAccessException e) {
+      return rabbitTemplate;
+    }
+  }
+
+  private MessagePostProcessor[] currentProcessorsWithTracing(
+          RabbitTemplate rabbitTemplate)
+          throws NoSuchFieldException, IllegalAccessException {
+    Field field = RabbitTemplate.class
+            .getDeclaredField("beforePublishPostProcessors");
+    Collection<MessagePostProcessor> processors =
+            (Collection<MessagePostProcessor>) field.get(rabbitTemplate);
+    List<MessagePostProcessor> newProcessors = new ArrayList<>();
+    if (!processors.contains(tracingMessagePostProcessor)) {
+      newProcessors.add(tracingMessagePostProcessor);
+    }
+    if (processors != null) {
+      newProcessors.addAll(processors);
+    }
+    return newProcessors.toArray(new MessagePostProcessor[] {});
+  }
+
   /**
    * Creates an instrumented SimpleRabbitListenerContainerFactory to be used to consume rabbit
    * messages.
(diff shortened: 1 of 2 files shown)


instrumentation/spring-rabbit/src/main/java/brave/spring/rabbit/SpringRabbitTracing.java
```diff
@@ -60,6 +68,36 @@ public RabbitTemplate newRabbitTemplate(ConnectionFactory connectionFactory) {
     return rabbitTemplate;
   }
 
+  /** Instruments an existing rabbit template. */
+  public RabbitTemplate fromRabbitTemplate(RabbitTemplate rabbitTemplate) {
```
adriancole: nit: maybe decorateRabbitTemplate?

instrumentation/spring-rabbit/src/main/java/brave/spring/rabbit/SpringRabbitTracing.java
```diff
@@ -60,6 +68,36 @@ public RabbitTemplate newRabbitTemplate(ConnectionFactory connectionFactory) {
...
+
+  private MessagePostProcessor[] currentProcessorsWithTracing(
+          RabbitTemplate rabbitTemplate)
+          throws NoSuchFieldException, IllegalAccessException {
+    Field field = RabbitTemplate.class
+            .getDeclaredField("beforePublishPostProcessors");
+    Collection<MessagePostProcessor> processors =
+            (Collection<MessagePostProcessor>) field.get(rabbitTemplate);
+    List<MessagePostProcessor> newProcessors = new ArrayList<>();
+    if (!processors.contains(tracingMessagePostProcessor)) {
```
adriancole: not sure this will work as we don't do any deep equals etc. Maybe we should just look for an instance of?

instrumentation/spring-rabbit/src/main/java/brave/spring/rabbit/SpringRabbitTracing.java
```diff
@@ -71,4 +109,28 @@ public SimpleRabbitListenerContainerFactory newSimpleMessageListenerContainerFac
...
+    return factory;
+  }
+
+  private Advice[] chainListWithTracing(Advice[] chain) {
+    List<Advice> currentChain = Arrays.asList(chain);
+    List<Advice> chainList = new ArrayList<>();
+    if (!currentChain.contains(tracingRabbitListenerAdvice)) {
+      chainList.add(tracingRabbitListenerAdvice);
+    }
+    if (chain != null) {
```
adriancole: if this were null would line 126 NPE?

instrumentation/spring-rabbit/src/main/java/brave/spring/rabbit/SpringRabbitTracing.java
```diff
@@ -71,4 +109,28 @@ public SimpleRabbitListenerContainerFactory newSimpleMessageListenerContainerFac
...
+      chainList.addAll(Arrays.asList(chain));
+    }
+    return chainList.toArray(new Advice[] {});
```
adriancole: neat trick: new Advice[0]

instrumentation/spring-rabbit/src/main/java/brave/spring/rabbit/SpringRabbitTracing.java
```diff
@@ -71,4 +118,39 @@ public SimpleRabbitListenerContainerFactory newSimpleMessageListenerContainerFac
...
+          SimpleRabbitListenerContainerFactory factory) {
+    Advice[] modifiedAdviceChain = chainListWithTracing(
+            factory.getAdviceChain());
+    factory.setAdviceChain(modifiedAdviceChain);
+    return factory;
+  }
+
+  private Advice[] chainListWithTracing(Advice[] chain) {
+    if (chain == null) {
+      return chain;
```
adriancole: do you want this? or do you want to return a singleton array with the tracingRabbitListenerAdvice?

adriancole:
will buff this on the way in. Thanks @marcingrzejszczak!

instrumentation/spring-rabbit/README.md
```diff
@@ -52,8 +54,10 @@ public SimpleRabbitListenerContainerFactory rabbitListenerContainerFactory(
     ConnectionFactory connectionFactory,
     SpringRabbitTracing springRabbitTracing
 ) {
-  return springRabbitTracing.newSimpleMessageListenerContainerFactory(connectionFactory);
+  return springRabbitTracing.newSimpleRabbitListenerContainerFactory(connectionFactory);
```
adriancole: @jonathan-lo heads up. fixing this typo in a patch on green

instrumentation/spring-rabbit/src/main/java/brave/spring/rabbit/SpringRabbitTracing.java
```diff
@@ -53,102 +51,98 @@ public SpringRabbitTracing build() {
...
     RabbitTemplate rabbitTemplate = new RabbitTemplate(connectionFactory);
     rabbitTemplate.setBeforePublishPostProcessors(tracingMessagePostProcessor);
     return rabbitTemplate;
   }
 
-  /** Instruments an existing rabbit template. */
+  /** Instruments an existing {@linkplain RabbitTemplate} */
   public RabbitTemplate decorateRabbitTemplate(RabbitTemplate rabbitTemplate) {
+    // Skip out if we can't read the field for the existing post processors
+    if (beforePublishPostProcessorsField == null) return rabbitTemplate;
```
marcingrzejszczak: why not just set the value of the field with the processor?

adriancole: if this is null, we had a reflection error which I'm not assuming will get better :)

PR #637, opened 2018-03-07
Clears up instructions for dubbo config

adriancole:
@xionghuiCoder @chickenlj @mercyblitz

### DIFF
--- a/instrumentation/dubbo-rpc/README.md
+++ b/instrumentation/dubbo-rpc/README.md
@@ -10,19 +10,60 @@ Note: A Dubbo Provider is a server, and a Dubbo Consumer is a client in
 Zipkin terminology.
 
 ## Configuration
+The filter "tracing" requires an extension of type `brave.Tracing` named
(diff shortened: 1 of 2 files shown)


kimmking:
awesome

chickenlj:
Hi, @adriancole 

Great job! And for the problems you met, i think maybe dubbo can do the following improvements:

1. Injection by type
2. Simple lifecycle support for SPIs.

I will raise a discussion on dubbo community, will let you know if any decisions.

adriancole:
@chickenlj thanks. the injection by type you may be able to solve with a platform-specific hook:
https://github.com/openzipkin/brave/blob/master/instrumentation/spring-webmvc/src/main/java/brave/spring/webmvc/WebMvcRuntime.java#L51

for lifecycle, possibly a shutdown hook mechanism internal so that users don't have to. nice conversation here:
https://github.com/census-instrumentation/opencensus-java/pull/1050

PR #643, opened 2018-03-11
Adds Spring WebMVC delegating filter and span customizing interceptors

adriancole:
This adds `DelegatingTracingFilter` so that Spring WebMVC tracing can
layer on top of normal servlet tracing. This reduces the amount of
concerns the code has to consider, by moving them down a layer.

This introduces two "span customizing" interceptors:
* `SpanCustomizingHandlerInterceptor` for Spring WebMVC 2.5
* `SpanCustomizingAsyncHandlerInterceptor` for Spring WebMVC 3+

The above just add controller tags and routing information. The
additional tags can be disabled by via the bean `HandlerParser.NOOP`

The `DelegatingTracingFilter` is a helper for configuring the normal
servlet `TracingFilter` as it gets its contents from the Spring
application context.

Finally, there are some fixes related to Spring 2.5 we didn't catch
earlier: this adds runtime detection to help keep that working.


adriancole:
this is the most important change for spring-cloud-sleuth cc @shakuzen @marcingrzejszczak

adriancole:
verified in spring mvc 2.5, 3 and 4 https://github.com/openzipkin/brave-webmvc-example/pull/23

PR #645, opened 2018-03-11
Adds Jersey Server span customizing event listener

adriancole:
This introduces `SpanCustomizingApplicationEventListener` which layers
over the servlet `TracingFilter` to add resource tags and routing
information.

The additional tags can be disabled by binding `EventParser.NOOP`


adriancole:
cc @jplock layering the new event listener over servlet isn't necessarily a better choice for dropwizard, but it is a good choice when the container layer is handled by servlet.

I saw your note "eventually using Jetty for the actual serving (see http://www.dropwizard.io/1.2.2/docs/manual/core.html#how-it-s-glued-together)" but didn't audit to see if DW actually requires routing through servlet, or whether it is conventional. Anyway both modes are still supported

PR #646, opened 2018-03-11
Deprecates JAX-RS TracingContainerFilter in favor of layering on Servlet

adriancole:
The `TracingContainerFilter` had a number of problems including error
handling and route information. This deprecates it and the
`TracingFeature` in favor of tracing at a lower level such as servlet or
Jersey. If there are no better choices for resource parsing the
`SpanCustomizingContainerFilter` can add tags correspondingly. These
can be modified or removed by providing `ContainerParser.NOOP`


adriancole:
cc @cschneider @reta @jplock This is a severe move to deprecate, but the ContainerFilter is actually rubbish as it doesn't trace reliably.

PR #647, opened 2018-03-11
Defaults to create dependency links for messaging spans

adriancole:
This adds default service names for kafka and rabbitmq spans so that
dependency links are created by default. Set to null to disable this.


adriancole:
cc @ImFlog @jonathan-lo @marcingrzejszczak @shakuzen

PR #650, opened 2018-03-12
Exposes Tracer.currentSpanCustomizer() and Span.customizer() for safety

adriancole:
Eventhough we made a safe user type `SpanCustomizer`, we didn't provide
a safe alternative to `Tracer.currentSpan()`. This led to routine
leaking of `brave.Span` to users (even if they need to cast to discover
this). While we did have `CurrentSpanCustomizer`, that relied on current
context to work. Particularly the new servlet instrumentation uses
explicit references, which are cheaper.

This adds a couple utilities and uses them:
`Tracer.currentSpanCustomizer()` - safer than `currentSpan()`
`Span.customizer()` - safely masks lifecycle methods from parsers

These are safer as they don't have the ability to abend the span, and
have a chance of holding less state (ex NoopSpanCustomizer is constant).

Future work will introduce `UnsafeSpan`, which can rely on these methods
to ensure higher performance in single-threaded scenarios such as
synchronous calls.

### DIFF
--- a/brave/README.md
+++ b/brave/README.md
@@ -484,7 +484,7 @@ operation.
 is generally a safe object to expose to third-party code to add tags.
 
 `Tracer.currentSpan()` should be reserved for framework code that cannot
-reference the span explicitly which to close or abandon an operation.
+reference the span it wants to finish, flush or abandon explicitly.
 
 `Tracer.nextSpan()` uses the "current span" to determine a parent. This
 creates a child of whatever is in-flight.
--- a/brave/README.md
+++ b/brave/README.md
@@ -478,9 +478,16 @@ noot cache the result. Instead, look them up each time you need them.
 ## Current Span
 
 Brave supports a "current span" concept which represents the in-flight
-operation. `Tracer.currentSpan()` can be used to add custom tags to a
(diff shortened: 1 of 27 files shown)


adriancole:
cc @shakuzen @zeagord @anuraaga @yschimke @wu-sheng @jonathan-lo @marcingrzejszczak 
this is mainly bulletproofing. Later cheaper instrumentation will layer over this.

yschimke:
In a number of cases you get a SpanCustomizer (potentially a NoopSpanCustomizer) and pass it down through other methods.  Is checking instanceof NoopSpanCustomizer the preferred way to avoid any additional work ala Log4j, isDebugEnabled?

yschimke:
Looks good, I'll adopt it once it lands.

adriancole:
@yschimke I don't mind making NoopSpanCustomizer public type. sadly we can't add an isNoop() method to customizer because it is an interface and we have a lot of <java8 code still

adriancole:
I will make a note for brave5 about customizer should have isNoop regardless

adriancole:
not sure timeline on follow-up for unsafe.. might be next release but anyway here's the description https://github.com/openzipkin/brave/issues/651

PR #652, opened 2018-03-12
fix typo in Tracing.java

discostu105:


brave/src/main/java/brave/Tracing.java
```diff
@@ -69,7 +69,7 @@ public final Clock clock(TraceContext context) {
   static volatile Tracing current = null;
 
   /**
-   * Returns the most recently created tracer iff its component hasn't been closed. null otherwise.
```
adriancole: heh this was if and only iff, but probably if is fine and less distracting :)

adriancole:
thanks

PR #653, opened 2018-03-13
Fixes timestamp race condition in rabbit tracing

adriancole:


adriancole:
fixed testing with sleuth

screen shot 2018-03-13 at 12 24 10 pm

Logic-32:
@adriancole , was the bug here that you finished a parent span prior to initializing the child span?  In doing so, the SpanMap got cleared out so the Clock was no longer associated with the TraceContext?

instrumentation/spring-rabbit/src/main/java/brave/spring/rabbit/TracingRabbitListenerAdvice.java
```diff
@@ -67,16 +67,19 @@
     TraceContextOrSamplingFlags extracted = extractTraceContextAndRemoveHeaders(message);
 
     // named for BlockingQueueConsumer.nextMessage, which we can't currently see
-    Span consumerSpan = tracer.nextSpan(extracted).kind(CONSUMER).name("next-message").start();
+    Span consumerSpan = tracer.nextSpan(extracted).kind(CONSUMER).name("next-message");
+    Span listenerSpan = tracer.newChild(consumerSpan.context()).name("on-message");
```
adriancole: @Logic-32 the bug was yeah a race.. the underlying recorder loses the clock for the parent on finish, so I create an unstarted span here to fix the race. It is a workaround really

Logic-32: Would it be worthwhile to have a SpanMap like equivalent for TraceContext which simply tracks the clock?  Would you be able to guarantee the same weakness of the Map while still maintaining consistency between arbitrarily created TraceContexts?

adriancole:
I thought about this and think we can "fix" messaging spans easier because
the consumer and processor are on the same thread.. always so far at least
both kafka and rabbit non batch processors. Thread local cache of last few
clocks (which are effectively 2 longs) is possible in other words.

Another option is letting trace contexts in the existing weak map go stale
a bit. For example, we mark them finished but dont eagerly purge them.

We need to be careful that we dont create too much overhead here as this is
fine tuning.

Issue #661, opened 2018-03-21
[BUG] DEBUG-trace is not always reported


tgeens:
Assumption: even using NEVER_SAMPLE, a debug trace should always be sampled. 

From the B3-Propagation spec:
> When Debug is set, the trace should be reported to the tracing system and also override any collection-tier sampling policy. Debug implies an accept sampling decision.

There is a bug under the following conditions:

* service tracer has sampler that is **not** ALWAYS_SAMPLE
* _X-B3-Flags_: 1
* _X-B3-Sampled_ is not set
* _X-B3-TraceId/SpanId/ParentSpanId_ is correctly set

Actual result: the `X-B3-Flags: 1` trace is NOT sampled.
Expected result: the `X-B3-Flags: 1` trace IS sampled.

According to the B3-Propagation spec, this set of propagation headers are expected in a downstream service:
> Debug is encoded as X-B3-Flags: 1. Debug implies an accept decision: don't also send X-B3-Sampled: 1.

In the method `B3Propagation#extract(C carrier)` the parameters are extracted from the carrier and a `TraceContextOrSamplingFlags` is returned. This object is created with `TraceContextOrSamplingFlags.create(result.sampled(sampledV).debug(debug).build())` which is a _TraceContext_ with field `flags = 8`.

This results in `traceContext.debug()` returning `true` and `traceContext.sampled()` returning `null`.  Strictly speaking, this is correct: the extractor only found a _debug_ flag and not a sampling decision. Please note that this is a bit of an anomaly: for `SamplingFlags.DEBUG` both the fields `sampled` and `debug` are set to `true`.

But let's assume that a _TraceContext_ with debug enabled, without a sampling decision set, is correct:

In `Tracer#joinSpan(TraceContext context)` the context is queried to find out a sampling decision has been made or not:

```
// If we are joining a trace, we are sharing IDs with the caller
// If the sampled flag was left unset, we need to make the decision here
if (context.sampled() == null) { // then the caller didn't contribute data
      context = context.toBuilder().sampled(sampler.isSampled(context.traceId())).build();
}
```

The sampler is asked to make a decision on wether this trace should be sampled or not. Note that `.debug()` is never checked. If the sampler makes the decision to not sample this trace, the new context has the flags: `sample=false` and `debug=true`.

[...]

PR #662, opened 2018-03-21
Fixes issue #661 - debug trace not sampled

tgeens:
The test `SamplingFlagsTest#debugImpliesSampled` uses a Builder to construct a SamplingFlags object. The problem is that the Builder contains the logic that implicitly enables sampling, when the debug flag is set. But not all sampling-flags are created using that Builder!

The solution provided here is moving the logic back to the `.sampled()` method, for both:
1. `SamplingFlagsImpl.sampled()` and
2. the `static Boolean sampled(int flags)` used by `TraceContext` and `TraceIdContext`

This should resolve issue #661

### DIFF
--- a/brave/src/main/java/brave/propagation/SamplingFlags.java
+++ b/brave/src/main/java/brave/propagation/SamplingFlags.java
@@ -61,7 +61,7 @@ public abstract class SamplingFlags {
     }
 
     @Override public Boolean sampled() {
-      return sampled;
+      return debug ? Boolean.TRUE : sampled;
     }
 
     @Override public boolean debug() {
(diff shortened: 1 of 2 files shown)


adriancole:
sorry I didn't see this. I appreciate your help, but since I've made https://github.com/openzipkin/brave/pull/663 with tests and such are you ok with me closing this?

adriancole:
otherwise you can merge my changes into yours.. your choice

adriancole:
thank you for this change. I fixed it in a different way in #663 which allows me to immediately release a patch for you. Please do help again next time!

PR #663, opened 2018-03-21
Fixes bug where debug did not imply sampled

### LINKED ISSUE #661: [BUG] DEBUG-trace is not always reported

adriancole:
It is true the b3-propagation spec suggests debug implies sampled.
This leniently parses a missing `X-B3-Sampled` header as long as
`X-B3-Flags: 1` is present.

Fixes #661

### DIFF
--- a/brave-tests/src/test/java/brave/propagation/B3PropagationTest.java
+++ b/brave-tests/src/test/java/brave/propagation/B3PropagationTest.java
@@ -72,4 +72,16 @@ public class B3PropagationTest extends PropagationTest<String> {
     assertThat(result)
         .isEqualTo(SamplingFlags.EMPTY);
   }
+
+  @Test public void extractTraceContext_debug_with_ids() {
+    MapEntry mapEntry = new MapEntry();
+    map.put("X-B3-TraceId", "463ac35c9f6413ad48485a3953bb6124"); // ok
+    map.put("X-B3-SpanId", "48485a3953bb6124"); // ok
+    map.put("X-B3-Flags", "1"); // accidentally missing sampled flag
+
+    TraceContext result = propagation().extractor(mapEntry).extract(map).context();
+
+    assertThat(result.sampled())
+        .isTrue();
+  }
 }
--- a/brave/src/main/java/brave/propagation/TraceIdContext.java
+++ b/brave/src/main/java/brave/propagation/TraceIdContext.java
@@ -224,9 +224,12 @@ public final class TraceIdContext extends SamplingFlags {
       return this;
     }
 
+    /** Ensures sampled is set when debug is */
     InternalBuilder debug(boolean debug) {
       if (debug) {
         flags |= FLAG_DEBUG;
+        flags |= FLAG_SAMPLED_SET;
+        flags |= FLAG_SAMPLED;
       } else {
         flags &= ~FLAG_DEBUG;
       }
--- a/brave/src/test/java/brave/propagation/TraceIdContextTest.java
+++ b/brave/src/test/java/brave/propagation/TraceIdContextTest.java
@@ -32,6 +32,20 @@ public class TraceIdContextTest {
         .isEqualTo("00000000000000de000000000000014d");
   }
 
+  @Test public void debugImpliesSampled() {
+    TraceIdContext primitives = context.toBuilder()
+        .debug(true)
+        .build();
+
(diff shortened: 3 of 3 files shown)


tgeens:
What is the reasoning to handle this in the builder and not in the sampled() method ?

Extraction from the B3-headers should work correctly with this PR, but I expect this assertion would fail:

```
assertThat(context.toBuilder().debug(true).sample(null).build().sample()).isTrue();
```

adriancole:
>
> What is the reasoning to handle this in the builder and not in the
> sampled() method ?
>
because the context is immutable, so there's no need to conditionally
consider this at runtime (the extra logic isn't needed). debug is very edge
case and it is also distracting to have the most used code (sampled) have
to make decisions for something rarely used and also implies a fixed
decision. So it is mainly about what we are optimizing for.. I don't like
optimizing for the odd things.

> Extraction from the B3-headers should work correctly with this PR, but I
> expect this assertion would fail:
>
> assertThat(context.toBuilder().debug(true).sample(null).build().sample()).isTrue();
>
> debug is already a very edge-case feature, one that almost no-one uses.
When it is used, it is often used alongside the sampled flag.
for example, the chrome plugin does this.. that's why this issue itself
wouldn't hit many. So, let's keep in mind the likelihood of it ever being
set in this way.
I added propagation tests, for the off-chance.

Now, in your test case, you are talking about a trace context manually
changed, in a strange order, which is even more edge case..

For example, sampled isn't set to null after a decision is made. A decision
is persistent in the trace once made. It basically doesn't make sense even
if it made sense to manually change a context.

Sorry this particular line is getting more attention than a lot of the
codebase :) I'm going to stop explaining now and hope I answered.. still
lots to do!

PR #666, opened 2018-03-30
Span Error Handler

devinsba:
This supports the case where certain backend systems support errors as a first class object.

XRay supports encoding errors in their model directly. Having a configurable error handler in brave will allow a `zipkin-aws` functionality for storing them for reporting to the backend


adriancole:
For the sake of conventions, lets switch this to line up with the other parsers and use a default implementation. We've been using classes to prevent subtle changes from affecting people.. and basically everywhere we use interfaces has been difficult. Here's something I just ripped from HttpParser. Mind swapping out your interface for that? I don't think we need a noop as we already have noop span and this isn't done in the other parsers.

```java
public class ErrorParser {
  /**
   * Override to change what data from the error are parsed into the span modeling it. By
   * default, this tags "error" as the message or simple name of the type.
   */
  public void error(Throwable error, SpanCustomizer customizer) {
    String message = message = error.getMessage();
    if (message == null) message = error.getClass().getSimpleName();
    customizer.tag("error", message);
  }
}
```

With this in place, the existing code can use it.. ex changing HttpParser, GrpcParser etc subtypes..
make sense?

adriancole:
thanks for taking a crack at this.. has been on the mind for a long time @devinsba!

adriancole:
discussed with @devinsba offline. I will take this on to try and help land this in time for sleuth's next RC. This means we can obviate more code there

adriancole:
made changes according to feedback so far. need some README updates etc

adriancole:
Note: I added a commit to make this work with the new scoped span without forcing the latter to implement SpanCustomizer https://github.com/openzipkin/brave/pull/666/commits/f1e3a0c2946d53f9620402c4745cb8c6dba23aef

adriancole:
here's WIP of sleuth consuming this feature https://github.com/spring-cloud/spring-cloud-sleuth/pull/928

adriancole:
https://github.com/openzipkin/brave/pull/684 for XML

Issue #656, opened 2018-03-13
Keep clocks around longer


adriancole:
Processor spans happen after consumer ones.. the current "clock sharing" is scoped to span objects. The problem is that a consumer span opens and closes before a processor one does. When not using JRE 9, this can create a small amount of distracting skew

<img width="965" alt="screen shot 2018-03-13 at 3 17 39 pm" src="https://user-images.githubusercontent.com/64215/37327546-b5a852d6-26d1-11e8-89a2-51f23c4780fe.png">

In rabbit, we can fake this out as we are simulating spans anyway https://github.com/openzipkin/brave/pull/653

In general, I think we need to change the implicit clock sharing to be based on the trace, not the parent. Also, it should try to keep clocks around a while in case there's some small lag before the next span starts.

Logic-32:
Looked at your commit @adriancole, was the bug here that you finished a parent span prior to initializing the child span?  In doing so, the SpanMap got cleared out so the Clock was no longer associated with the TraceContext?

I ask because, associating a clock with a TraceContext should mean that clock sharing is based on the trace.  But I can see how storing it in the SpanMap would make it not actually behave that way.

adriancole:
yep you got it. To do this for gaps (or even concurrent inbound requests
for the same trace), we'd need a different cache, based on traceId I think.

PR #675, opened 2018-04-04
Deprecates SpanCustomizer.annotate(long timestamp, String value)

adriancole:
This deprecates `SpanCustomizer.annotate(long timestamp, String value)`
as it is easy to create problematic timestamps, such as those skewed
within the same trace.

See #656

### DIFF
--- a/brave/src/main/java/brave/CurrentSpanCustomizer.java
+++ b/brave/src/main/java/brave/CurrentSpanCustomizer.java
@@ -33,8 +33,8 @@ public final class CurrentSpanCustomizer implements SpanCustomizer {
     return tracer.currentSpanCustomizer().annotate(value);
   }
 
-  /** {@inheritDoc} */
-  @Override public SpanCustomizer annotate(long timestamp, String value) {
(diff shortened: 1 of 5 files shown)


adriancole:
the deprecated api has no uses on github https://github.com/search?p=4&q=brave+SpanCustomizer+annotate&type=Code

devinsba:
Full disclosure, I have used this API in the past to add GC events to the timeline after the fact using some of the monitoring MBeans.

Currently that instrumentation is broken but I plan on fixing it and continuing to use this

adriancole:
>
> Full disclosure, I have used this API in the past to add GC events to the
> timeline after the fact using some of the monitoring MBeans.
>
> Currently that instrumentation is broken but I plan on fixing it and
> continuing to use this
>
do you use that with SpanCustomizer? or with Span.

The point here is just to move it off the "api type" SpanCustomizer, not
remove it completely. Span is like the advanced type and while not
suggested for casual use by user code, seems to still fit what you've
mentioned.

devinsba:
Ok. Yeah. I think I was using it on Span. Thanks for clarifying. I missed that detail

________________________________

Issue #676, opened 2018-04-04
Tracer.startScopedSpan


adriancole:
After #666, we require two operations to start a span and make it visible for synchronous tracing
```java
Span span = tracer.nextSpan().name(name).start();
try (SpanInScope ws = tracer.withSpanInScope(span)) {
  return pjp.proceed();
} catch (RuntimeException | Error e) {
  span.error(e);
  throw e;
} finally {
  span.finish();
}
```

This proposes a special-cased mechanism for synchronous interception that combines the apis:
```java
ScopedSpan span = tracer.startScopedSpan(name);
try {
  return pjp.proceed();
} catch (RuntimeException | Error e) {
  span.error(e);
  throw e;
} finally {
  span.finish();
}
```

### Notable features
The api of ScopedSpan is simply tag, error and annotate. While the lines of code are the same, the lines themselves are more concise, drawing attention to the flow.

* it doesn't support late setting of span name, as the call site should make this obvious
  * this also simplifies the amount of calls made to potentially none when using a pure function
* it doesn't support backdating timestamps, as the intention is to trace a block of code
* it doesn't implement closable.
  * try-with-resources has a race condition on error as described by [wingtips](https://github.com/Nike-Inc/wingtips/pull/43)
  * use of normal try/finally does not have this race condition, and encourages safe practice

The api is explicitly single-threaded
* this means the api is marked as not thread-safe (even if it happens to be)
* this makes sense as scopes are thread sensitive
* this allows us to make the implementation more efficient.

### Similar designs

Census has a scoped span which can be build from the same builder as a non-scoped one. The scope has no handle to the current span, you have to use the current span helper to do that. It does differentiate how to use in java 7 vs below, but it doesn't currently advice on the implications of try-with-resources on error handling.
```java
try (Scope ss = tracer.spanBuilder("MyChildSpan").startScopedSpan()) {
  doSomeWork();  // Here the new span is in the current Context, so it can be used
                             // implicitly anywhere down the stack. Anytime in this closure the span
                             // can be accessed via tracer.getCurrentSpan().
} finally {
  ss.close();
}
```

Elasticsearch APM's [Span](https://github.com/elastic/apm-agent-java/blob/master/apm-agent-api/src/main/java/co/elastic/apm/api/Span.java) api is auto-closable with minimal features. There's no advice on how to handle errors with the try-with-resources mechanism.
```java
try (Span span = tracer.startSpan()) {
  span.setName("SELECT FROM customer");
[...]

PR #678, opened 2018-04-04
WIP Scoped Span

adriancole:


brave/src/main/java/brave/RealScopedSpan.java
```diff
@@ -0,0 +1,46 @@
+package brave;
+
+import brave.internal.recorder.Recorder;
+import brave.propagation.TraceContext;
+import com.google.auto.value.AutoValue;
+
+/** This wraps the public api and guards access to a mutable span. */
+@AutoValue
+abstract class RealScopedSpan extends ScopedSpan {
```
adriancole: note.. you may realize this doesn't actually scope yet :) I raised early just to show the shape of the api

brave/src/main/java/brave/ScopedSpan.java
```diff
@@ -0,0 +1,49 @@
...
+ *   span.error(e); // Unless you handle exceptions, you might not know the operation failed!
+ *   throw e;
+ * } finally {
+ *   span.finish(); // finish - start = the duration of the operation in microseconds
+ * }
+ * }</pre>
+ *
+ * <p>Note: this type is only useful for local operations. For remote operations, use {@link Span}
+ */
+public abstract class ScopedSpan {
```
devinsba: The design in #676 implies that this would be marked as not thread safe. I would expect a comment to that affect here

adriancole: will do.. good catch

brave/src/main/java/brave/Tracer.java
```diff
@@ -129,22 +148,65 @@ public Tracer build() {
...
+   * Tracer tracer = tracing.tracer().withSampler(declarativeOverride);
+   *
+   * // Now, commands like startScopedSpan use a method-specific decision when starting a new trace
+   * }</pre>
+   *
+   * <p>See {@code brave.http.HttpSampler.toSampler()}
+   *
+   * @see DeclarativeSampler
+   */
+  public Tracer withSampler(Sampler sampler) {
```
adriancole: note: this hook is needed in order to permit ease of use when using aspectj or other declarative samplers for local originated traces

adriancole:
OK rebased this over #679 to clarify the changes central to scoped spans

brave/src/main/java/brave/ScopedSpan.java
adriancole: done. thanks

adriancole:
here's WIP of sleuth consuming this feature https://github.com/spring-cloud/spring-cloud-sleuth/pull/928

PR #680, opened 2018-04-05
Clarifies dubbo is 2.6+ and fixes test which tried to bind docker IP

adriancole:
We had feedback offline that our RPC tracing doesn't work well before
version 2.6. While we don't have details, safer to say that we only
work with 2.6+.

This also fixes a test hang, where Dubbo's IP detection can pick docker
bridge IP instead of link local.


chickenlj:
Dubbo will try to retrieve ip in the following orders:

outer configurations(e.g., system environment variables)
hosts mapping
guess by traversing the network` interface.
the interface that can connect to remote registry.
first interface has valid ip address.
As for guessing approach, Dubbo almost uses the same approach as zipkin/brave does: https://github.com/openzipkin/brave/blob/master/brave/src/main/java/brave/internal/Platform.java#L63, except for it does not check if (address.isSiteLocalAddress()) {. I am not sure if site-local addresses: 10/8, 172.16/12 and 192.168/16 the best choice for dubbo.

The deploy environment for dubbo varies, we highly recommend rely on the first two approaches, specifying the ip you want explicitly can be the best way, avoid guessing as it's unpredictable. In a docker environment, bind the docker ip or host ip can depends, leave the choice to users.

Issue #682, opened 2018-04-06
Consider immediately flushing certain "extra fields" to a log layer


adriancole:
https://github.com/spring-cloud/spring-cloud-sleuth/issues/927#issuecomment-378900051 discusses the understood problem of a race condition when implementing `CurrentTraceContext` hooks for tools like MDC. Thanks to @dhoffi for elaborating this.

The overal problem is that logging isn't immediately flushed when you set a variable. To do that, you need to know the underlying impl, like MDC: Ex redundant looking commands happen here:
```java
    @SendTo(SourceChannels.OUTPUT)
    public void timerMessageSource() {
        MDC.put(BUSINESS_PROCESS_NAME, ...);
        ExtraFieldPropagation.set(sleuthTracer.currentSpan().context(), BUSINESS_PROCESS_NAME, ...);
    }
```

would be to leave out the other either by default:
```java
    @SendTo(SourceChannels.OUTPUT)
    public void timerMessageSource() {
        ExtraFieldPropagation.set(sleuthTracer.currentSpan().context(), BUSINESS_PROCESS_NAME, ...);
    }
```

or by a special command like "flush" which attempts to set the thing immediately
```java
    @SendTo(SourceChannels.OUTPUT)
    public void timerMessageSource() {
        ExtraFieldPropagation.flush(sleuthTracer.currentSpan().context(), BUSINESS_PROCESS_NAME, ...);
    }
```

Eventhough #577 will solve the problem permanently, we might be able to get a tactical win by establishing some hook for immediately flushing certain (or all) keys to the current trace context.

This would only work in implementations that don't rely on scopes internally. For example, `MDC.put("key", "value")` is fine, but `try (Scope scope = context.scope("key", "value")` would not.

This has some cleanup implications, but we might be able to manage them in how we implement the scope field for `ExtraFieldsPropagation`. This all is only possible because "extra field" support is in the core library, so we can consider special-case it.

adriancole:
@dhoffi question: I assume you are ok if all extra fields are synced to MDC? or do you want to cherry-pick?

dhoffi:
well as the strong recommendation is not to put too much data in Propagation, I think it would be fine if exactly the fields of `spring.sleuth.baggage-keys` and `spring.sleuth.propagation-keys` would be synced to MDC, wouldn't it?

adriancole:
>
> well as the strong recommendation is not to put too much data in
> Propagation, I think it would be fine if exactly the fields of
[...]

PR #683, opened 2018-04-06
Adds ScopeDecorator and backports log correlation to use it

adriancole:
This is based on some work in Fukuoka where @anuraaga noticed we can stop some of the subclassing we do now

### Customizing in-process propagation (Ex. log correlation)

The `CurrentTraceContext` makes a trace context visible, such that spans
aren't accidentally added to the wrong trace. It also accepts hooks like
log correlation.

For example, if you use Log4J 2, you can copy trace IDs to your logging
context with [our decorator](https://github.com/openzipkin/brave/tree/master/context/log4j2):
```java
    .currentTraceContext(ThreadLocalCurrentTraceContext.newBuilder()
       .addScopeDecorator(ThreadContextScopeDecorator.create())
       .build()
    )
    ...
    .build();
```

Besides logging, other tools are available. `StrictScopeDecorator` can
help find out when code is not closing scopes properly. This can be
useful when writing or diagnosing custom instrumentation.

See #682

### DIFF
--- a/brave/src/main/java/brave/internal/propagation/CorrelationFieldScopeDecorator.java
+++ b/brave/src/main/java/brave/internal/propagation/CorrelationFieldScopeDecorator.java
@@ -15,6 +15,12 @@ import static brave.internal.HexCodec.lowerHexEqualsUnsignedLong;
  */
 public abstract class CorrelationFieldScopeDecorator implements ScopeDecorator {
 
+  /**
+   * When the input is not null "traceId", "parentId" and "spanId" correlation properties are saved
+   * off and replaced with those of the current span. When the input is null, these properties are
+   * removed. Either way, "traceId", "parentId" and "spanId" properties are restored on {@linkplain
+   * Scope#close()}.
+   */
   @Override public Scope decorateScope(@Nullable TraceContext currentSpan, Scope scope) {
     String previousTraceId = getIfString("traceId");
     String previousSpanId = getIfString("spanId");
@@ -39,6 +45,10 @@ public abstract class CorrelationFieldScopeDecorator implements ScopeDecorator {
     return new CorrelationFieldCurrentTraceContextScope();
   }
 
+  /**
+   * Idempotently sets correlation properties to hex representation of trace identifiers in this
+   * context.
+   */
   void maybeReplaceTraceContext(
       TraceContext currentSpan,
       String previousTraceId,
@@ -60,10 +70,15 @@ public abstract class CorrelationFieldScopeDecorator implements ScopeDecorator {
     if (!sameSpanId) put("spanId", HexCodec.toLowerHex(currentSpan.spanId()));
   }
 
+  /**
+   * Returns the correlation property of the specified name iff it is a string, or null otherwise.
+   */
   protected abstract @Nullable String getIfString(String key);
 
+  /** Replaces the correlation property of the specified name */
   protected abstract void put(String key, String value);
 
+  /** Removes the correlation property of the specified name */
   protected abstract void remove(String key);
 
   final void replace(String key, @Nullable String value) {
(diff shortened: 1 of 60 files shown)


brave/src/main/java/brave/internal/propagation/CorrelationFieldScopeDecorator.java
```diff
@@ -0,0 +1,76 @@
...
+    } else {
+      boolean sameParentId = lowerHexEqualsUnsignedLong(previousParentId, parentId);
+      if (!sameParentId) put("parentId", HexCodec.toLowerHex(parentId));
+    }
+
+    boolean sameSpanId = lowerHexEqualsUnsignedLong(previousSpanId, currentSpan.spanId());
+    if (!sameSpanId) put("spanId", HexCodec.toLowerHex(currentSpan.spanId()));
+  }
+
+  protected abstract @Nullable String getIfString(String key);
```
anuraaga: Javadoc? Also I'm not sure how to grok this `get` - at first I expected it to get the value from the current trace context. `getPreviouslyCorrelated` (a bit wordy though)?

brave/src/main/java/brave/propagation/CurrentTraceContext.java
```diff
@@ -21,23 +24,82 @@
...
+  public interface Factory {
+    CurrentTraceContext create(List<ScopeDecorator> scopeDecorators);
+  }
+
+  protected CurrentTraceContext() {
+    this.scopeDecorators = Collections.emptyList();
+  }
+
+  protected CurrentTraceContext(Builder<?> builder) {
+    this.scopeDecorators = (List<ScopeDecorator>) builder.scopeDecorators.clone();
```
anuraaga: I personally haven't seen `clone()` for copying anything other than an array. `new ArrayList<ScopeDecorator>(builder.scopeDecorators)`?

brave/src/main/java/brave/propagation/CurrentTraceContext.java
```diff
@@ -21,23 +24,82 @@
...
+   *     }
+   *     Scope result = new ThreadLocalScope();
+   *     // ensure scope hooks are attached to the result
+   *     return decorateScope(currentSpan, result);
+   *   }
+   * }</pre>
+   */
+  protected Scope decorateScope(@Nullable TraceContext currentSpan, Scope scope) {
+    int length = scopeDecorators.size();
+    if (length == 0) return scope;
```
anuraaga: I think the for loop duplicates this check so it doesn't really speed up the `0` case.

adriancole:
will be used by https://github.com/line/armeria/pull/1328

brave/src/main/java/brave/internal/propagation/CorrelationFieldScopeDecorator.java
adriancole: yep. it is internal, but no excuse :P

brave/src/main/java/brave/propagation/CurrentTraceContext.java
adriancole: agreed

brave/src/main/java/brave/internal/propagation/CorrelationFieldScopeDecorator.java
```diff
@@ -0,0 +1,91 @@
...
+    }
+
+    boolean sameSpanId = lowerHexEqualsUnsignedLong(previousSpanId, currentSpan.spanId());
+    if (!sameSpanId) put("spanId", HexCodec.toLowerHex(currentSpan.spanId()));
+  }
+
+  /**
+   * Returns the correlation property of the specified name iff it is a string, or null otherwise.
+   */
+  protected abstract @Nullable String getIfString(String key);
```
anuraaga: `ifString` seems like it was an implementation detail of `MDC`, for this abstract class, how about just `get`?

adriancole: ok with me!

PR #685, opened 2018-04-06
Locks down tests

adriancole:
I noticed some tests weren't closing Tracing


brave/src/test/java/brave/features/log4j2_context/Log4JThreadContextTest.java
```diff
@@ -1,76 +0,0 @@
...
-import brave.Tracer;
-import brave.Tracing;
-import brave.propagation.CurrentTraceContext;
-import brave.propagation.TraceContext;
-import org.apache.logging.log4j.ThreadContext;
-import org.junit.Test;
-
-import static org.assertj.core.api.Assertions.assertThat;
-
-public class Log4JThreadContextTest {
```
adriancole: we don't need this test as we ship log4j2 in the same repo. we would if we didn't

Issue #626, opened 2018-03-01
brave-census bridge


adriancole:
@jorgheymans currently uses brave-instrumentation-grpc and is looking to interop with census for the grpc layer. His application architecture looks like

incoming=rest -> mid tier grpc/kafka/ldap -> outgoing=rest 

So, if census-grpc uniformly handles that layer (still reporting to zipkin). Assuming brave doesn't implement the census api layer (which I suppose it could), the essence of this integration are the following:

* use grpc context as brave's CurrentTraceContext so that propagation works
* When reading the current trace context, look at census to see if there's an upstream span and convert that accordingly to a brave TraceContext.
* When writing the current trace context, make a fake Census span which delegates to the current brave span.

In looking at census grpc instrumentation, the parent is only read (not modified), so this *should* work. It won't work with advanced features like brave's extra field propagation, and it won't propagate things absent in census like the parent ID. However I still think it should work somewhat.

A design note: it would be cleaner if census didn't propagate a span rather just the trace context. Since it propagates the span, we have to create a span object to integrate, which is heavier (because it is stateful type) and makes the integration approach clunky due to this (like us needing to make a more expensive fake span just to get the 'current parent' code which only reads the reference to work). If census had a mode where it only propagated the trace context (impling a recording function based on trace context internally), this integration would be neater imho.

So the alternates I'm aware of are:
* brave implementing census tracer api (which makes sense if you are sending to zipkin anyway as the signal loss is the same as exporting to zipkin)
* brave implementing the same grpc tracer as census (much easier, but doesn't actually integrate with census)

adriancole:
PS what led to this was some curiosity about latency in a gRPC app. The brave grpc integration still uses interceptor approach (as we are back compat all the way till grpc 1.3). We haven't done a grpc tracer implementation for reasons including it seems the api isn't stable yet, and maybe census integration can solve this.

jorgheymans:
Thanks for drafting your thoughts on this, as said let me know if there is something you want me to test.

adriancole:
census no longer is a dependency of grpc, as of the latest version. I think this is not something we will progress as census is also supposed to be switched out with open telemetry once that's out of alpha

PR #688, opened 2018-04-11
Reads and writes gRPC trace and tags binary metadata

adriancole:
## gRPC Propagation Format (Census interop)

gRPC defines a [binary encoded propagation format](https://github.com/census-instrumentation/opencensus-specs/blob/master/encodings/BinaryEncoding.md) which is implemented
by [OpenCensus](https://opencensus.io/) instrumentation. When this is
the case, incoming requests will have two metadata keys "grpc-trace-bin"
and "grpc-tags-bin".

When enabled, this component can extract trace contexts from these
metadata and also write the same keys on outgoing calls. This allows
transparent interop when both census and brave report data to the same
tracing system.

To enable this feature, set `grpcPropagationFormatEnabled` which is off
by default:
```java
grpcTracing = GrpcTracing.newBuilder(tracing)
                         .grpcPropagationFormatEnabled(true).build();
```

Warning: the format of both "grpc-trace-bin" and "grpc-tags-bin" are
version 0. As such, consider this feature experimental.

## notes:

This allows interop with gRPC services that don't use B3 (for example,
the default census module). This also acknowledges that the OpenCensus
binary format is fairly contained within gRPC, so doesn't need to become
a top-level propagation format, for example used with non-gRPC services
as yet.

Redundantly encoding this format has some cost to it, notable 14
character header name "grpc-trace-bin" and the 38 character base64
encoded binary context value.

Note that this only allows black-box interop with Census. It does not
(yet) interop with census in the same process as Brave. However, this
should be a helpful step to those especially currently using Brave, but
trying out Census for non-java services.

See #626

### DIFF
--- a/instrumentation/grpc/README.md
+++ b/instrumentation/grpc/README.md
@@ -60,6 +60,28 @@ overrideSpanName = new GrpcClientParser() {
 };
 ```
 
+## gRPC Propagation Format (Census interop)
+
+gRPC defines a [binary encoded propagation format](https://github.com/census-instrumentation/opencensus-specs/blob/master/encodings/BinaryEncoding.md) which is implemented
+by [OpenCensus](https://opencensus.io/) instrumentation. When this is
+the case, incoming requests will have two metadata keys "grpc-trace-bin"
+and "grpc-tags-bin".
+
+When enabled, this component can extract trace contexts from these
+metadata and also write the same keys on outgoing calls. This allows
+transparent interop when both census and brave report data to the same
+tracing system.
+
+To enable this feature, set `grpcPropagationFormatEnabled` which is off
+by default:
+```java
+grpcTracing = GrpcTracing.newBuilder(tracing)
+                         .grpcPropagationFormatEnabled(true).build();
+```
+
+Warning: the format of both "grpc-trace-bin" and "grpc-tags-bin" are
+version 0. As such, consider this feature experimental.
+
 ## Development
 
 If you are working on this module, then you need to run `mvn install` to first compile the protos. Once the protos are compiled, then can be found in the directories:
--- a/instrumentation/grpc/src/main/java/brave/grpc/TraceContextBinaryMarshaller.java
+++ b/instrumentation/grpc/src/main/java/brave/grpc/TraceContextBinaryMarshaller.java
@@ -1,0 +1,115 @@
+package brave.grpc;
+
+import brave.propagation.TraceContext;
+import io.grpc.Metadata.BinaryMarshaller;
+import java.util.logging.Logger;
+
+import static com.google.common.base.Preconditions.checkNotNull;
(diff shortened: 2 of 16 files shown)


adriancole:
I need to add an interop test to prove this actually works cc @bogdandrutu

adriancole:
note this will be refactored to be optional behavior

instrumentation/grpc/src/test/java/brave/grpc/ITCensusInterop.java
```diff
@@ -0,0 +1,101 @@
...
+  @After public void close() throws Exception {
+    client.shutdown();
+    client.awaitTermination(1, TimeUnit.SECONDS);
+    server.shutdown();
+    server.awaitTermination(1, TimeUnit.SECONDS);
+    Tracing current = Tracing.current();
+    if (current != null) current.close();
+  }
+
+  @Test public void readsPropagatedSpan() throws Exception {
```
adriancole: here's the test that shows we are participating in the same trace

adriancole:
added censusPropagationEnabled property and client+server tests

adriancole:
going to add tags propagation as well, as we can lose it otherwise..

adriancole:
refactored to not drop tag context

adriancole:
not-dropping is probably not good enough, as the last upstream part of the tag context should be replaced..

adriancole:
ok rewrote everything. tags now works

instrumentation/grpc/src/main/java/brave/grpc/TraceContextBinaryMarshaller.java
```diff
@@ -0,0 +1,111 @@
...
+    } else {
+      logger.fine("Invalid input: expected trace ID at offset " + pos);
+      return null;
+    }
+    if (bytes[pos] == SPAN_ID_FIELD_ID) {
+      pos++;
+      spanId = readLong(bytes, pos);
+      pos += 8;
+    } else {
+      logger.fine("Invalid input: expected span ID at offset " + pos);
```
adriancole: self-note: kill all these string cats as they shouldn't happen unless fine is enabled

instrumentation/grpc/README.md
```diff
@@ -60,6 +60,28 @@ overrideSpanName = new GrpcClientParser() {
...
+and "grpc-tags-bin".
+
+When enabled, this component can extract trace contexts from these
+metadata and also write the same keys on outgoing calls. This allows
+transparent interop when both census and brave report data to the same
+tracing system.
+
+To enable this feature, set `grpcPropagationFormatEnabled` which is off
+by default:
+```java
```
adriancole: To give most flexibility I made this a feature flag similar to how tracing is still a flag in gRPC

adriancole:
going to release this tomorrow unless screams to the contrary

PR #693, opened 2018-04-20
WIP TraceContext propagation handler

adriancole:
I paired with @tylerbenson so thanks for the help.

Very early so don't read too much into how the code is shaped until comments suggest otherwise

See also https://github.com/openzipkin/zipkin4net/issues/200 cc @bgrainger

### DIFF
--- a/brave/src/main/java/brave/propagation/TraceContext.java
+++ b/brave/src/main/java/brave/propagation/TraceContext.java
@@ -518,7 +518,7 @@ public final class TraceContext extends SamplingFlags {
       String missing = "";
       if (traceIdHigh == 0L && traceId == 0L) missing += " traceId";
       if (spanId == 0L) missing += " spanId";
-      if (!"".equals(missing)) throw new IllegalArgumentException("Missing :" + missing);
+      if (!"".equals(missing)) throw new IllegalArgumentException("Missing:" + missing);
       return new TraceContext(
         flags, traceIdHigh, traceId, localRootId, parentId, spanId, ensureImmutable(extraList)
       );
--- a/brave/src/main/java/brave/propagation/TraceContext.java
+++ b/brave/src/main/java/brave/propagation/TraceContext.java
@@ -518,7 +518,7 @@ public final class TraceContext extends SamplingFlags {
       String missing = "";
       if (traceIdHigh == 0L && traceId == 0L) missing += " traceId";
       if (spanId == 0L) missing += " spanId";
-      if (!"".equals(missing)) throw new IllegalArgumentException("Missing:" + missing);
+      if (!"".equals(missing)) throw new IllegalArgumentException("Missing :" + missing);
       return new TraceContext(
         flags, traceIdHigh, traceId, localRootId, parentId, spanId, ensureImmutable(extraList)
       );
--- a/propagation/w3c/RATIONALE.md
+++ b/propagation/w3c/RATIONALE.md
@@ -1,0 +1,62 @@
+# brave-propagation-w3c rationale
+
+## Trace Context Specification
+
+### Why do we write a tracestate entry?
+We write both "traceparent" and a "tracestate" entry for two reasons. The first is due to
+incompatibility between the "traceparent" format and our trace context. This is described in another
+section.
+
+The other reason is durability across hops. When your direct upstream is not the same tracing
+system, its span ID (which they call parentId in section 3.2.2.4) will not be valid: using it will
+break the trace. Instead, we look at our "tracestate" entry, which represents the last known place
+in the same system.
+
+### What is supported by Brave, but not by traceparent format?
+
+#### `SamplingFlags` can precede a trace
+B3 has always had the ability to influence the start of a trace with a sampling decision. For
+example, you can propagate `b3=0` to force the next hop to not trace the request. This is used for
+things like not sampling health checks. Conversely, you can force sampling by propagating `b3=d`
+(the debug flag). `traceparent` requires a trace ID and a span ID, so cannot propagate this.
+
+#### `TraceIdContext` (trace ID only)
+Amazon trace format can propagate a trace ID prior to starting a trace, for correlation purposes.
+`traceparent` requires a trace ID and a span ID, so cannot a trace ID standalone.
+
+#### Not yet sampled/ deferred decision
+B3 and Amazon formats support assigning a span ID prior to a sampling decision. `traceparent` has no
+way to tell the difference between an explicit no and lack of a decision, as it only has one bit
+flag.
+
+#### Debug flag
+B3 has always had a debug flag, which is a way to force a trace even if normal sampling says no.
+`traceparent` cannot distinguish between this and a normal decision, as it only has one bit flag.
+
+#### Trace-scoped sampling decision
+`traceparent` does not distinguish between a hop-level or a trace scoped decision in the format.
+This means that traces can be broken as it is valid to change the decision at every step (which
+breaks the hierarchy). This is the main reason why we need a separate `tracestate` entry.
+
+### Why serialize the trace context in two formats?
+
+The "traceparent" header is only portable to get the `TraceContext.traceId()` and
+`TraceContext.spanId()`. Section 3.2.2.5.1, the sampled flag, is incompatible with B3 sampling. The
+format also lacks fields for `TraceContext.parentId()` and `TraceContext.debug()`. This requires us
+to re-serialize the same context in two formats: one for compliance ("traceparent") and one that
+actually stores the context (B3 single format).
+
+The impact on users will be higher overhead and confusion when comparing the sampled value of
+"traceparent" which may be different than "b3".
+
+### Why is traceparent incompatible with B3?
+
+It may seem like incompatibility between "traceparent" and B3 were accidental, but that was not the
+case. The Zipkin community held the first meetings leading to this specification, and were directly
+involved in the initial design. Use cases of B3 were well known by working group members. Choices to
+become incompatible with B3 (and Amazon X-Ray format) sampling were conscious, as were decisions to
+omit other fields we use. These decisions were made in spite of protest from Zipkin community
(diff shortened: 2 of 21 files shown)


adriancole:
I will move this to openzipkin-contrib repo at some point. It is too early to put this into brave and I also have less confidence the spec until there are others implementing it.

adriancole:
https://github.com/w3c/distributed-tracing is a decent place to consider

adriancole:
fyi this is on pause still as there's a discussion that significantly affects portability pending. We don't have the resources to thrash the impl so waiting on a decision to finalize with ideally a quorum of diverse input.

https://github.com/w3c/distributed-tracing/issues/167

adriancole:
@SergeyKanzhelev do you mind a quick mention of the status of the w3c tracecontext. I see it is working draft on the github repo, but then also see some pull requests for review. How stable are things? are there any public endpoints available to interop with?

adriancole:
I will rebase this again

adriancole:
rebased

adriancole:
some things that changed since this was done:
* tracestate encoding is more complex as characters such as `@` are now allowed. I've noticed some using a more expensive regex approach to split on account of this.
* there's a version in the header now, oddly 00 eventhough the spec is final now.
* there are a lot of MUST in the spec for things that were opposite before. For example, you have to handle mixed case, which will be silly in messaging for the reasons we already know about.
* odd things like traceparent field for span id renamed to parent id, which conflicts conceptually with B3 with no mention about B3 of course. super strange behavior as using simply "id" could have been done if the word span was problematic. anyway like all the above regressions, this is water under the bridge.

adriancole:
here's the published version, several days ago. we should watch out to see which errata are accepted (eg if version 00 is errata or intentional) https://www.w3.org/TR/trace-context-1/

iotw recommend if someone wants to pick this up, to wait a few days for errata

adriancole:
looks like people settled on version double-zero. Also the implementation expectations are extremely low, like just map b3 to traceparent and push tracestate https://github.com/jaegertracing/jaeger-client-java/pull/694

that in mind, I'll happily check this off and push the same parity.

adriancole:
I rebased this and renamed the thing to "w3c" as I can't imagine us wanting to add another module for correlation-context later. Also, I wrote in the README the implementation rationale, which is made to survive multiple systems.

later, I'll scan for spec drift and rally for merge.

adriancole:
backfilled traceparent tests as that format didn't change much. next is to manage the drift on tracestate which is a lot more permissive than last time.

adriancole:
I will redo the traceparent stuff following this as it is faster and has better logging https://github.com/openzipkin/brave/pull/1114

adriancole:
rewrote the traceparent parser based on recent work committed for b3 single (thanks to @anuraaga for the epic review on that).

This fully supports the traceparent spec now, including edge cases like forward versions. next step is to redo the tracestate parser.

propagation/w3c/src/main/java/brave/propagation/w3c/TraceparentFormat.java
```diff
@@ -0,0 +1,297 @@
...
+
+  static byte[] asciiToNewByteArray(char[] buffer, int length) {
+    byte[] result = new byte[length];
+    for (int i = 0; i < length; i++) {
+      result[i] = (byte) buffer[i];
+    }
+    return result;
+  }
+
+  static final ThreadLocal<WeakReference<TraceContext.Builder>> BUILDER_REF = new ThreadLocal<>();
```
adriancole: TODO: I've not analyzed benchmarks with vs without, yet, but it should help reduce allocations without resorting to internal code. Also, we need to add the classloader test which would break if you switch to non weak ref.

propagation/w3c/RATIONALE.md
```diff
@@ -0,0 +1,62 @@
...
+#### `TraceIdContext` (trace ID only)
+Amazon trace format can propagate a trace ID prior to starting a trace, for correlation purposes.
+`traceparent` requires a trace ID and a span ID, so cannot a trace ID standalone.
+
+#### Not yet sampled/ deferred decision
+B3 and Amazon formats support assigning a span ID prior to a sampling decision. `traceparent` has no
+way to tell the difference between an explicit no and lack of a decision, as it only has one bit
+flag.
+
+#### Debug flag
```
adriancole: TODO: I forgot to add the (actual) parent ID. the traceparent labels span ID parent ID (it is so confusing.. if span was a bad word, just say ID :P). traceparent does not record the parent ID of the context propagated, basically.

propagation/w3c/src/main/java/brave/propagation/w3c/TraceparentFormat.java
adriancole: it sincerely didn't help

adriancole:
started on tracestate key generation. it took a bit to understand the validation rules. hats off to @anuraaga for the tip to parse with a boolean array as it is faster than the usual approach https://github.com/netty/netty/pull/9896

adriancole:
tracestate key parsing numbers from latest commit:

```
Benchmark                                                     Mode  Cnt  Score   Error   Units
TracestateFormatBenchmarks.validateKey_brave_longest_basic   thrpt   15  5.168 ± 0.061  ops/us
TracestateFormatBenchmarks.validateKey_brave_longest_tenant  thrpt   15  4.056 ± 0.018  ops/us
TracestateFormatBenchmarks.validateKey_regex_longest_basic   thrpt   15  0.401 ± 0.013  ops/us
TracestateFormatBenchmarks.validateKey_regex_longest_tenant  thrpt   15  0.414 ± 0.011  ops/us
```

adriancole:
ok tracestatekey algo done.. need to go back and cite the sections of the spec and add logging for the various failure cases.

```
Benchmark                                                     Mode  Cnt  Score   Error   Units
TracestateFormatBenchmarks.validateKey_brave_longest_basic   thrpt   15  5.265 ± 0.035  ops/us
TracestateFormatBenchmarks.validateKey_brave_longest_tenant  thrpt   15  5.127 ± 0.021  ops/us
TracestateFormatBenchmarks.validateKey_regex_longest_basic   thrpt   15  0.375 ± 0.033  ops/us
TracestateFormatBenchmarks.validateKey_regex_longest_tenant  thrpt   15  0.425 ± 0.009  ops/us
```

adriancole:
added all the validation rules and assertions so that when a `tracestate` header arrives with a malformed key, there's a chance to know what is wrong.

Tomorrow I will work on the value parsing and integrating the logic written today (it isn't in the tracestate value parsing loop yet)

brave/src/main/java/brave/propagation/TraceContext.java
```diff
@@ -278,6 +278,17 @@ public String spanIdString() {
       extra = context.extra;
     }
 
+    public final Builder clear() {
```
adriancole: TODO: revert this as it is wasn't needed

samukce:
hi @adriancole, do you need a hand on this?

adriancole:
@samukce would be helpful, sure.. there's some things to follow-up on.. I think maybe need a checklist, but will mention for now.

1. somewhere there's a debate about how many `@` can be present or if it will be ok to allow any. We should delete all the special cases around "tenant" if they decided that way..
2. the next step here was parsing tracestate, however the code we'd make here is code we already need upstream for dynamic baggage. I'd be less inclined to progress until the other is done. One way you can hasten things is to review any PRs that mention baggage in any way, become acclimated with the codebase as it will be what powers tracestate. You may find opportunities to contribute to make that quicker also.

sg?

adriancole:
@samukce also in front of this is getting secondary sampling working on the baggage apis. the main thing is to get everything aligned so that the apis we use here also work for secondary sampling etc. a bit of up front work, but it is slowly coming together later https://github.com/openzipkin-contrib/zipkin-secondary-sampling

adriancole:
as far as I know there won't be a constraint on at signs, so it is far easier to parse now https://github.com/w3c/trace-context/pull/386

adriancole:
started reworking this over latest master. next step: take out the unnecessary `@` logic

propagation/w3c/src/test/java/brave/propagation/w3c/TraceContextPropagationTest.java
```diff
@@ -0,0 +1,123 @@
...
+    request.put("traceparent", validTraceparent);
+    request.put("tracestate", "b3=" + validB3Single + "," + otherState);
+
+    assertThat(extractor.extract(request)).hasToString(
+        "Extracted{"
+            + "traceContext=" + sampledContext + ", "
+            + "samplingFlags=SAMPLED_REMOTE, "
+            + "extra=[tracestate: " + otherState + "]"
+            + "}");
+  }
```
adriancole, line 91: thanks to @anuraaga for all the advice, this is much more intuitive now, and especially useful for partially deconstructed formats like this.

adriancole:
Annoyingly, trace-context defines some strange whitespace rules. Ex

`tracestate: key=          value` where `          value` is actually the value per the spec. 

https://github.com/w3c/trace-context/issues/409

I don't expect them to fix this so I wrote a parser that can deal with it. 

once this is in I can finish up https://github.com/openzipkin/brave/pull/1193

adriancole:
@anuraaga @samukce I think Brave now has what's needed to do this with the least cruft. As such, once 5.12 is out I'll cut a new openzipkin-contrib repository named brave-w3c with this as "initial commit" referencing this issue in order to retain history. After that, I'll close this PR.

In that repo, we can carry forward the rest of it decoupled from Brave releases, allow people to experiment while details in w3c are still fluid.

adriancole:
@anuraaga @samukce as people are often confused by HTTP encoding rules, I summarized the relationship with our APIs here https://github.com/openzipkin/brave/pull/1205

adriancole:
https://github.com/openzipkin-contrib/brave-propagation-w3c will carry this on

PR #695, opened 2018-04-28
Fixes bug where we didn't re-scope when a request is already processed

adriancole:
This is hard to test but I will add one after the fact. This is visible
when the following is used in Spring Boot 2 with Tomcat:

```java
  @Configuration
  static class CustomerErrorPageRegistrar implements ErrorPageRegistrar {

    @Override public void registerErrorPages(ErrorPageRegistry registry) {
      registry.addErrorPages(new ErrorPage(HttpStatus.NOT_FOUND, "/error404"));
    }

  }
```

Fixes https://github.com/spring-cloud/spring-cloud-sleuth/issues/948#issuecomment-382233251


marcingrzejszczak:
This starts to remind the initial code in Sleuth as far as the complexity of logic is concerned ;)

PR #700, opened 2018-05-11
Register remote endpoint in TracingStatementInterceptor if cannot parse IP

TanyaGaleyev:
As `Endpoint` can exist with ipv4 and ipv6 both equal to null it worth registering endpoint if `connection.getHost()` returns DNS name instead of IP.

### DIFF
--- a/instrumentation/mysql/src/main/java/brave/mysql/TracingStatementInterceptor.java
+++ b/instrumentation/mysql/src/main/java/brave/mysql/TracingStatementInterceptor.java
@@ -79,7 +79,7 @@ public class TracingStatementInterceptor implements StatementInterceptorV2 {
         }
       }
       Endpoint.Builder builder = Endpoint.newBuilder().serviceName(remoteServiceName).port(port);
-      if (!builder.parseIp(connection.getHost())) return;
+      builder.parseIp(connection.getHost());
       span.remoteEndpoint(builder.build());
     } catch (Exception e) {
       // remote address is optional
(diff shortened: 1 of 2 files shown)


devinsba:
@TanyaGaleyev if this is to get merged, you will need to fix the test that this change broke. There is a test validating the old behavior

TanyaGaleyev:
@devinsba thanks! Test fixed.

instrumentation/mysql/src/main/java/brave/mysql/TracingStatementInterceptor.java
```diff
@@ -79,7 +79,7 @@ static void parseServerAddress(Connection connection, Span span) {
         }
       }
       Endpoint.Builder builder = Endpoint.newBuilder().serviceName(remoteServiceName).port(port);
-      if (!builder.parseIp(connection.getHost())) return;
```
adriancole: I think this meant to be "return builder.build()" as the port isn't useful without the IP

TanyaGaleyev: I will explain. We have mysql connector configured with URL containing DNS name. With current behaviour we see traces like below:
```
api-server    : | GET /api/resource                         |
    api-server:     | SELECT * FROM table    |
```
Second "api-server" entry is related to mysql. It will be convenient to see mysql service name instead of "api-server" there. But because connector is configured without explicit IP new endpoint is not added to current Span.

adriancole: ya I mean the old code just did "return" I think you can get this behavior by only changing it to "return builder.build"

thanks for the clarification though!

TanyaGaleyev: Do you suggest to return `Endpoint` from `parseServerAddress` method? Currently it is `void`.

adriancole: ah I didn't notice that. your code is simplest

PR #703, opened 2018-05-16
Tests demonstrating assertion exceptions from TracingApplicationEventListener

gordz:
Adding tests demonstrating an assertion error thrown by TracingApplicationEventListener
for jax-rs resources that have a method with no @Path annotation.


adriancole:
thanks tons

PR #711, opened 2018-06-01
Implements GCP propagation

joaoandremartins:
This propagation re-uses B3Propagation's injector, while adding a new
CompositeExtractor concept that aims at extracting multiple trace IDs,
in order.

In this case, we're interested in extracting the trace ID from the
'x-cloud-trace-context' header, followed by 'X-B3-TraceId'.

Fixes https://github.com/spring-cloud/spring-cloud-gcp/issues/703

### DIFF
--- a/brave/src/main/java/brave/propagation/TraceIdContext.java
+++ b/brave/src/main/java/brave/propagation/TraceIdContext.java
@@ -161,7 +161,7 @@ public final class TraceIdContext extends SamplingFlags {
      * @return false if the input is null or malformed
      */
     // temporarily package protected until we figure out if this is reusable enough to expose
-    final boolean parseTraceId(String traceIdString, Object key) {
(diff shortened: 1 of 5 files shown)


joaoandremartins:
@adriancole FYI: this is intended as a very early, WIP prototype, just to get some early high-level feedback into what is right or wrong with this design. Thanks!

adriancole:
thanks for the start! note that in hindsight it was probably a mistake to
have the Amazon format here as opposed to zipkin-aws mavens relocations are
a bit of a pain so we left it. iirc @devinsba noted this and it was mostly
arbitrary. later other brave code was needed in zipkin-aws for example to
ensure the right data is collected.

I think probably the Google cloud one should be in the zipkin-gcp repo
especially knowing what we know now and also for version alignment with the
other things folks need like the sender and any possible future data policy
component.

PR #713, opened 2018-06-09
Add instrumentation for mysql-connector-java 8.x, the current GA vers…

anuraaga:
…ion.

The new version separates query interception and exception interception. As it still does not provide any context for propagating information between these two, we just have to rely on the thread-local span's propagation, meaning for exception logging there is a dependency on having the query interceptor enabled explicitly. Annoying, but at least we can validate the presence of both interceptors.

The differences between mysql-connector-6 is 

- TracingExceptionInterceptor
- When exception interceptor is enabled, the query interceptor does not close the span since it runs before the exception interceptor. Admittedly, I'm not sure if this order is guaranteed in future versions...
- Just from trying it out, it seems the new `Supplier<String>` of queries returns even for prepared statements, so removed the previous workaround that used `PreparedStatement`.


anuraaga:
Ah - added. Also at some point it may make sense to consider removing `mysql6` since it was a developmental release, so it seems like overkill for Brave to support it long-term.

adriancole:
Thanks for the tip!

Issue #561, opened 2017-12-28
Mysql calls are being attributed to the calling service and not treated as a remote one


rieske:
This code:
https://github.com/openzipkin/brave/blob/0b44ef88cb818a69d178f0c30e7e2ba8711b2966/instrumentation/mysql/src/main/java/brave/mysql/TracingStatementInterceptor.java#L82

expects an IP address to treat mysql call as a remote endpoint. Why?
I would like to use a domain name for my mysql connection and still have instrumentation treat it as an external service.

adriancole:
zipkin's data format stores ip information as they can change. There's no
field for hostname, which can resolve to many different IPs. The ways to
address this include some of the following:

* use the hostname as a service name (if it is stable)
* resolve the hostname into an IP (performance hit and possibly inaccurate)
* add a tag for the hostname

We have some outstanding work to have standard tags for sql things, but
anyway let me know if anything above sounds reasonable to you

TanyaGaleyev:
Hi,

I have the same problem. Are ipv4 and ipv6 fields of `Endpoint` required? I suspect that one of them can easily be null. If both could be null it seems that problem can be resolved by removing conditional return from `TracingStatementInterceptor` and leaving just:
```java
      builder.parseIp(connection.getHost());
      span.remoteEndpoint(builder.build());
```

devinsba:
Yes, bot IPs are allowed to be null, since sometimes you can't resolve the IP without a DNS lookup

TanyaGaleyev:
I expressed my suggestions in simple PR https://github.com/openzipkin/brave/pull/700

rieske:
Since the above PR was merged in, I ported the same fix for mysql connectors 6 and 8: https://github.com/openzipkin/brave/pull/726

PR #726, opened 2018-06-27
Register remote endpoint in TracingStatementInterceptor if cannot parse IP

rieske:
Related issue: https://github.com/openzipkin/brave/issues/561

This change ports this fix https://github.com/openzipkin/brave/pull/700
to mysql connectors 6 and 8

As `Endpoint` can exist with ipv4 and ipv6 both equal to null it worth registering endpoint if `connection.getHost()` returns domain name instead of IP.


adriancole:
thanks for noticing and putting your hands towards this

Issue #557, opened 2017-12-18
Support "firehose mode"


adriancole:
Full description pending, but this will describe a mode where brave collects data regardless of sampling decision. At the point where spanReporter is used today, a branch will occur where "firehose" reporter gets all spans, and the existing one, only sampled ones. This is in support of aggregators such as span metrics.

Coming out of https://github.com/openzipkin/brave/issues/529

adriancole:
(TraceContext, MutableSpan) will be the primary input for firehose handlers, whether they are metrics or service aggregators, or a secondary zipkin instance.

There will be another flag that will determine if we record regardless of the primary sampling status. Stay tuned.

adriancole:
finally got focus time on firehose https://github.com/openzipkin/brave/pull/791

adriancole:
https://github.com/openzipkin/brave/pull/796 < README

PR #730, opened 2018-07-21
Removes bytecode so we can add more

adriancole:
This removes the auto-value dependency and related for inlining the few
cases where it was used. It also trims the class hierarchy of sampling
flags. Finally, it moves out the in-flight localEndpoint as it is
constant.

Besides fixing a few bugs, these changes reduced the size of the jar by
12KiB and scraps 2 dependencies.

```bash
du -k /Users/acole/.m2/repository/io/zipkin/brave/brave/5.1.2/brave-5.1.2.jar /Users/acole/.m2/repository/io/zipkin/brave/brave/5.1.3-SNAPSHOT/brave-5.1.3-SNAPSHOT.jar
116	/Users/acole/.m2/repository/io/zipkin/brave/brave/5.1.2/brave-5.1.2.jar
128	/Users/acole/.m2/repository/io/zipkin/brave/brave/5.1.3-SNAPSHOT/brave-5.1.3-SNAPSHOT.jar
```

### DIFF
--- a/brave/src/main/java/brave/propagation/SamplingFlags.java
+++ b/brave/src/main/java/brave/propagation/SamplingFlags.java
@@ -3,41 +3,80 @@ package brave.propagation;
 import brave.internal.Nullable;
 
 //@Immutable
-public abstract class SamplingFlags {
-  public static final SamplingFlags EMPTY = new SamplingFlagsImpl(null, false);
-  public static final SamplingFlags SAMPLED = new SamplingFlagsImpl(true, false);
-  public static final SamplingFlags NOT_SAMPLED = new SamplingFlagsImpl(false, false);
-  public static final SamplingFlags DEBUG = new SamplingFlagsImpl(true, true);
+public class SamplingFlags {
+  static final int FLAG_SAMPLED = 1 << 1;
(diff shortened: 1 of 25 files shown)


adriancole:
FYI this is in support of #557 (firehose mode) which I have some work in progress on. First steps are to make room for code.

brave/src/main/java/brave/propagation/SamplingFlags.java
```diff
@@ -3,90 +3,100 @@
 import brave.internal.Nullable;
 
 //@Immutable
-public abstract class SamplingFlags {
-  public static final SamplingFlags EMPTY = new SamplingFlagsImpl(null, false);
-  public static final SamplingFlags SAMPLED = new SamplingFlagsImpl(true, false);
-  public static final SamplingFlags NOT_SAMPLED = new SamplingFlagsImpl(false, false);
-  public static final SamplingFlags DEBUG = new SamplingFlagsImpl(true, true);
```
adriancole: intellij showed warnings about classloader deadlock by referencing a subclass ctor in a super. the changes here not only make things smaller, but also remove that possibility.

PR #731, opened 2018-07-21
Attempts to clarify the intent of sampled status

adriancole:
cc @openzipkin/core While this is in brave, it is overriding the text lifted originally and unmodified since finagle.

### DIFF
--- a/brave/src/main/java/brave/propagation/SamplingFlags.java
+++ b/brave/src/main/java/brave/propagation/SamplingFlags.java
@@ -20,8 +20,23 @@ public class SamplingFlags {
   }
 
   /**
-   * Should we sample this request or not? True means sample, false means don't, null means we defer
-   * decision to someone further down in the stack.
+   * Sampled means send span data to Zipkin (or something else compatible with its data). It is a
+   * consistent decision for an entire request (trace-scoped). For example, the value should not
+   * move from true to false, even if the decision itself can be deferred.
+   *
+   * <p>Here are the valid options:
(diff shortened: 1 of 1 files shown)


brave/src/main/java/brave/propagation/SamplingFlags.java
```diff
@@ -20,13 +20,23 @@
...
    *
-   * <p>Note: this is a uniform decision for the entire trace. Advanced sampling patterns can
-   * overlay this via {@link Propagation.Factory#isNoop(TraceContext)}. For example, a noop context
-   * usually implies sampled is false or unset. However, you can collect data anyway, locally for
-   * metrics, or to an aggregation stream.
+   * <p>Here are the valid options:
+   * <pre><ul>
+   *   <li>True means the trace is reported, starting with the first span to set the value true</li>
+   *   <li>False means the trace should not be reported</li>
+   *   <li>Null means the decision should be deferred to another span</li>
```
basvanbeek: maybe nitpicking here... but in case of shared spans this is not completely true.

adriancole: no.. this is a good point. Originally, the text was "to the next hop" should we retain that?

basvanbeek: The next hop is indeed better.

adriancole: done

PR #732, opened 2018-07-23
Adds more defense when reporting spans on ideal and orphaned paths

adriancole:
This moves the crash-guard from only on orphaned spans to all spans.
This also ensures when an unrelated thread is flushing orphaned spans,
it only gets clock overhead once.


shakuzen:
CircleCI failure was https://github.com/openzipkin/brave/issues/733.

adriancole:
ignoring the flake :P

PR #734, opened 2018-07-24
Re-adds TraceContext.shared to partition joined span data

adriancole:
When loopback RPC requests occur, data can overlap, resulting in missing
spans data depending on whether the client or server completes first.

This fixes the situation by restoring the `TraceContext.shared` flag and
adding new tests that show what this supports.

This reverts commit 6d3322279de8d68b2c6cd83b7f8990bd316d2201.


adriancole:
going to cut this as it is a fairly big bug

adriancole:
releasing as brave 5.1.3

PR #738, opened 2018-07-27
Avoids dragons in javadoc by avoiding implicit lookups in servlet filter

### LINKED ISSUE #729: Potential NPE in TracingAsyncListener#adaptResponse
I'm currently also in the process of writing instrumentation for the async servlet stuff and noticed a potential bug in brave.

Because the request and response objects are not supplied when adding the `AsyncListener` to the `AsyncContext` here
https://github.com/openzipkin/brave/blob/49b91c05bf50c0837dfa26d6246e4d07cbdb8453/instrumentation/servlet/src/main/java/brave/servlet/ServletRuntime.java#L77

`AsyncEvent#getSuppliedRequest()` must return `null` per specification

> If the AsyncListener to which this AsyncEvent is being delivered was added using AsyncContext.addListener(AsyncListener, ServletRequest, ServletResponse), the returned ServletRequest will be the same as the one supplied to the above method. If the AsyncListener was added via AsyncContext.addListener(AsyncListener), this method must return null.

This can lead to NPEs here:
https://github.com/openzipkin/brave/blob/49b91c05bf50c0837dfa26d6246e4d07cbdb8453/instrumentation/servlet/src/main/java/brave/servlet/ServletRuntime.java#L121-L124

The funny thing is that most application servers don't actually implement it that way and are happy to return a request when calling `AsyncEvent#getSuppliedRequest()`, no matter how the `AsyncListener` is added. I think it was WebSphere in one specific version which actually implements this according to the spec. Can't remember which one that was but if you are interested I can find it out.

adriancole:
@felixbarny warned us about a subtle case noted in javadoc which can
result in null references.

Fixes #729

### DIFF
--- a/instrumentation/servlet/src/main/java/brave/servlet/ServletRuntime.java
+++ b/instrumentation/servlet/src/main/java/brave/servlet/ServletRuntime.java
@@ -36,7 +36,7 @@ abstract class ServletRuntime {
   abstract boolean isAsync(HttpServletRequest request);
 
   abstract void handleAsync(HttpServerHandler<HttpServletRequest, HttpServletResponse> handler,
-      HttpServletRequest request, Span span);
+      HttpServletRequest request, HttpServletResponse response, Span span);
 
   ServletRuntime() {
   }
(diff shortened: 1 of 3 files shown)


instrumentation/servlet/src/main/java/brave/servlet/ServletRuntime.java
```diff
@@ -72,38 +72,48 @@ private static ServletRuntime findServletRuntime() {
     }
 
     @Override void handleAsync(HttpServerHandler<HttpServletRequest, HttpServletResponse> handler,
-        HttpServletRequest request, Span span) {
+        HttpServletRequest request, HttpServletResponse response, Span span) {
       if (span.isNoop()) return; // don't add overhead when we aren't httpTracing
-      request.getAsyncContext().addListener(new TracingAsyncListener(handler, span));
+      TracingAsyncListener listener = new TracingAsyncListener(handler, request, response, span);
+      request.getAsyncContext().addListener(listener);
```
felixbarny: Instead of adding request and response to the listener itself, you can also do

```java
request.getAsyncContext().addListener(listener, request, response);
```

That way, `e.getSuppliedRequest()` does not return null.

The advantage of your current solution is, however, that you don't need to cast, as `e.getSuppliedRequest()` returns a `ServletRequest`.

adriancole:
will address any post merge comments. feel free to make them!

instrumentation/servlet/src/main/java/brave/servlet/ServletRuntime.java
felixbarny: Couldn't you call

```java
eventAsyncContext.addListener(this, event.getSuppliedRequest(), event.getSuppliedResponse())
```

when re-adding?

adriancole:
>
> Couldn't you call
>
> eventAsyncContext.addListener(this, event.getSuppliedRequest(), event.getSuppliedResponse())
>
> when re-adding?
>
suppose so, if we believe those won't be null by some nuance of servlet :P

the comment on the scary line is

If another async is created (ex via asyncContext.dispatch), this needs
to be re-attached

so basically I can't even test this, much less if when this is the case it
has no chance of not being null. I'm not trying to be FUD'y, but anyway
that was the reason.

I will add the approach you mention as I like it anyway as less references
to track.. if someone breaks later, we know what to revert :P

instrumentation/servlet/src/main/java/brave/servlet/ServletRuntime.java
adriancole: acb2ed62 done! thanks for the idea

nicmunroe:
@adriancole So the servlet rabbit hole never ends. When looking into the javadoc for `addListener(...)` I saw this:
```
The given ServletRequest and ServletResponse objects will be made available to the given AsyncListener via the getSuppliedRequest and getSuppliedResponse methods, respectively, of the AsyncEvent delivered to it. These objects should not be read from or written to, respectively, at the time the AsyncEvent is delivered, because additional wrapping may have occurred since the given AsyncListener was registered, but may be used in order to release any resources associated with them.
```
So there's actually a theoretical functional reason for not caching the request/response and using them later - additional wrapping might be done between when you register the listener and when the listener executes. If you used pre-cached versions of the request and response, they may be missing wrapping done as part of request/response processing before your listener executes (potentially missing request attributes needed for tagging, for example). At least theoretically...but since I can't find a servlet framework that even follows the javadoc on `addListener(someListener)` causing `event.getSupplied[Request/Response]()` to return null (they all just return the original request/response anyway), I'm not sure this matters in reality. 

You can trigger the listener reattaching behavior in a test - something like this: https://github.com/Nike-Inc/wingtips/blob/e17e7a0d69b94ef1c333687249a9f8da7848d26f/wingtips-servlet-api/src/test/java/com/nike/wingtips/componenttest/RequestTracingFilterComponentTest.java#L381-L395.

PR #739, opened 2018-07-27
Bumps versions to latest

adriancole:
awaiting zipkin-reporter 2.7.7


pom.xml
```diff
@@ -164,29 +165,6 @@
             <artifactId>zipkin</artifactId>
             <version>${zipkin.version}</version>
         </dependency>
-        <dependency>
```
adriancole: this was needed in the old brave stuff. removing it as not referenced anymore

Issue #399, opened 2017-05-09
Bootify spring instrumentation


adriancole:
It would be convenient to have spring boot magically configure rest template and spring MVC filter. For example, spring-boot-starter-web already includes spring-web and spring-web-mvc deps.

One way to do this is to make spring-boot an optional dep of the existing `brave-instrumentation-spring-web` and `brave-instrumentation-spring-web-mvc` packages. It could then add an autoconfiguration file which configures the filters (given an bean of type `HttpTracing` exists). In non-spring boot environments, this class wouldn't be loaded, so wouldn't crash or otherwise matter. 

The only downside is that if people have their own auto-configuration, they may need to exclude this. That said, this should be manageable via a `@ConditionalOnMissingBean` thing..

adriancole:
another option would be to make an autoconfig directory and jars.

devinsba:
I plan to do this some time in the next few weeks if no one beats me to it.
Whoever you are please let us know here if you do start working on it

devinsba:
cc @TilinC since you are using spring-boot with brave

devinsba:
I am going to try to start working on this this week. I'll link the PR here when I do

jorgheymans:
This issue can be closed, sleuth now incorporates brave/zipkin natively.

shakuzen:
> This issue can be closed, sleuth now incorporates brave/zipkin natively.

In the linked pull request there was some discussion about doing this for people using Spring Boot but not using Spring Cloud (Sleuth), or who are stuck on Spring Boot 1.x. That said, unless we're aware of users for this today, I think we're good to close this.

PR #740, opened 2018-07-27
[WIP] Bootify Spring tracing

ImFlog:
Intent to fix https://github.com/openzipkin/brave/issues/399

### DIFF
--- a/instrumentation/spring-web/README.md
+++ b/instrumentation/spring-web/README.md
@@ -12,3 +12,7 @@ it is in place before proceeding. Here's an example in [XML](https://github.com/
 Then, wire `TracingClientHttpRequestInterceptor` and add it with the
 `RestTemplate.setInterceptors` method. If you are using `AsyncRestTemplate` and Spring 4.3+, you can
 wire `AsyncTracingClientHttpRequestInterceptor` and add it via `AsyncRestTemplate.setInterceptors`.
+
+For Spring boot users, autoConfiguration is available when the `HttpTracing` bean is configured.
+Basically it creates a `RestClient` and an `AsyncRestClient` for you if no other bean of this type exist.
+
--- a/instrumentation/spring-web/pom.xml
+++ b/instrumentation/spring-web/pom.xml
@@ -1,5 +1,7 @@
 <?xml version="1.0"?>
-<project xmlns="http://maven.apache.org/POM/4.0.0" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 http://maven.apache.org/xsd/maven-4.0.0.xsd">
+<project xmlns="http://maven.apache.org/POM/4.0.0"
+    xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
+    xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 http://maven.apache.org/xsd/maven-4.0.0.xsd">
   <parent>
     <groupId>io.zipkin.brave</groupId>
     <artifactId>brave-instrumentation-parent</artifactId>
@@ -25,6 +27,12 @@
       <groupId>org.springframework</groupId>
       <artifactId>spring-web</artifactId>
       <version>${spring4.version}</version>
+      <scope>provided</scope>
+    </dependency>
+    <dependency>
+      <groupId>org.springframework.boot</groupId>
+      <artifactId>spring-boot-starter-web</artifactId>
+      <version>${spring.boot.version}</version>
       <scope>provided</scope>
     </dependency>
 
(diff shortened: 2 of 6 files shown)


adriancole:
cc some springy folks @jonathan-lo @marcingrzejszczak @zeagord @shakuzen @devinsba @llinder 

I think it is awesome to have this effort picked up. We need to tread carefully, but definitely we should tread!

ImFlog:
As I also said in Gitter, I do have second thoughts too. I am wondering if this is not overlapping with Sleuth for example.

To create a spring-boot-instrumentation module where autoconfiguration for all spring things would be done (using conditionalOnClass for example) is a good idea IMHO.

adriancole:
PS especially with rest template we should be careful about double-instrumentation. The dragons...

https://github.com/spring-cloud/spring-cloud-sleuth/blob/master/spring-cloud-sleuth-core/src/main/java/org/springframework/cloud/sleuth/instrument/web/client/TraceWebClientAutoConfiguration.java

If we do something like this, I would suggest it as a "best efforts" and kick out to sleuth for anything serious.

adriancole:
related is the playground app here: https://github.com/openzipkin/brave-webmvc-example/pull/25

spencergibb:
At this point isn't sleuth auto configuration of brave. Why the duplicate effort?

adriancole:
Some of the conversations have gone like this:

* Similarly to how we still support Spring 2.5, allow basics for those not
yet on Boot 2, or apps that may not upgrade for years.
* Some don't use sleuth even if we recommend it. Basics help reduce cost of
support and prevent us pushing people out of zipkin's ecosystem.
* Establish a pattern for things that may never be in sleuth (ex there are
things like camel which might be boot but not sleuth)

For example, if there is basic autoconfiguration, properties and otherwise
tightly coupled to the tracing component can be where the component is
defined. We do this in zipkin's server for example. We have buddy modules
for autoconfig to reduce the integration costs and coupling.

I don't expect autoconfig to all be usable in sleuth, for example, as it is
more advanced. Also, sleuth is more about molecules than atoms. For
example, there are interesting relationships between multiple components
which themselves require auto-configuration.

TL;DR; to the degree we pursue this...

I think we should do *only basics*, allowing those on custom spring boot
setups to not mess up basics, reducing cost of support on our channel
especially where english is second language.

We should define a pattern for remote instrumentation that isn't in sleuth,
or would make its dependency graph laborious. For example, cassandra and
camel are on different modules. There is some value in being able to drop
things in and not have to explain how to do autoconfiguration by hand.

Beyond this, especially to the "atoms not molecules" point, we recommend to
use sleuth for boot. For example, I'd expect no inter-relationships like
exist in sleuth to exist here. For example, no feign+ribbon+hystrix as it
is frankly not something we'd want to try to afford.

My 2p

adriancole:
another thing to consider

even if we don't configure instrumentation, configuring Tracing (naming
correctly setting up reporters and closing them etc) this would be a big
help. similar to brave-springbeans we could start with a copy/paste
removing thing which would be largely reusable also by sleuth. maybe this
isnt 80% of the way but at least I think it would reduce some of the
support cost.

ex a lot of the stuff here is undifferentiated and the code in sleuth to
pick which sender is also
https://github.com/openzipkin/brave-webmvc-example/pull/25/files

if we could at least get the bedding right, yeah that sounds quite a win.

shakuzen:
I think a separate module containing the auto-configuration would be better to avoid two auto-configurations for the same thing ending up on the classpath and having to rely on ordering or dependencies on one another to avoid funkiness.

The separate module should also help with migrating Sleuth auto-configuration to the Brave module while maintenance branches of Sleuth can still upgrade but not depend on the Brave auto-config module.

I do wonder if we will be making new support issues for ourselves by having more ways to do about the same thing. It might cause confusion on when to use what, but to the degree that users have varied setups, I'm not sure how much we can avoid that while supporting them.

jonathan-lo:
So what I'm hearing is that this is primarily catering who are using boot 1...?

IMHO we shouldn't need to be too smart about double instrumentation - it should be a case of adding instructions to use either sleuth or brave's autoconfig

ImFlog:
I think we agree that if we decide to autoconfigure, we will do it in a separate module.
But the question remains, do we want to provide basic autoconfiguration or do we just provide samples like https://github.com/openzipkin/brave-webmvc-example/pull/25 or else to redirect explicitly to Sleuth somewhere in the documentation ?

spencergibb:
There will be dependency issues since the brave module will now depend on spring-boot and spring-cloud does not have dependencies that depend on spring-boot. If this is provided, sleuth won't be able to depend on it and will, therefore, duplicate auto-configuration.

spencergibb:
ping @dsyer

adriancole:
My 2p is still start with one for the tracing things, like we have for
spring beans. We get a fair amount of support requests for basics like
setting up the tracer. People forget to close it and that costs a lot of
time. Let's start with that.. ex at least Tracing bean (likely also
HttpTracing and any future MessagingTracing) and the senders?

Then, the examples for those in ad-hoc or otherwise become small and
dependency mess won't be present.

adriancole:
this is nearing a year old.. should we close it?

jonathan-lo:
We won't be needing this in the near future.

adriancole:
okie dokie. thanks for the feedback

PR #747, opened 2018-08-04
Decouples Brave from zipkin2.Endpoint

### LINKED ISSUE #745: Decouple from zipkin2.Endpoint
One thing that hurt us updating from zipkin v1 library to zipkin v2 was the Endpoint type. It caused a fair amount of library dancing and a difficult to remove dependency. For example, assigning a remoteEndpoint caused an overload on a close, but still 3rd party type. Moreover, we know people want to change only specific pieces of the endpoint. For example, just the port for services. Sometimes, like in camel or sidecars, folks want to change the local service name. It can be tricky to do this without accidentally messing up the IP information. Finally, whenever this occurs, there's some overhead due to re-allocating Endpoint objects.

The idea here is to decouple the tracing apis from zipkin2.Endpoint types, instead using a brave cover type, which in the case of NoOp does nothing. Underneath a "MutableEndpoint" type would be used in recording which is only conditionally present when overriding defaults. We'd deprecate the existing apis which ask for `zipkin2.Endpoint` for removal in Brave 6.

adriancole:
This adds methods that decouple Brave apis from zipkin2.Endpoint:
* Tracing.Builder.localIp, localPort for global config
* Span.remoteServiceName, remoteIpAndPort for remote endpoint

Fixes #745


adriancole:
I owe some more tests and benchmarks, but currently happy with the shape of this.

brave/src/main/java/brave/Tracing.java
```diff
@@ -295,12 +324,21 @@ public Tracing build() {
       this.stringPropagation = builder.propagationFactory.create(Propagation.KeyFactory.STRING);
       this.currentTraceContext = builder.currentTraceContext;
       this.sampler = builder.sampler;
-      SpanReporter reporter = new SpanReporter(builder.reporter, noop);
+      MutableEndpoint localEndpoint = new MutableEndpoint();
+      localEndpoint.serviceName(builder.localServiceName);
+      localEndpoint.ip(builder.localIp);
+      localEndpoint.port(builder.localPort);
+      Endpoint localZipkinEndpoint = Endpoint.newBuilder()
```
adriancole: note: you'll see stuff like internal use of zipkin2.Endpoint. The first goal is removing that from the user api. Removing from internals will be low-priority refactoring post merge.

brave/src/main/java/brave/internal/Platform.java
```diff
@@ -178,6 +177,10 @@ static Jre7 buildIfSupported() {
       return null;
     }
 
+    @IgnoreJRERequirement @Override public String getHostString(InetSocketAddress socket) {
```
adriancole: subtle, but socket.getHostString was added in java 7. We still support lowest java level. For example, netty and dubbo are Java 6+.

brave/src/main/java/brave/internal/recorder/MutableEndpoint.java
```diff
@@ -0,0 +1,176 @@
+package brave.internal.recorder;
+
+import brave.internal.Nullable;
+
+/** The network context of a node in the service graph. */
+public final class MutableEndpoint {
```
adriancole: this type might go away, or become package private

brave/README.md
```diff
@@ -166,10 +166,8 @@ Here's an example of a client span:
 // before you send a request, add metadata that describes the operation
 span = tracer.nextSpan().name(service + "/" + method).kind(CLIENT);
 span.tag("myrpc.version", "1.0.0");
-span.remoteEndpoint(Endpoint.newBuilder()
-    .serviceName("backend")
-    .ip("172.3.4.1")
-    .port(8108).build());
+span.remoteServiceName("backend");
+span.parseRemoteIpAndPort("172.3.4.1", 8108);
```
jcchavezs: The name `parseRemoteIpAndPort` makes me think this function is returning something instead of being a void. What about just `remoteIpAndPort`?

brave/src/main/java/brave/internal/Platform.java
```diff
@@ -45,31 +45,30 @@
     }
   }
 
-  public Endpoint endpoint() {
+  @Nullable public abstract String getHostString(InetSocketAddress socket);
```
jcchavezs: Is the `String` suffix necessary? Isn't `getHostAddress` better since the type is string already?

brave/src/main/java/brave/internal/recorder/MutableEndpoint.java
```diff
@@ -0,0 +1,176 @@
...
+        hasColon = true;
+      } else if (notHex(c)) {
+        return IpFamily.Unknown; // Everything else must be a decimal or hex digit.
+      }
+    }
+
+    // Now decide which address family to parse.
+    if (hasColon) {
+      if (hasDot) {
+        int lastColon = ipString.lastIndexOf(':');
```
jcchavezs: `lastColonPos` or `lastColonIndex` is maybe a better name. WDYT?

brave/src/main/java/brave/internal/recorder/MutableSpan.java
```diff
@@ -47,6 +49,22 @@ public void kind(Kind kind) {
...
+  /** @see brave.Span#remoteServiceName(String) */
+  public void remoteServiceName(String remoteServiceName) {
+    if (remoteServiceName == null) throw new NullPointerException("remoteServiceName == null");
+    if (remoteEndpoint == null) remoteEndpoint = new MutableEndpoint();
+    remoteEndpoint.serviceName(remoteServiceName);
+  }
+
+  /** @see brave.Span#parseRemoteIpAndPort(String, int) */
+  public boolean parseRemoteIpAndPort(@Nullable String remoteIp, int remotePort) {
+    if (remoteIp == null) return false;
```
jcchavezs: I am curious about this interface, here you return `true/false` whereas in `removeServiceName` you throw an exception.

brave/src/main/java/brave/internal/recorder/MutableEndpoint.java
jcchavezs: I could not find a unit test for this class methods, I believe this one specially needs one due to the many ifs. What do you think?

adriancole: yep indeed this needs unit tests.. there area a couple other new methods that do as well. good point

brave/README.md
adriancole: the parse part is also used in zipkin2.Endpoint which is trying to be different. It returns true if it was successful. We don't want to trust the string values as one thing that's common is people put hostnames etc where IPs should be. So parse is implying this..

It might be better named validateAndSetRemoteIpAndPort?

brave/src/main/java/brave/internal/Platform.java
adriancole: this one is a language level guard on InetSocketAddress.getHostAddress, which isn't available until v1.7. So `getHostAddress(socket)` reads similarly to `socket.getHostAddress` intentionally. I'll add javadoc as yeah not clear.

brave/src/main/java/brave/internal/recorder/MutableEndpoint.java
adriancole: copy/pasted code, but agreed it is a better name. I'll switch in both places.

brave/src/main/java/brave/internal/recorder/MutableSpan.java
adriancole: if you look at the javadoc on Span.parseRemoteIpAndPort you'll notice that it is common to have to look in multiple places for ip information. Ex. here

```java
    InetAddress address = target.getAddress();
    if (address != null) {
      if (span.parseRemoteIpAndPort(address.getHostAddress(), target.getPort())) return;
    }
    span.parseRemoteIpAndPort(target.getHostName(), target.getPort());
```

This kindof code looks even worse if you are constantly checking for null. It is special cased similar to how code was done in zipkin2.Endpoint due to the crazy amount of null checking you'd need otherwise :P

remoteServiceName is like all the other fields on span.. we shouldn't set them to null. the only odd one is the remote socket info.

adriancole: for example, consider address.getHostAddress() can also return null and target.getPort() could conceivably return -1. When looking at all the call sites, it made this code least atrocious to be a validating call, notably as we don't depend on IP info anyway strictly. Also, in the case of no-op we "trick" by returning true always. This means a bit better performance when we are recording nothing. This is why there are tons of notes on the javadoc, but anyway the call sites are significantly less ugly this way imho.

adriancole:
sometimes we use the word maybe.. ex `maybeRemoteIpAndPort` which returns true if it is, could be more sensible than `parseRemoteIpAndPort`.. especially as we usually use the word parse when scraping types like HttpRequest, and in this case it is just a string and int. 

@jcchavezs @devinsba @marcingrzejszczak @llinder @devinsba @shakuzen @shimamoto any preferences? I agree this method is the odd-ball, definitely need to pick something sensible.

Here are some choices with real code. I didn't add "AndSet" as there are too many ands. Plus we don't use the word "set" in any of our code, but I could be convinced.:

Parse (current)
```java
span.remoteServiceName("backend");
span.parseRemoteIpAndPort("172.3.4.1", 8108);

  /**
   * This sets the client IP:port to the {@linkplain HttpServletRequest#getRemoteAddr() remote
   * address} if the {@link HttpServerAdapter#parseClientIpAndPort default parsing} fails.
   */
  @Override public boolean parseClientIpAndPort(HttpServletRequest req, Span span) {
    if (super.parseClientIpAndPort(req, span)) return true;
    return span.parseRemoteIpAndPort(req.getRemoteAddr(), req.getRemotePort());
  }
```

Maybe (different word than parse)
```java
span.remoteServiceName("backend");
span.maybeRemoteIpAndPort("172.3.4.1", 8108);
  /**
   * This sets the client IP:port to the {@linkplain HttpServletRequest#getRemoteAddr() remote
   * address} if the {@link HttpServerAdapter#parseClientIpAndPort default parsing} fails.
   */
  @Override public boolean parseClientIpAndPort(HttpServletRequest req, Span span) {
    if (super.parseClientIpAndPort(req, span)) return true;
    return span.maybeRemoteIpAndPort(req.getRemoteAddr(), req.getRemotePort());
  }
```

OrFalse? (allows you to read the methods lexicographically and have a suffix telling you it is different). Might be a different suffix choice
```java
span.remoteServiceName("backend");
span.remoteIpAndPortOrFalse("172.3.4.1", 8108);

  /**
   * This sets the client IP:port to the {@linkplain HttpServletRequest#getRemoteAddr() remote
   * address} if the {@link HttpServerAdapter#clientIpAndPortOrFalse default parsing} fails.
   */
  @Override public boolean clientIpAndPortOrFalse(HttpServletRequest req, Span span) {
    if (super.clientIpAndPortOrFalse(req, span)) return true;
    return span.remoteIpAndPortOrFalse(req.getRemoteAddr(), req.getRemotePort());
  }
```

adriancole:
PS so far I am still fine with the parse prefix as it is used everywhere else, although only zipkin2.Endpoint returns boolean when doing so.

adriancole:
So far, I think this is a performance win outside statistical error, even if only a few microseconds

Before
```
Benchmark                                            Mode  Cnt    Score   Error  Units
EndToEndBenchmarks.server_get                        avgt   15  142.404 ± 4.273  us/op
EndToEndBenchmarks.traced128Server_get               avgt   15  198.336 ± 2.636  us/op
EndToEndBenchmarks.traced128Server_get_resumeTrace   avgt   15  202.861 ± 4.952  us/op
EndToEndBenchmarks.tracedExtraServer_get             avgt   15  210.209 ± 5.718  us/op
EndToEndBenchmarks.tracedExtraServer_get_request_id  avgt   15  213.879 ± 3.727  us/op
EndToEndBenchmarks.tracedServer_get                  avgt   15  199.546 ± 3.893  us/op
EndToEndBenchmarks.tracedServer_get_resumeTrace      avgt   15  207.808 ± 5.562  us/op
EndToEndBenchmarks.unsampledServer_get               avgt   15  157.828 ± 3.170  us/op
```

After
```
Benchmark                                            Mode  Cnt    Score    Error  Units
EndToEndBenchmarks.server_get                        avgt   15  142.165 ±  3.012  us/op
EndToEndBenchmarks.traced128Server_get               avgt   15  193.668 ±  3.397  us/op
EndToEndBenchmarks.traced128Server_get_resumeTrace   avgt   15  195.480 ±  2.659  us/op
EndToEndBenchmarks.tracedExtraServer_get             avgt   15  206.897 ± 13.223  us/op
EndToEndBenchmarks.tracedExtraServer_get_request_id  avgt   15  210.645 ±  5.961  us/op
EndToEndBenchmarks.tracedServer_get                  avgt   15  191.812 ±  1.660  us/op
EndToEndBenchmarks.tracedServer_get_resumeTrace      avgt   15  197.930 ±  6.315  us/op
EndToEndBenchmarks.unsampledServer_get               avgt   15  152.701 ±  2.938  us/op
```

adriancole:
renamed Span.parseRemoteIpAndPort to remoteIpAndPort to avoid the distraction. lemme know if works for you @jcchavezs!

adriancole:
revised description

Issue #752, opened 2018-08-09
Feature request: Upgrade Kafka Clients API to 2.0.0


jeqo:
Latest version of Apache Kafka breaks compatibility by removing support for Java 7 and adding more methods to Kafka Consumers, which make it fail when using Kafka Clients brave instrumentation (e.g. `poll(Duration timeout)` is not implemented by `TracingKafkaConsumer`.

Kafka Streams v2.0 includes support for headers that could enable instrumentation for tracing but this can be discussed in other issue once this issue is resolved.

jeqo:
I have a working branch: https://github.com/jeqo/brave/tree/kafka-clients-v2 but requires `kafka-junit` to be upgraded also (https://github.com/charithe/kafka-junit/pull/38)

jeqo:
kafka-junit has been upgraded now :) https://github.com/charithe/kafka-junit/pull/38

ImFlog:
Agreed with Adrian, but from what I see in @jeqo branch, it is already done this way by not using @Override on new methods.
Waiting for the PR ;)

jeqo:
There you go https://github.com/openzipkin/brave/pull/753/

jeqo:
Closing as #753 got merged. Thanks for your help!

PR #753, opened 2018-08-09
Upgrade Kafka Clients API to 2.0.0

jeqo:
Changes: 
* Upgrading Kafka Clients versions to 2.0.0
* Upgrade Kafka JUnit dependency to 4.1.2

Solves: https://github.com/openzipkin/brave/issues/752

### DIFF
--- a/instrumentation/kafka-clients/src/main/java/brave/kafka/clients/TracingConsumer.java
+++ b/instrumentation/kafka-clients/src/main/java/brave/kafka/clients/TracingConsumer.java
@@ -6,6 +6,7 @@ import brave.internal.Nullable;
 import brave.propagation.TraceContext.Extractor;
 import brave.propagation.TraceContext.Injector;
 import brave.propagation.TraceContextOrSamplingFlags;
+import java.time.Duration;
 import java.util.Collection;
 import java.util.LinkedHashMap;
 import java.util.List;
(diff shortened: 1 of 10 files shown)


jcchavezs:
I am curious about the tests here. Should we cover https://github.com/openzipkin/brave/pull/753/files#diff-1256abdd71e2b447c0bd97db22a69e72R52 or https://github.com/openzipkin/brave/pull/753/files#diff-1256abdd71e2b447c0bd97db22a69e72R259 at least?

ImFlog:
You forgot to remove the `@Override` for new methods (to not break compatibility).
Sorry I said It was ok on the other thread but it is because I saw 
```
// Do not use @Override annotation to avoid compatibility issue version < 1.0
  public void subscribe(Pattern pattern) {
```

jeqo:
@jcchavezs I've added a unit test case for `poll(Duration)` and add it to integration test also. WDYT?

instrumentation/kafka-clients/src/main/java/brave/kafka/clients/TracingConsumer.java
```diff
@@ -49,12 +49,13 @@
     this.remoteServiceName = remoteServiceName;
   }
 
-  @Deprecated @Override public ConsumerRecords<K, V> poll(long timeout) {
-    return poll(Duration.ofMillis(timeout));
+  // Do not use @Override annotation to avoid compatibility issue version < 1.0
```
ImFlog: `version < 2.0`
Also can you put this comment everywhere you removed the annotation ? May avoid some mistakes in the future ;)

jeqo:
@ImFlog I have removed overrides.

instrumentation/kafka-clients/src/main/java/brave/kafka/clients/TracingConsumer.java
jeqo: Yeap

ImFlog:
For what it's worth, LGTM :)

adriancole:
So I will help on this, because I think it is important that we do a runtime check with kafka 1. It is somewhat tricky.

I will do this and rebase your branch over it (with maven invoker)

adriancole:
OK I added the invoker test squashed and force pushed the branch. If
works and no other feedback we can merge

PR #754, opened 2018-08-09
Kafka Streams support

jeqo:
Changes: 

* Adding a KafkaClientSupplier to be injected on Kafka Streams materialization, reusing Kafka Clients instrumentation.
* Editing `TracingProducer` to link incoming spans when sending data out from a Stream Process.


instrumentation/kafka-clients/src/main/java/brave/kafka/clients/TracingProducer.java
```diff
@@ -60,17 +63,26 @@
...
 
-  /** We wrap the send method to add tracing. */
-  @Override
-  public Future<RecordMetadata> send(ProducerRecord<K, V> record, @Nullable Callback callback) {
-    Span span = tracing.tracer().nextSpan();
+  /**
+   * We wrap the send method to add tracing.
+   */
+  @Override public Future<RecordMetadata> send(ProducerRecord<K, V> record, @Nullable Callback callback) {
+    TraceContextOrSamplingFlags traceContextOrSamplingFlags = extractor.extract(record.headers());
```
adriancole: hmm why are we doing this for producer? This adds overhead...

adriancole:
haha lots of merging huh :) @ImFlog wanna take a look?

instrumentation/kafka-clients/src/main/java/brave/kafka/clients/TracingProducer.java
jeqo: Agree, but the issue is when `input-topic -> process -> output-topic`, when producing records to `output-topic` we need to link it with the trace from consumer. This is the easiest approach I found to do it, but happy to re-evaluate the approach.

jeqo:
@adriancole Haha yes, I was back and forth if we need 2.0.0 for this, and we do xD

instrumentation/kafka-clients/src/main/java/brave/kafka/clients/TracingProducer.java
adriancole: in the previous code we require the processor to use a utility (nextSpan I think) to compose this. There was a fairly lengthy chat on gitter about this with @ImFlog so want to see where he got so that we can concentrate efforts.

ImFlog: In Kafka Clients we have this:
```java
  public Span nextSpan(ConsumerRecord<?, ?> record) {
    TraceContextOrSamplingFlags extracted = extractAndClearHeaders(record);
    Span result = tracing.tracer().nextSpan(extracted);
    if (extracted.context() == null && !result.isNoop()) {
      addTags(record, result);
    }
    return result;
  }
```
The idea is to create a span when we really start to treat records in Kafka, this had to be manually done.

If we keep the same idea, we need to call this somewhere before the first transformation on the stream, so probably a transformer to call after `builder.stream(XXX)`.

@jeqo you know well the KStream API, is there anything else than a Transformer that we could use to reduce the work for users ?

ImFlog:
Overall I think we need something like this to test the stream (warning, bulk code paste).
```java
StreamsBuilder builder = new StreamsBuilder();
    builder.stream(INPUT_TOPIC, Consumed.with(Serdes.String(), Serdes.String()))
        .mapValues(val -> val + "_foo")
        .to(OUTPUT_TOPIC);
    Topology topology = builder.build();

    // setup test driver
    Properties props = new Properties();
    props.put(StreamsConfig.APPLICATION_ID_CONFIG, "test");
    props.put(StreamsConfig.BOOTSTRAP_SERVERS_CONFIG, "dummy:1234");
    try (TopologyTestDriver testDriver = new TopologyTestDriver(topology, props)) {
      testDriver.pipeInput(fakeRecord);

      ProducerRecord<String, String> outputRecord = testDriver.readOutput(
          OUTPUT_TOPIC,
          new StringDeserializer(),
          new StringDeserializer());
```
This acts as integration test without the burden of starting an embedded Kafka broker. WDYT ?

instrumentation/kafka-clients/src/main/java/brave/kafka/clients/TracingProducer.java
jeqo: In a second thought, extending the Producer could not be an option, so instead of validating context from headers on the producer side always, we can only do it when there is no current span available.

jeqo:
@ImFlog checking the `TopologyTestDriver` in more detail, it actually mocks the Client interaction part, so it won't be possible to use it for unit-testing.

ImFlog:
Ok, so back to kafka junit :)

adriancole:
let me know if you want help with your branch. I could help rebase it over master and force push as a single commit to proceed with.. or just leave it be :P

instrumentation/kafka-clients/src/main/java/brave/kafka/clients/TracingProducer.java
```diff
@@ -59,10 +62,17 @@
...
-  /** We wrap the send method to add tracing. */
-  @Override
-  public Future<RecordMetadata> send(ProducerRecord<K, V> record, @Nullable Callback callback) {
-    Span span = tracing.tracer().nextSpan();
+  /**
+   * We wrap the send method to add tracing.
+   */
+  @Override public Future<RecordMetadata> send(ProducerRecord<K, V> record, @Nullable Callback callback) {
+    final Span span;
+    if (tracing.tracer().currentSpan() == null) {
```
adriancole: This style assumes someone didn't use the apis that we tell them to use.. probably needs more comment here (and tests if we decide to go with it) explaining why we are checking for someone not using things.

At any rate this will hurt perf.. if you want to hurt less, you can do this:

```java
TraceContext maybeParent = tracing.currentTraceContext().get();
if (maybeParent == null) { // check to see if someone forgot to call something
 ... calling extract will also hurt perf! that's why we should be careful not to call this unless we have good reason commented
} else {
... tracer.newChild(parent);
}
```

jeqo: thanks for your comments @adriancole 
Will add comments and test for this. Also, we can hurt a bit less if we only perform `extract` when there are headers available, then users that are not using them will not be affected.

instrumentation/kafka-clients/src/main/java/brave/kafka/clients/TracingProducer.java
```diff
@@ -59,10 +63,24 @@
...
+  /**
+   * We wrap the send method to add tracing.
+   * This method is expecting an existing current trace context to create a child span.
+   * If there is no current context, an additional operation will be performed to extract
+   * context from record headers. This could impact performance if no context is injected
+   * on record headers, but is required to propagate context on Kafka Streams instrumentation.
+   */
+  @Override public Future<RecordMetadata> send(ProducerRecord<K, V> record, @Nullable Callback callback) {
+    final Span span;
+    final TraceContext maybeParent = tracing.currentTraceContext().get();
```
adriancole: the key here is that the order is flipped. for example, usually we try to extract and then if that fails use the current span. Here we are looking for the current span first (because we are on the sender side!)

adriancole:
I'm going to merge this as it is good. I will do a follow-up with fixies. Really good stuff folks!

adriancole:
cleanups are here.. no logic change. https://github.com/openzipkin/brave/pull/770

let's carry forward thinking.. I think we have a week or so to hammer out details before next release

adriancole:
If we do a release tomorrow, should we keep this feature or pull it?

adriancole:
actually I'm going to pull it as the kafka-streams part is quite small (even if tests aren't) and easy to copy/paste until we flesh this out further.

adriancole:
#781 will be an easy thing to revert when we are ready

PR #763, opened 2018-08-21
Adds B3SinglePropagation for the "b3" header

adriancole:
This will be used for JMS and other propagation formats such as w3c
tracestate entry. It is notably more efficient than multi-header.

Example header: `b3: 4bf92f3577b34da6a3ce929d0e0e4736-00f067aa0ba902b7-1`

To aid in migration, this teaches the normal B3 to attempt to extract
single header variants as well.

See https://github.com/openzipkin/b3-propagation/issues/21

### DIFF
--- a/brave/src/main/java/brave/propagation/B3SingleFormat.java
+++ b/brave/src/main/java/brave/propagation/B3SingleFormat.java
@@ -52,11 +52,11 @@ final class B3SingleFormat {
 
   static @Nullable TraceContextOrSamplingFlags maybeB3SingleFormat(String b3) {
     int length = b3.length();
-    if (length == 1) {
+    if (length == 1) { // assume just tracing flag. ex "b3: 1"
       int flags = parseSampledFlag(b3, 0);
       if (flags == 0) return null;
       return TraceContextOrSamplingFlags.create(SamplingFlags.toSamplingFlags(flags));
-    } else if (length == 3) {
+    } else if (length == 3) { // assume tracing + debug flag. ex "b3: 1-1"
       int flags = parseSampledFlag(b3, 0);
       if (flags == 0) return null;
       flags = parseDebugFlag(b3, 2, flags);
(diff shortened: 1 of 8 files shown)


adriancole:
cc @openzipkin/instrumentation-owners in case you'd like to see an example

brave/src/main/java/brave/propagation/B3SingleFormat.java
```diff
@@ -0,0 +1,168 @@
...
+    } else if (length > FORMAT_MAX_LENGTH) {
+      logger.fine("Invalid input: too long");
+      return null;
+    }
+
+    int pos = 0;
+    long traceIdHigh, traceId;
+    if (b3.charAt(32) == '-') {
+      traceIdHigh = lenientLowerHexToUnsignedLong(b3, 0, 16);
+      traceId = lenientLowerHexToUnsignedLong(b3, 16, 32);
```
devinsba: You do `lenientLowerHexToUnsignedLong(b3, pos, pos + 16);` for parsing the span ID, did you mean to use `pos` here also to be consistent?

adriancole: I probably meant to do the following in order to save from doing extra math when there will be constant results (due to this being the initial parse). Maybe the distraction isn't worth it.. I can switch to `pos, pos + 16` NP

```java
int pos;
if (b3.charAt(32) == '-') {
  --snip--
  pos = 33; // traceid128-
} else {
  --snip--
  pos = 17; // traceid64-
}
```

adriancole: I moved initialization of pos further down to make it ideally more clear

adriancole:
review after commit welcome. merging to progress some messaging things which hinge on this

adriancole:
This is a follow-up change which gets somewhat obsessive about parsing tests in order to expose functions public cc @devinsba https://github.com/openzipkin/brave/pull/769

PR #764, opened 2018-08-22
Adds JMS 1.1/2.0 instrumentation

### LINKED ISSUE #584: JMS instrumentation
There's an indirect requirement for JMS instrumentation through spring-cloud-sleuth. Besides the multiple versions of JMS apis, there's an important concern with propagation due to naming constraints which exist in JMS, but not necessarily in the underlying transport.

JMS requires that message header names follow java naming conventions. Notably this excludes hyphens and dots. For example, [in Camel](http://people.apache.org/~dkulp/camel/jms.html), there's a naming policy to map to and from these constraints.

For example, many simply downcase B3 headers to `x-b3-traceid` for use in non-http transports (which might not be case insensitive). In JMS, this wouldn't work, even if rabbitmq is used underneath which has no such constraint. Ex, if using camel JMS, this header would map into `x_HYPHEN_b3_HYPHEN_traceid`.

In some cases, JMS headers are replaced based on constants or other patterns, like globally replacing hyphens with underscores. Interestingly opentracing decided on a [different pattern](https://github.com/opentracing-contrib/java-jms/pull/1) `_$dash$_`, though they have no pattern for dots: For example, if you were using the OT library and camel on the other, you'd get `x_HYPHEN_b3_HYPHEN_traceid` on one side and `x_$dash$_b3_$dash$_traceid` on the other, mutually unreadable.

The universal implication of using JMS is that it implies global coordination of these naming patterns, or there will be propagation incompatibility (ex traces will restart). This will also break anything else propagated that isn't a trace ID. This type of problem has [already been noticed](https://github.com/spring-cloud/spring-cloud-sleuth/issues/537) in early users of message tracing.

Importantly, even if we were to have static replacements for B3, we don't know if users will introduce propagated headers with dots or hyphens in them. In other words, I don't think a solution to this problem can be constant in nature, though constants could help.

Challenge here is to introduce support for JMS propagation, but without assuming we are using B3 format, and ideally inheriting a naming convention as opposed to introducing another dimension of naming convention complexity for others. Coming up with a sane default might be particularly tricky.

Calling for advice from @davsclaus @garyrussell @Xylus @jkowall @christian-schwarzbauer @CodingFabian (others feel free to contribute too!)

See also discussion on w3c trace context https://github.com/w3c/distributed-tracing/issues/35

adriancole:
# Brave JMS instrumentation
This module provides instrumentation for JMS 1.1 or 2.0 consumers,
producers and listeners. It works by wrapping connections or connection
factories.

Under the scenes:
* `TracingMessageProducer` - completes a producer span per message and propagates it via headers.
* `TracingMessageConsumer` - completes a consumer span on `receive`, resuming a trace in headers if present.
* `TracingMessageListener` - does the same as `TracingMessageConsumer`, and times the user-supplied listener.

## Setup
First, setup the generic Jms component like this:
```java
jmsTracing = JmsTracing.newBuilder(tracing)
                       .remoteServiceName("my-broker")
                       .build();
```

Now, just wrap your connection or connection factory
```java
ConnectionFactory tracingConnectionFactory = jmsTracing.connectionFactory(connectionFactory);

// When you later send a message, you'll notice "b3" as a message property
producer.send(message);
```

*NOTE* `JMSConsumer.receiveBodyXXX()` methods are not traced due to lack
of hooks. `JMSConsumer.receive()` followed by `Message.getBody()` is the
most compatible alternative. If you desire other routes, please raise an
issue or join https://gitter.im/openzipkin/zipkin

## What's happening?
Typically, there are three spans involved in message tracing:
* If a message producer is traced, it completes a PRODUCER span per message
* If a consumer is traced, receive completes a CONSUMER span based on the incoming queue or topic
* Message listener timing is in a child of the CONSUMER span.

## Custom Message Processing

If you are using Jms `MessageListener`, the following is automatic. If
you have custom code, or are using a different processing abstraction,
read below:

When ready for processing use `JmsTracing.nextSpan` to continue the trace.

```java
// Typically, poll is in a loop. the consumer is wrapped
while (running) {
  Message message = tracingMessageConsumer.receive();
  // either automatically or manually wrap your real process() method to use jmsTracing.nextSpan()
  messages.forEach(message -> process(message));
}
```

If you are in a position where you have a custom processing loop, you can do something like this
to trace manually or you can do similar via automatic instrumentation like AspectJ.
```java
void process(Message message) {
  // Grab any span from the queue. The queue or topic is automatically tagged
  Span span = jmsTracing.nextSpan(message).name("process").start();

  // Below is the same setup as any synchronous tracing
  try (SpanInScope ws = tracer.withSpanInScope(span)) { // so logging can see trace ID
    return doProcess(message); // do the actual work
    span.error(e); // make sure any error gets into the span before it is finished
    throw e;
  } finally {
    span.finish(); // ensure the span representing this processing completes.
  }
}
```

## Notes
* This instrumentation library works with JMS 1.1 and 2.0
* More information about "Message Tracing" [here](https://github.com/openzipkin/openzipkin.github.io/blob/master/pages/instrumenting.md#message-tracing)

Fixes #584


adriancole:
added producer but need tests and compatibility checks

adriancole:
looks like there's no appetite for JMS 2 in normal activemq, but there is in artemis.

http://activemq.2283324.n4.nabble.com/Discussion-JMS-2-0-API-support-tt4681089.html#a4697806

So, I'll swap the libraries and use normal activemq as an invoker test to ensure we don't accidentally depend on JMS 2.

Footnote: it is interesting so little interest in JMS 2, maybe this wasn't worth looking into, but anyway having gotten this far I'll complete it.

adriancole:
For those who regularly give back to zipkin community, in forms of support, advice or just being sympathetic.. here's a free look at how to instrument JMS! I've noticed none of you do, yet, so should save you quite a lot of time I hope.

@felixbarny from https://github.com/elastic/apm-agent-java
@nicmunroe from https://github.com/Nike-Inc/wingtips
@bogdandrutu from https://github.com/census-instrumentation/opencensus-java
@xylus from https://github.com/naver/pinpoint
@ivantopo from https://github.com/kamon-io/Kamon
@wu-sheng from https://github.com/apache/incubator-skywalking

adriancole:
next step is to pull out invoker test to force a clean classpath to check JMS 1.1 strictly, then write README, then pretty much done

adriancole:
invoker tests now run, but normal tests have junk in the log due to artemis config.. I'll sort that out and write readme after lunch

adriancole:
polished and ready to go. PTAL!

zeagord:
I can give it a go once it is ready in my local environments to see if there are any surprises.

adriancole:
self note: forgot unit tests for jms message header utility. forgot benchmarks

jeqo:
Tested on WebLogic JMS 12.2 https://github.com/jeqo/poc-brave-jms-weblogic and looking good!
Wondering if brave will also cover `JMSConsumer` and `JMSProducer` APIs introduced on JMS 2.0 also? reference: https://www.oracle.com/technetwork/articles/java/jms20-1947669.html

adriancole:
@jeqo sure will give a go. Thanks for the link

adriancole:
@jeqo I think 2.0 support implies wrapping the `JMSContext`, as otherwise when people call something like JMSConsumer.receiveBody() you might not know what the destination was..

adriancole:
switching "jms.destination" to "jms.topic" and "jms.queue" as destination has no common format which renders it useless for search

adriancole:
need to look into if there is any way to continue a tracing using `JMSConsumer.receiveBody()` as it isn't obvious how you'd hook into the message...

adriancole:
looks like internally activemq implements like this:

```java
   @Override
   public <T> T receiveBodyNoWait(Class<T> c) {
      try {
         Message message = consumer.receiveNoWait();
         context.setLastMessage(ActiveMQJMSConsumer.this, message);
         return message == null ? null : message.getBody(c);
      } catch (JMSException e) {
         throw JmsExceptionUtils.convertToRuntimeException(e);
      }
   }
```

doing the same is fine, but I don't think we can because it will interfere with the ack tracking..

adriancole:
or we can just ack on close like they do..

adriancole:
nah can't interfere as implementations like qpid have a lot more logic.. https://github.com/apache/qpid-jms/blob/052125c5c5f1f4ea43ff9de5481a7de2a81bff7a/qpid-jms-client/src/main/java/org/apache/qpid/jms/JmsMessageConsumer.java#L229

adriancole:
even if we did, we couldn't continue the trace. For example, receiveBody returns an object. There's no rendezvous we can easily use to continue a trace. There's no thread context (as this is a return value). Maybe we could stuff.. something in a thread local, but there's no reason to expect someone to check there later.

For now, best thing I can think of is document that the "receiveBody" helper methods are unsupported and will result in traces breaking. Someone who later figures out a solution will at least have notes to help along the way.

adriancole:
current thought is to discourage use of JMS 2.0 `receiveBodyX` methods if observability is a priority. There's no clean way to hook into this and it is not adding much value to call sites over `receive(), message.toBody()`. Considering many drivers are commercially licensed, we won't be able to OSS all the possible hooks.

Another way is to make a `receive(), message.toBody()` decorator that users can opt-into. Will wait for demand on that.

adriancole:
ok waiting for feedback on JMSxxx 2.0 stuff before backfilling tests and benchmarks cc @jorgheymans @jeqo

adriancole:
PS It might be less work to create a TracingSession.. vs wrapping producers and consumers individually. Have to see how this fits with things like spring-jms first..

adriancole:
added more wrap than...

anyway if someone would like to try using JmsTracing.connection or connectionFactory would be grand. I can do more gruntwork after this feels it works (tons of tests etc)

jeqo:
Tested with JMS 2.0 APIs.
I'm very happy with the current API :+1:  thanks @adriancole !!

adriancole:
internally there's still too much copy/paste in message production side... this is due to a large inheritance tree. I'll try an enum command dispatch approach to... dispatch this problem

adriancole:
started tree'ing out the tests. Note: I'm not going for 100% coverage, but will make sure the dodgy areas are covered, particularly method dispatch. Planning on wrapping this up today

adriancole:
@jeqo thanks for all the testing.. it is really important!

adriancole:
OK by the way.. the radio silence was due to realizing JMS libraries commonly implement types "flat" meaning for example, connection factory also implements queue connection factory etc. The previous code path of treeing out the hierarchy actually interfered with that, masking based on what was passed in. Now, the code works more like artemis (activemq) by having some dispatch approach instead.

adriancole:
OK I will merge work in progress so that folks can test using snapshots which is ever more easy than using custom builds. test and polishing is easy as all the types are hidden except "JmsTracing"

adriancole:
On our snapshot repo (http://oss.jfrog.org/oss-snapshot-local) 5.3.0-SNAPSHOT should publish soon, including this feature

PR #769, opened 2018-08-22
Exposes B3SingleFormat after getting meticulous about parsing

adriancole:
This adds a huge amount of state checks to understand the source of
malform problems. This exposes B3SingleFormat methods so they can be
used in other code, such as w3c tracestate or JMS directly.

See #763

### DIFF
--- a/brave/src/main/java/brave/propagation/B3SingleFormat.java
+++ b/brave/src/main/java/brave/propagation/B3SingleFormat.java
@@ -81,10 +81,10 @@ public final class B3SingleFormat {
   }
 
   /**
-   * Like {@link #writeB3SingleAsBytes(TraceContext)}, but for carriers with byte array or byte
-   * buffer values. For example, {@link ByteBuffer#wrap(byte[])} can wrap the result.
+   * Like {@link #writeB3SingleFormatAsBytes(TraceContext)}, but for carriers with byte array or
+   * byte buffer values. For example, {@link ByteBuffer#wrap(byte[])} can wrap the result.
    */
-  public static byte[] writeB3SingleAsBytes(TraceContext context) {
+  public static byte[] writeB3SingleFormatAsBytes(TraceContext context) {
     char[] buffer = getCharBuffer();
     int length = writeB3SingleFormat(context, context.parentIdAsLong(), buffer);
     return asciiToNewByteArray(buffer, length);
--- a/brave/src/main/java/brave/propagation/B3SingleFormat.java
+++ b/brave/src/main/java/brave/propagation/B3SingleFormat.java
@@ -2,6 +2,7 @@ package brave.propagation;
 
 import brave.internal.HexCodec;
 import brave.internal.Nullable;
+import java.nio.ByteBuffer;
 import java.util.Collections;
 import java.util.logging.Logger;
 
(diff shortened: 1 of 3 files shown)


adriancole:
fyi specifically for tracestate I think the parser needs to accept a (charsequence, pos, limit) or similar to avoid having to split strings.

and also here are the updated benchmark results so far:

```
B3SinglePropagationBenchmarks.extract:extract·p0.999                       sample             2.260            us/op
B3SinglePropagationBenchmarks.extract:·gc.alloc.rate.norm                  sample      15    80.010 ±  0.001    B/op
B3SinglePropagationBenchmarks.extract_malformed:extract_malformed·p0.999   sample             7.506            us/op
B3SinglePropagationBenchmarks.extract_malformed:·gc.alloc.rate.norm        sample      15     0.004 ±  0.001    B/op
B3SinglePropagationBenchmarks.extract_nothing:extract_nothing·p0.999       sample             0.149            us/op
B3SinglePropagationBenchmarks.extract_nothing:·gc.alloc.rate.norm          sample      15     0.001 ±  0.001    B/op
B3SinglePropagationBenchmarks.extract_unsampled:extract_unsampled·p0.999   sample             0.240            us/op
B3SinglePropagationBenchmarks.extract_unsampled:·gc.alloc.rate.norm        sample      15     0.002 ±  0.001    B/op
B3SinglePropagationBenchmarks.inject:inject·p0.999                         sample             9.232            us/op
B3SinglePropagationBenchmarks.inject:·gc.alloc.rate.norm                   sample      15   320.016 ±  0.003    B/op
```

adriancole:
@drolando @ivantopo @nicmunroe fyi I will give a one week "quiet time" then brief you on all this stuff. That way you don't have to thrash through various optimizations :P

brave/src/main/java/brave/propagation/B3SingleFormat.java
```diff
@@ -50,7 +74,7 @@ static String writeB3SingleFormat(TraceContext context) {
     return new String(result, 0, pos);
   }
 
-  static @Nullable TraceContextOrSamplingFlags maybeB3SingleFormat(String b3) {
+  @Nullable public static TraceContextOrSamplingFlags parseB3SingleFormat(String b3) {
```
adriancole: this part, changing to CharSequence is one idea.

PS @llinder @michaelsembwever I think we should switch cassandra to use this single-header propagation format, since it is a good time anyway (as currently trace propagation is turned off in brave-cassandra!)

brave/src/main/java/brave/propagation/B3SingleFormat.java
```diff
@@ -73,49 +97,83 @@ static String writeB3SingleFormat(TraceContext context) {
...
-      parentId = lenientLowerHexToUnsignedLong(b3, pos, pos + 16);
-      if (parentId == 0L) {
-        logger.log(FINE, "Invalid input: expected a 16 lower hex parent ID at offset {0}", pos);
+    if (length > pos) {
+      // If we are at this point, we have more than just traceId-spanId.
+      // If the sampling field is present, we'll have a delimiter 2 characters from now. Ex "-1"
+      // If it is absent, but a parent ID is (which is strange), we'll have at least 17 characters.
+      // Therefore, if we have less than two characters, the input is truncated.
+      if (length == pos + 1) {
+        logger.fine("Invalid input: truncated");
```
adriancole: another idea I have is to make a base type which accepts a "parse error" operation. by default, this can go to JUL, but it could also go into a place where it can be a span tag.. for example propagation fail to parse reason.

adriancole:
added ability to write byte array values. this is needed for cassandra (ByteBuffer) and kafka (byte[]) propagation

adriancole:
incidentally grpc also uses byte arrays

adriancole:
test coverage is 100% even edge cases.. will add the ability to parse ByteBuffer next. This will allow cassandra and kafka to not have to allocate arrays to read their data (ex ByteBuffer.wrap(array)  doesn't copy an array)

PR #773, opened 2018-08-24
Simplifies b3 single format by formalizing debug as a sampling status

adriancole:
As discussed, this makes debug use the letter 'd' in the same slot as
unsampled '0' or sampled '1'. The result is much simpler code and
ideally more a more intuitive header.

See https://github.com/openzipkin/b3-propagation/issues/21#issuecomment-415614217


adriancole:
I will revert this if the topic once settles goes in another direction. If we need to do a release prior to feedback, I'll make all the related types hidden until that's not the case.

PR #774, opened 2018-08-26
Ensures Kafka callback has span in scope

adriancole:
Callbacks should have a span in scope so that log IDs are assigned.


adriancole:
cc @ImFlog @jeqo

PR #775, opened 2018-08-27
Forces invoker to run the IT tests we thought it did

adriancole:
We meant to run tests named `IT*` which are typically executed with the failsafe plugin (not the surefire one as that looks for tests name `*Test`. The invoker plugin doesn't run the verify phase which would have invoked failsafe. Attempting various things to no avail, this fakes the plugin that does run (surefire) to run the tests we meant to run.


adriancole:
circleci looks bad, but it usually does. merging and will look into it if the error is still there on a master build (pull requests have been failing in circleci for a long time now)

PR #781, opened 2018-09-07
Temporarily stops publishing kafka-streams

adriancole:
kafka-streams has not been released yet, and it likely shouldn't until
we sort out making intermediate spans. Meanwhile, folks can copy/paste
the small amount of setup that exists.


jeqo:
I'll be back to work on this at the middle of next week.

adriancole:
awesome!

Issue #785, opened 2018-09-11
Ensure Brave types are unloadable


adriancole:
Issue #716 discusses a small leak where memory allocated to thread locals isn't "cleaned up" when brave's tracing instance is closed. There's another problem which this describes:

Thread locals that hold a reference to a Brave type will prevent those classes from being unloaded. This is a form of leak. We can mitigate this by not using Brave types in thread locals, and/or switch to an alternate thread local store like https://github.com/raphw/weak-lock-free

Here's the full story by @raphw (thanks tons for the scoop!)
> When an application is undeployed, it is usually garbage collected. But if one of your classes is stuck in a thread, the class cannot be collected and neither can its class loader which again references all of its loaded classes. You have created a leak, so to speak. Tomcat does however detect this and renews its thread objects in the server's thread pool. Other application servers cannot do this, though.

> To avoid this, you must reference a thread local value in one of your instances/classes that will be part of the application being undeployed. To create a thread local value, you are using the thread as a key of a map to the value which emulates thread locals behavior. Doing so, you do however now stop tomcat from renewing its threads as those threads stick in you map and cannot be collected.

> Yet, now you are in control and can store the thread references weakly. To avoid an overallocation of weak references, you need to mask the keys and reuse the weak references you created previously. To avoid all the clutter, you can just use this library: https://github.com/raphw/weak-lock-free - either with inline expunction (that is how Javas weak map works) or you start a dedicated thread to clear out the weak refs.

cc also @trustin for fun and good measure

adriancole:
this one has most of the core work, though it still breaks.. maybe I'll figure it out next flight

adriancole:
going out in 5.3.2 thanks @swankjesse @raphw for tips.. we now have tests to prove things are unloadable. ex: https://github.com/openzipkin/brave/blob/master/brave-tests/src/test/java/brave/TracingClassLoaderTest.java

PR #787, opened 2018-09-11
Make cores types unloadable

adriancole:
This changes internal code so that thread locals don't hold references to Brave types. After this change, applications that use brave in a child classloader will themselves be unloadable.

See #785

### DIFF
--- a/brave-tests/src/main/java/brave/test/util/ClassLoaders.java
+++ b/brave-tests/src/main/java/brave/test/util/ClassLoaders.java
@@ -10,6 +10,7 @@ import java.net.URL;
 import java.net.URLClassLoader;
 import java.util.function.Consumer;
 import java.util.function.Supplier;
+import java.util.logging.LogManager;
 
 import static org.assertj.core.api.Assertions.assertThat;
 
@@ -68,6 +69,11 @@ public final class ClassLoaders {
 
   /** Runs the type in a new classloader that recreates brave classes */
   public static void assertRunIsUnloadable(Class<? extends Runnable> runnable, ClassLoader parent) {
+    // We can't use log4j2's log manager. More importantly, we want to make sure loggers don't hold
+    // our test classloader from being collected.
+    System.setProperty("java.util.logging.manager", LogManager.class.getName());
+    assertThat(LogManager.getLogManager().getClass()).isSameAs(LogManager.class);
+
     WeakReference<ClassLoader> loader;
     try {
       loader = invokeRunFromNewClassLoader(runnable, parent);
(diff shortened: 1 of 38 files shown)


brave-tests/src/main/java/brave/test/util/ClassLoaders.java
```diff
@@ -0,0 +1,103 @@
...
+   * Creates a new classloader that reloads types matching the given prefix. This is used to test
+   * behavior such a leaked type in a thread local.
+   *
+   * <p>This works by using a bridge loader over the normal system one. The bridge loader always
+   * loads new classes when they are prefixed by brave types.
+   *
+   * <p>This approach is the same as what's used in Android's {@code tests.util.ClassLoaderBuilder}
+   * See https://android.googlesource.com/platform/libcore/+/master/support/src/test/java/tests/util/ClassLoaderBuilder.java
+   */
+  static ClassLoader reloadClassNamePrefix(ClassLoader parent, String prefix) {
```
adriancole: thanks @swankjesse for the link

brave-tests/src/test/java/brave/ThreadLocalSpanClassLoaderTest.java
```diff
@@ -0,0 +1,49 @@
...
+  }
+
+  @Test public void nextAndRemove_unloadable() {
+    assertRunIsUnloadable(getClass().getClassLoader(), CurrentTracerRemovesSpan.class);
+  }
+
+  static class CurrentTracerRemovesSpan implements Runnable {
+    @Override public void run() {
+      try (Tracing tracing = Tracing.newBuilder().build()) {
+        ThreadLocalSpan.CURRENT_TRACER.next();
```
adriancole: this test bombs.. I'm not sure I have the debugging skills to figure out what is holding up the classloader. General tricks suggested if folks know some libraries or tools that can print out what's holding something from being gc'd. cc @raphw

adriancole:
Got a lot of progress but stumped on something.. I'd like to make sure everything is covered test-wise, just one bozo code seems to still be uncollectable even when it appears there are none of our types still hooked up. https://github.com/openzipkin/brave/blob/core-threadlocal-fixies/brave/src/main/java/brave/propagation/ThreadLocalSpan.java#L152

adriancole:
Added a commit that ensures log4j2's log manager doesn't lead anyone in the future to a distracting wander :P

adriancole:
ok got the core code sorted. Later pull requests will ensure instrumentation don't also have similar problems.

PR #797, opened 2018-09-26
WIP - Add thread factory for ForkJoinPool

devinsba:
I've added in a WIP class that wraps a Java 7 class. Interested in thought around this idea


adriancole:
Incidentally I think it is ok to have this in the main module as long as there is no hard reference to it in critical paths java 6 code has to call. We have some other code that is Java 8 which is guarded prior to use (Ex in brave.internal.Platform), which implies we cannot compile on Java 7 anyway.

devinsba:
OK, I was hesitant to add new classes since I know that is a concern sometimes (android). I'll rework

adriancole:
next step will be unit tests I guess. We have some for the other executory things.


you can also take a peek at @talsma-ict https://github.com/talsma-ict/context-propagation which likely has some related work

devinsba:
Thanks for the pointers. I've only got time to work on things in the
mornings so I'll keep going tomorrow.

pom.xml
```diff
@@ -502,6 +502,9 @@
               <artifactId>${main.signature.artifact}</artifactId>
               <version>1.0</version>
             </signature>
+            <ignores>
+              <ignore>java.util.concurrent.ForkJoin*</ignore>
```
adriancole: yep that or there is an annotation I think we use elsewhere (in Platform.java). Might be better to use the annotation

devinsba: From gitter `@IgnoreJRERequirement`

rigsby-chi-activeviam:
`ForkJoinTask` also needs to be wrapped like wraping `Runnable` and `Callable`. However, methods in `ForkJoinTask` like `doExec()` and `exec()` are protected/private. Even if you can wrap it (I tried using reflection to invoke the method and it works), it can only help if the task is submitted using the `submit` method. It won't help if a task fork itself (e.g. `forkJoinTask.fork()`).

devinsba:
@rigsby-chi-activeviam Thanks, I have work in progress on the task portion as well, I will take a look. Do you by chance have the ability to make a simple little test app I could use for debugging?

rigsby-chi-activeviam:
@devinsba Please check the project [brave-with-forkjoinpool-example](https://github.com/rigsby-chi-activeviam/brave-with-forkjoinpool-example) which I created for my testing (so the code is a bit messy..sorry about that..). It also contains my current solution for your reference. Hope this help. 

@adriancole I borrowed your [brave-webmvc-example](https://github.com/openzipkin/brave-webmvc-example) and modified it for the ForkJoinPool example. Thanks a lot =]

rigsby-chi-activeviam:
@devinsba May I know if you have progress on the fork join task handling? I already gave up my "wrapping" solution. I found that there are many normal fork-join use cases it cannot handle. Below is one of the example:
```
wrappedForkJoinPool.submit(forkJoinTask)
forkJoinTask.get() // Dead here. wait indefinitely
```

adriancole:
hey rigsby are you in a position to help fill the gaps or add tests to this
work? trying to figure out if it is an implementation limitation or
something not possible.

adriancole:
@devinsba need a hand? continuing this? I can put some time into it if you like.

rigsby-chi-activeviam:
> hey rigsby are you in a position to help fill the gaps or add tests to this work? trying to figure out if it is an implementation limitation or something not possible.

I am quite sure that wrapping the ForkJoinPool like how `CurrentTraceContext` does for `ExecutorService` is a dead end. So far I am running out of idea on how to proceed.

devinsba:
I am going to close this for now. I don't think this approach will work, but I absolutely want to have this working. I'll take some time on this after the holidays

PR #798, opened 2018-09-26
Proposes ThreadLocal optimizations for the default current trace context

adriancole:
Follows https://github.com/kamon-io/Kamon/pull/553

Thanks @dpsoft @trask cc @trustin

### DIFF
--- a/brave/src/main/java/brave/propagation/CurrentTraceContext.java
+++ b/brave/src/main/java/brave/propagation/CurrentTraceContext.java
@@ -175,7 +175,8 @@ public abstract class CurrentTraceContext {
    */
   public static final class Default extends ThreadLocalCurrentTraceContext {
     // Inheritable as Brave 3's ThreadLocalServerClientAndLocalSpanState was inheritable
-    static final InheritableThreadLocal<TraceContext> INHERITABLE = new InheritableThreadLocal<>();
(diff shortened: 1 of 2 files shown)


adriancole:
Note I ran benchmarks a lot and it won't come out in normal single threaded instrumentation, but could be very helpful for instrumentation that schedules a lot cc also @koji @anuraaga

adriancole:
this breaks inheritable thread local tests :P we don't encourage them, but it is default.. have to think it through

brave/src/main/java/brave/propagation/CurrentTraceContext.java
```diff
@@ -175,7 +175,8 @@ public Scope maybeScope(@Nullable TraceContext currentSpan) {
    */
   public static final class Default extends ThreadLocalCurrentTraceContext {
     // Inheritable as Brave 3's ThreadLocalServerClientAndLocalSpanState was inheritable
-    static final InheritableThreadLocal<TraceContext> INHERITABLE = new InheritableThreadLocal<>();
+    static final InheritableThreadLocal<Object[]> INHERITABLE =
```
adriancole: one way to deal with this is to special-case an inheritable implementation. meaning if the thread local is inheritable, use normal non-wrapper approach

adriancole:
closing for now as it isn't going to work without reconsideration

PR #800, opened 2018-09-30
Kafka Streams instrumentation update

jeqo:
Changes:
- Added instrumentation for `ProcessorSupplier` and `TransformerSupplier`.
- Adding `KafkaClientSupplier` to a `KafkaStreams` will be exposed as a function by `KafkaStreamsTracing` instead of offering independently and expecting the user to use it. Same for `TracingProcessorSupplier` and `TracingTransformerSupplier`


instrumentation/kafka-streams/src/test/java/brave/kafka/streams/BaseTracingTest.java
```diff
@@ -0,0 +1,176 @@
...
+
+abstract class BaseTracingTest {
+    String TEST_APPLICATION_ID = "myAppId";
+    String TEST_TASK_ID = "0_0";
+    String TEST_TOPIC = "myTopic";
+    String TEST_KEY = "foo";
+    String TEST_VALUE = "bar";
+
+    Function<Headers, ProcessorContext> processorContextSupplier =
+            (Headers headers) ->
```
adriancole: maybe use mockito

instrumentation/kafka-streams/src/test/java/brave/kafka/streams/BaseTracingTest.java
```diff
@@ -0,0 +1,176 @@
...
+            .currentTraceContext(ThreadLocalCurrentTraceContext.newBuilder()
+                    .addScopeDecorator(StrictScopeDecorator.create())
+                    .build())
+            .spanReporter(spans::add)
+            .build();
+    KafkaStreamsTracing kafkaStreamsTracing = KafkaStreamsTracing.create(tracing);
+
+    ProcessorSupplier<String, String> fakeProcessorSupplier =
+            kafkaStreamsTracing.processorSupplier(
+                    "processor-1",
```
adriancole: maybe name "forward" to hint at decent name choices

instrumentation/kafka-streams/src/test/java/brave/kafka/streams/BaseTracingTest.java
```diff
@@ -0,0 +1,176 @@
...
+                    new Transformer<String, String, String>() {
+                        ProcessorContext context;
+                        @Override
+                        public void init(ProcessorContext context) {
+                            this.context = context;
+                        }
+
+                        @Override
+                        public String transform(String key, String value) {
+                            return "transformed";
```
adriancole: maybe return value? then it is more like a typical transform?

instrumentation/kafka-streams/src/test/java/brave/kafka/streams/KafkaStreamsTracingTest.java
```diff
@@ -1,28 +1,176 @@
...
+    Transformer processor = fakeTransformerSupplier.get();
+    processor.init(processorContextSupplier.apply(new RecordHeaders()));
+    processor.transform(TEST_KEY, TEST_VALUE);
+
+    assertThat(spans)
+            .flatExtracting(s -> s.tags().entrySet())
+            .containsOnly(
+                    entry("kafka.streams.application.id", TEST_APPLICATION_ID),
+                    entry("kafka.streams.task.id", TEST_TASK_ID),
+                    entry("kafka.streams.key", TEST_KEY));
```
adriancole: is key usually a good thing to know? not sure if it is or not just asking

instrumentation/kafka-streams/README.md
```diff
@@ -1,22 +1,59 @@
...
-                           .build();
+kafkaStreamsTracing = KafkaStreamsTracing.create(kafkaTracing);
 ```
 
-Next, create a streams component like this:
+To add a Tracing Processor to your application use the `TracingProcessorSupplier` provided by instrumentation:
 ```java
-kafkaStreamsTracing = KafkaStreamsTracing.create(kafkaTracing);
+builder.stream(inputTopic)
+       .processor(kafkaStreamsTracing.processorSupplier(
```
adriancole: sometimes I use same name as library. ex if they say "transform" I do too. ex in sparkjava we do this

instrumentation/kafka-streams/README.md
```diff
@@ -1,22 +1,59 @@
...
+            "processor-1",
+            new AbstractProcessor<String, String>() {
```
adriancole: in example maybe use a pre-canned processor if there is such a thing

instrumentation/kafka-streams/src/main/java/brave/kafka/streams/KafkaStreamsTags.java
```diff
@@ -0,0 +1,19 @@
...
+import org.apache.kafka.streams.processor.ProcessorContext;
+
+/**
+ * Tagging policy is not yet dynamic. The descriptions below reflect static policy.
+ */
+class KafkaStreamsTags {
+    /**
+     * Added on {@link KafkaStreamsTracing#nextSpan(ProcessorContext)} when the key not null or empty.
+     */
+    static final String KAFKA_STREAMS_KEY_TAG = "kafka.streams.key";
```
adriancole: nit I noticed files are in mixed format. please use square formatter

instrumentation/kafka-streams/src/test/java/brave/kafka/streams/KafkaStreamsTracingTest.java
jeqo: Could be if Processor/Transformer is used in a "non-traced" Topology, where `KafkaClientSupplier` is not used, and it just creates a individual `span` with the `key` available as a tag. I should test this scenario.

instrumentation/kafka-streams/README.md
jeqo: In this case we are using the same name as the parameter: e.g. `.processor(ProcessorSupplier)` -> `.processor(kafkaStreamsTracing.processorSupplier(...))` 
Do you think naming it the same as the operation will be more readable? e.g. `.processor(ProcessorSupplier)` -> `.processor(kafkaStreamsTracing.processor(...))`

instrumentation/kafka-streams/src/test/java/brave/kafka/streams/BaseTracingTest.java
jeqo: Agree

adriancole:
> In this case we are using the same name as the parameter: e.g. .processor(ProcessorSupplier) -> .processor(kafkaStreamsTracing.processorSupplier(...))
> Do you think naming it the same as the operation will be more readable? e.g. .processor(ProcessorSupplier) -> .processor(kafkaStreamsTracing.processor(...))

It is only a hunch really that if the result of a method only goes
into one other method, that making the method names match can be more
intuitive from a puzzle piece pov. So in the latter I think It
visually looks like it should be there eventhough it doesn't match the
types. my 2p

jeqo:
@adriancole After checking `sparkjava` API, I understood your point. Thanks for the pointer :)

instrumentation/kafka-streams/src/main/java/brave/kafka/streams/TracingProcessorSupplier.java
```diff
@@ -0,0 +1,63 @@
...
+  }
+
+  /** This wraps transform method to add tracing. */
+  @Override public Processor<K, V> get() {
+    return new Processor<K, V>() {
+      ProcessorContext processorContext;
+
+      @Override
+      public void init(ProcessorContext processorContext) {
+        this.processorContext = processorContext;
```
adriancole: add a comment that `init()` is done on the same thread as `supplier.get()` and which code in kafka does that..  assuming that's the case.

adriancole:
back to this one, about whether we should be tagging the key:

> Could be if Processor/Transformer is used in a "non-traced" Topology, where KafkaClientSupplier is not used, and it just creates a individual span with the key available as a tag. I should test this scenario.

I think what I meant to ask is if we should tag the key, especially as it might be a UUID or something which can add significant index load. We usually don't by default, but I don't know common practice. If we do by default, in the constant declaration, if for no other reason but memory, we should comment some practice by which adding the key is worth doing by default.. maybe referencing some kafka docs.

instrumentation/kafka-streams/src/main/java/brave/kafka/streams/KafkaStreamsTracing.java
```diff
@@ -1,23 +1,91 @@
...
-  public KafkaClientSupplier kafkaClientSupplier() {
+  KafkaClientSupplier kafkaClientSupplier() {
+    KafkaTracing kafkaTracing = KafkaTracing.create(tracing);
     return new TracingKafkaClientSupplier(kafkaTracing);
   }
+
+  Span nextSpan(ProcessorContext context) {
+    TraceContextOrSamplingFlags extracted = extractor.extract(context.headers());
+    Span result = tracing.tracer().nextSpan(extracted);
+    if (!result.isNoop()) {
```
adriancole: is there any case where this would skip the client instrumentation setup? Ex I would expect there to always be headers here. If there isn't then maybe someone skipped a step.

One way to tell is in your micro-integration tests, check that the root span has kind=CONSUMER

instrumentation/kafka-streams/src/main/java/brave/kafka/streams/TracingProcessorSupplier.java
jeqo: I've found that this does not happen:

ProcessorSupplier docs:
```
 * It is used in {@link Topology} for adding new processor operators, whose generated
 * topology can then be replicated (and thus creating one or more {@link Processor} instances)
 * and distributed to multiple stream threads.
```

Then, a new processor is created on potentially different threads as the `get` method. Should I then pass the tracer on the processor constructor, right? I'm planning to move the anonymous processor/transformer to a class `TracingProcessor/Transformer`:

On processorSupplier:
```
  @Override public Processor<K, V> get() {
    return new TracingProcessor<>(kafkaStreamsTracing, name, delegateProcessor);
  }
```
and `TracingProcessor`:
```
  public TracingProcessor(KafkaStreamsTracing kafkaStreamsTracing,
      String name, Processor<K, V> delegateProcessor) {
    this.kafkaStreamsTracing = kafkaStreamsTracing;
    this.tracer = kafkaStreamsTracing.tracing.tracer();
    this.name = name;
    this.delegateProcessor = delegateProcessor;
  }
```
Is this mechanism enough to trace correctly each processor instance?

instrumentation/kafka-streams/src/main/java/brave/kafka/streams/KafkaStreamsTracing.java
jeqo: Client instrumentation could be skipped it by creating a `KafkaStreams` directly without the builder from `KafkaStreamsTracing#kafkaStreams`, where clients are injected. In this case, processors will create independent spans, or potentially linked to a trace injected by a `TracingKafkaProducer` outside the Kafka Streams application. 

I have a test case for this scenario called: `should_create_spans_from_stream_without_tracing_with_tracing_transformer/processor`. I'll also add `kind=CONSUMER` validation on the input and `PRODUCER` on the output.

jeqo:
> I think what I meant to ask is if we should tag the key, especially as it might be a UUID or something which can add significant index load. We usually don't by default, but I don't know common practice. If we do by default, in the constant declaration, if for no other reason but memory, we should comment some practice by which adding the key is worth doing by default.. maybe referencing some kafka docs.

Agree. Kafka Producer/Consumer instrumentation already add tags for this, then I think we can remove it from Streams, and let users tag as needed inside `TracingProcessor/Transformer`. WDYT?

instrumentation/kafka-streams/src/main/java/brave/kafka/streams/KafkaStreamsTracing.java
adriancole: awesome sauce

adriancole:
> let users tag as needed inside TracingProcessor/Transformer. WDYT?

yep!

after that I'm good. wdyt @ImFlog?

ImFlog:
This looks very good ! GJ @jeqo.

I was wondering (maybe it could be another feature), if we could add spans annotation to Kafka streams.
Most of the time when using Kafka Streams, we use the high level DSL where we define our transformations before sending the result to a given topic. What would be interesting would be to add span annotation, either automatically (don't know if it's possible from a Kafka POV) or with a predefined transformer (not changing anything, just annotating).

Like this we get information about our different streaming steps without burden, and if we define custom processors, we can use what's in this PR.

WDYT @jeqo @adriancole

jeqo:
@ImFlog yes, I have been checking how to create spans from topology processors directly, here is the discussion on the Kafka mailing list: https://lists.apache.org/thread.html/f7a57528c056bb347454374fa231e6664925fdc70cfb812862e34b8e@%3Cdev.kafka.apache.org%3E 

The result was: to decorate every processor in a Topology we would have to propose changes to internal APIs. Changing this will require a KIP.
Another option was to create something like interceptors in Client APIs ("ProcessorInterceptor"), that will enable add some code before and after a processor runs - I like this idea better.

As result, I've added a `TracingProcessor` and `TracingTransformer` to wrap processors that you want to add to the trace. Even though we don't get the whole topology as a trace, we can add the meaningful steps via these processors.

adriancole:
with regards to tagging things the Kafka client instrumentation already does, I would yes yes.. let's opt out for now.

I think there is one last topic open here which was a response to @ImFlog but not yet verified by him. once this closes we can merge!

adriancole:
@coolbetm @dragontree101 @jpiechowka  this is for tracing kafka streaming. are any of you in a position to review this?

ImFlog:
Just to bounce back on my previous comment, what I was saying is that we could expose something like this : 
`kafkaStreamsTracing.transformerSupplier("transformer-1")`
Where implicitely it would only do spans operations and do not transform the inflight message (may have to be named differently)
Something like this would be valid:
```
builder.stream(inputTopic)
     .map(...)
     .transform(kafkaStreamsTracing.processorSupplier("map-1"))
     .flatMap(...)
     .transform(kafkaStreamsTracing.processorSupplier("flatmap-1"))
     .to();
```
I think that people use mainly the high level API and that if they want simple tracing, people could just add some markers in their current topology.
That's why I talked about using annotation first because if someone just want to add some simple markers like this, a simple hint on the zipkin-UI with the annotation may be enough.

I hope i'm clear enough but for me everything looks good, and all of the above comment could end up in a new task.
Unfortunately I will not have enough time to compile this and try it on some sample code, but I saw @jeqo sandbox and I'm pretty sure all is good :)

adriancole:
you mean something like

     .transform(kafkaStreamsTracing.annotateStartAndFinish("my-map"))

which does span.annotate("my-map.start") and
span.annotate("my-map.finish") around the task?

jeqo:
@ImFlog thanks for the details! 

I understand that these transformers will mark the end (or the beginning) of a task, right? Then you will have a span from(topic).....to(topic)

```
+------------------stream-processor---------------------------------------+
 +---+
   poll
                  +                                             +
                map                                         flatmap
                                                                        +------+
                                                                           send
```

Seems like a good idea to have a custom annotation transformer. I could give a try.

jeqo:
@ImFlog I have played a bit with the current API and I found one limitation to create this annotations: trace  is created by 2 related spans: One created by the consumer on the `build.stream(...)` task, and another at the end, in the `.to(...)`. As there is not active span in between, I have nothing to annotate. 

I'd need to create an "emptySpan" in the middle to without latency and use it as an annotation. I'm not sure about this approach. Wdyt?

ImFlog:
i forgot about the fact that there is no span between the consumer and the producer. I guess it is indeed strange to create empty spans as marker.
Do you have a screenshot of what it looks like in the UI ? To see if it worth doing this.

ImFlog:
Yeah maybe not worth doing this for now, this is kinda misleading on what this span represent (start ? end ?)
Let's wait for people to instrument their first streams and see if this kind of feature arise.
There is still ongoing discussion for Kafka API changes also.

jeqo:
@ImFlog @adriancole just checking if we agree that this PR is complete to be prepared for next release :)

instrumentation/kafka-streams/src/main/java/brave/kafka/streams/KafkaStreamsTracing.java
```diff
@@ -1,23 +1,105 @@
...
+   */
+  public <K, V> ProcessorSupplier<K, V> processor(String name, Processor<K, V> processor) {
+    return new TracingProcessorSupplier<>(this, name, processor);
   }
 
-  final KafkaTracing kafkaTracing;
+  /**
+   * Create a tracing-decorated {@link TransformerSupplier}
+   */
+  public <K, V, R> TransformerSupplier<K, V, R> transformer(String name,
```
adriancole: will help to add javadoc code example, nicely formatted showing example usage.

While not relevant, here is the style I'm hoping for on these public methods.

```java
  /**
   * Use this to temporarily override the sampler used when starting new traces. This also serves
   * advanced scenarios, such as {@link DeclarativeSampler declarative sampling}.
   *
   * <p>Simple example:
   * <pre>{@code
   * // Ensures new traces are always started
   * Tracer tracer = tracing.tracer().withSampler(Sampler.ALWAYS_SAMPLE);
   * }</pre>
   *
   * @see DeclarativeSampler
   */
```

instrumentation/kafka-streams/src/main/java/brave/kafka/streams/KafkaStreamsTracing.java
```diff
@@ -1,23 +1,105 @@
...
+  KafkaStreamsTracing(Builder builder) { // intentionally hidden constructor
+    this.tracing = builder.tracing;
+    this.extractor = tracing.propagation().extractor(KafkaStreamsPropagation.GETTER);
+  }
+
+  public static KafkaStreamsTracing create(Tracing tracing) {
+    return new KafkaStreamsTracing.Builder(tracing).build();
+  }
+
+  static void addTags(ProcessorContext processorContext, SpanCustomizer result) {
```
adriancole: please move this down with the other package private

instrumentation/kafka-streams/src/main/java/brave/kafka/streams/KafkaStreamsTracing.java
```diff
@@ -1,23 +1,105 @@
...
-  KafkaStreamsTracing(KafkaTracing kafkaTracing) { // intentionally hidden constructor
-    this.kafkaTracing = kafkaTracing;
+  public <K, V, VR> ValueTransformerWithKeySupplier<K, V, VR> valueTransformerWithKey(String name,
+      ValueTransformerWithKey<K, V, VR> valueTransformerWithKey) {
+    return new TracingValueTransformerWithKeySupplier<>(this, name, valueTransformerWithKey);
   }
 
   /** Returns a client supplier which traces send and receive operations. */
-  public KafkaClientSupplier kafkaClientSupplier() {
+  KafkaClientSupplier kafkaClientSupplier() {
```
adriancole: intentionally package private? mention why we need an accessor like this if package private

instrumentation/kafka-streams/src/main/java/brave/kafka/streams/TracingProcessor.java
```diff
@@ -0,0 +1,53 @@
+package brave.kafka.streams;
+
+import brave.Span;
+import brave.Tracer;
+import org.apache.kafka.streams.processor.Processor;
+import org.apache.kafka.streams.processor.ProcessorContext;
+
+public class TracingProcessor<K, V> implements Processor<K, V> {
```
adriancole: does this type need to be public?

instrumentation/kafka-streams/src/main/java/brave/kafka/streams/TracingTransformer.java
```diff
@@ -0,0 +1,55 @@
+package brave.kafka.streams;
+
+import brave.Span;
+import brave.Tracer;
+import org.apache.kafka.streams.kstream.Transformer;
+import org.apache.kafka.streams.processor.ProcessorContext;
+
+public class TracingTransformer<K, V, R> implements Transformer<K, V, R> {
```
adriancole: does this type need to be public

instrumentation/kafka-streams/src/main/java/brave/kafka/streams/TracingValueTransformer.java
```diff
@@ -0,0 +1,55 @@
+package brave.kafka.streams;
+
+import brave.Span;
+import brave.Tracer;
+import org.apache.kafka.streams.kstream.ValueTransformer;
+import org.apache.kafka.streams.processor.ProcessorContext;
+
+public class TracingValueTransformer<V, VR> implements ValueTransformer<V, VR> {
```
adriancole: does this type need to be public

instrumentation/kafka-streams/src/main/java/brave/kafka/streams/TracingValueTransformerWithKey.java
```diff
@@ -0,0 +1,55 @@
+package brave.kafka.streams;
+
+import brave.Span;
+import brave.Tracer;
+import org.apache.kafka.streams.kstream.ValueTransformerWithKey;
+import org.apache.kafka.streams.processor.ProcessorContext;
+
+public class TracingValueTransformerWithKey<K, V, VR> implements ValueTransformerWithKey<K, V, VR> {
```
adriancole: does this type need to be public

adriancole:
@llofberg how does this feel to you? Are you happy with it with no changes?

instrumentation/kafka-streams/src/main/java/brave/kafka/streams/KafkaStreamsTracing.java
jeqo: ClientSupplier is only used by `kafkaStreams()` method. I will move this logic to that method.

instrumentation/kafka-streams/src/main/java/brave/kafka/streams/TracingProcessor.java
jeqo: no. fixing it.

instrumentation/kafka-streams/src/main/java/brave/kafka/streams/TracingValueTransformerWithKey.java
jeqo: no. fixing it.

instrumentation/kafka-streams/src/main/java/brave/kafka/streams/TracingValueTransformer.java
jeqo: no. fixing it.

instrumentation/kafka-streams/src/main/java/brave/kafka/streams/TracingTransformer.java
jeqo: no. fixing it.

jeqo:
Seems working :) 
image

instrumentation/kafka-streams/src/main/java/brave/kafka/streams/KafkaStreamsTracing.java
jeqo: @adriancole I've added comments with samples for all public API methods, wdyt?

jeqo:
Awesome! Just pushed latest changes. Looking forward to the next release!! 🎉

adriancole:
5.5.0 on the way

PR #801, opened 2018-10-01
Adds concept of "local root" used to partition spans by entry point

adriancole:
Subgraphs are often "squashed" when processing dependency links.
Usually, we have to skip data server-side to achieve this, for example,
"skipping up" the tree until we find the root-most span in that process.

Moreover, several conversations led to a desire for "skeletal spans"
which contain no intermediate info between inbound and outbound
requests. This allows for 100% sampling in edge cases such as surges or
very large amounts of traffic.

Finally, some APM systems require reporting that groups together entry
points for reasons of squashing or post-processing a trace. For example,
Amazon X-Ray have a type Segment which is only for exit spans, reserving
SubSegment for local ones. Having some means to partition data allows
post-processing such as this, for example bundling.

"local root" is the solution to this problem and similar. By adding a
property: `localRootId` to the trace context, we can track spans by
entry point. This means we can re-write parents to squash intermediates.
We can also expose this in logging contexts to accomodate correlation.

### DIFF
--- a/brave/src/main/java/brave/propagation/TraceContext.java
+++ b/brave/src/main/java/brave/propagation/TraceContext.java
@@ -1,5 +1,6 @@
 package brave.propagation;
 
+import brave.Span;
 import brave.internal.InternalPropagation;
 import brave.internal.Nullable;
 import brave.internal.Platform;
@@ -9,6 +10,7 @@ import java.util.List;
 
 import static brave.internal.HexCodec.lenientLowerHexToUnsignedLong;
 import static brave.internal.HexCodec.writeHexLong;
+import static brave.internal.InternalPropagation.FLAG_LOCAL_ROOT;
 import static brave.internal.InternalPropagation.FLAG_SAMPLED;
 import static brave.internal.InternalPropagation.FLAG_SAMPLED_LOCAL;
 import static brave.internal.InternalPropagation.FLAG_SAMPLED_SET;
@@ -77,6 +79,32 @@ public final class TraceContext extends SamplingFlags {
   /** Unique 8-byte identifier for a trace, set on all spans within it. */
   public long traceId() {
     return traceId;
+  }
+
+  /**
+   * Returns the first {@link #spanId()} in a partition of a trace: otherwise known as an entry
+   * span. This could be a root span or a span representing incoming work (ex {@link
+   * Span.Kind#SERVER} or {@link Span.Kind#CONSUMER}. Unlike {@link #parentIdAsLong()}, this value
+   * is inherited to child contexts until the trace exits the process. This value is inherited for
+   * all child spans until the trace exits the process. This could also be described as an entry
+   * span.
+   *
+   * <p>When {@link #isLocalRoot()}, this ID will be the same as the {@link #spanId() span ID}.
+   *
+   * <p>The local root ID can be used for dependency link processing, skipping data or partitioning
+   * purposes. For example, one processor could skip all intermediate (local) spans between an
+   * incoming service call and any outgoing ones.
+   *
+   * <p>This does not group together multiple points of entry in the same trace. For example,
+   * repetitive consumption of the same incoming message results in different local roots.
+   */
+  // This is the first span ID that became a Span or ScopedSpan
+  public long localRootId() {
+    return localRootId;
+  }
+
+  public boolean isLocalRoot() {
+    return (flags & FLAG_LOCAL_ROOT) == FLAG_LOCAL_ROOT;
   }
 
   /**
@@ -181,12 +209,14 @@ public final class TraceContext extends SamplingFlags {
 
   public static final class Builder {
     long traceIdHigh, traceId, parentId, spanId;
+    long localRootId; // intentionally only mutable by the copy constructor in order to control usage.
     int flags;
     List<Object> extra = Collections.emptyList();
 
     Builder(TraceContext context) { // no external implementations
       traceIdHigh = context.traceIdHigh;
       traceId = context.traceId;
+      localRootId = context.localRootId;
       parentId = context.parentId;
       spanId = context.spanId;
       flags = context.flags;
@@ -376,7 +406,7 @@ public final class TraceContext extends SamplingFlags {
       if (spanId == 0L) missing += " spanId";
       if (!"".equals(missing)) throw new IllegalStateException("Missing: " + missing);
       return new TraceContext(
-          flags, traceIdHigh, traceId, parentId, spanId, ensureImmutable(extra)
+          flags, traceIdHigh, traceId, localRootId, parentId, spanId, ensureImmutable(extra)
       );
     }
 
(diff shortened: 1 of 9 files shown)


adriancole:
cc'ing folks who may have related works @ivantopo  @wu-sheng @felixbarny @tylerbenson @narayaruna @drolando @cwensel

also @williewheeler who is likely to have some near real time graph aggregations in a bit. This sort of handling can send connectable spans to such a pipeline https://github.com/ExpediaDotCom/haystack-adaptive-alerting

brave/src/main/java/brave/propagation/TraceContext.java
```diff
@@ -79,6 +81,32 @@ public long traceId() {
     return traceId;
   }
 
+  /**
+   * Returns the first {@link #spanId()} in a partition of a trace: otherwise known as an entry
```
adriancole: here's the important docs

felixbarny:
FWIW, in Elastic APM, we have a dedicated domain object for entry spans - they are called `Transaction`s. In the Java agent, the TraceContext also has a [field for the `transactionId`](https://github.com/elastic/apm-agent-java/blob/d8b583f30c6b406ba8f81e63f0ce71af22dc2469/apm-agent-core/src/main/java/co/elastic/apm/impl/transaction/TraceContext.java#L57) so that each span knows which transaction it belongs to.

adriancole:
> the TraceContext also has a field for the transactionId
> <https://github.com/elastic/apm-agent-java/blob/d8b583f30c6b406ba8f81e63f0ce71af22dc2469/apm-agent-core/src/main/java/co/elastic/apm/impl/transaction/TraceContext.java#L57>
> so that each span knows which transaction it belongs to.
>
interesting.. in what case would this be different than if you re-used the
span ID at the entry point of the transaction as the transaction id?

>

felixbarny:
Not sure I understand your question.

Entry spans are transactions in our data model. `Transaction extends AbstractSpan` and `Span extends AbstractSpan`. I.e. Transactions are a special kind of span which represents an entry in a service.

adriancole:
right sorry. I mean in the link you pasted, the transactionId is generated
at initialization and there is also a separate id field. was just curious
if there would have been impact if the same id value was shared.

felixbarny:
> I mean in the link you pasted, the transactionId is generated at initialization 

It's only initialized with zeros in the field declaration. The reason is that the `TraceContext` object is `Recyclable`. The actual ID is generated in `asRootSpan` method and copied from the parent id in the `asChildOf` or method.

adriancole:
ah ok now I understand! thanks felix

felixbarny:
Just as a suggestion, an alternative naming for `localRootId` could be `entrySpanId`. This is just based on the fact that I read of entry spans before but never about local roots. There is no standard or de-facto-standard name for that concept, but if you also like entry span it would make sense to align. That would kind of make this term a de-facto standard. Choose whatever you feel fits best but be aware that this might have a big impact in the tracing community/terminology in general which already has quite a lot of jargon.

adriancole:
one other way could be to remove the local root span condition from this
property (ex by redoing the code to see if heuristically we can accomplish
the same by looking for no parent id).

In that case entrySpanId would fit perfectly.

adriancole:
ps it wont work (constraining towards the term entry span by eliminating
root). we need the partition id to always be present and inherited. root is
a frequent case and you have no other way of knowing the root span ID as
order isnt guaranteed and trace id is not guaranteed to be a function of
root span id...

so we are back to conflating terminology eventhough it doesnt match or
choosing something else or picking something different.

basvanbeek:
Tricky one. I would only prefer `entrySpanId` if it indeed would exclude locally originated root spans. However I'm not sure we want to exclude those from the logic.

I have no objection to localRoot which in case of locally originated root spans would mean localRoot == root span. Because they happen to be both.

felixbarny:
What are some of the use cases you have encountered where the local root is not an entry span? Does it even make sense to track those? I guess it probably depends...

adriancole:
>
> What are some of the use cases you have encountered where the local root
> is not an entry span? Does it even make sense to track those? I guess it
> probably depends...
>
scheduled tasks start traces for example and so aren't entry spans.

>

adriancole:
another is lambda invocations. you can't always tell what triggered the
function eventhough you can read an env variable if there is an Amazon
trace in progress.

felixbarny:
Another one which comes to my mind is JDBC calls during container startup, although you'd probably not want to monitor these unless you create a span which wraps the whole startup.

> scheduled tasks start traces for example and so aren't entry spans.

Depending on how you define it, these could still be considered entry spans, even though there is no network interaction. You could argue that the scheduling of the task is the entry.

felixbarny:
Thanks for the history lesson :)
Seems like `localRoot` fits right in the Zipkin terminology. But it also seems to have slightly different semantics than the Transaction concept we have in Elastic APM. We also have the concept of what you called skeleton traces (I like that name), as we do report non-sampled transactions (metadata like tags are removed on those).

drolando:
In my opinion both `localRoot` and `entrySpanId` are a bit ambiguous and could be confusing for users. Between the 2 I'd choose `localRoot` as it seems to fit better with brave terminology.

adriancole:
PS I have been trying to wrap brain around entrySpanId, to figure a
way to have it not be confusing. We'd have to define it like this:

The entry span is the first span in a branch of a trace that is
visible to an instance of a tracer. IOTW, it is an entrypoint into a
trace, possibly its initial entry.

It could be a root span, that originated locally like:
* an android client application creating a root span in response to a gesture
* a watcher that tripped due to a file change, invoking a workflow
* a CLI application like git pull
* a scheduled task, either process scoped or internal to one
* a lambda function invoked from an unknown source

More typically, if could be a branch in an existing span, like:
* a server side RPC invocation (even if it reuses the same ID)
* a message consumption event (regardless of bulk or not, or if reprocessing)

If we are ok with defining entry span to describe all of the above,
I'm ok with calling it that. I still don't like the word entry, but
anyway, let's all sleep on it?

Notes about how we could possibly be tentative..

One possible way out is to hide the field completely on the trace
context and make it a special function to access it for its only call
site (a finished span handler).

Then it won't be a public method, therefore not even visible to users,
so can't confuse them. As I mentioned before, users don't have to know
about this at all unless it is added to a logging context. Ex
InternalPropagation.localRootId(context) would access the field.

Only issue is that it is a bit weird and I don't want to encourage
people to use internal methods, even non-users aka power users.
Another way is to make this still named localRootId, but make it a
function of the FinishedSpanHandler (ex
FinishedSpanHandler.localRootId(context) ). I'm not sure though if it
would be needed later in internal propagation logic... and I'd hate to
move it twice. We've never done that with any field before.

So, maybe what's best is to let folks sleep on it a few days. I'm
taking the weekend off anyway whah hah hah.

adriancole:
ps added this thread in case others outside our ecosystem have anything to add. It will be input, not a democratic vote though :) as I don't think everyone in the world are brave developers (sniff sniff), so can't necessarily comment on how it fits here.

https://groups.google.com/forum/#!topic/distributed-tracing/HWOD3zdWD3s

wu-sheng:
@adriancole I am on vacation now, so I will catch you up after the days off.

PR #805, opened 2018-10-15
Migrates off rxjava2 Fuseable base types as they broke inheritance

adriancole:
This inlines code we needed from Fuseable types in order to allow the
same library to work with RxJava 2.2 and 2.3+

See https://github.com/ReactiveX/RxJava/pull/6129/files#diff-173446df25e9a6926e322f1849bf6c35R31

### DIFF
--- a/context/rxjava2/pom.xml
+++ b/context/rxjava2/pom.xml
@@ -21,7 +21,7 @@
     <dependency>
       <groupId>io.reactivex.rxjava2</groupId>
       <artifactId>rxjava</artifactId>
-      <version>2.2.0</version>
+      <version>2.2.2</version>
       <scope>provided</scope>
     </dependency>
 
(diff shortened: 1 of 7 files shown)


adriancole:
PS This was our fault, not RxJava's. We used internal types without realizing it as I guess we focused on the type names not the package. Ideally one day we can have a errorprone checker for accidental use of types in packages named internal or internals

cc @anuraaga as well

anuraaga:
If I recall correctly, armeria's javadoc build fails when internal classes get referenced on a public one. I believe a combination of enabling all doclint and excluding internal packages can help?

https://github.com/line/armeria/blob/d4f13e4c4dc90826d001bfb73d6ca05dce19689c/build.gradle#L120

/cc @trustin

trustin:
And this: https://github.com/line/armeria/blob/d4f13e4c4dc90826d001bfb73d6ca05dce19689c/site/build.gradle#L71-L136

adriancole:
hi all https://github.com/openzipkin/brave/pull/809 tried to insulate further

PR #807, opened 2018-10-17
Fixes length bug in b3 single format

adriancole:
Thanks to @zyfjeff for finding this!

See https://github.com/envoyproxy/envoy/pull/4712#issuecomment-430634088


adriancole:
cc @drolando @sfackler @fedj @dpsoft just in case you copy/pasted my mistake by looking at brave's code!

adriancole:
Impact of this bug: Those who switched Tracing.propagationFactory to B3 single *and also* Tracing.traceId128 would have failed badly. Also, any receipt of a 128-bit "b3" header which also propagated the parent ID would have restarted a trace (this could happen in http, but not JMS as for example we don't propagate parent ID in JMS).

PR #809, opened 2018-10-19
Insulates RxJava2 code from future drift in internal packages

adriancole:
The "fuseable" package came out in RxJava 2.0 and is as yet still
internal. This puts insulation around it in case it ever becomes not
internal.

Meanwhile there is still risk that some parts of these internal code
drift, for example renaming a method inside `ScalarCallable` or
`ConditionalSubscriber`. If this happens, the code will still break.

See #805

### DIFF
--- a/context/rxjava2/src/main/java/brave/context/rxjava2/internal/fuseable/TraceContextSubscriber.java
+++ b/context/rxjava2/src/main/java/brave/context/rxjava2/internal/fuseable/TraceContextSubscriber.java
@@ -1,9 +1,9 @@
-package brave.context.rxjava2;
+package brave.context.rxjava2.internal.fuseable;
 
+import brave.context.rxjava2.internal.Util;
 import brave.propagation.CurrentTraceContext;
 import brave.propagation.CurrentTraceContext.Scope;
 import brave.propagation.TraceContext;
-import io.reactivex.internal.subscriptions.SubscriptionHelper;
 import io.reactivex.plugins.RxJavaPlugins;
 import org.reactivestreams.Subscriber;
 import org.reactivestreams.Subscription;
@@ -26,13 +26,12 @@ final class TraceContextSubscriber<T> implements Subscriber<T> {
   }
 
   @Override public final void onSubscribe(Subscription s) {
-    if (SubscriptionHelper.validate(upstream, s)) {
+    if (Util.validate(upstream, s)) {
       downstream.onSubscribe((upstream = s));
     }
   }
 
-  @Override
-  public void onNext(T t) {
+  @Override public void onNext(T t) {
     Scope scope = currentTraceContext.maybeScope(assemblyContext);
     try { // retrolambda can't resolve this try/finally
       downstream.onNext(t);
@@ -41,8 +40,7 @@ final class TraceContextSubscriber<T> implements Subscriber<T> {
     }
   }
 
-  @Override
-  public void onError(Throwable t) {
+  @Override public void onError(Throwable t) {
     if (done) {
       RxJavaPlugins.onError(t);
       return;
@@ -57,8 +55,7 @@ final class TraceContextSubscriber<T> implements Subscriber<T> {
     }
   }
 
-  @Override
-  public void onComplete() {
+  @Override public void onComplete() {
     if (done) return;
     done = true;
 
--- a/context/rxjava2/src/test/java/brave/context/rxjava2/CurrentTraceContextAssemblyTrackingTest.java
+++ b/context/rxjava2/src/test/java/brave/context/rxjava2/CurrentTraceContextAssemblyTrackingTest.java
@@ -18,6 +18,8 @@ import io.reactivex.Single;
 import io.reactivex.SingleObserver;
 import io.reactivex.flowables.ConnectableFlowable;
 import io.reactivex.functions.Predicate;
+import io.reactivex.internal.fuseable.ConditionalSubscriber;
+import io.reactivex.internal.fuseable.QueueFuseable;
 import io.reactivex.internal.fuseable.ScalarCallable;
 import io.reactivex.internal.operators.completable.CompletableEmpty;
 import io.reactivex.internal.operators.completable.CompletableFromCallable;
@@ -44,6 +46,7 @@ import org.junit.Assert;
 import org.junit.Before;
 import org.junit.Test;
 import org.reactivestreams.Subscriber;
+import org.reactivestreams.Subscription;
 
 import static org.assertj.core.api.Assertions.assertThat;
 import static org.assertj.core.api.Assertions.failBecauseExceptionWasNotThrown;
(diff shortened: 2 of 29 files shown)


akarnokd:
You have a few options: opt-out from fusion support in brave or make a commitment to keep updating code that depends on internals. I don't know your release policy nor if your users tend to update RxJava much frequently than brave.

adriancole:
@akarnokd thanks for the insight. Yeah I am comfortable with keeping up with updates (on internals). This is in our core repo anyway, and we have a tendency to.  For example, we still support gRPC 1.3 (via maven invoker tests) all the way to latest. if/when fuseable becomes non-internal, it will break our build and we'll deal. The instructions to users would be to update to latest brave which would sort it (and back-insulating the internal variant with an invoker test).

cc also @koji @trustin @anuraaga sound good?

adriancole:
I added hardening in our tests to show if our conditional.. conditionalsubscriber support is actually working. This shows I don't know how to idiomatically get fusion to happen :P

I will re-read this and see if I can figure out a non-hacky way to ensure `tryOnNext` is called in my subject

adriancole:
ok made the test less awkward after reading

context/rxjava2/src/test/java/brave/context/rxjava2/CurrentTraceContextAssemblyTrackingTest.java
```diff
@@ -294,6 +295,48 @@ public void parallelFlowable_subscribesUnderScope() {
     subscribesAndRunsInScope2(source, errorSource).assertResult(1, 2, 3);
   }
 
+  /**
+   * This is an example of "micro fusion" where use use a source that supports fusion: {@link
```
adriancole: @akarnokd do you mind reviewing this part? I think this relatively good exercise of the fuseable things I meant to exercise. I really appreciate your blogs. If I got anything conceptually wrong here.. please comment!

context/rxjava2/src/main/java/brave/context/rxjava2/TraceContextObserver.java
```diff
@@ -25,7 +25,7 @@
   }
 
   @Override public final void onSubscribe(Disposable d) {
-    if (DisposableHelper.validate(upstream, d)) {
+    if (Util.validate(upstream, d)) {
       downstream.onSubscribe((upstream = d));
```
akarnokd: The upstream disposable in an Observer should never just passed along as it may cause problems with fusion.

context/rxjava2/src/main/java/brave/context/rxjava2/internal/fuseable/TraceContextSubscriber.java
```diff
@@ -26,13 +26,12 @@
   }
 
   @Override public final void onSubscribe(Subscription s) {
-    if (SubscriptionHelper.validate(upstream, s)) {
+    if (Util.validate(upstream, s)) {
       downstream.onSubscribe((upstream = s));
```
akarnokd: The upstream subscription should not just be passed along.

context/rxjava2/src/test/java/brave/context/rxjava2/CurrentTraceContextAssemblyTrackingTest.java
akarnokd: This is a conditional micro fusion case btw, but the test looks okay.

context/rxjava2/src/main/java/brave/context/rxjava2/internal/fuseable/TraceContextSubscriber.java
akarnokd: Also `TraceContextSubscriber` should implement `FlowableSubscriber` instead, otherwise RxJava will add extra boundaries around it and fusion will not work.

akarnokd:
Btw, [assembly tracking is impemented this way](https://github.com/akarnokd/RxJava2Extensions/tree/master/src/main/java/hu/akarnokd/rxjava2/debug).

adriancole:
thank you for all the feedback. I will try to sort the other problems you
noticed tomorrow. kind of wandering my way through this so appreciate your
flashlight!

adriancole:
I will open a new PR for next fixes

adriancole:
this renames fields so it is easier to compare behavior https://github.com/openzipkin/brave/pull/812

context/rxjava2/src/main/java/brave/context/rxjava2/TraceContextObserver.java
adriancole: can you point me to a line of doing this right? I'm concerned about inheriting a tree of internal classes, but it would still be useful to know what correct is. Also, if you have any insight on how to prove this will be wrong, that would help, too.

context/rxjava2/src/main/java/brave/context/rxjava2/internal/fuseable/TraceContextSubscriber.java
adriancole: > Also TraceContextSubscriber should implement FlowableSubscriber instead, otherwise RxJava will add extra boundaries around it and fusion will not work.

I assume I should only do this when downstream is also a FlowableSubscriber correct?

context/rxjava2/src/main/java/brave/context/rxjava2/TraceContextObserver.java
akarnokd: https://github.com/ReactiveX/RxJava/blob/2.x/src/main/java/io/reactivex/internal/observers/BasicFuseableObserver.java#L66

Try instrumenting something like this: `Observable.range(1, 10).concatMap(Observable::just)`. If I understand brave correctly, the `TraceContextObserver` is sandwitched between `range` and `concatMap`, `concatMap` will receive the `QueueDisposable` of `range` and poll it directly, thus `TraceContextObserver` will never get its `onXXX` methods called.

context/rxjava2/src/main/java/brave/context/rxjava2/internal/fuseable/TraceContextSubscriber.java
akarnokd: You should never encounter a plain `Subscriber` inside operators.

context/rxjava2/src/main/java/brave/context/rxjava2/TraceContextObserver.java
adriancole: thanks for your patience.. I'm moving this stuff to a new PR where I can hopefully address it. I will read into this and try to understand https://github.com/openzipkin/brave/pull/814

context/rxjava2/src/main/java/brave/context/rxjava2/internal/fuseable/TraceContextSubscriber.java
adriancole: I guess I just don't quite understand as there seem to be dozens of types that implement Subscriber and not FlowableSubscriber

context/rxjava2/src/main/java/brave/context/rxjava2/TraceContextObserver.java
adriancole: ps I think you are right.. I will address over there

context/rxjava2/src/main/java/brave/context/rxjava2/internal/fuseable/TraceContextSubscriber.java
akarnokd: > there seem to be dozens of types

Which ones?

adriancole:
in your extensions repo.. 3rd party

adriancole:
that and others, yes. one problem with instrumentation is that it is hard
to control things like this. if it is possible to have a Subscriber that
isnt a FlowableSubscriber we likely will need to instrument it even if
eventually there are no more of the former. That is because often people
have indirect dependencies they may not even know about.

PR #811, opened 2018-10-20
Ensures ScalarCallable.call() isn't instrumented

adriancole:
After reading more about scalar callable, it seems to explicitly refer
to constants. Since this is the case, even optimized instrumentation is
unnecessary overhead as returning a pre-computed value will never result
in a new span or log entry.

See http://akarnokd.blogspot.com/2016/04/operator-fusion-part-2-final.html


adriancole:
@akarnokd @koji @anuraaga @trustin am I correct in interpreting ScalarCallable.call() as a constant function? Similar to the work on fusion itself, we really want to avoid amplifying overhead, even small overhead.

adriancole:
both javadocs and wiki work for me to imply wrapping around call (in a
ScalableCallable) is unnecessary as call occurs at runtime, and the value
happens at assembly time (prior to that). Thanks for the tip!

call() be constant or simple computation that can be legally executed
during assembly time.

PR #814, opened 2018-10-21
Cleans up RxJava handling removing temporarily any fusion functionality

adriancole:
This addresses the following
> downstream.onSubscribe(this) is mandatory, even if you don't plan to support fusion. Without this, operators trying to detect the fuseable feature of their immediate upstream would interface with the wrong operator.

This also removes support for fusion until we can better handle it.

Finally, this backfills a bunch of tests

Original scope was more and descoped due to complexity of fusion mixed with the complexity of working with internal types that implement it. We can revisit later.. original scope below

----

This PR will eventually address feedback on #809 by @akarnokd including:
* The upstream subscription and disposable should not just be passed along
* Lack of FlowableSubscriber instrumentation which causes problems with fusion

I was referred to https://github.com/akarnokd/RxJava2Extensions/tree/master/src/main/java/hu/akarnokd/rxjava2/debug

For example, `OnAssemblyObserver` extends `BasicFuseableObserver`. If we
go this route, we will become even more sensitive to internal types as
`BasicFuseableObserver` is itself internal. Currently, we only have
exposure to two internal method signatures.

For example, if we go this route we end up exposed also to
`QueueDisposable` even for types that are not fuseable in nature.

What I'd like to know is how we can instrument an Observable which does
not include "fuseable" support directly. In other words, what are the
"right ways"? and do any of the right ways avoid internal types?

Finally, how could we evaluate if the right ways are in place or not?
At the moment, we have no failing tests, and that's not to say our
tests are good, just we don't know how to assess a problem in fusion vs
a problem in non-fusion, else our tests would have broken.


adriancole:
cc @rmichela @koji .. this is an interesting thing for me.. hopefully learning will happen in my brain, if it already happened in yours feel free to give pointers as well

adriancole:
question: we don't currently instrument: `org.reactivestreams.Processor` what would be a good test for this?

adriancole:
feedback from the other PR from @akarnokd:

> You should never encounter a plain `Subscriber` inside operators.

I'm trying to grok this.. obviously there are a lot of types that implement `Subscriber` and not `FlowableSubscriber`.  For example, I see many in my classpath from `hu.akarnokd.rxjava2.basetypes`. So I suppose I would like to know if it is an antipattern to instrument `Subscriber`, or if the main goal is to also instrument `FuseableSubscriber` (which this PR adds support for?

akarnokd:
Any non-`FlowableSubscriber` is wrapped into a `StrictSubscriber` to ensure the more strict Reactive Streams requirements of `Subscriber`: https://github.com/ReactiveX/RxJava/blob/2.x/src/main/java/io/reactivex/Flowable.java#L14424

`StrictSubscriber` can't support fusion so it will force the operators to work in the traditional way. This can be a problem when you try to instrument sequences that now work on non-fused paths only whereas without instrumentation, you'd gett the fused paths.

`subscribeActual` always receives a `FlowableSubscriber` so there is no need to check for this interface.

akarnokd:
Also `downstream.onSubscribe(this)` is mandatory, even if you don't plan to support fusion. Without this, operators trying to detect the fuseable feature of their immediate upstream would interface with the wrong operator.

adriancole:
I added a test based on a good guess from @akarnokd  
https://github.com/openzipkin/brave/pull/809#discussion_r226855487

So If we create an observable in a specific scope, and later apply a fusion operator to it (ex concatMap), it doesn't push the initial assembly's trace context. This could be bad.. I think.

adriancole:
the XMap thing will require some thinking/code redo as eventhough assembly hooks are called, XMap is not a function of "this" rather a result extracted from it. This implies we have to take a different approach to propagate the input:

```java
    public final <R> Observable<R> switchMapDelayError(Function<? super T, ? extends ObservableSource<? extends R>> mapper, int bufferSize) {
        ObjectHelper.requireNonNull(mapper, "mapper is null");
        ObjectHelper.verifyPositive(bufferSize, "bufferSize");
        if (this instanceof ScalarCallable) {
            @SuppressWarnings("unchecked")
            T v = ((ScalarCallable<T>)this).call(); // << the assembly context is stashed in the current observable!            
            if (v == null) {
                return empty();
            }
            return ObservableScalarXMap.scalarXMap(v, mapper); // << v is derived so lost the assembly context
        }
        return RxJavaPlugins.onAssembly(new ObservableSwitchMap<T, R>(this, mapper, bufferSize, true));
    }
```

adriancole:
actually I take it back.. XMap is by definition happening in the assembly context anyway, so the inner call to `RxJavaPlugins.onAssembly(new ScalarXMap...` can catch it. I'll add a test to prove this.

adriancole:
added a commit. the scalar.Xmap problem is real, but only exists when the assembly is split across multiple scopes. It is so hard to fix I might just punt it to docs unless someone else can think a way around it.

adriancole:
OK I think I've solved the instrumentation gap during fusion. There are still other problems, but I'd like to sanity check the last commit through @akarnokd

adriancole:
I'm not sure why github is not seeing the latest commit on this branch, but it is here https://github.com/openzipkin/brave/commit/ca7ec9837875d7d5aec8f36720e5f1410608b5fe

@akarnokd I *believe* this solves the instrumentation gap of micro fusion between the time ScalarCallable.call is invoked and the next onAssembly hook, but not extending inside subscription. I think this is correct as micro fusion is as far as I know something that is assembly in nature.

adriancole:
I'm not sure why github is not seeing the latest commit on this branch, but it is here https://github.com/openzipkin/brave/commit/ca7ec9837875d7d5aec8f36720e5f1410608b5fe

@akarnokd I *believe* this solves the instrumentation gap of micro fusion between the time ScalarCallable.call is invoked and the next onAssembly hook, but not extending inside subscription. I think this is correct as micro fusion is as far as I know something that is assembly in nature.

adriancole:
I'm not sure why github is not seeing the latest commit on this branch, but it is here https://github.com/openzipkin/brave/commit/ca7ec9837875d7d5aec8f36720e5f1410608b5fe

@akarnokd I *believe* this solves the instrumentation gap of micro fusion between the time ScalarCallable.call is invoked and the next onAssembly hook, but not extending inside subscription. I think this is correct as micro fusion is as far as I know something that is assembly in nature.

adriancole:
PS by correcting this part:

""
Also downstream.onSubscribe(this) is mandatory, even if you don't plan to support fusion. Without this, operators trying to detect the fuseable feature of their immediate upstream would interface with the wrong operator.
""

This broke all the huge code that made it work(ish) in the first place. Fixing this again will cost too much time, so I will just revert scalar support and we can try again some other time.

adriancole:
so basically will remove all support for "io.reactivex.internal.fuseable" as we can't afford it right now. The better way is to work in rx itself until it can be done without code amplification or inevitable lock in to internal types and consequent drift.

adriancole:
OK the code is now reverted to a cleanup of existing things. This unlocks us so we can safely release it. If we get questions on fusion from sites using this, we can reconsider adding effort.

PR #816, opened 2018-10-24
WIP: Adds verification of logs on b3 single fail

adriancole:
Coming out of https://github.com/envoyproxy/envoy/pull/4712#issuecomment-432499819

Thanks to @zyfjeff cc @shakuzen


adriancole:
this fails while I backfill all the things.. raised early as wanted to show how to do this

adriancole:
ok done.. interestingly we did miss one case here, too

adriancole:
cc @drolando @sfackler @fedj @dpsoft again.. just in case you copy/pasta'd an error

drolando:
@adriancole I don't really understand what `not-yet-sampled header` or your example mean.
Is a b3 header containing `{x-b3-traceid}-{x-b3-spanid}-{x-b3-parentspanid}` (missing sampled flag) valid?

adriancole:
I am not extremely creative, but one possibility for parent ID w/o sampled yet is someone doing log based IDs prior to entering zipkin. For example, if you had a few load balancers stacked. Again I don't think this is something tracers should think about.. the change here was more about the logging vs the edge case.

bplotnick:
I can't remember specifically. Based on my concern here: https://github.com/openzipkin/b3-propagation/issues/21, I might've been referencing one particular issue we had internally with an incorrect custom implementation (we've had a few such cases of custom internal implementations that have made mistakes by not following the b3 standard exactly). Or i could've just wanted more clarity on an ambiguity.

adriancole:
Thanks for piping in @bplotnick FWIW I'm happy to be extra lenient for now. We can make a different change to be un-lenient.. the former code said it addressed this and didn't hence me calling it a bug. the pattern is still questionable at best, and kindof a separate issue I suppose.

PR #819, opened 2018-10-31
Adds Rate limiting sampler

devinsba:
The rate-limited sampler allows you to choose an amount of traces to accept on a per-second interval. The minimum number is 0 and the max is 2,147,483,647 (max int).

For example, to allow 10 traces per second, you'd initialize the following:
```java
tracingBuilder.sampler(RateLimitingSampler.create(10));
```

### Appropriate Usage
If the rate is 10 or more traces per second, an attempt is made to distribute the accept decisions equally across the second. For example, if the rate is 100, 10 will pass every decisecond as opposed to bunching all pass decisions at the beginning of the second.

This sampler is efficient, but not as efficient as the `BoundarySampler`. However, this sampler is insensitive to the trace ID and will operate correctly even if they are not perfectly random.

### Implementation
The implementation uses `System.nanoTime()` and tracks how many yes decisions occur across a second window. When the rate is at least 10/s, the yes decisions are equally split over 10 deciseconds, allowing a roll-over of unused yes decisions up until the end of the second.

### Prior art
`RateLimitingSampler` was made to allow Amazon X-Ray rules to be expressed in Brave. We considered their [Reservoir design](https://github.com/aws/aws-xray-sdk-java/blob/2.0.1/aws-xray-recorder-sdk-core/src/main/java/com/amazonaws/xray/strategy/sampling/reservoir/Reservoir.java). Our implementation differs as it removes a race condition and attempts to be more fair by distributing accept decisions every decisecond. If https://github.com/aws/aws-xray-sdk-java/pull/47 is merged, we'll have the same implementation!

### Thanks!
Thanks @devinsba for spiking the implementation and @anuraaga for suggesting how to be more fair when sampling large numbers of requests. Also appreciate review from @huydx and @zeagord


brave/src/main/java/brave/sampler/RateLimitingSampler.java
```diff
@@ -0,0 +1,57 @@
...
+  private AtomicLong nextUpdate;
+
+  RateLimitingSampler(int tracesPerSecond) {
+    this.tracesPerSecond = tracesPerSecond;
+    this.usage = new AtomicLong(0);
+    this.nextUpdate = new AtomicLong(getNextUpdateValue(System.nanoTime()));
+  }
+
+
+  @Override public boolean isSampled(long traceId) {
```
zeagord: Samll thing. The `traceId` is not used anywhere. Is it there for any future implementations for the WIP?

devinsba: It has to be there from the `Sampler` class we are extending. Maybe I could rename the variable to signal that a bit, but something has to be there

adriancole: if you notice the counting sampler ignores this. we can make a javadoc on the base type that says implementations that don't consider trace ID can ignore the input.

brave/src/main/java/brave/sampler/RateLimitingSampler.java
```diff
@@ -0,0 +1,57 @@
...
+    long now = System.nanoTime();
+    long updateAt = nextUpdate.get();
+
+    // now is past update time OR
+    //      (update time is positive (near Long.MAX_VALUE) AND now is negative (long overflow))
+    if (now > updateAt || (updateAt > 0 && now < 0)) {
+      usage.set(1);
+      nextUpdate.set(getNextUpdateValue(updateAt));
```
adriancole: I think you can use a compare and set to solve the race condition here

devinsba: :+1:

brave/src/main/java/brave/sampler/RateLimitingSampler.java
devinsba: I've added javadoc to the abstract method, and changed the variable name to signal that it is ignored

adriancole:
@anuraaga @trustin fun change to look at. We are trying to include a java 6 compat dependency free non-blocking rate limited sampler. For simplicity this would allow the bucket to empty immediately (ex if 1000 requests/second we don't partition it into 10millis buckets, rather one big bucket and so if all requests deplete in the first millis, so be it).

That said, what thoughts have you on the impl and where we go from here. This is one of the most requested features and is a gap as folks try to bridge from the AWS sdk which has a similar impl but using clock time instead.

adriancole:
@devinsba question around license headers.. yeah this can wait. we do package the LICENSE file since a long time.

brave/src/main/java/brave/sampler/RateLimitingSampler.java
```diff
@@ -0,0 +1,58 @@
...
+
+  @Override public boolean isSampled(long ignoredTraceId) {
+    long now = System.nanoTime();
+    long updateAt = nextUpdate.get();
+
+    // now is past update time OR
+    //      (update time is positive (near Long.MAX_VALUE) AND now is negative (long overflow))
+    if (now > updateAt || (updateAt > 0 && now < 0)) {
+      if (nextUpdate.compareAndSet(updateAt, getNextUpdateValue(updateAt))) {
+        usage.set(1);
```
anuraaga: Would use 0 and `incrementAndGet` since it's a bit weird for empty to have a non-zero value.

adriancole:
I would say add decrementBy to brave.internal.Platform

in jre less than 8 it would be a loop, but that's ok as many will use the
normal method.

adriancole:
added a cleanup commit

adriancole:
here are bench results. It does take time to issue nanoTime, so some overhead is normal. Keep in mind this only applies to first span in a trace and we are talking about operations per microsecond

```
Benchmark                                         (traceId)   Mode  Cnt    Score    Error   Units
SamplerBenchmarks.sampler_boundary     -9223372036854775808  thrpt   15  512.167 ± 20.785  ops/us
SamplerBenchmarks.sampler_boundary      1234567890987654321  thrpt   15  515.468 ± 26.426  ops/us
SamplerBenchmarks.sampler_counting     -9223372036854775808  thrpt   15   25.422 ±  1.467  ops/us
SamplerBenchmarks.sampler_counting      1234567890987654321  thrpt   15   26.166 ±  1.278  ops/us
SamplerBenchmarks.sampler_rateLimited  -9223372036854775808  thrpt   15   14.433 ±  0.650  ops/us
SamplerBenchmarks.sampler_rateLimited   1234567890987654321  thrpt   15   14.650 ±  0.336  ops/us
```

adriancole:
I take back the advice for "I would say add decrementBy to brave.internal.Platform" it isn't needed.

brave/src/main/java/brave/sampler/RateLimitingSampler.java
```diff
@@ -0,0 +1,43 @@
...
+    this.tracesPerSecond = tracesPerSecond;
+    long now = System.nanoTime();
+    this.nextUpdate = new AtomicLong(now + NANOS_PER_SECOND);
+  }
+
+  @Override public boolean isSampled(long ignoredTraceId) {
+    long now = System.nanoTime();
+    long updateAt = nextUpdate.get();
+
+    if (now - updateAt >= 0) {
```
devinsba: Whelp, that's a much better way of doing the negative value correction. Nice

brave/src/main/java/brave/sampler/RateLimitingSampler.java
```diff
@@ -0,0 +1,43 @@
...
+import java.util.concurrent.atomic.AtomicLong;
+
+/**
+ * Based on https://github.com/aws/aws-xray-sdk-java/blob/2.0.1/aws-xray-recorder-sdk-core/src/main/java/com/amazonaws/xray/strategy/sampling/reservoir/Reservoir.java
+ *
+ * Uses only Java 6 APIs to be compatible with Brave Core requirements. Accounts for nanoTime()
+ * overflow, as unlikely as that might be.
+ */
+public class RateLimitingSampler extends Sampler {
+  public static Sampler create(int tracesPerSecond) {
```
huydx: is it better use duration (pass long and TimeUnit), it will make the implementation become a bit complex but we have more friendly interface.

devinsba: I was considering that this morning. That logic could all be handled in the constructor so that the sample logic is the same. I'll give it a shot in the morning now that Adrian has provided benchmark scripts to check with

adriancole: I think we should add long and timeunit eventhough some timeunits are weird :) I have a different reason, though. You'll notice the base type Sampler has a create method. you can easily accidentally end up with a different impl if you use the wrong numeric type.

Changing the arity of the method would also solve that even if only units at or larger than second are allowed. One of the issues with having raw units though is that you can end up with fractional conversion if someone expresses something like 10/minute :P

IOTW I can see it both ways, I definitely think per second is easiest, and that is also the same as amazon accept fwiw

adriancole: @trustin @anuraaga any thoughts on this one?

adriancole: another thought is we use a builder.. one thing left dangling here is that the reset interval is default to second which is very easy impl as it is same unit as rate. reset interval of smaller can be helpful for people with high throughput, but again you end up with how to split the fractions if things are not aligned.

ex 35span/second reset interval of 10millis, you have to choose to have some buckets include zero and others one. This can be a separate change, as the logic would grow quite a bit.

adriancole:
was thinking about the main problem with the "all samples at the beginning of the second" problem @devinsba @anuraaga. What if we setup some basic rules with a "rollover plan"

Ex.
if < 10/s allow all to happen at any time in the second
if < 100/s bucket <=10/decisecond with a rollover.

ex rate is 99 and if none until the last decisecond, there are 99 available.

then some function over that as the rate increases, either a higher magnitude per decisecond or smaller bucket than decisecond. Note: fixed buckets, especially few fixed buckets can allow for some interesting impls. Ex our counting sampler is cute due to known bounds https://github.com/openzipkin/brave/blob/master/brave/src/main/java/brave/sampler/CountingSampler.java

Thoughts?

adriancole:
one reminder on the bucket thing is that the minimum rate in my suggestion is 1/s. it is easier code and matches Amazon's ReservoirSize https://docs.aws.amazon.com/xray/latest/api/API_SamplingRule.html slow rates like per minute or hour is complicated and I don't think we want to handle that until/unless explicitly requested

anuraaga:
That SGTM - not thinking about per minute / hour seems fine since they would rarely sample.

instrumentation/benchmarks/src/main/java/brave/sampler/SamplerBenchmarks.java
```diff
@@ -0,0 +1,108 @@
...
+import org.openjdk.jmh.annotations.State;
+import org.openjdk.jmh.annotations.Threads;
+import org.openjdk.jmh.annotations.Warmup;
+import org.openjdk.jmh.runner.Runner;
+import org.openjdk.jmh.runner.RunnerException;
+import org.openjdk.jmh.runner.options.Options;
+import org.openjdk.jmh.runner.options.OptionsBuilder;
+
+/**
+ * <p>Brave uses before-the-fact sampling. This means that the decision to keep or drop the trace
```
adriancole: @wu-sheng some text

devinsba:
If someone else has the time, energy, and know how to get this over the finish line I'd appreciate it. I've been sick and time has been tight.

adriancole:
I have a long flight tomorrow.. my intention is to help.. unless they
have 6 liam neeson films..

brave/src/main/java/brave/sampler/RateLimitingSampler.java
```diff
@@ -0,0 +1,103 @@
...
+  }
+
+  static final long NANOS_PER_SECOND = TimeUnit.SECONDS.toNanos(1);
+  static final int NANOS_PER_DECISECOND = (int) (NANOS_PER_SECOND / 10);
+
+  final MaxFunction maxFunction;
+  final AtomicInteger usage = new AtomicInteger(0);
+  final AtomicLong nextReset;
+
+  RateLimitingSampler(int tracesPerSecond) {
```
huydx: @adriancole personally I may want this variable to be reset-able, which mean that we could dynamically change it from outside. We may just need an `apply` function which take new tracesPerSecond and replace maxFunction with new one, HDYT?

adriancole:
if we do this we need to synchronize the value then also re adjust. since
not everyone will set this dynamic and also we can set dynamic in a
different way... what if instead the sampler itself is replaced?

devinsba:
To Note: I had intended to do the "replace the sampler" approach for the XRay wrapper of this

adriancole:
reason I suggest doing dynamic adjustment externally is that it is in
practice hard to guess how you will update..  would you pass AtomicInteger?
a callback with a new rate? or we expect a synchronous call to setRate?
there are multiple ways to update a value. Then as the change is defined
someone also wants to compose a 5% overflow on top! in other words I would
prefer to help solve the dynamic adjustment thing externally until we know
how the rate is adjusted is actually shared.

huydx:
@adriancole replace the sampler itself sounds better, totally forgot about that. Anw the implementation looks good for me.

brave/src/main/java/brave/sampler/RateLimitingSampler.java
huydx: do we gonna make a builder?

adriancole:
> do we gonna make a builder?
>

nice thing about builder is we can add late. right now there is only one
parameter so we can defer until there is more than one and still not break
simple case

>

instrumentation/benchmarks/src/main/java/brave/sampler/SamplerBenchmarks.java
```diff
@@ -0,0 +1,108 @@
...
+
+  // Convenience main entry-point
+  public static void main(String[] args) throws RunnerException {
+    Options opt = new OptionsBuilder()
+        .include(".*" + SamplerBenchmarks.class.getSimpleName())
+        .build();
+
+    new Runner(opt).run();
+  }
+}
```
huydx: better add new line

adriancole:
I think all this needs is javadoc polishing and readme. @anuraaga are you happy with impl?

brave/src/main/java/brave/sampler/RateLimitingSampler.java
adriancole: I think for now it is ok to use as-is. It prevents people from making units less than one second

adriancole:
ran benchmarks against another tracer which won't be mentioned.. our results are an order of magnitude more efficient in terms of allocation and runtime.

brave/src/main/java/brave/sampler/RateLimitingSampler.java
```diff
@@ -0,0 +1,103 @@
+package brave.sampler;
+
+import java.util.concurrent.TimeUnit;
+import java.util.concurrent.atomic.AtomicInteger;
+import java.util.concurrent.atomic.AtomicLong;
+
+/**
+ * Inspired by on https://github.com/aws/aws-xray-sdk-java/blob/2.0.1/aws-xray-recorder-sdk-core/src/main/java/com/amazonaws/xray/strategy/sampling/reservoir/Reservoir.java
```
adriancole: TODO remove this as it is not really like that impl in any way. the readme should say that the intent is to accept the same parameter as x-ray though.

adriancole:
added https://github.com/aws/aws-xray-sdk-java/pull/47 to give back to amazon

brave/src/main/java/brave/sampler/RateLimitingSampler.java
adriancole: as mentioned, we'll do this externally

brave/src/main/java/brave/sampler/RateLimitingSampler.java
```diff
@@ -0,0 +1,122 @@
+package brave.sampler;
+
+import java.util.concurrent.TimeUnit;
+import java.util.concurrent.atomic.AtomicInteger;
+import java.util.concurrent.atomic.AtomicLong;
+
+/**
+ * The rate-limited sampler allows you to choose an amount of traces to accept on a per-second
+ * interval. The minimum number is 1 and there is no maximum.
```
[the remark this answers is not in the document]
huydx: and the implement seems said that it accept 0 too, should we throw exception for that?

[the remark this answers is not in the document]
adriancole: umm we can clarify that zero is the same as never sample. or just say minimum is zero

devinsba:
:+1: Thanks for picking this up @adriancole, I'll see if I can find time to fix that XRay test today so your PR can get merged

adriancole:
I will look into the test failure in aws/aws-xray-sdk-java#47 before merge. thanks!

adriancole:
Amazon's test case found a legit glitch. Fixed in both places

PR #823, opened 2018-11-03
Allow adding annotations with a relative timestamp.

anuraaga:
In https://github.com/line/armeria/pull/1423 I am working on adding a timestamp for wire send / receive using `System.nanoTime()` and would like to be able to annotate the span with a relative timestamp. I guess this may be reasonably common usage as many systems will have their own logging facade which is interfaced with Brave in some way, so hopefully it makes sense to add some API to `Span` for it.

### DIFF
--- a/brave/src/main/java/brave/NoopSpan.java
+++ b/brave/src/main/java/brave/NoopSpan.java
@@ -46,6 +46,10 @@ final class NoopSpan extends Span {
     return this;
   }
 
+  @Override public Span annotateRelative(long microsFromStart, String value) {
+    return this;
+  }
+
   @Override public Span remoteServiceName(String remoteServiceName) {
     return this;
   }
--- a/brave/src/main/java/brave/RealSpan.java
+++ b/brave/src/main/java/brave/RealSpan.java
@@ -101,6 +101,16 @@ final class RealSpan extends Span {
     return this;
   }
 
+  @Override public Span annotateRelative(long microsFromStart, String value) {
+    synchronized (state) {
+      long startTimestamp = state.startTimestamp();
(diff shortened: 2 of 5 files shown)


adriancole:
heh I hate to say it, but internally brave already does relative timestamps with TickClock :P You don't need to do anything special to use nanoTick for annotations (as just doing span.annotate("ws") will use a tick)

Good news is we think alike!... or I misunderstand you and you really do need to supply a delta?

adriancole:
hmm this is interesting.. I think you are looking at explicit relative time. How do spans work now in armeria? you use implicit start and finish, but explicit timestamps for annotations? I suppose without this api you'd need to save off the timestamp first...

There is some code in brave that tries to do alignment like this.. the api that might not be obvious is that you can get a clock for a context.. with that you can fix an explicit timestamp and do other things later.. https://github.com/openzipkin/brave/blob/master/instrumentation/spring-rabbit/src/main/java/brave/spring/rabbit/TracingRabbitListenerAdvice.java#L64

anuraaga:
Yeah we currently use implicit start and finish, so we get to use Brave's microseconds clock. Was hoping to use our other nanotime timestamps to take advantage of the microseconds clock, since wall time is only measured with milliseconds precision in Armeria's logging facade. 

But actually, looking at the implementation of Platform.java in Brave, I realized it wouldn't be that hard to update Armeria's logging to use micros instead I guess that could be better than adding this API.

adriancole:
still desired?

anuraaga:
Nope guess there'd never be a point for it since either you track all time separately or all time with Brave, but mixing wouldn't work well.

PR #834, opened 2018-11-26
fix typo

trayo:


brave/src/main/java/brave/sampler/CountingSampler.java
```diff
@@ -5,7 +5,7 @@
 
 /**
  * This sampler is appropriate for low-traffic instrumentation (ex servers that each receive <100K
- * requests), or those who do not provision random trace ids. It not appropriate for collectors as
+ * requests), or those who do not provision random trace ids. It is not appropriate for collectors as
```
devinsba: It looks like our line lengths were fairly normal, can you move the `as` down on the next line.
```suggestion
 * requests), or those who do not provision random trace ids. It is not appropriate for collectors
 * as the sampling decision isn't idempotent (consistent based on trace id).
```

brave/src/main/java/brave/sampler/CountingSampler.java
```diff
@@ -5,7 +5,8 @@
 
 /**
  * This sampler is appropriate for low-traffic instrumentation (ex servers that each receive <100K
- * requests), or those who do not provision random trace ids. It not appropriate for collectors as
+ * requests), or those who do not provision random trace ids. It is not appropriate for collectors
+ * as the sampling decision isn't idempotent (consistent based on trace id).
  * the sampling decision isn't idempotent (consistent based on trace id).
```
devinsba: Well, i guess that feature is in beta....
```suggestion
```

devinsba: I wonder how to do a multiline suggestion?

trayo: Heh, I thought I would give it a try. Seemed convenient. :)

devinsba:
Looks like a flaky test in gRPC land is failing the build. I'll create an issue to address it. Thanks

PR #835, opened 2018-11-26
Update broken link in documentation

trayo:
I updated the link pointing to Twitter finagle. Included is a fix to the folder which changed from `finagle-zipkin` to `finagle-zipkin-core`. Also included in the url is the commit blob to ensure the link with the line number doesn't change unintentionally.

I'm not sure if the line number is still the correct, and I would appreciate if a maintainer could double check. Perhaps @adriancole ?


adriancole:
yep good guess

adriancole:
thanks!

PR #837, opened 2018-11-29
update vert.x tracing README

liushunqiu:
update vert.x tracing README

### DIFF
--- a/instrumentation/vertx-web/README.md
+++ b/instrumentation/vertx-web/README.md
@@ -21,3 +21,17 @@ router.route("/foo").handler(ctx -> {
     ctx.response().end("bar");
 });
 ```
+The exception is that the vert.x tracing order cannot be set to -1 
+when the application needs to handle requests such as POST 
+and needs to place the vert.x traing after
+ the `router.route().Handler(BodyHandler.create())`.<br/>
+ (例外，当应用需要处理POST等请求，需要把vert.x traing 放到
+ `router.route().handler(BodyHandler.create())`之后，vert.x tracing order 不能设置为-1)
+```java
+router.route().handler(BodyHandler.create());
+Handler<RoutingContext> routingContextHandler = VertxWebTracing.create(httpTracing).routingContextHandler();
+router.route()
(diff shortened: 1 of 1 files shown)


instrumentation/vertx-web/README.md
```diff
@@ -9,15 +9,16 @@ also sent to Zipkin.
 To enable tracing you need to set `order`, `handler` and `failureHandler`
 hooks:
 ```java
+//tracing setup must happen after any configuration like .handler(BodyHandler.create()) or application routes
 vertxWebTracing = VertxWebTracing.create(httpTracing);
 routingContextHandler = vertxWebTracing.routingContextHandler();
 router.route()
-      .order(-1) // applies before routes
```
adriancole: only concern is removing this as it breaks all our tests

adriancole:
let's put this on hold until https://github.com/openzipkin/brave/pull/837 is done?

adriancole:
fyi as per my comment https://github.com/openzipkin/brave/pull/837/files#r237379958

this isn't going to be merged until we understand why it is ok to recommend a setting that breaks all of our tests. Someone today said that this works if you change to 1, not -1

Maybe @vietj can recommend something.. If I remove the order completely, it breaks our tests. I haven't tried changing it to 1. If we change the README, we should also change our test, and then it shouldn't break as a result.

https://github.com/openzipkin/brave/blob/master/instrumentation/vertx-web/src/test/java/brave/vertx/web/ITVertxWebTracing.java#L91

```
@anuj_iitd1_twitter
20:54
Can anyone merge this pull request
https://github.com/openzipkin/brave/pull/837/commits/f40cff39f8ef254deef4efbcda721d55a55e8d15

It mentions if you are using
router.route().Handler(BodyHandler.create())
then route order should be 1 and not -1 i.e.
router.route() .order(1)
as it is mentioned as -1 in the ReadMe . I wasted a hell lot of time figuring out this.
```

adriancole:
I'm going to close this until someone responds

PR #843, opened 2018-12-11
Adds JDK Flight Recorder context correlation

adriancole:
This adds trace and span IDs to JDK Flight Recorder "Scope"
events so that you can correlate with Zipkin UI or logs accordingly.

This currently requires JDK 11. JVMs that use this need to add the below
to their VM start arguments:
```bash
-XX:+UnlockCommercialFeatures -XX:+FlightRecorder
```

To view the flight recordings, you need [JDK Mission Control 7](http://jdk.java.net/jmc/).

With the above in place, you can configure `brave.Tracing` with
`JfrScopeDecorator` like so:

```java
    .currentTraceContext(ThreadLocalCurrentTraceContext.newBuilder()
       .addScopeDecorator(JfrScopeDecorator.create())
       .build()
    )
    ...
    .build();
```

After a flight is recorded, you can look for "Zipkin/Scope" in the
Event browser like so:

flight recording

Users could then copy/paste the trace ID into the zipkin UI, or use log
correlation to further debug a problem.

Thanks to @thegreystone for hints on how to get this working!


adriancole:
PS screen shots are from the https://github.com/openzipkin/brave-webmvc-example which uses @anuraaga's nice scope decorator design as it allows you to add several types of correlation, in this case SLF4J and also JFR without wrapping the tracer or a current context component.

context/jfr/src/main/java/brave/context/jfr/JfrScopeDecorator.java
```diff
@@ -0,0 +1,70 @@
...
+ *                    .addScopeDecorator(JfrScopeDecorator.create())
+ *                    .build()
+ *                  )
+ *                  ...
+ *                  .build();
+ * }</pre>
+ */
+public final class JfrScopeDecorator implements ScopeDecorator {
+
+  @Category("Zipkin")
```
adriancole: super cool that from here down is less than 50loc. it is very easy to implement this, almost copy/pasteable!

adriancole:
PS the phrase "Scope Event" looks awkward in the list as other events have nicer names and don't say "event" either. Ex "Socket Read" "Allocation in new TLAB". I only used that name from the opentracing example.

It might be best to name it something like "Span in Scope", which matches the actual operation.

adriancole:
@nicmunroe FYI this is fun and doesn't rely on out-of-band data. It might work well for you in wingtips.

adriancole:
I will rename "Scope Event" to "Span in Scope" unless someone says they hate it :P

adriancole:
This is more reaction than most features we've done lately :P will merge soon!

schrepfler:
Being a ThreadLocal, I suppose this would be of limited applicability to things like Akka, Twitter Future or Monix Task? Can we somehow make it generic/pluggable to be usable with other concurrency mechanisms (also Fibers will be coming our way).

adriancole:
@schrepfler the implementation is not dependent on thread local, it is actually a scoping api (try/finally). if the scoping doesn't work you have bigger problems than JFR :P

adriancole:
ex https://github.com/openzipkin/brave/blob/master/context/rxjava2/src/main/java/brave/context/rxjava2/internal/TraceContextSubscriber.java#L35

thegreystone:
Is there a problem with keeping this as part of the OpenTracing community? Taking the idea wholesale (http://hirt.se/blog/?p=1081, https://oracle.rainfocus.com/widget/oracle/oow18/catalogcodeone18?search=hirt) and jamming it into Brave, apart from making me feel slightly violated, makes dual events possible if running with the OpenTracing contrib part. Even worse, the JMC specific extensions we're planning will be for the open tracing events.

devinsba:
So,to clarify, works with wrapped executor services?


________________________________

adriancole:
@devinsba anything derived from `CurrentTraceContext` will work including executors.

adriancole:
@thegreystone I cited your work, maybe not all of the links you just did, but attempted to!

The code here does not prevent opentracing from doing anything, unless I somehow misunderstand. We are using a separate category, clearly marked as Zipkin. How does that interfere?

The below is the only code similar to what you have. Are you saying that using the `jdk.jfr` package in this way is somehow proprietary?

```java
  @Category("Zipkin")
  @Label("Span In Scope")
  @Description("Zipkin event representing a span being placed in scope")
  static final class ScopeEvent extends Event {
    @Label("Trace Id") String traceId;
    @Label("Parent Id") String parentId;
    @Label("Span Id") String spanId;
  }
```

thegreystone:
> The code here does not prevent opentracing from doing anything, unless I somehow misunderstand. >We are using a separate category, clearly marked as Zipkin. How does that interfere?

It does not, but it produces redundant data, but perhaps that is okay.

> The below is the only code similar to what you have. Are you saying that using the `jdk.jfr` package > in this way is somehow proprietary?

No, it is fine. I should probably take a close look at the code. The problem starts on the JMC side of things - we were planning on building nice support for distributed tracers in JMC itself. If every tracer runs off and does custom integration, it will be hard for us to do this nicely. I _wish_ there was a standardised nice decorator API that was shared among the distributed tracers, but there is not.

adriancole:
@thegreystone brave isn't exactly "any tracer" it is the most used library, far more used than opentracing. We do have users that use the opentracing bridge, but it will never be an entrypoint for all apps due to various reasons mentioned in their issues list the last few years, plus not everyone wants to use it.

Ideally, JMC will like to work with zipkin as it is the largest ecosystem... if you'd prefer not to, that's ok to.. your call! Meanwhile, people here have already engaged in your ideas far more than the other project. So, take the win?

adriancole:
I was going to merge this today, but since there's some mild drama will wait a day

thegreystone:
> I was going to merge this today, but since there's some mild drama will wait a day

Sorry for providing some drama. I can also provide review feedback. The latter will hopefully be useful.

thegreystone:
Want me to leave the review comments here, like I did with the unlock commercial features one?

thegreystone:
> I will rename "Scope Event" to "Span in Scope" unless someone says they hate it :P

Could you just name it Scope?

adriancole:
@thegreystone comments here are good. line notes even better, but anywhere is fine. thanks for taking the time!

thegreystone:
Don't have access to commenting directly. Perhaps that could be fixed?

adriancole:
@thegreystone I have no idea why you can't make a line comment.. I quickly tried to find if there's even a setting to restrict this and couldn't. We've never restricted line comments intentionally. sorry it isn't working for whatever reason.

adriancole:
renamed to Scope, that's a better name. Thanks @thegreystone. I'll wait until other feedback is in before re-doing the screen shot.

thegreystone:
Also fix the comment about commercial features - for JDK 11 Oracle donated flight recorder etc. The flag will give you a warning now, and since you only support JDK 11 with this patch (not JDK 8) the comment should be changed.

adriancole:
wrt "about commercial features" will nix.. easier anyway!

thegreystone:
We're being stupid in our implementation too...

For the scope event, just call commit:
No need for this:
         if (!event.shouldCommit()) return;
         event.end();
         event.commit();

Also, in JDK 11 - no need to end the event first. Just call commit.

.travis.yml
```diff
@@ -10,7 +10,7 @@ cache:
 language: java
 
 jdk:
-  - oraclejdk9
+  - oraclejdk11
```
anuraaga: Just wondering, should we make this openjdk11? I guess it might be causing the confusion about commercial features. Targeting the baseline seems reasonable

thegreystone: Yes. Definitely.

adriancole: ah yeah that was unintentional.. I just needed something 11 :P can try to change to openjdk11 presuming travis has it

thegreystone:
Good thing about this is that I get to clean up my code. ;)

thegreystone:
Any chance all you tracer folks could get together and agree on a nice vendor neutral decorator API? ;)

context/jfr/src/main/java/brave/context/jfr/JfrScopeDecorator.java
```diff
@@ -0,0 +1,68 @@
...
+    }
+
+    event.begin();
+
+    class JfrCurrentTraceContextScope implements Scope {
+      @Override public void close() {
+        scope.close();
+        if (!event.shouldCommit()) return;
+        event.end();
+        event.commit();
```
thegreystone: No need to check shouldCommit, no need to end before commit (in JDK 11). Simply just call event.commit()

.travis.yml
thegreystone: openjdk11

adriancole:
> Any chance all you tracer folks could get together and agree on a nice vendor neutral decorator API? ;)
you mean besides ours? :P anyway feel free to propose our design to opentracing. they sometimes accept change requests.

thegreystone:
> @thegreystone I have no idea why you can't make a line comment

Seems to work now. Weird.

adriancole:
ok I think I got all the feedback in, but check me. I need to be away for a 90m meeting then sleep. Anything else will address tomorrow.

Thanks @thegreystone this is better thanks to you!

thegreystone:
>you mean besides ours? :P anyway feel free to propose our design to opentracing. they sometimes >accept change requests.

Well, it would need to be a common one shared among all participating tracers. So, in a namespace that you could all agree on. But it would certainly make life easier.

thegreystone:
> Thanks @thegreystone this is better thanks to you!

Aside from "existing", but I'll take it. ;) I still think it would be better if we could somehow find a vendor independent way of doing this.

thegreystone:
Hey, I'm not a tracing guy, so I don't care how you all go about standardising and/or agreeing on standards. :) My life would just be easier if you do.

thegreystone:
> It's cool connecting profilers and tracing like this. 

That said, JFR is much more than a profiler.

adriancole:
out in 5.6.0 https://github.com/openzipkin/brave/tree/master/context/jfr (added explicit credits, though we should also not forget that @thegreystone is also to thank for JMC itself!)

felixbarny:
This is really cool :)

@thegreystone Is it possible to correlate events with method profiling? I'm thinking about implementing a feature to create a flame graph of a particular trace/span.

@adriancole Is it guaranteed that the activation and corresponding deactivation is on the same thread? If so, you might be able to reuse the event object in a thread local. Because spans can be nested, it would have to be something like a `ThreadLocal<List<ScopeEvent>>`. And then there is the problem of class loader leaks with thread locals...

schrepfler:
Would be good to cover also Monix `TaskLocal` (https://medium.com/pragmatic-scala/local-variable-propagation-with-monix-part-1-106b80e652bb ) and related Twitter's `Contexts` https://twitter.github.io/finagle/guide/Contexts.html CC'ng @alexandru and @cacoco for awareness.

nicmunroe:
> @nicmunroe FYI this is fun and doesn't rely on out-of-band data. It might work well for you in wingtips.

Thanks for the heads up! I'll definitely need to take a look.

thegreystone:
> @thegreystone Is it possible to correlate events with method profiling? I'm thinking about implementing a feature to create a flame graph of a particular trace/span.

https://github.com/thegreystone/jmc-flame-view

thegreystone:
So Flame View, plus techniques hinted at in this article, is what you are looking for:
hirt.se/blog/?p=1081

Then you can simply select a Trace ID looking interesting, "Store and set as active selection", and then you can look at a Flame View over not only method profiling events (though you can do that too), but a Flame View of the aggregated stack traces for the allocation hotspots, or whatever else you fancy.

Though, honestly, I did the Flame View mostly for fun. I find the normal stack trace view in JMC to be much faster to work with.

PR #846, opened 2018-12-11
Reduces allocations when using ID correlation by caching hex values

### LINKED ISSUE #844: consider string helpers for parent span id and span id
We have a lot of code like this:

```java
      if (parentId != 0L) event.parentSpanId = HexCodec.toLowerHex(parentId);
      event.spanId = HexCodec.toLowerHex(currentSpan.spanId());
```

If we had `spanIdString()` and `parentSpanIdString()` we could cache the hex and likely save a lot of GC thrash. Also, end users won't need to reference an internal type.

adriancole:
Before, whenever hex operations were needed, a new string was allocated.
This meant if a span was scoped multiple times, all ids would be
generated multiple times. This removes that problem and also caches
hashCode as it is used as a key.

Fixes #844

### DIFF
--- a/brave-tests/src/main/java/brave/test/propagation/PropagationTest.java
+++ b/brave-tests/src/main/java/brave/test/propagation/PropagationTest.java
@@ -166,7 +166,7 @@ public abstract class PropagationTest<K> {
       injector.inject(ctx, map);
 
       assertThat(extractor.extract(map).context())
-          .isEqualToComparingFieldByField(ctx);
+          .isEqualToIgnoringGivenFields(ctx, "traceIdString", "spanIdString");
     }
   }
 }
--- a/brave-tests/src/test/java/brave/propagation/B3PropagationTest.java
+++ b/brave-tests/src/test/java/brave/propagation/B3PropagationTest.java
@@ -1,6 +1,5 @@
 package brave.propagation;
 
-import brave.internal.HexCodec;
 import brave.internal.Nullable;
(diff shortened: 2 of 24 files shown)


adriancole:
Added benchmarks that show if multiple scope decorators are present, this has a big impact:

before

```
EndToEndBenchmarks.tracedCorrelatedServer_get:tracedCorrelatedServer_get·p0.999   sample           1291.336               us/op
EndToEndBenchmarks.tracedCorrelatedServer_get:·gc.alloc.rate.norm                 sample      15  50474.876 ± 21568.930    B/op
EndToEndBenchmarks.tracedServer_get:tracedServer_get·p0.999                       sample            621.432               us/op
EndToEndBenchmarks.tracedServer_get:·gc.alloc.rate.norm                           sample      15  28093.729 ±  9335.425    B/op
```

after

```
EndToEndBenchmarks.tracedCorrelatedServer_get:tracedCorrelatedServer_get·p0.999   sample            612.352               us/op
EndToEndBenchmarks.tracedCorrelatedServer_get:·gc.alloc.rate.norm                 sample      15  49501.385 ± 21044.690    B/op
EndToEndBenchmarks.tracedServer_get:tracedServer_get·p0.999                       sample            496.646               us/op
EndToEndBenchmarks.tracedServer_get:·gc.alloc.rate.norm                           sample      15  27977.567 ±  9267.475    B/op
```

shakuzen:
This is awesome. Thanks Adrian!

PR #863, opened 2019-01-30
grpc instrumentation close spans in halfClose #862

### LINKED ISSUE #862: grpc instrumentation error
When an error occurs in a grpc service, the span is not finished correctly.

It's related with this issue: https://github.com/grpc/grpc-java/issues/1552

The solution could be something like this, in class `brave.grpc.TracingServerInterceptor#ScopingServerCallListener`:

```
@Override public void onHalfClose() {
      SpanInScope scope = tracer.withSpanInScope(span);
      try { // retrolambda can't resolve this try/finally
          delegate().onHalfClose();
          span.error(e);
          span.finish();
          throw e;
      }  finally {
        scope.close();
      }
}
```

oburgosm:
Fixes #862

### DIFF
--- a/instrumentation/grpc/src/main/java/brave/grpc/TracingServerInterceptor.java
+++ b/instrumentation/grpc/src/main/java/brave/grpc/TracingServerInterceptor.java
@@ -136,6 +136,10 @@ final class TracingServerInterceptor implements ServerInterceptor {
       SpanInScope scope = tracer.withSpanInScope(span);
       try { // retrolambda can't resolve this try/finally
         delegate().onHalfClose();
+      } catch (RuntimeException | Error e) {
+          span.error(e);
+          span.finish();
+          throw e;
       } finally {
         scope.close();
       }
--- a/instrumentation/grpc/src/main/java/brave/grpc/TracingServerInterceptor.java
+++ b/instrumentation/grpc/src/main/java/brave/grpc/TracingServerInterceptor.java
@@ -137,9 +137,11 @@ final class TracingServerInterceptor implements ServerInterceptor {
       try { // retrolambda can't resolve this try/finally
         delegate().onHalfClose();
       } catch (RuntimeException | Error e) {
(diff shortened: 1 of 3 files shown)


instrumentation/grpc/src/main/java/brave/grpc/TracingServerInterceptor.java
```diff
@@ -136,6 +136,10 @@ public void sendMessage(RespT message) {
       SpanInScope scope = tracer.withSpanInScope(span);
       try { // retrolambda can't resolve this try/finally
         delegate().onHalfClose();
+      } catch (RuntimeException | Error e) {
```
adriancole: is the intent to close the span on half-close only on error? maybe add a comment why

oburgosm: I have replicated the behavior of the 'close' method, but I think there would be no problem in capturing Throwable instead.

adriancole: I guess then if half-close occurs then close wouldn't? sorry I'm not aware of the state diagram.

adriancole: So, I suppose a comment would be something like..

```
// if there was an exception executing onHalfClose, we don't expect other lifecycle commands
// Accordingly, we close the span
```

oburgosm:
```suggestion
      } catch (RuntimeException | Error e) {
      // if there was an exception executing onHalfClose, we don't expect other lifecycle commands
      // Accordingly, we close the span
```

adriancole: I will edit the file directly

PR #865, opened 2019-02-05
Ensures http responses served from cache don't end up marked remote

### LINKED ISSUE #864: [BUG][brave-instrumentation-httpclient] TracingCachingHttpClientBuilder breaks http client caching
TracingCachingHttpClientBuilder's override of decorateMainExec fails to add a CacheExecution to the http client execution chain. This prevents http caching from working.

Here is a [Wiremock](https://mvnrepository.com/artifact/com.github.tomakehurst/wiremock/2.20.0) test that reproduces the issue:

```java
public class TracingCachingHttpClientBuilderTest extends ITHttp {

    @Rule
    public WireMockRule wireMockRule = new WireMockRule();

    private HttpTracing httpTracing= HttpTracing.create(tracingBuilder(Sampler.NEVER_SAMPLE).build());

    // Tracing builder breaks caching:
    @Test 
    public void tracingCachingHttpClientShouldCacheResponse() throws Exception {
        String path = "/test/123";
        HttpClient httpClient = new TracingCachingHttpClientBuilder(httpTracing).build();

        stubFor(get(urlEqualTo(path))
                .willReturn(aResponse()
                        .withStatus(200)
                        .withHeader("Content-Type", "text/plain")
                        .withHeader("Cache-Control", "max-age=600, stale-while-revalidate=1200")
                        .withBody("Hello")));

        HttpGet httpGet = new HttpGet("http://localhost:8080" + path);

        httpClient.execute(httpGet);
        httpClient.execute(httpGet);

        verify(exactly(1), getRequestedFor(urlEqualTo(path)));
    }

    //Vanilla Apache builder is ok.
    @Test
    public void cachingHttpClientShouldCacheResponse() throws Exception {
        String path = "/test/123";
        HttpClient httpClient = CachingHttpClients.createMemoryBound();

        stubFor(get(urlEqualTo(path))
                .willReturn(aResponse()
                        .withStatus(200)
                        .withHeader("Content-Type", "text/plain")
                        .withHeader("Cache-Control", "max-age=600, stale-while-revalidate=1200")
                        .withBody("Hello")));

        HttpGet httpGet = new HttpGet("http://localhost:8080" + path);
        httpClient.execute(httpGet);
        httpClient.execute(httpGet);

        verify(exactly(1), getRequestedFor(urlEqualTo(path)));
    }
}
```

A simple fix like:

`return new TracingMainExec(httpTracing, super.decorateMainExec(exec));`

will lead to a span being started even when the request is served from the cache. If that's not desirable, then you'll probably need to subclass TracingMainExec, and avoid sending a span if HttpCacheContext.CACHE_RESPONSE_STATUS = CACHE_HIT.

adriancole:
Thanks to @amccarthy-atlas for the easy to follow solution

Fixes #864


adriancole:
while we dont have anything standard for this, the same concern exists in okhttp. probably recording a cache.hit tag or similar could reduce confusion. @shakuzen @devinsba @anuraaga any thoughts?

shakuzen:
> while we dont have anything standard for this, the same concern exists in okhttp. probably recording a cache.hit tag or similar could reduce confusion

I wonder if it would it be worthwhile to distinguish between HTTP cache hit and some other kind of cache hit? `http.cache.hit`?

adriancole:
as our namespaces are currently flat within a classifier, I'd mildly
suggest http.cache_hit

adriancole:
as I think the tag name is somewhat bikeshed.. I'll merge and someone can pipe in, in the next day or two, if they care to have something besides http.cache_hit :P

PR #868, opened 2019-02-26
Correct statement about BOM properties

shakuzen:
Removed mention of properties defined in the Brave BOM since they cannot be used in POMs that import the Brave BOM. Only the properties defined in the project's POM or its parents can be used.


shakuzen:
A change in the same vein as https://github.com/openzipkin/zipkin-reporter-java/pull/129

PR #876, opened 2019-03-11
Fixes bug where we misclassified 1xx range as error (4.13.x)

adriancole:
Same as #875, but for the 4.13.x branch. Note, we don't have a current
schedule to do a last 4.13.x release, but we should prior to moving to
ASF at least to fix dependency versions.


adriancole:
there is no CI on this branch as we almost never do backports

PR #880, opened 2019-03-19
add alibaba druid support more database tracing sql query

IAMTJW:
This is a tracing filter for Alibaba Druid 1.1.14 (https://github.com/alibaba/druid), support more databases。

### DIFF
--- a/instrumentation/alibaba-druid/README.md
+++ b/instrumentation/alibaba-druid/README.md
@@ -1,0 +1,24 @@
+# brave-instrumentation-alibaba-druid
+
+## Druid Filter
+
+This is a tracing filter for  [Alibaba Druid 1.1.14 ](https://github.com/alibaba/druid)，support more databases(mysql、postgresql、oracle、Microsoft、DB2、sqlite etc.) to tracing sql query and can replace brave-instrumentation-mysql.
+
+## Configuration
+
+    <bean id="tracingStatementFilter" class="brave.druid.TracingStatementFilter">
+	<property name="zipkinServiceName" value="dbServer" /></bean>
+
+  	<bean id="dataSource"  class="com.alibaba.druid.pool.DruidDataSource"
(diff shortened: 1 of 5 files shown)


devinsba:
To remove ambiguity with Apache Druid (incubating) it would be good to change the name of this slightly, perhaps: `brave-instrumentation-druidcp`

adriancole:
tough call as I also thought about Apache druid. most of our libs associate
with the jar name and the only problem (bikeshed as it may be) besides that
is I dont see druid being named druidcp anywhere.. another option would be
to include Alibaba but if this goes to Apache then that would drift.
however.. heh it would have to as druid is already in Apache as you mention
;)

so I guess I would suggest adding alibaba into the name unless druid is
called druidcp but I just missed noticing that.

adriancole:
Hi, @IAMTJW I do believe that this feature will be diversely used, but we do have a policy of establishing that before we merge. Otherwise known as rule of three https://github.com/openzipkin/brave/blob/master/HACKING.md#forks-are-welcome

Please find 3 people who will use this feature to 👍 or otherwise comment on this change so we know that people besides you will help support this long term. Often a mix of interest from maintainers and end users are helpful.

IAMTJW:
> Hi, @IAMTJW I do believe that this feature will be diversely used, but we do have a policy of establishing that before we merge. Otherwise known as rule of three https://github.com/openzipkin/brave/blob/master/HACKING.md#forks-are-welcome
> 
> Please find 3 people who will use this feature to  or otherwise comment on this change so we know that people besides you will help support this long term. Often a mix of interest from maintainers and end users are helpful.

It will be useful and Now we use postgresql as database service and we need it to tracing pg sql query.

adriancole:
hi @IAMTJW I understand you are interested in this. I meant to say can you help find others who are also interested?

PR #882, opened 2019-03-21
add alibaba druid support more database tracing sql query

IAMTJW:
This is a tracing filter for Alibaba Druid 1.1.14 (https://github.com/alibaba/druid), support more databases。

### DIFF
--- a/instrumentation/alibaba-druid/README.md
+++ b/instrumentation/alibaba-druid/README.md
@@ -1,0 +1,24 @@
+# brave-instrumentation-alibaba-druid
+
+## Druid Filter
+
+This is a tracing filter for  [Alibaba Druid 1.1.14 ](https://github.com/alibaba/druid)，support more databases(mysql、postgresql、oracle、Microsoft、DB2、sqlite etc.) to tracing sql query and can replace brave-instrumentation-mysql.
+
+## Configuration
+
+    <bean id="tracingStatementFilter" class="brave.druid.TracingStatementFilter">
+	<property name="zipkinServiceName" value="dbServer" /></bean>
+
+  	<bean id="dataSource"  class="com.alibaba.druid.pool.DruidDataSource"
+       init-method="init" destroy-method="close">
+      <property name="url" value="jdbc:derby:memory:spring-test;create=true" />
+      <property name="initialSize" value="1" />
+      <property name="maxActive" value="20" />
+      <property name="proxyFilters">
+          <list>
+              <ref bean="tracingStatementFilter" />
+          </list>
+      </property>
+  	</bean>
+
+By default the service name corresponding to your database uses the format `${database}`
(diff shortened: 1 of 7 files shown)


adriancole:
@IAMTJW maybe you can post this to your QQ group to see if others want this integration? If we have more thumbsup here we can proceed. https://github.com/alibaba/druid/wiki/FAQ#23-questions-and-suggestions-are-welcome

instrumentation/alibaba-druid/README.md
```diff
@@ -0,0 +1,24 @@
+# brave-instrumentation-alibaba-druid
+
+## Druid Filter
+
+This is a tracing filter for  [Alibaba Druid 1.1.14 ](https://github.com/alibaba/druid)，support more databases(mysql、postgresql、oracle、Microsoft、DB2、sqlite etc.) to tracing sql query and can replace brave-instrumentation-mysql.
+
+## Configuration
+
+    <bean id="tracingStatementFilter" class="brave.druid.TracingStatementFilter">
```
adriancole: wrap in ```xml and please format so indents are consistent

instrumentation/alibaba-druid/pom.xml
```diff
@@ -0,0 +1,76 @@
...
+        <dependency>
+            <groupId>com.alibaba</groupId>
+            <artifactId>druid</artifactId>
+            <version>1.1.14</version>
+            <scope>provided</scope>
+        </dependency>
+
+        <dependency>
+            <groupId>mysql</groupId>
+            <artifactId>mysql-connector-java</artifactId>
```
adriancole: probably choose one in-memory for test scope, such as derby. you can look at p6spy for an example.

instrumentation/alibaba-druid/src/main/java/brave/druid/TracingStatementFilter.java
```diff
@@ -0,0 +1,230 @@
...
+
+    }
+
+
+    @Override
+    protected void statementExecuteUpdateBefore(StatementProxy statement, String sql) {
+
+        super.statementExecuteUpdateBefore(statement,sql);
+
+        try {
```
adriancole: there are some concerns in this file. I can see you are doing some defensive code, but possibly there are ways to do this in less code. for example, if your before and after blocks did the defense, the delegation methods would not need to.

instrumentation/alibaba-druid/pom.xml
IAMTJW: thanks

instrumentation/alibaba-druid/src/main/java/brave/druid/TracingStatementFilter.java
IAMTJW: ok

instrumentation/alibaba-druid/README.md
IAMTJW: thanks

wangrangzhi:
wonderful support for sql！

adriancole:
I currently see two people interested in this. We need to see more to justify the review of this change and future maintenance.

adriancole:
in over one year only 2 people say interest. Unless we hear more I will close at the end of the week. @IAMTJW sorry just we have to make sure we use time maintaining things that a lot of people use.

jorgheymans:
If really still interested they could also develop and maintain it as a standalone module in https://github.com/openzipkin-contrib perhaps.

adriancole:
agree @jorgheymans , but even there should be at least 3 people I think as setting up the project is even more work than another module here.

Issue #886, opened 2019-03-29
TracingKafkaClientSupplier public constructor.


neetkee:
I'm trying to instrument my Spring Kafka Streams application with Brave. I can't directly create KafkaStreams object, because it creates internally in spring,  but I can add a KafkaClientSupplier to it.
However, [TracingKafkaClientSupplier](https://github.com/openzipkin/brave/blob/65e2909533eb504f8da6485f81aa823aa7b080ab/instrumentation/kafka-streams/src/main/java/brave/kafka/streams/TracingKafkaClientSupplier.java#L23) has no public constructor. Can you provide a way to instantiate this object, maybe via builder or something else?

I've created an **example project** to this issue. You can see details of streams configuration in this file: https://github.com/neetkee/brave-example/blob/master/src/main/java/com/neetkee/example/ExampleConfig.java

adriancole:
possibly a method like this added to `KafkaStreamsTracing` would fit in best, as it would allow us to adjust internals later with the same component configuration (`KafkaStreamsTracing.Builder`) 

```java
/** some important docs */
// exposed as Spring Kafka Streams controls creation of the KafkaStreams object, but accepts KafkaClientSupplier
public KafkaClientSupplier clientSupplier();
```

would that work? cc @ImFlog @jeqo

jeqo:
Agree, I'm in favor of exposing this as part of the `KafkaStreamsTracing` API. It will make this available for other frameworks with same constraints as well.

I'd be happy to work on this or help @neetkee if he already has made some progress.

jeqo:
Fixed by #887

PR #887, opened 2019-03-29
Kafka client supplier

jeqo:
Solve #886 

CC @neetkee


neetkee:
thanks! that will help

PR #888, opened 2019-03-30
RateLimitingSampler parallel theory didn't run by default

adriancole:
This fixes it. cc @anuraaga @trustin for advice on how to make
this better.

### DIFF
--- a/brave/src/test/java/brave/sampler/RateLimitingSamplerSoakTest.java
+++ b/brave/src/test/java/brave/sampler/RateLimitingSamplerSoakTest.java
@@ -1,0 +1,88 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one or more
+ * contributor license agreements.  See the NOTICE file distributed with
+ * this work for additional information regarding copyright ownership.
+ * The ASF licenses this file to You under the Apache License, Version 2.0
+ * (the "License"); you may not use this file except in compliance with
+ * the License.  You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package brave.sampler;
+
+import java.util.ArrayList;
+import java.util.List;
+import java.util.concurrent.ExecutorService;
+import java.util.concurrent.Executors;
+import java.util.concurrent.Future;
+import java.util.concurrent.TimeUnit;
+import java.util.concurrent.atomic.AtomicBoolean;
+import java.util.concurrent.atomic.AtomicLong;
+import org.junit.experimental.theories.DataPoints;
+import org.junit.experimental.theories.Theories;
+import org.junit.experimental.theories.Theory;
+import org.junit.runner.RunWith;
+
+import static com.google.common.util.concurrent.Uninterruptibles.sleepUninterruptibly;
+import static org.assertj.core.api.Assertions.assertThat;
+import static org.assertj.core.api.Assumptions.assumeThat;
+
+@RunWith(Theories.class)
+public class RateLimitingSamplerSoakTest {
+
+  @DataPoints public static final int[] SAMPLE_RESERVOIRS = {1, 11, 101, 1001, 1_000_001};
+
+  /** This test will take a little over a second per reservoir */
+  @Theory public void retainsPerSampleRate(int reservoir) throws Exception {
+    Sampler sampler = RateLimitingSampler.create(reservoir);
+
+    // We want to make sure we fill up the entire second, so
+    long startTick = System.nanoTime();
+    long lastDecisecond = startTick + TimeUnit.MILLISECONDS.toNanos(990);
+    // Because exacts don't work with Thread.sleep + operation overhead, set a deadline just under a second
+    long deadline = startTick + TimeUnit.MILLISECONDS.toNanos(998);
+
+    AtomicBoolean hitLastDecisecond = new AtomicBoolean();
+
+    AtomicLong passed = new AtomicLong();
+    Runnable sample = () -> {
+      long tick = System.nanoTime();
+      if (tick > deadline) return;
+      if (tick >= lastDecisecond) hitLastDecisecond.set(true);
+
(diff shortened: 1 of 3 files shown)


brave/src/test/java/brave/sampler/RateLimitingSamplerTest.java
```diff
@@ -92,17 +88,6 @@
     assertThat(sampler.isSampled(0L)).isTrue();
   }
 
-  @DataPoints public static final int[] SAMPLE_RESERVOIRS = {1, 10, 100};
-
-  @Theory public void retainsPerSampleRate(int reservoir) {
```
adriancole: this didn't run as it needed a test runner. I guess we added the powermock runner later and didn't notice

brave/src/test/java/brave/sampler/RateLimitingSamplerSoakTest.java
```diff
@@ -0,0 +1,66 @@
...
+import static org.assertj.core.api.Assertions.assertThat;
+import static org.assertj.core.api.Assumptions.assumeThat;
+
+@RunWith(Theories.class)
+public class RateLimitingSamplerSoakTest {
+
+  @DataPoints public static final int[] SAMPLE_RESERVOIRS = {11, 101, 1001, 1_000_001};
+
+  /** This test will take a little over a second per reservoir */
+  @Theory public void retainsPerSampleRate(int reservoir) throws Exception {
```
adriancole: I refactored this a little bit as I noticed our previous test didn't guarantee it covered the second. Still I think maybe there is a hidden problem we aren't able to see, yet.

advice on how to test this better welcome!

brave/src/test/java/brave/sampler/RateLimitingSamplerSoakTest.java
```diff
@@ -0,0 +1,66 @@
...
+    // Because exacts don't work with Thread.sleep + operation overhead, set a deadline just under a second
+    long deadline = startTick + TimeUnit.MILLISECONDS.toNanos(998);
+
+    AtomicBoolean hitLastDecisecond = new AtomicBoolean();
+
+    AtomicLong passed = new AtomicLong();
+    Runnable sample = () -> {
+      long tick = System.nanoTime();
+      if (tick > deadline) return;
+      if (tick >= lastDecisecond) hitLastDecisecond.set(true);
```
adriancole: note: this doesn't ensure we took a sample in the last decisecond, only that we executed within it..

brave/src/test/java/brave/sampler/RateLimitingSamplerSoakTest.java
anuraaga: Would it help to inject a clock? I notice the guava tests are easy to reason about since they don't use real time. They probably run way faster too

https://github.com/google/guava/blob/0cd4e9faa1360da4a343f84cb275d6eda0c5e732/android/guava-tests/test/com/google/common/util/concurrent/RateLimiterTest.java

adriancole:
not really thought about a concurrency test with a fake clock.. we have a
separate test case for unit tests.. maybe the boundaries would be easier to
test but then again they are already tested with unit tests. perhaps there
is no reason to be concerned about the boundaries and the logic here isnt
that important..

one reason I looked into this is that during voltdb testing I noticed the
rate wasnt honored properly when saturated and trying to figure out where
it breaks. we got too many samples.. more than the rate.

iotw, I used the synchronized counting impl and received exact results but
using this I got over the rate expected. I was looking to see if there's a
way to detect a concurrency flaw in our impl which visual looking and unit
tests haven't yet caught. I not sure it is a big deal but yeah...

brave/src/test/java/brave/sampler/RateLimitingSamplerSoakTest.java
```diff
@@ -0,0 +1,66 @@
...
+
+      if (sampler.isSampled(0L)) passed.incrementAndGet();
+    };
+
+    ExecutorService service = Executors.newFixedThreadPool(10);
+    List<Future<?>> futures = new ArrayList<>();
```
anuraaga: I think the timing might be more reliable with ten loops instead of one since submitting to an executor has reasonable overhead, something like

```java
futures.add(service.submit(() -> {
  do {
    sample();
    Thread.sleep(9);
  } while (System.nanoTime() < deadline);
});
```

brave/src/test/java/brave/sampler/RateLimitingSamplerSoakTest.java
```diff
@@ -0,0 +1,66 @@
...
+      }
+
+      for (Future<?> future : futures) future.get();
+      futures.clear(); // block instead of filling up the executor
+      Thread.sleep(9);
+    } while (System.nanoTime() < deadline);
+    service.shutdown();
+    service.awaitTermination(1, TimeUnit.SECONDS);
+
+    assertThat(passed.get()).isEqualTo(reservoir);
```
anuraaga: I'm surprised it's possible to have an exact match here given timing issues due to thread scheduling. Is it ok to do an exact match?

adriancole: this is why the threads skip out if they notice past a deadline

brave/src/test/java/brave/sampler/RateLimitingSamplerSoakTest.java
adriancole: yeah I like this better

adriancole: great! this finally breaks now (without me having to run a soak test externally)

adriancole:
ok breaks locally now, which is a good thing. There's some opportunity for extra samples which we aren't catching..

adriancole:
maybe I'm thinking too much, but I wonder if we shouldn't have the decisecond sampler implemented as 10 normal ones. An outer loop chooses which of the 10 normal ones we are in. these ones would refresh their permits when they recognize they are in a new second (as each will be abandoned for 9/10ths of a second)

adriancole:
ok I will re-integrate this (manually) with voltdb now and see if it is any better..

adriancole:
ps I'm currently unconvinced that there's an issue anymore in this. I currently get a rate lower than expected, but the voltdb impl is working off a processing queue that lags, which might explain it. I will try with saturating a brave direct example.

adriancole:
using a brave-specific saturation test over 100 seconds, I noticed slightly less traces leaked but not significantly. In other words, the bug caught here is extremely minor.

PR #894, opened 2019-04-08
Upgrade Kafka dependencies to 2.2.0

### LINKED ISSUE #893: Upgrade Kafka dependencies to 2.2.0
**Feature**

Update `kafka-clients` and `kafka-streams` dependencies to version 2.2.0.

**Rationale**

Support latest methods added to Producer and Consumer API, and test that new transformer added to Kafka Streams `KStream#flatTransform` works with current implementation.

jeqo:
Fixes #893

### DIFF
--- a/instrumentation/kafka-clients/src/main/java/brave/kafka/clients/TracingProducer.java
+++ b/instrumentation/kafka-clients/src/main/java/brave/kafka/clients/TracingProducer.java
@@ -131,7 +131,8 @@ final class TracingProducer<K, V> implements Producer<K, V> {
     delegate.close(timeout, unit);
   }
 
-  @Override public void close(Duration duration) {
+  // Do not use @Override annotation to avoid compatibility issue version < 2.0
+  public void close(Duration duration) {
     delegate.close(duration);
   }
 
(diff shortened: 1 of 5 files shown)


jeqo:
Waiting for https://github.com/charithe/kafka-junit/pull/42 to be merged to update snapshot dependency.

instrumentation/kafka-clients/src/main/java/brave/kafka/clients/TracingProducer.java
```diff
@@ -130,6 +131,10 @@
     delegate.close(timeout, unit);
   }
 
+  @Override public void close(Duration duration) {
```
adriancole: guessing this was prior java 8 bytecode anyway, ya? I think you might want to remove Override though. 

See "Tracing multiple versions of an interface with one instrumentation class" here https://cwiki.apache.org/confluence/display/ZIPKIN/Java++Instrumentation+Techniques

it would be cool to have an invoker test like grpc to prove older versions still work

jeqo: True, I forgot this. 

We have and `it` to test this module against `kafka-clients:1.0.0` but is not triggering an exception when `@Override`. Any clue on how to test this?

adriancole: maybe it is because we aren't compiling against that method in the IT? it is a good question!

PR #896, opened 2019-04-21
feat: use p6spy's log filtering settings when tracing

high-stakes:
p6spy provides sql query filtering options which were not used during span
creation, added support for them.

p6spy tends to produce enormous amount of logs due to including all queries, sending this huge amount of spans is usually counterproductive ,adds noise to the analysis and increases costs

### DIFF
--- a/instrumentation/p6spy/README.md
+++ b/instrumentation/p6spy/README.md
@@ -42,3 +42,11 @@ jdbc:mysql://127.0.0.1:3306/mydatabase?zipkinServiceName=myServiceName
 This will override the `remoteServiceName` set in `spy.properties`.
 
 The current tracing component is used at runtime. Until you have instantiated `brave.Tracing`, no traces will appear.
+
+### Filtering spans
+
+By default, all statements are recorded as client spans. 
+You may wish to exclude statements like `set session` from tracing. This library reuses p6spy's log filtering for this purpose.
+Filtering options are picked up from `spy.properties`, so you can blacklist/whitelist what type of statements to record as spans.
+
+For configuration details please see p6spy's log filtering documentation.
--- a/instrumentation/p6spy/src/main/java/brave/p6spy/TracingP6SpyOptions.java
+++ b/instrumentation/p6spy/src/main/java/brave/p6spy/TracingP6SpyOptions.java
@@ -13,25 +13,47 @@
  */
 package brave.p6spy;
 
+import com.p6spy.engine.logging.P6LogLoadableOptions;
+import com.p6spy.engine.logging.P6LogOptions;
 import com.p6spy.engine.spy.P6SpyOptions;
 import com.p6spy.engine.spy.option.P6OptionsRepository;
+
+import java.util.LinkedHashMap;
 import java.util.Map;
 
 final class TracingP6SpyOptions extends P6SpyOptions {
+
   static final String REMOTE_SERVICE_NAME = "remoteServiceName";
   static final String INCLUDE_PARAMETER_VALUES = "includeParameterValues";
 
-  final P6OptionsRepository optionsRepository;
+  private final P6OptionsRepository optionsRepository;
+  private final P6LogLoadableOptions logLoadableOptions;
 
   TracingP6SpyOptions(P6OptionsRepository optionsRepository) {
     super(optionsRepository);
+    logLoadableOptions = new P6LogOptions(optionsRepository);
     this.optionsRepository = optionsRepository;
   }
 
-  @Override public void load(Map<String, String> options) {
+  @Override
+  public void load(Map<String, String> options) {
     super.load(options);
+    logLoadableOptions.load(options);
     optionsRepository.set(String.class, REMOTE_SERVICE_NAME, options.get(REMOTE_SERVICE_NAME));
-    optionsRepository.set(Boolean.class, INCLUDE_PARAMETER_VALUES, options.get(INCLUDE_PARAMETER_VALUES));
+    optionsRepository.set(Boolean.class, INCLUDE_PARAMETER_VALUES,
+        options.get(INCLUDE_PARAMETER_VALUES));
+  }
+
+  @Override
+  public Map<String, String> getDefaults() {
+    Map<String, String> allDefaults = new LinkedHashMap(super.getDefaults());
+    allDefaults.putAll(logLoadableOptions.getDefaults());
+    allDefaults.put(INCLUDE_PARAMETER_VALUES, Boolean.FALSE.toString());
+    return allDefaults;
+  }
+
+  P6LogLoadableOptions getLogOptions() {
+    return logLoadableOptions;
   }
 
   String remoteServiceName() {
@@ -39,7 +61,6 @@ final class TracingP6SpyOptions extends P6SpyOptions {
   }
 
   Boolean includeParameterValues() {
-    Boolean logParameterValues = optionsRepository.get(Boolean.class, INCLUDE_PARAMETER_VALUES);
-    return logParameterValues == null ? false : logParameterValues;
+    return optionsRepository.get(Boolean.class, INCLUDE_PARAMETER_VALUES);
   }
 }
--- a/instrumentation/p6spy/src/test/java/brave/p6spy/TracingJdbcEventListenerTest.java
+++ b/instrumentation/p6spy/src/test/java/brave/p6spy/TracingJdbcEventListenerTest.java
@@ -23,6 +23,10 @@ import java.sql.Connection;
 import java.sql.DatabaseMetaData;
 import java.sql.SQLException;
 import java.util.ArrayList;
+
+import com.p6spy.engine.logging.P6LogOptions;
+import com.p6spy.engine.spy.option.P6OptionsRepository;
+import org.junit.Before;
 import org.junit.Test;
 import org.junit.runner.RunWith;
 import org.mockito.Mock;
@@ -39,7 +43,7 @@ public class TracingJdbcEventListenerTest {
   @Mock Connection connection;
   @Mock DatabaseMetaData metaData;
   @Mock StatementInformation statementInformation;
-  @Mock ConnectionInformation ci;
+  @Mock ConnectionInformation connectionInformation;
 
   @Mock Span span;
   String url = "jdbc:mysql://1.2.3.4:5555/mydatabase";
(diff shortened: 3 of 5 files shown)


jorgheymans:
:+1:

perhaps missing is a small note in the readme that this is now supported ?

instrumentation/p6spy/src/main/java/brave/p6spy/TracingP6SpyOptionsMBean.java
```diff
@@ -0,0 +1,11 @@
+package brave.p6spy;
+
+import com.p6spy.engine.logging.P6LogLoadableOptions;
+
+public interface TracingP6SpyOptionsMBean {
+    P6LogLoadableOptions getLogOptions();
```
adriancole: is this a stable api?

instrumentation/p6spy/src/test/java/brave/p6spy/TracingJdbcEventListenerTest.java
```diff
@@ -10,36 +10,48 @@
...
 import org.junit.runner.RunWith;
 import org.mockito.Mock;
 import org.mockito.junit.MockitoJUnitRunner;
 
 import static brave.p6spy.ITTracingP6Factory.tracingBuilder;
 import static org.assertj.core.api.Assertions.assertThat;
-import static org.mockito.Mockito.verify;
-import static org.mockito.Mockito.verifyNoMoreInteractions;
-import static org.mockito.Mockito.when;
+import static org.mockito.Mockito.*;
```
adriancole: please use square formatter it will protect from accidental wildcards etc

adriancole:
ps this looks like 2 changes JMX and also the logging thing

instrumentation/p6spy/src/main/java/brave/p6spy/TracingP6SpyOptions.java
```diff
@@ -1,32 +1,61 @@
...
 import com.p6spy.engine.spy.P6SpyOptions;
 import com.p6spy.engine.spy.option.P6OptionsRepository;
+
+import javax.management.StandardMBean;
+import java.util.Collections;
+import java.util.HashMap;
 import java.util.Map;
 
-final class TracingP6SpyOptions extends P6SpyOptions {
+public class TracingP6SpyOptions extends StandardMBean implements P6LoadableOptions, TracingP6SpyOptionsMBean {
```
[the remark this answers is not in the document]
high-stakes: in line with the other options classes of p6spy

instrumentation/p6spy/src/main/java/brave/p6spy/TracingP6SpyOptions.java
```diff
@@ -1,32 +1,61 @@
...
-    super.load(options);
+    logLoadableOptions.load(options);
     optionsRepository.set(String.class, REMOTE_SERVICE_NAME, options.get(REMOTE_SERVICE_NAME));
     optionsRepository.set(Boolean.class, INCLUDE_PARAMETER_VALUES, options.get(INCLUDE_PARAMETER_VALUES));
   }
 
-  String remoteServiceName() {
+  @Override
+  public Map<String, String> getDefaults() {
+    Map<String, String> allDefaults = new HashMap();
```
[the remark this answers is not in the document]
high-stakes: any specific reason or just preference?

instrumentation/p6spy/src/main/java/brave/p6spy/TracingP6SpyOptionsMBean.java
```diff
@@ -0,0 +1,11 @@
+package brave.p6spy;
+
+import com.p6spy.engine.logging.P6LogLoadableOptions;
+
+public interface TracingP6SpyOptionsMBean {
```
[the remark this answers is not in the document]
high-stakes: not sure, I followed the p6spy convention

instrumentation/p6spy/src/main/java/brave/p6spy/TracingP6SpyOptionsMBean.java
high-stakes: p6spy requires all options classes to have an mbean interface, previously this was implicitly already added because P6SpyOptions was extended (I think incorrectly because none of its options were used) and it already exposed an mbean interface

high-stakes:
> another option is a finished span handler that drops these. such wouldnt be as subject to api drift. you wouldnt need to change the code here at all or conflate logging with span recording..

Sounds cool, maybe I could add both? I will look into this on the weekend

instrumentation/p6spy/src/main/java/brave/p6spy/TracingP6SpyOptions.java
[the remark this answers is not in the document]
adriancole: nothing should be public unless it is required to. each public type adds a possibility for us to break api. there's a place in effective java about this, but it is something followed (or accidentally not followed) in this codebase even if p6spy doesn't follow the same coding rules we do. In this repo, we prefer our coding rules.

instrumentation/p6spy/src/main/java/brave/p6spy/TracingP6SpyOptionsMBean.java
adriancole: I mean to as if this P6LogLoadableOptions has been around for a while.. ex what version added etc. If we add this we could break others. Not all libraries track compatibility.

adriancole:
FYI if adding a change to directly implement JMX, such a PR will need a unit test to prove what we support changing can be done via JMX.

regardless, thanks for the help!

high-stakes:
> FYI if adding a change to directly implement JMX, such a PR will need a unit test to prove what we support changing can be done via JMX.
> 
> regardless, thanks for the help!

instrumentation/p6spy/src/main/java/brave/p6spy/TracingP6SpyOptionsMBean.java
high-stakes: hasn't been broken since the initial github release in 2013, 2.0.0

high-stakes:
> FYI if adding a change to directly implement JMX, such a PR will need a unit test to prove what we support changing can be done via JMX.
> 
> regardless, thanks for the help!

I mean the previous usage was incorrect (it would do nothing best case, worst case replace the officially registered MBean for P6SpyOptions which is always implicitly loaded), I just removed it to avoid bugs. Is it enough if i add an empty interface and not expose any methods?

instrumentation/p6spy/src/main/java/brave/p6spy/TracingP6SpyOptions.java
```diff
@@ -1,32 +1,58 @@
...
+  private final P6OptionsRepository optionsRepository;
+  private final P6LogLoadableOptions logLoadableOptions;
+
+  static {
+    defaults.put(INCLUDE_PARAMETER_VALUES, Boolean.FALSE.toString());
+  }
 
-  TracingP6SpyOptions(P6OptionsRepository optionsRepository) {
-    super(optionsRepository);
+  public TracingP6SpyOptions(P6OptionsRepository optionsRepository) {
```
adriancole: revert public here. note: please use the format on all files changed as I notice some missing or too many spaces around. if you look at the diff you can notice.

instrumentation/p6spy/src/main/java/brave/p6spy/TracingP6SpyOptions.java
```diff
@@ -1,32 +1,58 @@
...
     optionsRepository.set(String.class, REMOTE_SERVICE_NAME, options.get(REMOTE_SERVICE_NAME));
-    optionsRepository.set(Boolean.class, INCLUDE_PARAMETER_VALUES, options.get(INCLUDE_PARAMETER_VALUES));
+    optionsRepository.set(Boolean.class, INCLUDE_PARAMETER_VALUES,
+        options.get(INCLUDE_PARAMETER_VALUES));
+  }
+
+  @Override
+  public Map<String, String> getDefaults() {
+    Map<String, String> allDefaults = new LinkedHashMap();
+    allDefaults.putAll(defaults);
```
adriancole: inline`defaults` vs a static initializes for the hashmap which only has one attribute.

instrumentation/p6spy/src/main/java/brave/p6spy/TracingP6SpyOptions.java
```diff
@@ -1,32 +1,58 @@
...
+    allDefaults.putAll(logLoadableOptions.getDefaults());
+    return allDefaults;
+  }
+
+  public P6LogLoadableOptions getLogOptions() {
```
adriancole: PS If you want to add an interface, just make the interface itself package private. Then, you can add `@Override` tag for this and below. As long as types are package private, we're good. 

f we are broken on JMX and don't know it because there is no test, a good fix would be to add a test vs us guessing if it is broke or not. I suspect there's code out there somewhere that has a unit test for JMX registration even if we don't have any in this repo. I'd consider this part optional for this PR even if we are broken as ensuring JMX works can be separate

instrumentation/p6spy/src/main/java/brave/p6spy/TracingP6SpyOptionsMBean.java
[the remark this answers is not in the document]
adriancole: either remove this or add a unit test to prove it works. right now it is a distracting file

high-stakes:
> as p6spy can work from properties (it has so far) please either remove all JMX related changes or add a unit test to show that they work.
> 
> separately, I also left comments with some style-related fuzz. The core change here about log options is done, it is just in the middle of a bunch of other things about JMX which are incomplete and distracting. If you can revert that into a different PR that has a unit test to show it works, we'll all save time.

Makes sense, added the changes and reverted the jmx part.

high-stakes:
Hi @adriancole , could you merge this? The changes discussed are added. Thanks

instrumentation/p6spy/README.md
```diff
@@ -42,3 +42,10 @@ jdbc:mysql://127.0.0.1:3306/mydatabase?zipkinServiceName=myServiceName
 This will override the `remoteServiceName` set in `spy.properties`.
 
 The current tracing component is used at runtime. Until you have instantiated `brave.Tracing`, no traces will appear.
+
+### Log filtering
```
adriancole:
```suggestion
### Filtering spans
```

instrumentation/p6spy/README.md
```diff
@@ -42,3 +42,10 @@ jdbc:mysql://127.0.0.1:3306/mydatabase?zipkinServiceName=myServiceName
...
+
```
adriancole:
```suggestion

By default, all statements are recorded as client spans. You may wish to exclude statements like `set session` from tracing. This library reuses p6spy's log filtering for this purpose.

```

instrumentation/p6spy/README.md
```diff
@@ -42,3 +42,10 @@ jdbc:mysql://127.0.0.1:3306/mydatabase?zipkinServiceName=myServiceName
...
+The instrumentation also supports picking up p6spy's log filtering options from `spy.properties`, so you can filter 
+what type of statements to record.
```
adriancole:
```suggestion
what type of statements to record as spans.
```

high-stakes:
done!

adriancole:
many thanks @high-stakes and @jorgheymans for the review

Issue #899, opened 2019-04-23
support for filter() in KafkaStreamsTracing


jorgheymans:
Similar to #889 , it would be nice if the filter operation would be made observable.

Since a filter operation is not a `Processor` but a `Predicate`, it might require a bit more work as current tracing code uses the `ProcessorContext` to access the current trace headers. Any ideas ? I'ld be happy to pilot a PR given some coaching :-) @jeqo

jeqo:
@jorgheymans thanks for taking this one!

We can access headers only by using `Processor` or `Transformer` interfaces, not by "wrapping" existing DSL operations (e.g. KStream#filter). The approach we have choose is to re-implement most common operations as map/peek/foreach.
We can do the same for filter, if you check the internals, filter is implemented by: https://github.com/apache/kafka/blob/trunk/streams/src/main/java/org/apache/kafka/streams/kstream/internals/KStreamFilter.java

You could use https://github.com/openzipkin/brave/blob/master/instrumentation/kafka-streams/src/main/java/brave/kafka/streams/KafkaStreamsTracing.java#L221 as guidance to add a filter transfomer. In this case we will receive the Predicate instead of the Mapper as input. Looking forward to your PR!

jorgheymans:
Thanks for the pointers, that's helpful. I'll have a go at implementing this.

Do you think it's useful to tag the span with the result of the filtering (ie filtered=true/false) ?

jeqo:
I think so! you could add a tag: `kafka.streams.filtered`. Without it will be hard to understand why is the last span.

PR #901, opened 2019-04-25
implement tracing streams.filter()

jorgheymans:
fixes https://github.com/openzipkin/brave/issues/899


instrumentation/kafka-streams/src/main/java/brave/kafka/streams/TracingPredicate.java
```diff
@@ -0,0 +1,43 @@
...
+  TracingPredicate(KafkaStreamsTracing tracing, String spanName,
+      Predicate<K, V> delegatePredicate) {
+    this.kafkaStreamsTracing = tracing;
+    this.spanName = spanName;
+    this.delegatePredicate = delegatePredicate;
+  }
+
+  @Override
+  public KeyValue<K, V> transform(K key, V value) {
+    Span span = kafkaStreamsTracing.tracing.tracer().nextSpan();
```
jorgheymans: @jeqo initially i was using `kafkaStreamsTracing.nextSpan(processorContext)` here but Kafka did not like it. `AbstractProcessorContext` barfed with `throw new IllegalStateException("This should not happen as headers() should only be called while a record is processed");` . Can you shed more light on this ?

instrumentation/kafka-streams/src/main/java/brave/kafka/streams/TracingPredicate.java
```diff
@@ -0,0 +1,43 @@
+package brave.kafka.streams;
+
+import brave.Span;
+import brave.Tracer;
+import org.apache.kafka.streams.KeyValue;
+import org.apache.kafka.streams.kstream.Predicate;
+
+class TracingPredicate<K, V> extends AbstractTracingTransformer<K, V, KeyValue<K, V>> {
```
jeqo: nit: instead of `TracingPredicate` consider `TracingFilterTransformer`, as predicate is just the condition object. 

Similar to https://github.com/apache/kafka/blob/trunk/streams/src/main/java/org/apache/kafka/streams/kstream/internals/KStreamFilter.java#L39

instrumentation/kafka-streams/src/main/java/brave/kafka/streams/TracingPredicate.java
jeqo: Take a look to https://github.com/openzipkin/brave/blob/master/instrumentation/kafka-streams/src/main/java/brave/kafka/streams/TracingTransformer.java#L33. Instrumentation need to access headers, and headers are part of the `processorContext`, initialized on `init`.

instrumentation/kafka-streams/src/main/java/brave/kafka/streams/TracingPredicate.java
```diff
@@ -0,0 +1,43 @@
...
+    if (!span.isNoop()) {
+      span.name(spanName);
+      span.start();
+    }
+
+    try (Tracer.SpanInScope ws = kafkaStreamsTracing.tracing.tracer().withSpanInScope(span)) {
+      if (delegatePredicate.test(key, value)) {
+        return KeyValue.pair(key, value);
+      } else {
+        span.annotate(KafkaStreamsTags.KAFKA_STREAMS_FILTERED_TAG);
```
jeqo: We should be careful about `filter` semantics. I would say a tag with filtered=true/false will be better aligned with KStreams#filter semantics: https://github.com/apache/kafka/blob/trunk/streams/src/main/java/org/apache/kafka/streams/kstream/KStream.java#L59

Also, I'd recommend to have a filterNot variable, similar to https://github.com/apache/kafka/blob/trunk/streams/src/main/java/org/apache/kafka/streams/kstream/internals/KStreamFilter.java#L27 to support `filterNot` as well.

instrumentation/kafka-streams/src/test/java/brave/kafka/streams/ITKafkaStreamsTracing.java
```diff
@@ -208,6 +208,37 @@ public void process(String key, String value) {
     streams.cleanUp();
   }
 
+  @Test
+  public void should_create_spans_from_stream_with_tracing_filter() throws Exception {
```
jeqo: Consider to add counter example as well, `filter=true`.

instrumentation/kafka-streams/src/main/java/brave/kafka/streams/TracingPredicate.java
jorgheymans: :+1:

instrumentation/kafka-streams/src/main/java/brave/kafka/streams/TracingPredicate.java
jorgheymans: > We should be careful about `filter` semantics. I would say a tag with filtered=true/false will be better aligned with KStreams#filter semantics: https://github.com/apache/kafka/blob/trunk/streams/src/main/java/org/apache/kafka/streams/kstream/KStream.java#L59

If we go for `filtered=true|false` that means that every span running through this filter will be tagged. Should we be concerned with the amount of data this generates ? Previously only the records that get filtered out would get annotated, seemed more logical to me.

instrumentation/kafka-streams/src/test/java/brave/kafka/streams/ITKafkaStreamsTracing.java
jorgheymans: :+1:

jorgheymans:
One more thing, TracingFilterTransformer is an AbstractTracingTransformer that handles its own `transform` logic because it needs to tag the span. Because of this, the wrapping in `TracingTransformerSupplier#get()` is perhaps not needed for this case ? It adds an extra span unnecessarily IMO.

instrumentation/kafka-streams/src/main/java/brave/kafka/streams/TracingPredicate.java
jeqo: That will depends on the nature of the stream and messages, in some scenarios number of messages with `filtered=false` > `filtered=true`, others `filtered=true` > `filtered=false`.

Based on this, tagging every span is ok IMO.

jeqo:
> One more thing, TracingFilterTransformer is an AbstractTracingTransformer that handles its own `transform` logic because it needs to tag the span. Because of this, the wrapping in `TracingTransformerSupplier#get()` is perhaps not needed for this case ? It adds an extra span unnecessarily IMO.

True, we need a Supplier for TracingFilterTransformer as well.

instrumentation/kafka-streams/src/main/java/brave/kafka/streams/TracingPredicate.java
jorgheymans: :+1:

jorgheymans:
> True, we need a Supplier for TracingFilterTransformer as well.

Would you be ok with `TracingFilterTransform` implementing `TransformerSupplier` directly or separate class ?

jorgheymans:
The last commit adds the supplier and fixes the integration tests. Also somehow the error about accessing the kafka headers outside of message processing is gone now, maybe it had something to do with the FilterTransformer being wrapped .. not sure.

So from my part this is good to go, let me know if you have any further comments i'd be happy to address them :+1:

PR #904, opened 2019-05-08
Messaging adapter

jeqo:
initial draft for messaging adapter.


adriancole:
cc @jorgheymans @marcingrzejszczak

instrumentation/messaging/src/main/java/brave/messaging/MessagingProducerHandler.java
```diff
@@ -0,0 +1,56 @@
...
+      span = tracer.newChild(maybeParent);
+      adapter.clearPropagation(message);
+    }
+
+    if (!span.isNoop()) {
+      span.kind(Span.Kind.PRODUCER).name("send");
+      parser.message(adapter, message, span);
+      String remoteServiceName = adapter.remoteServiceName(message);
+      if (remoteServiceName != null) span.remoteServiceName(remoteServiceName);
+      span.start();
```
devinsba: Elsewhere we've done: `span.start().finish();` though I'm not sure I like that pattern a ton, since this will end up flushed anyway, maybe we should flush explictly

adriancole:
flush wont set a end timestamp which we need eventually. flushing before
finish will make reporting chatty and analysis more heavy as multiple docs
per span

jeqo:
Return boolean on `remoteServiceName`.

jeqo:
Creating an Adapter and Parser for Consumer and Producer.

jeqo:
Tagging move specific tagging to values: instead of `kafka.topic`=`name` -> `channel.name` = `name` + `channel.type` = `topic` + `channel.protocol` = `kafka`.

instrumentation/messaging/src/main/java/brave/messaging/MessagingParser.java
```diff
@@ -0,0 +1,43 @@
...
+      SpanCustomizer customizer) {
+    String channel = adapter.channel(chan);
+    if (chan != null) customizer.tag(adapter.channelTagKey(chan), channel);
+  }
+
+  public <Msg> void identifier(MessageAdapter<Msg> adapter, Msg message,
+      SpanCustomizer customizer) {
+    String identifier = adapter.identifier(message);
+    if (identifier != null) {
+      customizer.tag(adapter.identifierTagKey(), adapter.identifier(message));
```
llinder: Small nit but I think this could just be `customizer.tag(adapter.identifierTagKey(), identifier);`.  Also should we do a null and empty string check on `adapter.identifierTagKey()` just to ensure a bad adapter doesn't cause issues?

instrumentation/jms/src/main/java/brave/jms/JmsAdapter.java
```diff
@@ -0,0 +1,123 @@
...
+    @Override public String identifier(Message message) {
+      try {
+        return message.getJMSMessageID();
+      } catch (JMSException e) {
+        // don't crash on wonky exceptions!
+      }
+      return null;
+    }
+
+    @Override public void clearPropagation(Message message) {
```
llinder: My personal preference would be to return the `Message`.  Reasoning is that some libraries or coding choices may prefer immutable data structures where copying and returning a new object is required.

instrumentation/messaging/src/main/java/brave/messaging/MessageConsumerAdapter.java
```diff
@@ -0,0 +1,20 @@
...
+
+  /**
+   * Message identifier, e.g. kafka record key, jms message correlation id.
+   */
+  String identifier(Msg message);
+
+  /**
+   * Removes propagation context from Message context carrier.
+   */
+  void clearPropagation(Msg message);
```
llinder: My personal preference would be to return the Message. Reasoning is that some libraries or coding choices may prefer immutable data structures where copying and returning a new object is required.

instrumentation/messaging/src/main/java/brave/messaging/MessageProducerAdapter.java
```diff
@@ -0,0 +1,20 @@
...
+
+  /**
+   * Message identifier, e.g. kafka record key, jms message correlation id.
+   */
+  String identifier(Msg message);
+
+  /**
+   * Removes propagation context from Message context carrier.
+   */
+  void clearPropagation(Msg message);
```
llinder: My personal preference would be to return the Message. Reasoning is that some libraries or coding choices may prefer immutable data structures where copying and returning a new object is required.

instrumentation/messaging/src/main/java/brave/messaging/MessagingConsumerHandler.java
```diff
@@ -0,0 +1,119 @@
...
+      TraceContext.Injector<Msg> injector) {
+    return new MessagingConsumerHandler<>(delegate, tracing, channelAdapter, messageAdapter,
+        extractor, injector);
+  }
+
+  public final C delegate;
+  final Tracing tracing;
+
+  public MessagingConsumerHandler(
+      C delegate,
```
llinder: Are there plans to use `delegate` inside of `MessagingConsumerHandler` in the future?  If not should it even be included here?  Also curious if implementations are expected to extend this class for most cases?  So far it looks like JMS and Kafka examples extend it.  If thats the case maybe it should be abstract to help emphasis intent?

adriancole: I personally don't think we should make anything that hints these should be wrappers as this part isn't important (ex whether we are wrapping something or extending etc)

instrumentation/messaging/src/main/java/brave/messaging/MessageProducerAdapter.java
adriancole: good and even better with a comment reminding that

instrumentation/messaging/src/main/java/brave/messaging/MessagingParser.java
adriancole:
```suggestion
      customizer.tag(adapter.identifierTagKey(), identifier);
```

instrumentation/messaging/src/main/java/brave/messaging/MessagingParser.java
```diff
@@ -0,0 +1,43 @@
...
+  public <Chan> void channel(ChannelAdapter<Chan> adapter, Chan chan,
+      SpanCustomizer customizer) {
+    String channel = adapter.channel(chan);
+    if (chan != null) customizer.tag(adapter.channelTagKey(chan), channel);
+  }
+
+  public <Msg> void identifier(MessageAdapter<Msg> adapter, Msg message,
+      SpanCustomizer customizer) {
+    String identifier = adapter.identifier(message);
+    if (identifier != null) {
```
adriancole: my 2p is that the type of identifier should be its own value. for example, if it is important that the type is correlation id. "message.idenifier_type=correlation" In this way, FinishedSpanHandler can choose to take whatever the identifier is without having to first see what the key could be.. for example, it might never know the name of the key returned here. Stable keys can help with portable data policy iotw

instrumentation/messaging/src/main/java/brave/messaging/MessagingTracing.java
```diff
@@ -0,0 +1,50 @@
...
+  MessagingTracing(Builder builder) {
+    this.tracing = builder.tracing;
+    this.parser = builder.parser;
+  }
+
+  public Tracing tracing() {
+    return tracing;
+  }
+
+  public MessagingParser parser() {
```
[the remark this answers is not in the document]
jeqo: Fixed.

instrumentation/messaging/pom.xml
```diff
@@ -0,0 +1,51 @@
...
+         xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
+         xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 http://maven.apache.org/xsd/maven-4.0.0.xsd">
+  <modelVersion>4.0.0</modelVersion>
+  <parent>
+    <groupId>io.zipkin.brave</groupId>
+    <artifactId>brave-instrumentation-parent</artifactId>
+    <version>5.6.4-SNAPSHOT</version>
+  </parent>
+
+  <artifactId>brave-instrumentation-messaging</artifactId>
```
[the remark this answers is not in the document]
jeqo: added

instrumentation/messaging/pom.xml
```diff
@@ -0,0 +1,51 @@
...
+  <name>Brave Instrumentation: Messaging</name>
+
+  <properties>
+    <main.basedir>${project.basedir}/../..</main.basedir>
+    <main.java.version>1.8</main.java.version>
```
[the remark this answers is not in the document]
jeqo: fixed.

instrumentation/messaging/pom.xml
```diff
@@ -0,0 +1,51 @@
...
+    <main.signature.artifact>java18</main.signature.artifact>
+
+    <!-- disable errorprone override warning as we do this intentionally to allow old clients -->
```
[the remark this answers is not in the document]
jeqo: removed.

instrumentation/messaging/pom.xml
```diff
@@ -0,0 +1,51 @@
...
+      <plugin>
+        <groupId>net.orfjackal.retrolambda</groupId>
+        <artifactId>retrolambda-maven-plugin</artifactId>
+        <executions>
+          <execution>
+            <phase>none</phase>
+          </execution>
+        </executions>
+      </plugin>
+      <plugin>
```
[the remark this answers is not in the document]
jeqo: done

jeqo:
@llinder I'm wondering if the comments on `cleanPropagation` still holds after considering [this idea from pow-wow](https://cwiki.apache.org/confluence/pages/viewpage.action?pageId=113709067):

> One idea is that we use only idempotent header formats. meaning that there are no headers that need to be read before written. ex lower b3 single format

My understanding of idempotency in this case is that we "don't care" about cleaning context carrier, but just override value if already exist. 

If this is correct, then we should not be a `clearPropagation`; or at least not as part of the Parser API. 
For instance, I know that for JMS we actually *need* to clear message properties before updating value, otherwise it fails with read-only properties. In this case we would have to clear on injector setter, I think.

I'd like your feedback on:
- Does this interpretation of idempotency make sense to everyone?
- Should we constraint to B3 single header format moving forward?

adriancole:
@jeqo I strongly prefer we gate this on clean formats and use that to our advantage. for many sites this will mean b3 single only. for those using extra fields we may want to define something like a single header form that concatenates map values similar to how Amazon and w3c tracestate does. we can take extra field propagation out of scope until solved

adriancole:
I think we need to add a copy of sqs instrumentation perhaps in the src/test tree in order to verify the design and then remove it later. rule of 3 of abstraction.. and we know there were some kinks. if it is vendored into the src/test tree we can easily collaborate on it and change the design. @llinder if you have some time to move your wip here much oblidged

jeqo:
@adriancole I could recreate this PR with the Apache branch instead of my personal to enable easier collaboration, WDYT?

adriancole:
ah yes.. I forgot this wasn't already the case.

jeqo:
following discussion here: #914

PR #907, opened 2019-05-11
Reuse buffer when converting long to hex.

### LINKED ISSUE #905: reduce amount of array allocation
in `B3SingleFormat`, we do something like https://github.com/apache/incubator-zipkin/pull/2582 to reduce array allocations. However, we don't at `TagContextBinaryMarshaller.readAsciiString` and `HexCodec.toLowerHex`.

Note: we don't at `TraceContext.toString` and `TraceIdContext.toString` either, but these I don't think are called in the critical path of anything.

cc @anuraaga

anuraaga:
Fixes #905

### DIFF
--- a/brave/src/main/java/brave/internal/HexCodec.java
+++ b/brave/src/main/java/brave/internal/HexCodec.java
@@ -52,9 +52,9 @@ public final class HexCodec {
 
   /** Inspired by {@code okio.Buffer.writeLong} */
   public static String toLowerHex(long v) {
-    char[] data = new char[16];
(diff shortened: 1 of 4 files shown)


anuraaga:
Ah - looks like tests here mock out `Platform`. As this isn't really platform-dependent, I'm thinking of a separate class `RecycledBuffers` to avoid changing a lot. Does that make sense?

adriancole:
it is up to you.. we aren't using Platform always in platform specific ways.. what you mention is more consistent and intuitive I think.

PR #908, opened 2019-05-11
Fixes some edge cases around sampler resetting

adriancole:
I found these working on https://github.com/adriancole/zipkin-voltdb/pull/18

I don't have unit tests for the changes, but integrated looks a bit right now.

I think @pburm might be interested.

### DIFF
--- a/brave/src/main/java/brave/sampler/RateLimitingSampler.java
+++ b/brave/src/main/java/brave/sampler/RateLimitingSampler.java
@@ -132,8 +132,9 @@ public class RateLimitingSampler extends Sampler {
     }
 
     @Override int max(long nanosUntilReset) {
-      // Check to see if we are in the first interval
+      // Check to see if we are in the first or last interval
       if (nanosUntilReset > NANOS_PER_SECOND - NANOS_PER_DECISECOND) return max[0];
+      if (nanosUntilReset < NANOS_PER_DECISECOND) return max[9];
 
       // Choose a slot based on the remaining deciseconds
       int decisecondsUntilReset = (int) (nanosUntilReset / NANOS_PER_DECISECOND);
--- a/brave/src/main/java/brave/sampler/RateLimitingSampler.java
+++ b/brave/src/main/java/brave/sampler/RateLimitingSampler.java
@@ -92,7 +92,6 @@ public class RateLimitingSampler extends Sampler {
   }
 
   static abstract class MaxFunction {
-    /** @param nanosUntilReset zero if was just reset */
     abstract int max(long nanosUntilReset);
   }
 
@@ -104,7 +103,7 @@ public class RateLimitingSampler extends Sampler {
       this.tracesPerSecond = tracesPerSecond;
     }
 
-    @Override int max(long nanosUntilReset) {
+    @Override int max(long nanosUntilResetIgnored) {
       return tracesPerSecond;
     }
   }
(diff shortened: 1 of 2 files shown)


adriancole:
I used https://github.com/adriancole/zipkin-voltdb to integration test this and sent 4 bursts of 10 seconds of 3K+ requests/second. Using a throttled rate of 20/second it ended up quite close to correct (778 instead of 800). Note most of the gap seems warmup related and I don't completely understand that. Ex the last burst had 196 traces (20/second * 10 seconds) so quite close.

Screen Shot 2019-05-11 at 10 42 36 PM

(note: this is based on export/tail sampling.. we should probably bundle this in zipkin lib so we don't have to copy/paste for collectors to use it)

brave/src/main/java/brave/sampler/RateLimitingSampler.java
```diff
@@ -55,14 +55,17 @@ public static Sampler create(int tracesPerSecond) {
...
     long nanosUntilReset = -(now - updateAt); // because nanoTime can be negative
-    boolean shouldReset = nanosUntilReset <= 0;
-    if (shouldReset) {
-      if (nextReset.compareAndSet(updateAt, updateAt + NANOS_PER_SECOND)) {
-        usage.set(0);
-      }
+    if (nanosUntilReset <= 0) {
+      // Attempt to move into the next sampling interval.
+      // nanosUntilReset is now invalid regardless of race winner, so we can't sample based on it.
+      if (nextReset.compareAndSet(updateAt, updateAt + NANOS_PER_SECOND)) usage.set(0);
```
pburm: Only hole that I see here is that since we are ONLY adding one second to updateAt, when the sampler is long lived and there have not been any sampling decisions in several seconds......it will sample at 100% until updateAt has caught up. We forked at added + -(nanosUntilReset) in addition to NANOS_PER_SECOND when we update updateAt. That might push it too far for somebody's taste, but it cured the 100% sampling in our implementation.

devinsba: The recursion below should also solve that, but having not seen that problem (`updateAt` being more than a second in the past) in my first reading, the recursion is now screaming stack overflow to me so I think that needs to be addressed

adriancole: test case should cover this.. thanks for pointing out the potential

adriancole: I think maybe the below is the easier way to get to the point (we should be working off "now"), I will also add a unit test so recursion worries float away.

```diff
--- a/brave/src/main/java/brave/sampler/RateLimitingSampler.java
+++ b/brave/src/main/java/brave/sampler/RateLimitingSampler.java
@@ -74,7 +74,7 @@ public class RateLimitingSampler extends Sampler {
     if (nanosUntilReset <= 0) {
       // Attempt to move into the next sampling interval.
       // nanosUntilReset is now invalid regardless of race winner, so we can't sample based on it.
-      if (nextReset.compareAndSet(updateAt, updateAt + NANOS_PER_SECOND)) usage.set(0);
+      if (nextReset.compareAndSet(updateAt, now + NANOS_PER_SECOND)) usage.set(0);
 
       // recurse as it is simpler than resetting all the locals.
       // reset happens once per second, this code doesn't take a second, so no infinite recursion.
```

adriancole: just to make sure my jetlag math was ok, I added this assert temporarily and it passed :)

`assert now == updateAt + -nanosUntilReset;`

adriancole:
Thanks @devinsba and @pburm for finding some bugs!

adriancole:
here's a change that uses this https://github.com/adriancole/zipkin-voltdb/pull/24

PR #910, opened 2019-05-16
rename test helper method

jorgheymans:
fixes https://github.com/apache/incubator-zipkin-brave/issues/902


adriancole:
as they say in belgium, gracias!

PR #916, opened 2019-05-31
WIP: refine secondary sampling feature test to be realistic

adriancole:
@narayaruna is one of the first folks to look at this code carefully. I
think the intent wasn't clear. I'm changing the example to look more
realistic and also include dynamic configuration hints.


adriancole:
ps later not tonight I will try to work this into sleuth-webmvc-example

PR #917, opened 2019-06-02
Refactors to reduce injector complexity and duplicate code

adriancole:
this adds messaging processor code and some cleanups. there's
more to do and tests don't pass.


instrumentation/messaging/src/main/java/brave/messaging/MessagingOperation.java
```diff
@@ -16,16 +16,9 @@
...
-
-  /**
-   * Messaging broker service, e.g. kafka-cluster, jms-server.
-   */
-  String remoteServiceName(Channel channel);
+public enum MessagingOperation {
+  SEND,
+  BULK_SEND,
+  RECEIVE,
+  BULK_RECEIVE;
```
jeqo: Should we also consider `POLL` and `BULK_POLL`, to match Kafka semantics?

instrumentation/messaging/src/main/java/brave/messaging/MessagingParser.java
```diff
@@ -6,7 +6,7 @@
  * (the "License"); you may not use this file except in compliance with
  * the License.  You may obtain a copy of the License at
  *
- *     http://www.apache.org/licenses/LICENSE-2.0
+ *     messaging://www.apache.org/licenses/LICENSE-2.0
```
jeqo: typo.

```suggestion
  *     http://www.apache.org/licenses/LICENSE-2.0 
```

instrumentation/messaging/src/main/java/brave/messaging/MessagingAdapter.java
```diff
@@ -0,0 +1,73 @@
...
+   * <p>Conventionally associated with the key "message.correlation_id"
+   */
+  @Nullable public abstract String correlationId(Msg message);
+
+  /**
+   * Message broker name. {@code null} if unreadable.
+   *
+   * <p>Conventionally associated with {@link Span#remoteServiceName(String)}
+   */
+  @Nullable public abstract String brokerName(Chan channel);
```
jeqo: Is the `brokerName` meant to differentiate between protocols (e.g. jms or kafka or amqp) or do we need an additional method for it?

instrumentation/messaging/src/main/java/brave/messaging/MessagingAdapter.java
```diff
@@ -0,0 +1,73 @@
...
+   * <p>Conventionally associated with the key "message.channel"
+   */
+  @Nullable public abstract String channel(Chan channel);
+
+  /**
+   * Type of channel, e.g. queue or topic. {@code null} if unreadable.
+   *
+   * <p>Conventionally associated with the key "message.channel_type"
+   */
+  // TODO: naming is arbitrary.. we once used "kind" for Span but only because stackdriver did..
```
jeqo: I'm +1 on kind as it is used for `span`. Unless we have `type` somewhere else and match better the concept here.

instrumentation/messaging/src/main/java/brave/messaging/MessagingAdapter.java
```diff
@@ -0,0 +1,73 @@
...
+  // Not sure we should re-use kind for parity with that or not..
+  @Nullable public abstract String channelType(Chan channel);
+
+  /**
+   * Key used to identity or partition messages. {@code null} if unreadable.
+   *
+   * <p>Conventionally associated with the key "message.key"
+   */
+  @Nullable public abstract String messageKey(Msg message);
```
jeqo: I'm wondering if we should use `key` or `id` here. We have `correlation_id` as well, then `message.id` might fit better.

instrumentation/messaging/src/main/java/brave/messaging/ProcessorHandler.java
```diff
@@ -0,0 +1,71 @@
...
+   * destination</li>
+   * <li>A child span with the duration of the delegated listener</li>
+   * </ol>
+   *
+   * <p>{@code addConsumerSpan} should only be set when the message consumer is not traced.
+   */
+  // channel is nullable as JMS could have an exception getting it from the message
+  public Span startProcessor(@Nullable Chan channel, Msg message, boolean addConsumerSpan) {
+    if (!addConsumerSpan) {
+      Span result = tracer.nextSpan().start();
```
jeqo: Should we extract context from carrier before creating this span? If not, child span does not have a parent unless we create one before calling `nextSpan`

jeqo: This one fixes some of the issues on the Kafka Client tests.

instrumentation/messaging/src/main/java/brave/messaging/MessagingOperation.java
adriancole: imho the point of abstraction is to eliminate unnecessary naming differences. there's no difference between receive and poll. Later in this change, you'll see how someone can go back and match the names of kafka.

Why is this important? well, is the span any different in remote nature if we had one named "poll" and another "receive" if not, why would we split aggregations on that. Basically from a service perspective, I don't think we intend to have different names by default on accident of what libraries choose to name things.

iotw, I would prefer to rename RECEIVE to POLL of receive isn't a good name :) however, we used to use send and receive in our v1 messaging model, so thought maybe this is a better default name.

instrumentation/messaging/src/main/java/brave/messaging/MessagingAdapter.java
adriancole: probably we need a different method/tag. probably protocol is the correct choice, similar to what @trustin did here for http https://github.com/line/armeria/pull/1762

We could ask if it is important enough to include in the default tags, considering it will add index pressure to add "message.protocol"="jms" 10k per second :P though in fact jms is not a protocol anyway.

if we are looking for which library type is in use, probably "lc" local component is the better thing. kafka and amqp are protocols, for example, jms not.

instrumentation/messaging/src/main/java/brave/messaging/MessagingAdapter.java
adriancole: sgtm

instrumentation/messaging/src/main/java/brave/messaging/MessagingAdapter.java
adriancole: maybe we can use some examples to pin this down. kafka uses the word "key" I think. what does amqp and rocketmq use?

instrumentation/messaging/src/main/java/brave/messaging/ConsumerHandler.java
```diff
@@ -0,0 +1,143 @@
...
+  /** Returns a started processor span if {@code createProcessor} is true */
+  @Nullable Span handleReceive(Chan channel, Msg message, C carrier, String remoteServiceName,
+    TraceContextOrSamplingFlags extracted, boolean createProcessor) {
+    Span span = tracing.tracer().nextSpan(extracted);
+    // Creating the processor while the consumer is not finished ensures clocks are the same. This
+    // allows the processor to start later, but not be subject to clock drift relative to the parent.
+    Span processorSpan = createProcessor ? tracing.tracer().newChild(span.context()) : null;
+    if (!span.isNoop()) {
+      span.kind(Span.Kind.CONSUMER);
+      parser.start("receive", adapter, channel, message, span.context(), span.customizer());
```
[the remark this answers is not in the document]
adriancole: I started with this, but stopped because type complexity and using an enum hints at a closed amount of operations.. we could use this for other operations .. not sure yet really that's why at least temporarily open (ex http.method is open)

instrumentation/messaging/src/main/java/brave/messaging/ProcessorHandler.java
adriancole: thanks for finding a bug :P

adriancole:
Thanks for all the review feedback @jeqo !

adriancole:
ok rebased.. still need to address comments

instrumentation/messaging/src/main/java/brave/messaging/MessagingOperation.java
adriancole: lol this file was unused

instrumentation/messaging/src/main/java/brave/messaging/MessagingAdapter.java
adriancole: added notes

instrumentation/messaging/src/main/java/brave/messaging/MessagingAdapter.java
adriancole: added TODO

instrumentation/messaging/src/main/java/brave/messaging/ConsumerHandler.java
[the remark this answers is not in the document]
adriancole: yeah leaving open for now

adriancole:
JMS 1.1 tests are failing.. somehow we're accessing a JMS 2.0 type.. I will try to figure it out

adriancole:
worked around the JMS test failure which was a super strange classloader quirk

merging this to unlock the messaging branch.

PR #918, opened 2019-06-02
Updates dependencies, notable ones from Apache Zipkin (incubating)

jeqo:


pom.xml
```diff
@@ -117,7 +117,7 @@
 
     <!-- use the same values in bom/pom.xml -->
     <zipkin.version>2.12.5</zipkin.version>
```
adriancole: might as well update zipkin also, as we were waiting on that also..

adriancole: to match reporter would be 2.14.0.. PS there's probably some find/replacing to do here and elsewhere notably around group ID switches. I use the `ag` tool like.. `ag io.zipkin` to find references.

adriancole:
after switching versions in pom.xml and bom/pom.xml for zipkin and zipkin-reporter, make sure `ag io.zipkin` comes back clean. One thing we should have earlier deleted is the `travis` directory as we no longer publish that way.

adriancole:
ok updated all the things we could

Issue #920, opened 2019-06-07
brave.flush annotation reported for already finished span


lambcode:
## Problem
Brave Version: 5.6.0

I am encountering unexpected behavior when using Tracer.currentSpan() to get a span that has previously been finished. It appears that calling currentSpan() restarts the span to a degree in that the TraceContext is re-registered with PendingSpans. This in turn causes a brave.flush annotation to be reported when that TraceContext gets GC'd even though in reality the span has already been finished.
![Untitled Diagram(2)](https://user-images.githubusercontent.com/5820125/59121923-9561a100-8916-11e9-982d-953b5337b597.png)

This could probably be avoided by calling abandon() on the Span that is returned from currentSpan(), but in our case sometimes this code lives in libraries that are not aware whether or not the current span needs to be abandoned() (it was finished previously) or if it will be finished later.

This is also causing odd behavior in zipkin-ui (not lens) such as spans going missing presumably because it doesn't expect multiple spans with the same spanId that aren't shared spans. (For example I found one span had three documents in elasticsearch: a client span, a servier span, and another span with the brave.flush annotation)

## Proposed Fix
I wonder if instead of removing PendingSpan entries from the HashMap when PendingSpans.remove() is invoked, these entries could be marked as finished somehow. In this manner PendingSpans will not forget about a TraceContext until it has been GC'd meaning that the context can never be registered with pending spans more than once.

If this approach is pursued, special care will need to be taken when dealing with multiple Tracer instances, as each Tracer has it's own instance of PendingSpans that needs to remain in sync.

I have attached a sample project with unit tests for both a single tracer and multiple tracer test that exhibits this behavior.
[bugjar.zip](https://github.com/apache/incubator-zipkin-brave/files/3267116/bugjar.zip)

adriancole:
can you clarify if this is only accessing the span which causes the flush?
eg not adding data, simply accessing it? Perhaps an example json would be
helpful.

If it is as you say and there is no data except traceid/spanid and the
flush annotation (quite possible) I would recommend comparison against a
these fields, keep the logging statement, but not send to zipkin. This
would be cheaper and less impact than trying to track all completed traces,
though if we had to do the latter at some point, we could revisit.

Regardless, can you make the tests with existing unit tests in this repo,
as it is less work for us to evaluate and this would also be how a fix
would occur.

adriancole:
pps as always, use latest zipkin (lens UI) and brave when doing tests so we
[...]

PR #921, opened 2019-06-10
Add tests for #920. "brave.fush" annotation should not be reported for finished spans

lambcode:
These tests are currently failing, but show same bug as described in #920. See issue for more details.

### DIFF
--- a/brave/src/test/java/brave/TracerTest.java
+++ b/brave/src/test/java/brave/TracerTest.java
@@ -34,8 +34,10 @@ import java.util.LinkedHashMap;
 import java.util.List;
 import java.util.Map;
 import java.util.function.Function;
+import java.util.stream.Collectors;
 import org.junit.After;
 import org.junit.Test;
+import zipkin2.Annotation;
 import zipkin2.Endpoint;
 import zipkin2.reporter.Reporter;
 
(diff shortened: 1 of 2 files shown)


adriancole:
thanks will be looking at this shortly and next release will address whatever we can without breaking something else :)

brave/src/test/java/brave/TracerTest.java
```diff
@@ -693,6 +695,52 @@
...
+      .flatMap(span -> span.annotations().stream())
+      .map(Annotation::value)
+      .collect(Collectors.toList())).doesNotContain("brave.flush");
+  }
+
+  /**
+   * Must be a separate method from the test method to allow for local variables to be garbage
+   * collected
+   */
+  private static void simulateInProcessPropagation(Tracer tracer1, Tracer tracer2) {
```
adriancole: this is an odd use case. the tracer (tracing etc) is a singleton and meant to be used as such.

In-process propagation involves moving the Span or TraceContext across threads, not switching tracer instances per thread.

Can you elaborate why you are doing this? It might cause other problems then the scenario noted here.

lambcode: I was not aware that the tracer/tracing was meant to be a singleton. The API lets one create different instances via Tracing.newBuilder(). Is the intent that this builder is used only once per application? If so, why does the API allow you to create multiple instances?

As a little bit of background, we have different tracing instances each with a different serviceName. As an example lets say I have webservice A<sub>ws</sub> and webservice B<sub>ws</sub>. A<sub>ws</sub> requests information from B<sub>ws</sub> through a library B<sub>client</sub> that makes http calls to B<sub>ws</sub>. The B<sub>client</sub> library uses a tracing instance with serviceName "B" whereas the A<sub>ws</sub> has a tracing instance with serviceName "A". A trace will start in A<sub>ws</sub> with a span that has service name "A". This service will eventually call into the B<sub>client</sub> library code which will create child spans (via a different tracer) with serviceName "B".

We've actually had some discussions internally recently on whether or not we should be doing this, so some guidance on when different tracing instances should be used would be appreciated. 

As for the in-process propagation, I think you might be confused that the test does not actually use multiple threads. I found that this is not strictly necessary because the behavior is actually caused by using multiple tracing instances and scoped spans. I just found the in-process propagation example to be an easy one to understand.

adriancole: I will try to answer your questions:

Q: why let people make instances if should be a singleton?
A: many frameworks handle singletons differently, and there are classloader concerns, too. brave does not constrain you to using spring, guice, or otherwise singleton management.

Q: should we be using different instances per endpoint in order to control the service name?
A: this is complex and likely more fine grained than most use. For most sites endpoints are actually span names. We have a tool called FinishedSpanHandler which has the concept of localRoot which would allow you to rewrite data consistently per fraction of a trace. Again, this is more complex that what most sites do. if you want to discuss that further better use chat as it is less spammy to watchers https://gitter.im/openzipkin/zipkin

On the test talking about threads then not using them etc. I personally found this confusing and it took me a few hours to guess what was supposed to be checked. (think time not pure work). In general, being precise like your diagram is best as when comments lie they are distracting to current and future maintainers. Your tests helped, just it mentioned things that wouldn't have been tested in that code. thx regardless!

lambcode: Thanks for the answers! This is super helpful. And sorry for any confusion those tests caused.

PR #924, opened 2019-06-12
 Makes Tracer.currentSpanCustomizer() lazy 

### LINKED ISSUE #920: brave.flush annotation reported for already finished span

adriancole:
@lambcode noticed that calling `Tracer.currentSpan()` without using the
result can lead to orphaned spans. The simplest way to prevent this is
to defer overhead until data is added with the result.

Fixes #920

Added benchmarks..

before

```
TracerBenchmarks.currentSpan:currentSpan·p0.999                          sample             8.560             us/op
TracerBenchmarks.currentSpan:·gc.alloc.rate.norm                         sample      15    88.012 ±   0.005    B/op
TracerBenchmarks.currentSpan_tag:currentSpan_tag·p0.999                  sample            20.256             us/op
TracerBenchmarks.currentSpan_tag:·gc.alloc.rate.norm                     sample      15    88.065 ±   0.004    B/op
TracerBenchmarks.currentSpan_unsampled:currentSpan_unsampled·p0.999      sample             4.995             us/op
TracerBenchmarks.currentSpan_unsampled:·gc.alloc.rate.norm               sample      15    24.005 ±   0.001    B/op
```

after

```
TracerBenchmarks.currentSpan:currentSpan·p0.999                          sample              7.974             us/op
TracerBenchmarks.currentSpan:·gc.alloc.rate.norm                         sample      15     24.015 ±   0.008    B/op
TracerBenchmarks.currentSpan_tag:currentSpan_tag·p0.999                  sample             10.192             us/op
TracerBenchmarks.currentSpan_tag:·gc.alloc.rate.norm                     sample      15     72.050 ±  12.511    B/op
TracerBenchmarks.currentSpan_unsampled:currentSpan_unsampled·p0.999      sample              4.198             us/op
TracerBenchmarks.currentSpan_unsampled:·gc.alloc.rate.norm               sample      15     24.009 ±   0.007    B/op
```

### DIFF
--- a/brave/src/main/java/brave/LazySpan.java
+++ b/brave/src/main/java/brave/LazySpan.java
@@ -1,0 +1,143 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one or more
+ * contributor license agreements.  See the NOTICE file distributed with
+ * this work for additional information regarding copyright ownership.
+ * The ASF licenses this file to You under the Apache License, Version 2.0
+ * (the "License"); you may not use this file except in compliance with
+ * the License.  You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package brave;
+
+import brave.handler.MutableSpan;
+import brave.propagation.TraceContext;
+
+/**
+ * This defers creation of a span until first public method call.
+ *
+ * <p>This type was created to reduce overhead for code that calls {@link Tracer#currentSpan()},
+ * but without ever using the result.
+ */
+final class LazySpan extends Span {
+  final Tracer tracer;
+  final TraceContext context;
+  volatile Span delegate;
+
+  LazySpan(Tracer tracer, TraceContext context) {
+    this.tracer = tracer;
+    this.context = context;
+  }
+
+  @Override public boolean isNoop() {
+    return false;
+  }
+
+  @Override public TraceContext context() {
+    return context;
+  }
+
+  @Override public SpanCustomizer customizer() {
+    return new SpanCustomizerShield(this);
+  }
+
+  @Override public Span start() {
+    return span().start();
+  }
+
+  @Override public Span start(long timestamp) {
+    return span().start(timestamp);
+  }
+
+  @Override public Span name(String name) {
+    return span().name(name);
+  }
+
+  @Override public Span kind(Kind kind) {
+    return span().kind(kind);
+  }
+
+  @Override public Span annotate(String value) {
+    return span().annotate(value);
+  }
+
+  @Override public Span annotate(long timestamp, String value) {
+    return span().annotate(timestamp, value);
+  }
+
+  @Override public Span tag(String key, String value) {
+    return span().tag(key, value);
+  }
+
+  @Override public Span error(Throwable throwable) {
+    return span().error(throwable);
+  }
+
+  @Override public Span remoteServiceName(String remoteServiceName) {
+    return span().remoteServiceName(remoteServiceName);
+  }
+
+  @Override public boolean remoteIpAndPort(String remoteIp, int remotePort) {
+    return span().remoteIpAndPort(remoteIp, remotePort);
+  }
+
+  @Override public void finish() {
+    span().finish();
+  }
+
+  @Override public void finish(long timestamp) {
+    span().finish(timestamp);
+  }
+
+  @Override public void abandon() {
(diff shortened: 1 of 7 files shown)


adriancole:
I built this over the commits in #921 to ensure tests would pass considering the use case described

adriancole:
ps the use cases for calling tracer.currentSpan and not using the result are one of..
* checking null from `tracer.currentSpan()` as a signal for "is current"
* defensively doing some parsing based on whether or not `Span.isNoop()`

just guessing.. I think usually it would be a bug to call `Tracer.currentSpan()` and drop the result.

brave/src/main/java/brave/LazySpan.java
```diff
@@ -0,0 +1,126 @@
...
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package brave;
+
+import brave.propagation.TraceContext;
+
+/** This wraps the public api and guards access to a mutable span. */
```
anuraaga: Don't think this "guards" from someone - how about explicitly saying "ensuring a `Span` is only constructed on first API method call"?

brave/src/main/java/brave/LazySpan.java
```diff
@@ -0,0 +1,126 @@
...
+    } else if (o instanceof RealSpan) {
+      return context.equals(((RealSpan) o).context);
+    }
+    return false;
+  }
+
+  Span span() {
+    Span result = delegate;
+    if (result != null) return result;
+    return delegate = tracer.toSpan(context);
```
anuraaga: I'm not sure what the thread-safety guarantees of `Span` are, but I think normally we'd want to double-lock here to prevent multiple spans being created if accessing from different threads at the same time.

```java
if (result != null) return result;
synchronized {
  result = delegate;
  if (result != null) return result;
  return delegate = tracer.toSpan(context);
}
```

anuraaga: `abandon` and `flush` would need similar too if the thread-safety matters.

brave/src/main/java/brave/LazySpan.java
```diff
@@ -0,0 +1,126 @@
...
+    span().flush();
+  }
+
+  @Override public String toString() {
+    return "LazySpan(" + context + ")";
+  }
+
+  @Override public boolean equals(Object o) {
+    if (o == this) return true;
+    if (o instanceof LazySpan) {
```
anuraaga: How about a package-private method that returns the context?

adriancole: I don't know if there are any other callers that would use them... usually we don't add methods for single call sites.

brave/src/main/java/brave/LazySpan.java
adriancole: I forgot a comment I guess. Multiple copies of Span are not a problem apart from overhead. It is probably not worth optimizing the code as that would hint there is a problem. I will add a comment, thanks!

brave/src/main/java/brave/LazySpan.java
adriancole: this doc was meant for another type :) will fix!

brave/src/main/java/brave/LazySpan.java
adriancole: hope the comments help

adriancole:
I caused a bug https://github.com/apache/incubator-zipkin-brave/pull/927

PR #926, opened 2019-06-15
add unit test for brave-instrumentation-kafka-clients

jang675953:


instrumentation/kafka-clients/src/test/java/brave/kafka/clients/ITKafkaTracing.java
```diff
@@ -114,6 +114,40 @@ public void close() {
...
+    producer = createTracingProducer();
+    consumer = kafkaRule.helper().createStringConsumer(); // not traced
+
+    for (int i = 0; i < 2; i++) {
+      producer.send(new ProducerRecord<>(topic1, TEST_KEY, TEST_VALUE)).get();
+    }
+    consumer.subscribe(Collections.singletonList(topic1));
+    ConsumerRecords<String, String> records = consumer.poll(10000);
+    assertThat(producerSpans).hasSize(2);
+    assertThat(records).hasSize(2);
```
adriancole: here is ok, just you have a race condition. producerSpans is a blocking queue. You can use take() to block until the span is around.

```java
    ConsumerRecords<String, String> records = consumer.poll(10000);
    assertThat(records).hasSize(2);

    producerSpans.take();
    producerSpans.take();
```

instrumentation/kafka-clients/src/test/java/brave/kafka/clients/ITKafkaTracing.java
```diff
@@ -114,6 +114,40 @@ public void close() {
...
+    String topic1 = testName.getMethodName() + "1";
+
+    producer = kafkaRule.helper().createStringProducer(); // not traced
+    consumer = createTracingConsumer(topic1);
+
+    for (int i = 0; i < 2; i++) {
+      producer.send(new ProducerRecord<>(topic1, TEST_KEY, TEST_VALUE)).get();
+    }
+
+    ConsumerRecords<String, String> records = consumer.poll(10000);
```
adriancole: we only expect a single consumer span as it represents the bulk receive. again. you need to block.

```java
    ConsumerRecords<String, String> records = consumer.poll(10000);
    assertThat(records).hasSize(2);

    consumerSpans.take();
```

adriancole:
I don't think there is a bug here.. maybe a misunderstanding. the instrumentation approach is described in the README of kafka-clients. I think perhaps you may expect a consumer span for each message, when it is one consumer span per topic/poll.

PR #927, opened 2019-06-15
Fixes assertion by making equals the same between lazy and real span

adriancole:
We had an assertion in `ThreadLocalSpan` which broke because our equals
comparison was one way. This fixes it.


adriancole:
eventhough this only hurt when assertions are on, I will cut a release as it is pretty dire to always fail when assertions are on (ex unit tests that use brave can bork)

Issue #849, opened 2018-12-14
Context decorators like Scope decorators


adriancole:
Right now, we have `ScopeDecorator` for doing things special when a span is placed in scope. However, we still have a wrapping concern for propagation.. notably propagation can control the "extra" data added to spans. For example, to add extra fields, we have to wrap B3 or something else with ExtraFields. If you wanted to add something else, like a flag carrier or something to catch a service name for readback... you'd have to wrap that whole thing as well.

It would be nice to have a decorator option instead, where the actual propagation (ex B3 or X-Ray) is separate from add-ons like extra fields or correlation fields. That way, the latter can be more easily manipulated or controlled, and at least independently so.

It is possible to address this feature well would be a re-write, but it is also possible we can address it well enough. Since propagation is a very expensive part of the code, performance wise, we should take a lot of care when trying to retrofit this to the Brave 5 codebase.

Another note is that controlling "extra" is actually somewhat complex due to inheritance.. for example, do you inherit the data from parent to child, what happens when incoming context only has "extra" and you want to add to it? That's why we don't do this now actually as it is pretty complex (see the baggage paper from Brown for the whole story). Here's the undocumented impl that underpins extra fields for example https://github.com/openzipkin/brave/blob/master/brave/src/main/java/brave/internal/ExtraFactory.java

cc @anuraaga @koji who remember this from last time. @devinsba who is adding some integration to spring and @tdanylchuk who frequently asks for the ability to easily customize the context

PR #931, opened 2019-06-26
Adds ability to redact fields from downstream propagation

### LINKED ISSUE #929: [Feature Request] Store misc information within TraceContext which are not propagated
**Feature:**
Store misc information within TraceContext which are not propagated

**Rational**
Currently, based on experiments and reading the source code, the TraceContext.extra() is limited to a pre-defined key/value set that is designed to be propagated. However, it may be usable to store additional data which is local to the trace context which are not going to be propagated but consumed by Propagation.Factory.decorate(TraceContext).

**Example Scenario**
Provided there is HTTP header that contains a JWT of the identity of the user used to create a flow in system, this HTTP header is marked to be propagated to all calls that are part of this transaction.

There are claims in the JWT such as sub, name, email which should be written to logs per that transaction, exactly the same as the trace id and span id. These are calculated values out of JWT. There is no reason to propagate these as we already propagate the JWT, and there is no reason to calculate them over and over.

The solution we thought is to parse the token at a filter and add the required claims into the TraceContext.extra, implement Propagation.Factory.decorate(TraceContext) to load the information into MDC. However, the TraceContext.extra is not friendly for custom/generic input as implementation expects a concrete classes and structure [NOTE: it may well that I do not fully understand the multiple class handling within the list, however, it is not simple to register a new class post construction].

Will you consider a patch that will enable a generic Map<String, Object> that can be attached to the TraceContext for this kind of use case?

**Prior Art**
 * #390 - not sure if relevant
 * #693 - not sure if relevant

Thanks!

adriancole:
This lets you use the extra field propagation mechanism to propagate in,
but not out of process. This can be used to correlate properties with
logs in a way that doesn't add network overhead.

This design was chosen as it incurs very little in-process overhead and
has the least burden on the code base. It can be used as a stop-gap
until a more fully featured correlation context feature is developed.

Fixes #929
See also #577

### DIFF
--- a/brave/src/main/java/brave/propagation/ExtraFieldPropagation.java
+++ b/brave/src/main/java/brave/propagation/ExtraFieldPropagation.java
@@ -36,8 +36,7 @@ import java.util.Set;
 /**
  * Allows you to propagate predefined request-scoped fields, usually but not always HTTP headers.
  *
- * <p>For example, if you are in a Cloud Foundry environment, you might want to pass the request
- * ID:
+ * <p>For example, if you are in a Cloud Foundry environment, you might want to pass the request ID:
  * <pre>{@code
  * // when you initialize the builder, define the extra field you want to propagate
  * tracingBuilder.propagationFactory(
@@ -51,7 +50,20 @@ import java.util.Set;
  * ExtraFieldPropagation.get("x-country-code", "FO");
  * }</pre>
  *
- * <p>You may also need to propagate a trace context you aren't using. For example, you may be in
+ * <h3>Appropriate usage</h3>
+ * It is generally not a good idea to use the tracing system for application logic or critical code
+ * such as security context propagation.
+ *
+ * <p>Brave is an infrastructure library: you will create lock-in if you expose its apis into
+ * business code. Prefer exposing your own types for utility functions that use this class as this
+ * will insulate you from lock-in.
+ *
+ * <p>While it may seem convenient, do not use this for security context propagation as it was not
+ * designed for this use case. For example, anything placed in here can be accessed by any code, in
+ * the same classloader!
+ *
+ * <h3>Passing through alternate trace contexts</h3>
+ * <p>You may also need to propagate an second trace context transparently. For example, when in
  * an Amazon Web Services environment, but not reporting data to X-Ray. To ensure X-Ray can co-exist
  * correctly, pass-through its tracing header like so.
  *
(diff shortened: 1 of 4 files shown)


brave/src/main/java/brave/propagation/ExtraFieldPropagation.java
```diff
@@ -143,25 +150,20 @@ public FactoryBuilder addPrefixedFields(String prefix, Collection<String> fieldN
     }
 
     public Factory build() {
-      if (prefixedNames.isEmpty()) {
```
adriancole: most of the red here was just me trying to pay for the new code by removing old code

brave/src/main/java/brave/propagation/ExtraFieldPropagation.java
```diff
@@ -143,25 +150,20 @@ public FactoryBuilder addPrefixedFields(String prefix, Collection<String> fieldN
...
-        List<String> validated = Arrays.asList(ensureLowerCase(fieldNames));
-        for (int i = 0, length = validated.size(); i < length; i++) {
-          String nextFieldName = validated.get(i);
-          fields.add(nextFieldName);
-          keys.add(nextFieldName);
-          keyToFieldList.add(i);
-        }
+      int i = 0;
+      for (String fieldName : fieldNames) {
+        if (redactedFieldNames.contains(fieldName)) redacted.set(i);
```
adriancole: this is the only line that matters.

adriancole:
will add some text tomorrow @devinsba probably on the extra fields propagation type itself and the README. I've also noticed in sleuth support people abusing extra fields a lot just to get things done in process. Not sure if folks are always aware of the options out there.

Aside, I wanted to mention that anything more custom than this is possible with composition. An approach to the same could be making  a wrapping PropagationFactory like this:

```java
  @Override public <C> Injector<C> injector(Setter<C, K> setter) {
    return delegate.injector((carrier, key, value) -> {
      if (key.equals("redacted key")) return;
      setter.put(carrier, key, value);
    });
  }
```

However, a bunch of wrapped things itself can get gnarly for the simple case, hence  #849.

alonbl:
@devinsba: the spring security requires an explicit delegation while the brave/sleuth tracking the transaction implicitly, even to cases in which are not spring aware. this is why using a single mechanism to propagate data in the scope of transaction is preferable.

The Authentication object is authoritative object rather than informative. Of course we use this to set subject and authorities, however, it is not have event processing of assigning into new threads or when calling 3rd parties.

brave/src/main/java/brave/propagation/ExtraFieldPropagation.java
```diff
@@ -143,25 +150,24 @@ public FactoryBuilder addPrefixedFields(String prefix, Collection<String> fieldN
     }
 
     public Factory build() {
```
anuraaga: How about validating all names in redacted field names are in field names, to catch typos easier.

adriancole: right now adding a redacted field implies adding a field, so that should be covered.

adriancole:
@devinsba PTAL I added a disclaimer

PR #933, opened 2019-07-01
Fix bug in SpanCustomizerShield

devinsba:
https://stackoverflow.com/questions/56838165/h2-console-stopped-working-after-updating-from-spring-cloud-greenwich-sr1-to-sr2

cc @marcingrzejszczak


devinsba:
cc @adriancole you might feel like improving this test when you wake up. I am going to merge and release to get the fix out to users

adriancole:
wow what a bug. thanks!

PR #934, opened 2019-07-03
Properly implements old call-path of status codes

adriancole:
Eventhough new instrumentation don't use it, old could.

See https://github.com/spring-cloud/spring-cloud-sleuth/pull/1387


adriancole:
WIP as needs tests

adriancole:
I looked at all the javadoc to see what could return what, then double-checked source code and got a little defensive about method chaining too

adriancole:
backfilled all the tests

PR #937, opened 2019-07-04
Backfills context tests for sampled property

adriancole:
We had some missing tests


context/log4j12/src/test/java/brave/context/log4j12/MDCCurrentTraceContextTest.java
```diff
@@ -58,13 +58,17 @@ public void isnt_inheritable() throws Exception {
           .isEqualTo(context.parentIdString());
       assertThat(MDC.get("spanId"))
           .isEqualTo(context.spanIdString());
+      assertThat(MDC.get("sampled"))
```
adriancole: note log4j 1.2 is the only api of the 3 that kindof supports object properties. slf4j and log4j2 do not. for consistency we are using string value for sampled eventhough we could have used object in log4j 1.2

adriancole:
I will test all the mvc examples again prior to merge and release

Issue #867, opened 2019-02-20
is there a plan to change alibaba-dubbo dependency to apache-dubbo?


geercode:
- [x] Question
since all projects have been transfered to apache incubator,is there any paln to support dubbo 2.7+  ?
I have tried using dubbo-spring-boot-starter 2.7,the dependency of alibaba LoggerFactory is nologger supported.

adriancole:
in general we dont have plans unless users ask for something. are you able
to help migrate this to apache dubbo? is it possible to make it work with
both libraries (we dont want to break people who haven't moved yet)

geercode:
I think dubbo will be at a unsatble circumstance for a very long time.Between 2.6 and 2.7,it's very much like just some packages' name changing of the same content.I suppose using osgi to make them work together is a better way.At the same time,dubbo is just under maintainance.The rest protocol which 2.6 is already supported is a key module.Before dubbo 3.0,there would not be too many developers would use it.

adriancole:
feel free to chat with me here I am ready to have a look at brave stuff soon https://gitter.im/openzipkin/zipkin

adriancole:
this is a problem as the code is very different.. 2.6.5 and 2.7.0 removed classes.. that can be ok but now we have to figure out what to do.

adriancole:
It appears the dubbo team are deciding whether to stop working with us based on the fact that will still support their old clients. I am very disappointed at this behavior. I hope they change https://lists.apache.org/thread.html/4ab3582245cb7daa79becaa7f644089f93da08b3a8ae0c3c90e89a1c@%3Cdev.dubbo.apache.org%3E

adriancole:
we can't break compat of the old dubbo users because we are nice. to figure out what to do can be easy or hard depending on bytecode compat. For example, grpc we were able to make a wide version range work because the core classes are still compatible from version 1.3.

However, I don't think this is the case with dubbo. version 2.7 seems not compatible with 2.6 unless you revert to reflection. this likely means we need a separate artifact.

For example, we can start `brave-instrumentation-apache-dubbo` or a different scheme like `brave-instrumentation-dubbo-2.7` as the dubbo team also plan to break compatibility again in version 3.

The challenge here is thinking and planning only.. which route do we take? If so, which naming convention do we use? we will be stuck with it, so it is worthwhile having the thinking exercise.

ralf0131:
Hi, 

[...]

PR #938, opened 2019-07-04
Upgrade dubbo to 2.7.3

chickenlj:
related issue https://github.com/openzipkin/brave/issues/867.

Currently, this patch uses 2.7.3-SNAPSHOT, I will update to 2.7.3 once it's available then we can merge.


adriancole:
we can't do this as it will break alibaba dubbo. We need a less breaking solution for the transition

adriancole:
closing this as breaking everyone is not a good first step

PR #939, opened 2019-07-04
Begins work to bridge from alibaba to apache dubbo

### LINKED ISSUE #867: is there a plan to change alibaba-dubbo dependency to apache-dubbo?

adriancole:
Starting with the 2.7.3 all of our main code can compile with
org.apache.dubbo:dubbo and also com.alibaba:dubbo

There is some strange work here because the spring integrations were not
backported. Until such time as it is, we need reflection to actually
load the extensions. As this is fragile, this change shouldn't merge
until a maven invoker test is setup to prove the tests also pass with
com.alibaba:dubbo still. Also, we need to cleanup documentation around
spring as it will mention types that don't exist.

Fixes #867

### DIFF
--- a/instrumentation/dubbo-rpc/src/main/java/brave/dubbo/rpc/TracingFilter.java
+++ b/instrumentation/dubbo-rpc/src/main/java/brave/dubbo/rpc/TracingFilter.java
@@ -23,8 +23,6 @@ import brave.propagation.TraceContext;
 import brave.propagation.TraceContextOrSamplingFlags;
 import com.alibaba.dubbo.common.Constants;
 import com.alibaba.dubbo.common.extension.Activate;
-import com.alibaba.dubbo.common.extension.ExtensionLoader;
-import com.alibaba.dubbo.config.spring.extension.SpringExtensionFactory;
 import com.alibaba.dubbo.remoting.exchange.ResponseCallback;
 import com.alibaba.dubbo.rpc.Filter;
 import com.alibaba.dubbo.rpc.Invocation;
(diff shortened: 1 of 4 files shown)


instrumentation/dubbo-rpc/src/main/java/brave/dubbo/rpc/TracingFilter.java
```diff
@@ -48,10 +46,14 @@
   TraceContext.Injector<Map<String, String>> injector;
 
   /**
-   * {@link ExtensionLoader} supplies the tracing implementation which must be named "tracing". For
-   * example, if using the {@link SpringExtensionFactory}, only a bean named "tracing" will be
-   * injected.
+   * <h3>Notes for com.alibaba:dubbo</h3>
+   *
+   * {@code com.alibaba.dubbo.common.extension.ExtensionLoader} supplies the tracing implementation
```
adriancole: TODO: I think maybe we can just say the word ExtensionLoader and SpringExtensionFactory here and leave out the package name as they likely exist. We should also check the README for drift

adriancole:
this is still broken, and I've put a suggestion to give up on #867.

idea is to close this PR and leave brave-instrumentation-dubbo-rpc pinned to alibaba dubbo. make a new brave-instrumentation-dubbo that is compatible with apache dubbo. That way we don't break binary compatibility here which would surprise people. Also, this acknowledges that eventhough we tried it is too much effort attempting a bridge, and it always leads to nothing. By making a new module at least new people can use it. cc @openzipkin/brave

PR #957, opened 2019-08-06
Kafka Streams: Propagate context between stream operations

### LINKED ISSUE #956: Kafka Streams extra field propagation
**Feature:**
I'd like to use `ExtraFieldPropagation` with Kafka Streams, (mostly) to set field values based on message details, such that they become available in Zipkin and our log MDC (through Sleuth).

**Rational**
`ExtraFieldPropagation` is part of Brave, and I could see other users expecting this to work.

**Example Scenario**
Without this integration one can attempt to set extra fields, but they are not propagated yet. This is somewhat confusing and unexpected. My unsuccessful attempts thus far had used:
```java
KStream<String, String> numbersStream = builder.stream("numbers");
numbersStream.transformValues(kafkaStreamsTracing.peek("set", (k, v) -> {
	ExtraFieldPropagation.set(tracer.currentSpan().context(), "messageid", v);
	log.info("set for {} -> {}", k, v);
}))
.transformValues(kafkaStreamsTracing.peek("get", (k, v) -> {
	String value = ExtraFieldPropagation.get(tracer.currentSpan().context(), "messageid");
	log.info("got: {} for {} -> {}", value, k, v);
}));
```
But this results in the following log lines:
```
[kafka-streams-log-correlation,08516934550150511522640890549786,a1856e8c1acf9d47,true] 31504 \
    --- [-StreamThread-1] StreamsConfig: set for extra-propagation-IjD0xhR0ld -> 7
[kafka-streams-log-correlation,08516934550150511522640890549786,4cd66730012a3af8,true] 31504 \
    --- [-StreamThread-1] StreamsConfig: got: null for extra-propagation-IjD0xhR0ld -> 7
```
You'll notice that both log lines are from separate spans; with the extra field set in the first not present when retrieved in the second span.

I've thus far been unable to work this into a clean unit test for Brave; I hope it's clear what I'm trying to achieve.

**Prior Art**
 * Instrumentation at present through https://github.com/openzipkin/brave/tree/master/instrumentation/kafka-streams

jeqo:
Fix #956 

Current implementation is treating each stream operation as a child of the span created at initial consumption.

This leads to issues like #956, as there is no propagation between stream operations.

This PR fix this by injecting context before leaving each operation.

### DIFF
--- a/instrumentation/kafka-clients/src/test/java/brave/kafka/clients/BaseTracingTest.java
+++ b/instrumentation/kafka-clients/src/test/java/brave/kafka/clients/BaseTracingTest.java
@@ -14,7 +14,11 @@
 package brave.kafka.clients;
 
 import brave.Tracing;
+import brave.internal.PropagationFieldsFactory;
+import brave.propagation.B3Propagation;
 import brave.propagation.CurrentTraceContext;
+import brave.propagation.ExtraFieldPropagation;
+import brave.propagation.Propagation;
 import brave.propagation.StrictScopeDecorator;
 import brave.propagation.ThreadLocalCurrentTraceContext;
 import com.google.common.base.Charsets;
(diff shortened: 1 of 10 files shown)


adriancole:
excellent work!

adriancole:
5.6.10 has this and also I updated all versions of sleuth to have it. so whatever the next version will definitely have it :P

timtebeek:
Awesome! Thanks a lot! :)

PR #960, opened 2019-08-12
Tightens up code and tests around FinishedSpanHandler prior to updates

adriancole:
We will soon add features that use FinishedSpanHandler. This tightens up
code around it so that it is more stable when multiple handlers exist.

### DIFF
--- a/brave/src/main/java/brave/Tracing.java
+++ b/brave/src/main/java/brave/Tracing.java
@@ -26,9 +26,9 @@ import brave.propagation.Propagation;
 import brave.propagation.TraceContext;
 import brave.sampler.Sampler;
 import java.io.Closeable;
-import java.util.ArrayList;
-import java.util.List;
+import java.util.LinkedHashSet;
 import java.util.Locale;
+import java.util.Set;
 import java.util.concurrent.atomic.AtomicBoolean;
 import java.util.logging.Level;
 import java.util.logging.Logger;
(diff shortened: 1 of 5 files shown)


brave/src/main/java/brave/Tracing.java
```diff
@@ -145,7 +145,7 @@ public final Clock clock(TraceContext context) {
     boolean traceId128Bit = false, supportsJoin = true;
     Propagation.Factory propagationFactory = B3Propagation.FACTORY;
     ErrorParser errorParser = new ErrorParser();
-    List<FinishedSpanHandler> finishedSpanHandlers = new ArrayList<>();
+    Set<FinishedSpanHandler> finishedSpanHandlers = new LinkedHashSet<>(); // dupes not ok
```
anuraaga: I think I'd expect an `add` API like this one to accept duplicates. While there probably are very few use cases, if seems more natural.

adriancole: this intentionally prevents it. The issue is that in spring we may have implemented multiple interfaces and do not want to accidentally report data twice. I would be happy to hear complaints vs the otherwise.

adriancole: PS all that's needed if one really really wants to send the same data with a function that has the same identity, is to wrap it. When someone comes up with that use case, point the anger at me :P

adriancole:
meh I will remove the deduper eventhough I don't understand a use case for it. I can try to redesign things such that there won't be a chance of inheritance clash. using abstract class always can guarantee this.

adriancole:
this will result in a bunch of long docs about how to avoid accidental dupes.

PR #963, opened 2019-08-15
Update TracingFilter.java to solve a thread-safe problem

lyghzh:
the code is not safe when executed in several threads simultaneously.for example, if "tracer" is null, now threadA execute "setTracing" method, and tracer is not null. but "extractor" and "injector" has not been initialized, just at that moment threadB come, threadB try to execute "invoke" method,because tracer is not null,threadB will go to 72 line or 74 line.because injector and extractor is null,the code will throw NullPointerException,and the code didn't return "invoker.invoke(invocation)".this will cause dubbo service not invoked.
so i use a volatile flag to solve this problem,and make the code thread-safe.

### DIFF
--- a/instrumentation/dubbo-rpc/src/main/java/brave/dubbo/rpc/TracingFilter.java
+++ b/instrumentation/dubbo-rpc/src/main/java/brave/dubbo/rpc/TracingFilter.java
@@ -46,6 +46,7 @@ public final class TracingFilter implements Filter {
   Tracer tracer;
   TraceContext.Extractor<Map<String, String>> extractor;
   TraceContext.Injector<Map<String, String>> injector;
+  volatile boolean isInit = false;
 
   /**
    * {@link ExtensionLoader} supplies the tracing implementation which must be named "tracing". For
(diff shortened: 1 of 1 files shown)


lyghzh:
Update TracingFilter.java to solve a thread-safe problem

adriancole:
do you have a stack trace from the NullPointerException?

I'm curious if it is that ExtensionLoader is lazy or if this is just a thread visibility issue.

instrumentation/dubbo-rpc/src/main/java/brave/dubbo/rpc/TracingFilter.java
```diff
@@ -56,11 +57,12 @@ public void setTracing(Tracing tracing) {
     tracer = tracing.tracer();
     extractor = tracing.propagation().extractor(GETTER);
     injector = tracing.propagation().injector(SETTER);
+    isInit = true;
   }
 
   @Override
   public Result invoke(Invoker<?> invoker, Invocation invocation) throws RpcException {
-    if (tracer == null) return invoker.invoke(invocation);
+    if (isInit == false) return invoker.invoke(invocation);
```
adriancole: since about extension loader. how about.. `if (!loaded) return invoker.invoke(invocation);`

adriancole:
good enough even without the name change. thanks

instrumentation/dubbo-rpc/src/main/java/brave/dubbo/rpc/TracingFilter.java
lyghzh: yeh, it also works.

lyghzh:
you are welcome.

PR #965, opened 2019-08-15
Stops JmsTracing from clearing headers on the producer side

### LINKED ISSUE #860: [BUG][brave-instrumentation-jms] removes existing message properties (OracleAQ)
Hi there
I did a trial with brave-instrumentation-jms in my spring-boot app which uses Oracle AQ as JMS Implementation.
When I send a message, the b3 tracing header is submitted to the db successfully. But all other message properties I set, where gone.
What I found, was when sending the message the message properties are filtered and there is an exception in https://github.com/openzipkin/brave/blob/master/instrumentation/jms/src/main/java/brave/jms/PropertyFilter.java#L86 when setting the properties again.

oracle.jms.AQjmsException
JMS-110: Property JMSXDeliveryCount not writable

I guess JMSXDeliveryCount  is an oracle specific message property, which can't be modified on client side at all. I've read the comment in the code, with the link to the jms documentation, and I wonder if it is really necessary to clearProperties() before I can set new ones. Or if this ist just true on the consumer side.

adriancole:
Before, we were clearing properties before modifying a message anywhere.
Now, we only do it upon receiving a message.

It is subtle, but JMS spec phrasing allows us to be more precise:

[JMS Spec: Message Properties Section]
> When a client receives a message, its properties are in read-only mode. If a client attempts to set properties at this point, a MessageNotWriteableException is thrown. If clearProperties is called, the properties can now be both read from and written to.

While we currently have no means to verify, this *should* fix #860


adriancole:
@WillemJiang you understand this change? I think it is ok or at least possibly better choice. before we were scared so we cleared everywhere. here we just do it on receive side.

adriancole:
I will rebase over https://github.com/openzipkin/brave/pull/966

WillemJiang:
LGTM.

adriancole:
rebasing now

adriancole:
rebased

instrumentation/jms/src/main/java/brave/jms/PropertyFilter.java
```diff
@@ -100,51 +83,6 @@ static void filterProperties(Message message, Set<String> namesToClear, List<Obj
     }
   }
 
-  static class JMS2 { // sneaky way to delay resolution of JMSRuntimeException
```
adriancole: goodbye classloader hack!

PR #966, opened 2019-08-20
Adds debug logging to JMS call-sites that can fail provider specifically

adriancole:
This adds FINE messages on all call sites where we invoke JMS commands
implicitly, for reasons of parsing or propagation.

This is to investigate a few recent concerns, notably about bytes message mentioned by @Liodene

### DIFF
--- a/instrumentation/jms/src/test/java/brave/jms/ITJms_1_1_TracingMessageConsumer.java
+++ b/instrumentation/jms/src/test/java/brave/jms/ITJms_1_1_TracingMessageConsumer.java
@@ -15,6 +15,7 @@ package brave.jms;
 
 import java.util.Collections;
 import java.util.Map;
+import javax.jms.BytesMessage;
 import javax.jms.JMSException;
 import javax.jms.Message;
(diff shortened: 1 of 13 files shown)


instrumentation/jms/src/test/java/brave/jms/ITJms_1_1_TracingMessageConsumer.java
```diff
@@ -114,27 +127,40 @@ JmsTestRule newJmsTestRule(TestName testName) {
...
+    assertThat(listenerSpan.tags())
+      .hasSize(1) // no redundant copy of consumer tags
+      .containsEntry("b3", "false"); // b3 header not leaked to listener
   }
 
   @Test public void messageListener_resumesTrace() throws Exception {
     messageListener_resumesTrace(() -> messageProducer.send(message), messageConsumer);
   }
 
+  @Test(expected = ComparisonFailure.class) // TODO: https://github.com/openzipkin/brave/issues/967
```
adriancole: finally this is the failing test @Liodene

PR #971, opened 2019-08-23
Orphaned spans leak sensitive data

lambcode:
Found a bug where tags/annotations that are meant to be redacted could leak if the span is orphaned. I think the best way to fix this is to use the regular finishedSpanHandler when reporting orphaned spans. This pull request includes a unit test and the aforementioned fix.

### DIFF
--- a/brave/src/main/java/brave/Tracing.java
+++ b/brave/src/main/java/brave/Tracing.java
@@ -420,7 +420,7 @@ public abstract class Tracing implements Closeable {
         builder.clock,
         builder.propagationFactory,
         FinishedSpanHandlers.noopAware(finishedSpanHandler, noop),
-        new PendingSpans(clock, zipkinHandler, noop),
+        new PendingSpans(clock, finishedSpanHandler, noop),
         builder.sampler,
         builder.currentTraceContext,
         builder.traceId128Bit || propagationFactory.requires128BitTraceId(),
--- a/brave/src/test/java/brave/features/handler/RedactingFinishedSpanHandlerTest.java
+++ b/brave/src/test/java/brave/features/handler/RedactingFinishedSpanHandlerTest.java
@@ -95,4 +95,29 @@ public class RedactingFinishedSpanHandlerTest {
       "SSN=xxx-xx-xxxx"
     );
   }
+
+  @Test public void showRedactionInOrphanedSpan() throws InterruptedException {
+    createAndLeakSpan(); // done in separate method so we can force gc
+    blockOnGC();
+    tracing.tracer().nextSpan().abandon(); //create another span to trigger PendingSpans.reportOrphanedSpans()
+
+    assertThat(spans.get(0).tags()).containsExactly(
+      // SSN tag was nuked
+    );
+    assertThat(spans.get(0).annotations()).flatExtracting(Annotation::value).containsExactly(
+      "SSN=xxx-xx-xxxx",
+      "brave.flush"
+    );
+  }
+
+  private void createAndLeakSpan() {
+    brave.Span span = tracing.tracer().nextSpan();
+    span.tag("b", "912-23-1433");
+    span.annotate("SSN=912-23-1433");
+  }
+
+  private static void blockOnGC() throws InterruptedException {
+    System.gc();
+    Thread.sleep(200L);
+  }
 }


adriancole:
We intentionally didn't use the finished span handler on this path because as the name suggests, these are not finished spans. If we do it this way, it ends up messing with things like statistics which intend to see complete data. Let's find another way

adriancole:
basically using the type could be ok, just we shouldn't use exactly the same collection used for finished spans. make sense?

lambcode:
Thanks for the quick response! Before getting into implementation details I should note that in my use case I have a FinishedSpanHandler that changes the serviceName reported based on some other information. It happened that one of our applications wasn't properly managing spans and we got some abandoned spans reported with an unexpected serviceName. This led me to the discovery that the redaction wouldn't work for these abandoned spans. So, in reality, I am hoping that a user supplied FinishedSpanHandler can also be applied to these abandoned spans. I mention this to set priority expectations appropriately. (i.e. I don't think this super critical issue, at least for my use case)

> we shouldn't use exactly the same collection used for finished spans

As far as I can tell, the collection of FinishedSpanHandlers contains 2 distinct sets of things. All the user-defined FinishedSpanHandlers, and the ZipkinFinishedSpanHandler. It looks like the ZipkinFinishedSpanHandler is what does the actual reporting of the spans to zipkin, so that needs to be included. Any redaction will be defined in a user-defined FinishedSpanHandler, so those look to be required as well.

I see a couple of options going forward
1. Add a method to FinishedSpanHandler to denote whether it should be used for abandoned spans. Something like `handleAbandondSpans()`. Any redacting handlers can then return true and statistic/metric handlers can return false. I like this because it sounds flexible. It would add complexity though. I'm thinking the default value should be false for backwards compatibility.
2. Potentially filtering down the list of FinishedSpanHandlers by type. This would require that any statistic/metric handlers be defined in brave itself. I couldn't find any, so I'm guessing this isn't a viable option.
3. ~~Add a parameter (boolean?) to the `FinishedSpanHandler.handle()` method that can be used to determine whether or not a span was abandoned.~~ Or now that I think about it, just looking for the "brave.flush" annotation would work without adding a parameter. Downside to this updating existing statistic/metric handlers and tribal knowledge needed for correctly creating any new ones.
4. Just use all the span handlers. This is basically my original fix. Maybe metrics being off isn't such a bad thing?

As noted above, I'm leaning toward option 1. I'm open to discussion though ;-)

adriancole:
because the use case are very wide for FinishedSpanHandler instances we can't really assume that because some of the use cases make sense all of them will. So, not comfortable with 4.

I like 1 btw.. maybe have to polish the name, but straight-forward.

5. is allow a separate collection for the abandon path.. mentioning for completeness

If there's a FinishedSpanHandler option on the builder for abandoned spans, someone can choose to add the same collection there. internally we can look if the two collections are identical, then optimize, which gets to first change, just a different way in.

a configuration layer on top (like sleuth) could for example, choose to redundantly add all finished span handlers onto the abandon path, while still allowing the option to not do that.

adriancole:
since we use "supports" prefix elsewhere, if we did a flag, probably `supportsOrphanedSpans()` or something.. abandon is actually a method on span, so we should avoid that.

whatever the case is, if we do an option 1, the metrics example should be refactored to use it, and markdown added like whatever we put on javadoc

anuraaga:
It might also make sense to have a separate handle method `handleOrphaned` instead of adding the flag, and running all handlers. That gives the user a chance to differentiate orphaned if they want to handle them differently. We could add a static method that wraps a handler, delegating that method to `handle` to make it simple to write handlers with the same logic for both types.

adriancole:
I like the idea, though not sure why we would need to have a dispatch helper..

are you thinking same sig as this? except name changed?
```java
  public abstract boolean handle(TraceContext context, MutableSpan span);

  /** Override if you want orphaned spans to be handled as well, for example calling {@link #handle}. */
  public boolean handleOrphaned(TraceContext context, MutableSpan span) { // can't make abstract as breaks api
  }
```

I don't see why a helper is needed to dispatch to handle..

anuraaga:
Yeah thinking that. The dispatch helper would just be some possible icing but indeed probably not so helpful.

adriancole:
I'm having second thoughts.. maybe @lambcode initial design is best. a metrics handler can just guard on duration being present. having one handler chain also solves a later routing problem.. ex you want orphaned data to be routed consistently when there are secondary systems. the simplest thing here seems what @lambcode started with.. and the problems are easier to reason with vs adding variables.

@jeqo @devinsba do you also have thoughts?

lambcode:
Correct me if I'm wrong, but I don't think that looking for a duration would work for spans that are flushed (instead of finished). Maybe we could just add a method such as `isOrphaned()` to MutableSpan?

lambcode:
It's sounding like any direction we go, some FinishedSpanHandlers will need to be updated. Either the metrics handlers will need to be updated, or the redacting handlers will need to be updated. Whatever solution we pursue I think that it should default to running all existing handlers for abandoned spans so that any existing redacting handlers will no longer have this potential to leak information.

Can someone point out where these metric handlers live? I couldn't find any in my brief search yesterday. FWIW, I'll probably be able to update this pull request early next week.

adriancole:
good point about the signal. we already have brave.flushed annotation, but
it would be annoying to look for it.

there's another problem which is the trace context in flushed spans is
partial. this means some finished span handlers that look for extra fields
will spend cycles unnecessarily. for example the routing thing I mentioned
might not work for this reason.

finally the zipkin reporting currently happens on the GC thread. as soon as
we do finished span handling we introduce some additional risk.

it is important to think through all these things. a simple solution is
always best but we need a lot more edge cases handled if we literally
change the behavior of existing stuff. not all handlers are in our repo or
even on github right? this is a public api.

safest way is literally to add a second set of handlers for orphaned spans
and that's it. after sleeping on it I really dont like the amount of if
statements we have to consider otherwise and dont like the complexity on
the main path or introducing new surprises.

anyone who wants to punt handlers for both.. all they need to do is double
register. we dont need to dream up and or inherit complexity for someone
only one person has asked for. I do think it is good to allow something
though

adriancole:
I spent most time writing the javadoc as the code is way simpler. Let's carry on here.. you can feel free to make pull requests against this branch https://github.com/openzipkin/brave/pull/972

PR #972, opened 2019-08-25
Adds FinishedSpanHandler.supportsOrphans()

adriancole:
This allows FinishedSpanHandlers to declare themselves capable of orphaned spans. The high-level goal is to facilitate PII cleaning. Data scrubbing needs to happen regardless of whether data is reported on success or orphan paths.

This continues work of #971 on a shared branch so that others can contribute to finishing it.

Closes #971


brave/src/main/java/brave/internal/recorder/PendingSpans.java
```diff
@@ -120,21 +120,22 @@ void reportOrphanedSpans() {
...
 
+      boolean isEmpty = value.state.isEmpty();
+      Throwable caller = value.caller;
+
       TraceContext context = InternalPropagation.instance.newTraceContext(
         InternalPropagation.FLAG_SAMPLED_SET | InternalPropagation.FLAG_SAMPLED,
         contextKey.traceIdHigh, contextKey.traceId,
         contextKey.localRootId, 0L, contextKey.spanId,
-        Collections.emptyList()
+        caller != null && !isEmpty ? Collections.singletonList(caller) : Collections.emptyList()
```
adriancole: untested change here is adding the calling stack trace into the context, so that a finished span handler could processes it in a way different than logging to hunt down who created the bug.

adriancole:
ps the least complex, and side-effect causing alternative to this so far is adding a method `supportsOrphans()` to the FinishedSpanHandler, defaulting to false.

If we did that route, we'd have to change how composites work to make sure only handlers marked `supportsOrphans()` are added. It also assumes a relationship that orphan handling is a subset of normal handling. I don't like this conflation because it causes tension around operations that only apply to orphans.

For example, in okhttp you have the same interceptor type for network and application layer. having two collections allows you to write interceptors in a specific way for the call site, only considering the circumstances that apply there, and without risk of being invoked the wrong way.

it may seem that configuration would be easier if we had only one collection for `FinishedSpanHandler`, but the cost is complexity in other areas. Especially knowing this is only about bugs, it seems a prudent choice to push that up. Ex in spring, a qualifying annotation could mark a handler for the abandon path, and yeah it can be annoying to add the same handler, like redaction, twice, but this also prevents inlining those decisions. In other words the decision about which handlers do what are purely config time and therefore cannot cause late bugs or runtime complexity.

my 2p and hope it helps.

adriancole:
one last thought.. by adding `mutableSpan.hasAnnotation(String)`, someone can do their own layering to make one handler act differently according to if it was flushed or not, and without adding more heft to the mutableSpan object which has already collected a fair amount of state.

adriancole:
something we could do to make this more effective at big catching is to reverse implement the abandon span logging as a finished span handler. that way we can have the abandoned handler users provide get a hook regardless of if there was data or not. for example you could have a counter for call sites that abandon spans without writing any data. things that leak are important to track, regardless of if they produce data or not. the way things are currently plumbed, only by looking at logs can you tell if leaks occur. if we rejigged the abandon site a little more. you could count all leaks.

adriancole:
Thanks for the detailed notes, which I agree with, @lambcode.

`supportsOrphans()` seems the way out, but keep in mind this assumes that in order to handle orphans you also have to also handle non-orphans. In the rare case that exists, an added method `MutableSpan.hasAnnotation(String)` should work that out without adding more state.

The end goal is that no-one cares about or has to be aware of `supportsOrphans()` when writing a span handler. they don't need to go back and change existing handlers etc as it is false by default. 

Later, the propagation thing about keys not being visible at abandon, we can try something different which is to make it easier to make tags based on custom propagation. At that point you don't need as much state at the end anyway.

Do you agree this approach?

adriancole:
I think I have a way to retain extra fields.. that's the only thing we can safely retain in FinishedSpanHandler as we control the type used, it is package private so unlikely someone will stash a reference and hold up GC.

adriancole:
ok I will do some work to pay off the debt. I think I can redesign the state of MutableSpan to use similar overhead for multiple booleans (shared and orphaned). this will allow a simple check to see if something is orphaned or not. I will also add benchmark comparing GC profile of retaining a reference to extra fields in PendingSpans. If both of these succeed, we can have cake and eat it too.

Firstly, I will change the design to `FinishedSpanHandler.supportsOrphans()` as that allows others to review the api (who may not care about internals)

adriancole:
@lambcode very good point about the maintenance stealing work elsewhere (not on GC thread). this actually makes testing leak detection tricky because you have to use the tracer to trigger maintenance.

brave/src/test/java/brave/features/handler/RedactingFinishedSpanHandlerTest.java
```diff
@@ -86,13 +91,36 @@ static String maybeUpdateValue(String value) {
...
       entry("a", "1"),
       // SSN tag was nuked
       entry("c", "3")
     );
-    assertThat(spans.get(0).annotations()).flatExtracting(Annotation::value).containsExactly(
+    assertThat(finished.annotations()).flatExtracting(Annotation::value).containsExactly(
       "SSN=xxx-xx-xxxx"
     );
+
+    // Leak some data by adding a tag using the same context after the span was finished.
```
adriancole: @narayaruna this is how hard it is to test this :)

adriancole:
test coverage TODO before polishing state around how to decide if you are handling an orphan or not:

* FinishedSpanHandlersTest
 * show throwable doesn't crash

* PendingSpansTest
 * callable present on orphan path when log level fine

* TracingTest
  * add coverage of zipkinReportingFinishedSpanHandler
  * mixed handlers test (supportsOrphans true and false)
    * both Tracing and PendingSpans handlers are no-op aware
    * Tracing handler composes both
    * PendingSpans handler only includes the supports orphans one

adriancole:
@lambcode I think we'll leave extra field handling out of scope. it keeps the change cheaper and less risky to release anyway

adriancole:
@lambcode PTAL I tried to complete the minimal changes to make this work. I don't think you were adding a negative handler (one that exclusively works with abandoned spans), but I made it possible with a new method `MutableSpan.containsAnnotation(String)`

Hopefully, this can get you going as I'd like to divert attention asap to some other in-flight PRs.

Cheers and thanks for all the guidance.

brave-tests/src/test/java/brave/internal/recorder/PendingSpansClassLoaderTest.java
```diff
@@ -43,16 +43,16 @@
...
+    assertRunIsUnloadable(OrphanedContext.class, getClass().getClassLoader());
   }
 
-  static class ErrorReporting implements Runnable {
+  static class OrphanedContext implements Runnable {
     @Override public void run() {
       PendingSpans pendingSpans =
         new PendingSpans(Platform.get().clock(), new FinishedSpanHandler() {
           @Override public boolean handle(TraceContext context, MutableSpan span) {
-            throw new RuntimeException();
```
adriancole: pendingspans no longer takes responsibility of error-guarding. that's the duty of NoopAware... which is used both on finish and abandon paths now.

adriancole:
cc @jonathan-lo as probably this is interesting to you. Basically if there are any hooks trying to scrub data, before there was a path where scrubbing wouldn't apply. This aims to close that gap by allowing scrubbers to tag themselves capable of dealing with orphaned data.

brave/src/main/java/brave/handler/FinishedSpanHandler.java
```diff
@@ -64,9 +65,60 @@
...
 
+  /**
+   * Normally, {@link #handle(TraceContext, MutableSpan)} is only called upon explicit termination
+   * of a span: {@link Span#finish()}, {@link Span#finish(long)} or {@link Span#flush()}. When this
+   * method returns true, the callback will also receive data orphaned due to spans being never
+   * terminated or data added after termination. It is important to understand this, especially if
+   * your handler is performing work like redaction. This sort of work needs to happen on all data,
+   * not just the success paths.
+   *
+   * <p><h3>What is an orphaned span?</h3>
```
anuraaga: If this is zipkin style it's fine to be consistent, but would normally expect to not put the `h3` in the `p` (think browsers move the h3 out of p tags)

```java
<h3>Hello?</h3>

<p>World
```

brave/src/main/java/brave/handler/FinishedSpanHandler.java
```diff
@@ -64,9 +65,60 @@
...
+   * Brave adds an {@link Span#annotate(String) annotation} "brave.flush" when data remains
+   * associated with a span when it is garbage collected. This is almost always a bug. For example,
+   * calling {@link Span#tag(String, String)} after calling {@link Span#finish()}, or calling {@link
+   * Tracer#nextSpan()} yet never using the result. To track down bugs like this, set the logger
+   * {@link PendingSpans} to FINE level.
+   *
+   * <p><h3>Why handle orphans?</h3>
+   * Use cases for handling orphans include redaction, trimming the "brave.flush" annotation,
+   * logging a different way than default, or incrementing bug counters. For example, you could use
+   * the same SSN cleaner here as you do on the success path.
```
anuraaga: Think SSN is social security number but it's a pretty unfamiliar term for most I think. credit card number or password might be more global.

brave/src/main/java/brave/handler/FinishedSpanHandler.java
```diff
@@ -64,9 +65,60 @@
...
+   * When this method returns true, the {@link #handle(TraceContext, MutableSpan) handle method} is
+   * both invoked for normal spans and also orphaned ones. The following apply when handling
+   * orphans:
+   *
+   * <p>The {@link TraceContext} parameter contains minimal information, including lookup ids
+   * (traceId, spanId and localRootId) and sampling status. {@link TraceContext#extra() "extra"}
+   * will be empty.
+   *
+   * <p>The {@link MutableSpan} parameter {@link MutableSpan#containsAnnotation(String) includes
+   * the annotation} "brave.flush", and whatever state was orphaned (ex a tag).
```
anuraaga: I might not be familiar enough with the terms, but what does it mean for a tag to be orphaned? Aren't spans the orphaned ones?

brave/src/test/java/brave/features/handler/RedactingFinishedSpanHandlerTest.java
```diff
@@ -86,13 +91,36 @@ static String maybeUpdateValue(String value) {
...
+
+    // GC only clears the reference to the leaked data. Normal tracer use implicitly handles orphans
+    tracing.tracer().nextSpan().start().finish();
+
+    Span leaked = spans.take();
+    assertThat(leaked.tags()).containsExactly(
+      // SSN tag was nuked
+      entry("d", "SSN=xxx-xx-xxxx")
+    );
+    assertThat(leaked.annotations()).flatExtracting(Annotation::value).containsExactly(
```
anuraaga: Maybe use the new method

brave/src/main/java/brave/handler/FinishedSpanHandler.java
adriancole: this was just me fighting intellij's quick documentation thing recently. not the best thing to optimize for. you are definitely right!

brave/src/main/java/brave/handler/FinishedSpanHandler.java
adriancole: ack better idea. will change

brave/src/main/java/brave/handler/FinishedSpanHandler.java
adriancole: terminology problem maybe you can help solve.

it is orphaned data associated with a span id. doesn't matter what kind, whether it is a single tag, a late name change, a redundant finish timestamp (though we guard against that)

so..
orphaned data
orphaned span

meh maybe better to be close as possible to right and change consistently to "orphaned span" at least we can defined it once. sgtm!

brave/src/test/java/brave/features/handler/RedactingFinishedSpanHandlerTest.java
adriancole: agreed

adriancole: the method is on mutablespan not zipkin2.span :P lemme see if it makes sense to refactor the test..

adriancole: meh creates some busy work to do that as most assertions work more naturally against zipkin2.Span

adriancole:
thanks for the instant review @anuraaga!

brave/src/main/java/brave/Tracing.java
```diff
@@ -393,34 +393,30 @@ public Tracing build() {
...
-        case 0:
-          finishedSpanHandler = FinishedSpanHandler.NOOP;
-          break;
-        case 1:
-          finishedSpanHandler = finishedSpanHandlers.iterator().next();
-          break;
-        default:
-          finishedSpanHandler = FinishedSpanHandlers.compose(finishedSpanHandlers);
+      FinishedSpanHandler orphanedSpanHandler = finishedSpanHandler;
+      if (!builder.finishedSpanHandlers.equals(orphanedSpanHandlers)) {
```
anuraaga: It took a double-take (literally a look in a second review pass) to properly understand this comparison for some reason. Maybe clearer to set a boolean flag `notAllHandlersSupportOrphans` or something in the above loop instead.

brave/src/main/java/brave/internal/handler/NoopAwareFinishedSpanHandler.java
```diff
@@ -0,0 +1,117 @@
...
+    if (handlers.isEmpty()) return FinishedSpanHandler.NOOP;
+
+    if (handlers.size() == 1) {
+      FinishedSpanHandler onlyHandler = handlers.get(0);
+      return onlyHandler == FinishedSpanHandler.NOOP ? onlyHandler : new Single(onlyHandler, noop);
+    }
+
+    int i = 0;
+    boolean alwaysSampleLocal = false, supportsOrphans = false;
+    FinishedSpanHandler[] copy = new FinishedSpanHandler[handlers.size()];
```
anuraaga: May as well use toArray to make the copy and avoid the mysterious `i`

brave/src/test/java/brave/features/handler/RedactingFinishedSpanHandlerTest.java
```diff
@@ -59,40 +63,69 @@ static String maybeUpdateValue(String value) {
...
-  List<Span> spans = new ArrayList<>();
+  BlockingQueue<Span> spans = new LinkedBlockingQueue<>();
+  FinishedSpanHandler redacter = new FinishedSpanHandler() {
+    @Override public boolean handle(TraceContext context, MutableSpan span) {
+      span.forEachTag(ValueRedactor.INSTANCE);
+      span.forEachAnnotation(ValueRedactor.INSTANCE);
+      return true;
+    }
+
+    @Override public boolean supportsOrphans() {
```
anuraaga: Just to make sure do we have a test that does gc and makes sure orphaned spans aren't processed by non-supporting handlers?

brave/src/test/java/brave/features/handler/RedactingFinishedSpanHandlerTest.java
```diff
@@ -59,40 +63,69 @@ static String maybeUpdateValue(String value) {
...
+      entry("d", "cc=xxx-xx-xxxx")
     );
+    assertThat(leaked.annotations()).flatExtracting(Annotation::value).containsExactly(
+      "brave.flush"
+    );
+  }
+
+  static void blockOnGC() throws InterruptedException {
+    System.gc();
+    Thread.sleep(200L);
```
anuraaga: To make sure the test doesn't take longer than it needs to, can consider `await` on a weak reference becoming null

adriancole:
Thanks @anuraaga and @devinsba for the tips.. I'll work them in.

Much appreciated @lambcode for the spike and all the review passes, especially testing it out. Feel free to raise follow-up issues. Even if we don't action it, please do open an issue describing the scenario for extra, but not extra fields on the orphan path.

brave/src/test/java/brave/features/handler/RedactingFinishedSpanHandlerTest.java
```diff
@@ -59,40 +63,69 @@ static String maybeUpdateValue(String value) {
...
+      "cc=xxx-xx-xxxx"
+    );
+
+    // Leak some data by adding a tag using the same context after the span was finished.
+    tracing.tracer().toSpan(span.context()).tag("d", "cc=4121-2319-1483-3421");
+    span = null; // Orphans are via GC, to test this, we have to drop any reference to the context
+    blockOnGC();
+
+    // GC only clears the reference to the leaked data. Normal tracer use implicitly handles orphans
+    tracing.tracer().nextSpan().start().finish();
```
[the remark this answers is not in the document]
adriancole: agreed

brave/README.md
```diff
@@ -534,6 +534,10 @@ tracingBuilder.addFinishedSpanHandler(new FinishedSpanHandler() {
     );
     return true; // retain the span
   }
+
```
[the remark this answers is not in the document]
adriancole: good catch

brave/src/test/java/brave/features/handler/RedactingFinishedSpanHandlerTest.java
```diff
@@ -59,40 +63,69 @@ static String maybeUpdateValue(String value) {
...
+    assertThat(leaked.tags()).containsExactly(
+      // credit card tag was nuked
+      entry("d", "cc=xxx-xx-xxxx")
     );
+    assertThat(leaked.annotations()).flatExtracting(Annotation::value).containsExactly(
+      "brave.flush"
+    );
+  }
+
+  static void blockOnGC() throws InterruptedException {
```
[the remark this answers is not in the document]
adriancole: it is across two different modules, which would mean two more top-level types to hold it. feel free if you like, but I don't really feel like doing this right now.

brave/src/test/java/brave/features/handler/RedactingFinishedSpanHandlerTest.java
adriancole: I'm moving this and similar to a new help-wanted issue as this isn't new, but would be nice.

adriancole: https://github.com/openzipkin/brave/issues/975

brave/src/test/java/brave/features/handler/RedactingFinishedSpanHandlerTest.java
adriancole: we have a test that the Tracing object will not put unsupported handlers in the wrong place. https://github.com/openzipkin/brave/pull/972/files#diff-bce4e281981560a5111f2416f5f3413bR265

brave/src/main/java/brave/internal/handler/NoopAwareFinishedSpanHandler.java
adriancole: yeah good idea.. this was more drift from previous versions than anything else.

PR #976, opened 2019-08-28
Dubbo 2.7.x

icom2014:
1.support dubbo 2.7.x
2.fix the issuse :
     when A service invoke B service,B service then invoke C service,the parentId  of C service span  is A. because   invocation add attachments from rpcContext in org.apache.dubbo.rpc.protocol.AbstractInvoker(line 141),so what we do will be override


adriancole:
This is more about apache dubbo which is basically so different version should be called 3.0

See this comment https://github.com/openzipkin/brave/pull/939#issuecomment-520223131

I'm happy to review and merge without breaking old code. Please make a new module exactly same but called "brave-instrumentation-dubbo" using package "brave.dubbo" instead of "brave.dubbo.rpc"

adriancole:
later we simply stop publishing "brave-instrumentation-dubbo-rpc" (alibaba api) and continue publishing  "brave-instrumentation-dubbo" apache api

shakuzen:
I agree making a separate module seems like the best way forward. This will unblock people wanting Brave instrumentation of the new Dubbo and not break those still using the previous version of Dubbo with the latest versions of Brave.

icom2014:
submit later

adriancole:
want to make sure the problem described here is solved in apache dubbo also (with test) https://github.com/openzipkin/brave/pull/985

instrumentation/dubbo/src/main/java/brave/dubbo/TracingFilter.java
```diff
@@ -0,0 +1,169 @@
...
+
+    RpcContext rpcContext = RpcContext.getContext();
+    Kind kind = rpcContext.isProviderSide() ? Kind.SERVER : Kind.CLIENT;
+    final Span span;
+    if (kind.equals(Kind.CLIENT)) {
+      span = tracer.nextSpan();
+      // if use  invocation.getAttachments(),when A service invoke B service,B service then invoke C service,the parentId  of C service span  is A
+      // because   invocation add attachments from rpcContext in org.apache.dubbo.rpc.protocol.AbstractInvoker(line 141),so what we do will be override
+      //injector.inject(span.context(), invocation.getAttachments());
+      Map<String,String> temp=new HashMap<>();
```
adriancole: Instead make SETTER use Attachment.put

instrumentation/pom.xml
```diff
@@ -36,6 +36,7 @@
     <module>http</module>
     <module>http-tests</module>
     <module>dubbo-rpc</module>
+	  <module>dubbo</module>
```
adriancole:
```suggestion
    <module>dubbo</module>
```

instrumentation/dubbo/README.md
```diff
@@ -0,0 +1,69 @@
+# brave-instrumentation
```
adriancole:
```suggestion
# brave-instrumentation-dubbo
```

instrumentation/dubbo/README.md
```diff
@@ -0,0 +1,69 @@
...
+This is a tracing filter for RPC providers and consumers in [Dubbo 2.7+](http://dubbo.apache.org/en-us/docs/dev/impls/filter.html)
```
adriancole:
```suggestion
This is a tracing filter for RPC providers and consumers in [Apache Dubbo 2.7+](http://dubbo.apache.org/en-us/docs/dev/impls/filter.html)
```

instrumentation/dubbo/pom.xml
```diff
@@ -26,8 +26,8 @@
 
   <properties>
     <main.basedir>${project.basedir}/../..</main.basedir>
-    <main.java.version>1.6</main.java.version>
-    <main.signature.artifact>java16</main.signature.artifact>
+    <main.java.version>1.8</main.java.version>
```
adriancole: is dubbo itself minimum java 1.8 now? please check

adriancole: Nevermind I checked https://github.com/apache/dubbo/commit/6cab4ef1777f2f9d11d70848db2d3d4445603fa6

instrumentation/dubbo/src/main/java/brave/dubbo/TracingFilter.java
```diff
@@ -0,0 +1,165 @@
...
+import org.apache.dubbo.rpc.Filter;
+import org.apache.dubbo.rpc.Invocation;
+import org.apache.dubbo.rpc.Invoker;
+import org.apache.dubbo.rpc.Result;
+import org.apache.dubbo.rpc.RpcContext;
+import org.apache.dubbo.rpc.RpcException;
+import org.apache.dubbo.rpc.protocol.dubbo.FutureAdapter;
+import org.apache.dubbo.rpc.support.RpcUtils;
+import java.net.InetSocketAddress;
+import java.util.HashMap;
```
adriancole: remove
```suggestion
```

adriancole:
thanks. we can do more as separate pull requests!

adriancole:
I had to revert this as tests do not pass in travis. if you want, you can try to make pass f384add9832aad4d74102c174bc03dec6f4b5523

PR #977, opened 2019-08-28
Extract blockOnGc utilities that wait for weak references to clear.

### LINKED ISSUE #975: refactoring of gc related tests
two modules zipkin and zipkin-tests have these lines a few places

```java
  static void blockOnGC() throws InterruptedException {
    System.gc();
    Thread.sleep(200L);
  }
```

@lambcode noticed an opportunity to possibly dedupe for (2?) public utilities, and @anuraaga noticed an opportunity to block on a weak reference clearing to save some millis which would eventually accumulate to second across all call sites.

I it is possible that in order to have this do both, we'd need a couple methods, or perhaps two identical types that deal with the PendingSpans object.

I'm raising this issue as help-wanted as currently there are both much slower tests and also adding a couple new public types is arguably not terribly different that a few copy pasta. That said, someone can do it if someone else also willing to spend time on review and polish.

anuraaga:
Fixes #975

/cc @lambcode

### DIFF
--- a/brave-tests/pom.xml
+++ b/brave-tests/pom.xml
@@ -44,6 +44,10 @@
       <groupId>org.assertj</groupId>
       <artifactId>assertj-core</artifactId>
     </dependency>
+    <dependency>
+      <groupId>org.awaitility</groupId>
+      <artifactId>awaitility</artifactId>
+    </dependency>
   </dependencies>
 
   <build>
--- a/brave-tests/pom.xml
+++ b/brave-tests/pom.xml
@@ -44,10 +44,6 @@
       <groupId>org.assertj</groupId>
       <artifactId>assertj-core</artifactId>
     </dependency>
-    <dependency>
-      <groupId>org.awaitility</groupId>
-      <artifactId>awaitility</artifactId>
-    </dependency>
   </dependencies>
 
   <build>
--- a/brave-tests/src/main/java/brave/test/util/ClassLoaders.java
+++ b/brave-tests/src/main/java/brave/test/util/ClassLoaders.java
@@ -24,7 +24,6 @@ import java.net.URLClassLoader;
 import java.util.function.Consumer;
 import java.util.function.Supplier;
 import java.util.logging.LogManager;
-import org.awaitility.core.ConditionTimeoutException;
 
 import static org.assertj.core.api.Assertions.assertThat;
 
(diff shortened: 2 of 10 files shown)


brave-tests/pom.xml
```diff
@@ -44,6 +44,10 @@
       <groupId>org.assertj</groupId>
       <artifactId>assertj-core</artifactId>
     </dependency>
+    <dependency>
+      <groupId>org.awaitility</groupId>
```
adriancole: I really don't like adding this as a new dep everyone depends on this now needs to pull in. I'm aware you are a fan of this, but personally I'm not.

brave-tests/src/main/java/brave/test/util/ClassLoaders.java
```diff
@@ -94,23 +95,17 @@ public static void assertRunIsUnloadable(Class<? extends Runnable> runnable, Cla
       throw new AssertionError(e);
     }
 
-    blockOnGC();
+    try {
+      GarbageCollectors.blockOnGC(loader);
+    } catch (ConditionTimeoutException t) {
```
adriancole: this exception bites me many times which is one of the reasons I dislike awaitility. it crashes the test instead of failing it.

brave-tests/pom.xml
adriancole: another option is to see if this is needed in the main deps. if this can be moved to test scope, ok by me.

anuraaga:
Removed awaitility - I can remove the polling too if we're nervous about it.

brave-tests/src/test/java/brave/internal/recorder/PendingSpansClassLoaderTest.java
```diff
@@ -60,18 +61,16 @@
       pendingSpans.getOrCreate(context, true);
       context = null; // orphan the context
 
+      System.gc();
       try {
-        blockOnGC();
+        // We usually block on a weak reference being cleared, but here we would end up loading
+        // the WeakReference class from the classloader preventing it from unloading.
```
adriancole: non-blocking comment: is the WeakReference class the problem here? or is it the type it holds

brave/src/test/java/brave/features/handler/RedactingFinishedSpanHandlerTest.java
```diff
@@ -76,8 +78,17 @@ static String maybeUpdateValue(String value) {
     }
   };
 
+  // Only handles non-orphans
+  FinishedSpanHandler markFinished = new FinishedSpanHandler() {
+    @Override public boolean handle(TraceContext context, MutableSpan span) {
+      span.tag("oliver-twist", "false");
```
adriancole: good tag

brave-tests/src/test/java/brave/internal/recorder/PendingSpansClassLoaderTest.java
anuraaga: Pretty sure it's the `WeakReference` class itself - it's a bit of conjecture on my part, though, not exactly sure why trying to block on weak reference never works here.

PR #978, opened 2019-08-28
Adds TracingCustomizer and HttpTracingCustomizer

### LINKED ISSUE #973: add TracingCustomizer, HttpTracingCustomizer, and in future MessagingTracingCustomizer
In dependency injection such as spring boot (cc @marcingrzejszczak),  guice/dropwizard (cc @jplock) and dagger (cc @anuraaga), there is a common pattern for sharing changes to a builder. `XCustomizer`. This allows decoupling of configuration, where a combination of user and framework change sets can combine.

This also, and importantly in the case of @lambcode's recent issue #970, allows you to change some configuration that has no "hole punched yet" in the configuration layer. For example, if we added a new method on Tracing.Builder, otherwise you have an immediate wrestling match about which versions of the config layer need to immediately update.

I suggest we package java 6 bytecode compatible add TracingCustomizer, HttpTracingCustomizer, and in future MessagingTracingCustomizer interfaces, which are a function of their respective builders. This will allow anything as early as spring 2.5 to immediately take advantage of any new changes allowed by brave, reducing tension on feature normalization.

cc'ing a few usual suspects also @devinsba @shakuzen @jeqo @llinder

adriancole:
Fixes #973

### DIFF
--- a/brave/src/main/java/brave/TracingCustomizer.java
+++ b/brave/src/main/java/brave/TracingCustomizer.java
@@ -19,9 +19,9 @@ import zipkin2.reporter.Reporter;
 /**
  * This allows configuration plugins to collaborate on building an instance of {@link Tracing}.
  *
- * <p>Notably, this decouples framework-level configuration (such as {@link
- * Tracing.Builder#spanReporter(Reporter)}) from site-specific configuration (such as {@link
- * Tracing.Builder#addFinishedSpanHandler(FinishedSpanHandler) additional finished span handlers}.
+ * <p>For example a customizer can configure {@link Tracing.Builder#addFinishedSpanHandler(FinishedSpanHandler)
+ * finished span handlers} without having to also configure {@link Tracing.Builder#spanReporter(Reporter)
+ * span reporting}.
  *
  * <h3>Integration examples</h3>
  *
--- a/brave/src/main/java/brave/TracingCustomizer.java
+++ b/brave/src/main/java/brave/TracingCustomizer.java
@@ -19,7 +19,7 @@ import zipkin2.reporter.Reporter;
 /**
  * This allows configuration plugins to collaborate on building an instance of {@link Tracing}.
  *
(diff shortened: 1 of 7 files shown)


brave/src/main/java/brave/TracingCustomizer.java
```diff
@@ -0,0 +1,43 @@
...
+ */
+package brave;
+
+import brave.handler.FinishedSpanHandler;
+import zipkin2.reporter.Reporter;
+
+/**
+ * This allows configuration plugins to collaborate on building an instance of {@link Tracing}.
+ *
+ * <p>Notably, this decouples framework-level configuration (such as {@link
```
anuraaga: It's not so obvious how these decouple the configuration. Maybe `this can be used to decouple blah and blah by having each layer create its own customizer and applying them all to construct the final Tracing`

adriancole: sgtm!

adriancole: I think I might add a code example, as I'm struggling with words that don't also repeat the first sentence. Thanks for helping me think through this.

adriancole:
I forgot to implement this in spring-beans also..

brave/src/main/java/brave/TracingCustomizer.java
adriancole: I changed the wording instead. lemme know if this works.

adriancole:
I added spring integration and tests.

adriancole:
PS I think will need to follow up with an ExtraFieldsCustomizer for the same reason cc @marcingrzejszczak 

For example, one should be able to assign additional extra fields without changing the underlying default propagation mechanism

adriancole:
https://github.com/openzipkin/brave/pull/981 for extra fields

adriancole:
fyi https://github.com/spring-cloud/spring-cloud-sleuth/pull/1436

PR #979, opened 2019-08-29
Compare tags to actual exception to decouple from gRPC implementation…

anuraaga:
… details.

Noticed in https://travis-ci.org/openzipkin/brave/builds/578265563?utm_medium=notification&utm_source=email what appears to be a flaky gRPC status code. Having a comparison against a fixed code seems to be more coupled to gRPC than we need anyways, so I tried decoupling here. I removed onTransportException_reportsSpan which didn't seem to do much to make this change simpler.


adriancole:
just checked.. yeah it uses this seems win (knock on wood)

PR #981, opened 2019-08-30
Adds ExtraFieldCustomizer for ... customizing extra fields

adriancole:
This will allow you to change field setup without accidentally
overriding the trace propagation format.

Like #978, except for extra fields. I realize this dramatically
simplifies our spring 2.5 variant of brave-webmvc-example along the way.

PR #983, opened 2019-08-31
Adds CurrentTraceContextCustomizer for flexible MDC integration

adriancole:
Currently, we have tools which simultaneously configure extra fields
and also integrate them with MDC (for example Sleuth does this). This
adds `CurrentTraceContextCustomizer` to allow integrations like this
to be self-contained.

See https://github.com/openzipkin/brave/pull/981#pullrequestreview-282036426

### DIFF
--- a/brave/src/main/java/brave/propagation/CurrentTraceContext.java
+++ b/brave/src/main/java/brave/propagation/CurrentTraceContext.java
@@ -46,7 +46,11 @@ public abstract class CurrentTraceContext {
   public abstract static class Builder {
     ArrayList<ScopeDecorator> scopeDecorators = new ArrayList<>();
 
-    /** Implementations call decorators in order to add features like log correlation to a scope. */
+    /**
+     * Implementations call decorators in order to add features like log correlation to a scope.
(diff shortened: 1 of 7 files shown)


adriancole:
merging for easier feedback in sleuth (use existing snapshot repo instead of setting up jitpack). feedback later all good.

adriancole:
fyi https://github.com/spring-cloud/spring-cloud-sleuth/pull/1436

brave/src/main/java/brave/propagation/CurrentTraceContext.java
```diff
@@ -46,7 +46,11 @@
   public abstract static class Builder {
     ArrayList<ScopeDecorator> scopeDecorators = new ArrayList<>();
 
-    /** Implementations call decorators in order to add features like log correlation to a scope. */
+    /**
+     * Implementations call decorators in order to add features like log correlation to a scope.
+     *
+     * @since 5.2
```
shakuzen: yay for `@since` JavaDocs! 👍

Issue #982, opened 2019-08-30
kafka-streams instrumentation should not reuse the delegates it wraps


jorgheymans:
The Tracing* variants of `ProcessorSupplier`, `TransformerSupplier`, `ValueTransformerWithKeySupplier` etc all accept the respective delegate components in their constructor. 

Following https://docs.confluent.io/current/streams/faq.html#why-do-i-get-an-illegalstateexception-when-accessing-record-metadata this can cause issues

> If you attach a new Processor/Transformer/ValueTransformer to your topology using a corresponding supplier, you need to make sure that the supplier returns a new instance each time get() is called.

It surfaced in my use case due to `context().headers()` throwing NPE when called *inside* of `transform()`, something that should not happen.

So I would suggest that we accept the respective Suppliers in the constructors instead of the stream components itself. Would 'break' API but since this module is experimental ... I can provide a PR if you think this is the right direction, let me know.

jorgheymans:
Issue confirmed by Kafka peeps: https://lists.apache.org/thread.html/b6d76b67f34c5caf699bb6306c9b8427fdf2ebf03f0f72fd4ce50d54@%3Cusers.kafka.apache.org%3E

PR #984, opened 2019-09-01
using suppliers

jorgheymans:
Proposal fix for https://github.com/openzipkin/brave/issues/982 

@jeqo WDYT ?

### DIFF
--- a/instrumentation/kafka-streams/src/main/java/brave/kafka/streams/KafkaStreamsTracing.java
+++ b/instrumentation/kafka-streams/src/main/java/brave/kafka/streams/KafkaStreamsTracing.java
@@ -19,6 +19,7 @@ import brave.Tracing;
 import brave.kafka.clients.KafkaTracing;
 import brave.propagation.TraceContext;
 import brave.propagation.TraceContextOrSamplingFlags;
+import java.util.Properties;
 import org.apache.kafka.common.header.Headers;
 import org.apache.kafka.streams.KafkaClientSupplier;
 import org.apache.kafka.streams.KafkaStreams;
@@ -37,8 +38,6 @@ import org.apache.kafka.streams.kstream.ValueTransformerWithKeySupplier;
 import org.apache.kafka.streams.processor.AbstractProcessor;
 import org.apache.kafka.streams.processor.ProcessorContext;
 import org.apache.kafka.streams.processor.ProcessorSupplier;
-
-import java.util.Properties;
 
 /** Use this class to decorate Kafka Stream Topologies and enable Tracing. */
 public final class KafkaStreamsTracing {
--- a/instrumentation/kafka-streams/src/main/java/brave/kafka/streams/KafkaStreamsTracing.java
+++ b/instrumentation/kafka-streams/src/main/java/brave/kafka/streams/KafkaStreamsTracing.java
@@ -19,7 +19,6 @@ import brave.Tracing;
 import brave.kafka.clients.KafkaTracing;
 import brave.propagation.TraceContext;
 import brave.propagation.TraceContextOrSamplingFlags;
-import java.util.Properties;
 import org.apache.kafka.common.header.Headers;
 import org.apache.kafka.streams.KafkaClientSupplier;
 import org.apache.kafka.streams.KafkaStreams;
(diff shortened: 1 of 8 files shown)


instrumentation/kafka-streams/src/test/java/brave/kafka/streams/ITKafkaStreamsTracing.java
```diff
@@ -191,17 +191,17 @@ public void should_create_spans_from_stream_input_and_output_topics() throws Exc
   public void should_create_spans_from_stream_with_tracing_processor() throws Exception {
```
jeqo: would love to have an additional case--probably on the current master--reproducing the current issue, so we are sure it is fixed and we don't step into that again.

instrumentation/kafka-streams/src/main/java/brave/kafka/streams/KafkaStreamsTracing.java
```diff
@@ -98,13 +95,13 @@ public KafkaStreams kafkaStreams(Topology topology, Properties streamsConfig) {
...
    * builder.stream(inputTopic)
-   *        .process(kafkaStreamsTracing.processor("my-processor", myProcessor);
+   *        .process(kafkaStreamsTracing.processor("my-processor", myProcessorSupplier);
    * }</pre>
    *
    * @see TracingKafkaClientSupplier
    */
-  public <K, V> ProcessorSupplier<K, V> processor(String spanName, Processor<K, V> processor) {
-    return new TracingProcessorSupplier<>(this, spanName, processor);
+  public <K, V> ProcessorSupplier<K, V> processor(String spanName, ProcessorSupplier<K, V> processorSupplier) {
```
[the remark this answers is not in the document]
jorgheymans: Yes good idea we should explain this, but maybe in the release notes ? If put on the sources it might look out of date after a while.

instrumentation/kafka-streams/src/test/java/brave/kafka/streams/ITKafkaStreamsTracing.java
jorgheymans: I tried accessing headers from the integration tests before the fix but did not get the exception maybe due to the way the junit rule sets up the kafka instance? It would be nice to be able to get a test case indeed.

instrumentation/kafka-streams/src/main/java/brave/kafka/streams/KafkaStreamsTracing.java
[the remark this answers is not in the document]
jorgheymans: So how about we put this in the release notes @adriancole just so you remember when you're drafting them.

> kafka-streams instrumentation now correctly handles the components (Processors, Transformers, ...) it wraps. Previously the instrumentation would reuse the wrapped component across instances, Kafka mandates however that a new instance is created each time 'get()' is called on the Supplier interface. 
> Note that this fix is not backwards compatible and will cause compilation errors if you are currently tracing Processors or Transformers. Most of the time the fix will consist of just adding a lambda function for the Supplier.

instrumentation/kafka-streams/src/test/java/brave/kafka/streams/ITKafkaStreamsTracing.java
jeqo: I tried to reproduce it as well but I can't manage to make it fail. Could you replicate your failing scenario in a test-case to debug it?  I'm wondering if `new TracingProducessor(delegate)` on our suppliers already protect us from this issue

jorgheymans: this reproduces it, it's really trivial just point it to a cluster (pre-create the topics) and let it run https://github.com/jorgheymans/kafka-streams-supplier-testcase

adriancole:
fyi I'm thinking we should settle whether this does or doesn't fix things soon, as I can help do a release today but not later this weekend as I'm out.

jorgheymans:
> fyi I'm thinking we should settle whether this does or doesn't fix things soon, as I can help do a release today but not later this weekend as I'm out.

it definitely fixes the issue for me, just waiting for @jeqo to confirm he can reproduce the issue.

jeqo:
@jorgheymans I've tested your test project, but haven't managed to reproduce the issue :( 

Just added some comments on the PR: https://github.com/jorgheymans/kafka-streams-supplier-testcase/pull/1

jorgheymans:
here is the stacktrace:

```
Exception in thread "streams-test-case-StreamThread-1" java.lang.IllegalStateException: This should not happen as headers() should only be called while a record is processed
	at org.apache.kafka.streams.processor.internals.AbstractProcessorContext.headers(AbstractProcessorContext.java:149)
	at org.apache.kafka.streams.processor.internals.ForwardingDisabledProcessorContext.headers(ForwardingDisabledProcessorContext.java:147)
	at my.KafkaConfig$1.transform(KafkaConfig.java:75)
	at my.KafkaConfig$1.transform(KafkaConfig.java:63)
	at brave.kafka.streams.TracingValueTransformer.transform(TracingValueTransformer.java:53)
	at org.apache.kafka.streams.kstream.internals.AbstractStream$1.transform(AbstractStream.java:119)
	at org.apache.kafka.streams.kstream.internals.KStreamTransformValues$KStreamTransformValuesProcessor.process(KStreamTransformValues.java:56)
	at org.apache.kafka.streams.processor.internals.ProcessorNode.process(ProcessorNode.java:117)
	at org.apache.kafka.streams.processor.internals.ProcessorContextImpl.forward(ProcessorContextImpl.java:183)
	at org.apache.kafka.streams.processor.internals.ProcessorContextImpl.forward(ProcessorContextImpl.java:162)
	at org.apache.kafka.streams.processor.internals.ProcessorContextImpl.forward(ProcessorContextImpl.java:122)
	at org.apache.kafka.streams.processor.internals.SourceNode.process(SourceNode.java:87)
	at org.apache.kafka.streams.processor.internals.StreamTask.process(StreamTask.java:364)
	at org.apache.kafka.streams.processor.internals.AssignedStreamsTasks.process(AssignedStreamsTasks.java:199)
	at org.apache.kafka.streams.processor.internals.TaskManager.process(TaskManager.java:420)
	at org.apache.kafka.streams.processor.internals.StreamThread.runOnce(StreamThread.java:890)
	at org.apache.kafka.streams.processor.internals.StreamThread.runLoop(StreamThread.java:805)
	at org.apache.kafka.streams.processor.internals.StreamThread.run(StreamThread.java:774)
```

The linenumbers in KafkaConfig (63 and 75) are a bit off from the testcase here because it's from my local project, but as you can see from below screenshot it's within the transformer:

image

adriancole:
for posterity we can open an issue to backfill test with help-wanted and a link to the stack trace and reproduce scenario.

PR #985, opened 2019-09-02
fix dubbo async invoke has no correct trace data

glmapper:
In the Dubbo 2.6.x release series, when ResponseCallBack is used, ResponseCallBack in Brave may be overwritten。[Why doesn't ResponseCallback support multiple?](https://github.com/apache/dubbo/issues/4968)

So, When Dubbo is temporarily unable to provide an onResponse similar to the 2.7.x series, I am trying to provide a new solution.

### DIFF
--- a/instrumentation/dubbo-rpc/src/main/java/brave/dubbo/rpc/TracingFilter.java
+++ b/instrumentation/dubbo-rpc/src/main/java/brave/dubbo/rpc/TracingFilter.java
@@ -18,6 +18,7 @@ import brave.Span.Kind;
 import brave.Tracer;
 import brave.Tracing;
 import brave.internal.Platform;
+import brave.propagation.CurrentTraceContext;
 import brave.propagation.Propagation;
 import brave.propagation.TraceContext;
 import brave.propagation.TraceContextOrSamplingFlags;
@@ -45,6 +46,7 @@ import java.util.concurrent.Future;
 // public constructor permitted to allow dubbo to instantiate this
 public final class TracingFilter implements Filter {
 
+  CurrentTraceContext current;
   Tracer tracer;
   TraceContext.Extractor<Map<String, String>> extractor;
   TraceContext.Injector<Map<String, String>> injector;
@@ -56,14 +58,14 @@ public final class TracingFilter implements Filter {
    * injected.
    */
   public void setTracing(Tracing tracing) {
+    current = tracing.currentTraceContext();
     tracer = tracing.tracer();
     extractor = tracing.propagation().extractor(GETTER);
     injector = tracing.propagation().injector(SETTER);
     isInit = true;
   }
 
-  @Override
-  public Result invoke(Invoker<?> invoker, Invocation invocation) throws RpcException {
+  @Override public Result invoke(Invoker<?> invoker, Invocation invocation) throws RpcException {
     if (isInit == false) return invoker.invoke(invocation);
 
     RpcContext rpcContext = RpcContext.getContext();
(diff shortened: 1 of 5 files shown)


adriancole:
is there any test that can be added to show this works?

glmapper:
> works

Ok, I will provide test cases.

glmapper:
@adriancole you can use this test project [glmapper-boot-dubbo](https://github.com/glmapper/glmapper-boot-dubbo) , there are two test module prefix with dubbo-2.6.x . 

steps:

* run dubbo-2.6.x-provider
* run dubbo-2.6.x-comsumer

debug :
* if you use 5.6.11, debug TracingFilter#FinishSpanCallback#done, This method will not be executed when called asynchronously.
* checkout to my branch, and install , do again. This method works fine.

instrumentation/dubbo-rpc/src/main/java/brave/dubbo/rpc/TracingFilter.java
```diff
@@ -111,6 +115,63 @@ public Result invoke(Invoker<?> invoker, Invocation invocation) throws RpcExcept
     }
   }
 
+  static class FinishSpanCallback implements ResponseCallback {
```
adriancole: can you move this back to same place so easier to tell what has changed?

adriancole:
thank you. can you try to reproduce in ITTracingFilter_Consumer
ITTracingFilter_Producer
so that the tests run every time? This way it won't break later.

instrumentation/dubbo-rpc/src/main/java/brave/dubbo/rpc/TracingFilter.java
glmapper: > can you move this back to same place so easier to tell what has changed?

done

glmapper:
> thank you. can you try to reproduce in ITTracingFilter_Consumer ITTracingFilter_Producer so that the tests run every time? This way it won't break later.



> thank you. can you try to reproduce in ITTracingFilter_Consumer ITTracingFilter_Producer so that the tests run every time? This way it won't break later.

see [e57b4d1](https://github.com/openzipkin/brave/pull/985/commits/e57b4d1d82b96441ab06812b4af8cf5dd5d4830a)

instrumentation/dubbo-rpc/src/test/java/brave/dubbo/rpc/ITTracingFilter_Consumer.java
```diff
@@ -186,4 +188,20 @@
...
+
+
+  /**
+   * It will not be throws `Span was not reported` message, But before that, it will be
+   * @throws Exception
+   */
+  @Test public void test_async_invoke() throws Exception {
+    client.setAsync(true);
+    String jorge = client.get().sayHello("jorge");
+    Assert.assertTrue(jorge == null);
```
adriancole: yes.. it is true that @jeqo is not null.

adriancole:
Thank you. I will change some style and formatting when merging. The code looks good!

instrumentation/dubbo-rpc/src/main/java/brave/dubbo/rpc/TracingFilter.java
```diff
@@ -96,7 +98,9 @@ public Result invoke(Invoker<?> invoker, Invocation invocation) throws RpcExcept
       Future<Object> future = rpcContext.getFuture(); // the case on async client invocation
       if (future instanceof FutureAdapter) {
         deferFinish = true;
-        ((FutureAdapter) future).getFuture().setCallback(new FinishSpanCallback(span));
```
adriancole: I'm concerned about this change. Before, we set a callback. This ensures the span is finished in case user never sets a callback. Now, we don't. I think this is dangerous as it could lead to leaked spans. I will push a commit to your branch for review.

adriancole: tests were not running. this is why we didn't notice. I will fix the build. I've already fixed the code on this branch.

adriancole:
so I think this code is good to merge

PR #986, opened 2019-09-02
Adds HttpServerRequest and HttpServerResponse types

adriancole:
Initially, we had a partial function of (adapter, request) to parse attributes
like the HTTP path from a request. This made sense in the http abstraction, as
all types that needed these attributes, such as parsers and samplers, were in
the same package and library. It also saved an object allocation otherwise
needed to wrap a native type, and the indirection of unwrapping as needed.

This worked, but especially on the response path this broke down. We found that
attributes in the request, such as the method and matched route, are needed at
response time, but aren't themselves a part of the response. Over time, we
accumulated response wrappers or thread locals to pass these attributes to the
response as many response objects lacked a property bag.

Then, we noticed people wanting to do more during extraction than look at
headers. For example, Netflix wanted to inspect the http path to make a
secondary sampling decision. Without a known http request type, this couldn't
be accomplished portably.

Finally, we noticed zipkin-go side-step this problem by defining a top-level
type for http requests and responses. This allows the same object to be used
regardless of purpose, whether that is primary or secondary sampling, or tag
parsing.

All of this led to the introduction of `HttpServerRequest` and
`HttpServerResponse` types, retrofitted into the original `HttpServerAdapter`
so as to not break API.

Thanks @basvanbeek for the design help and @jeqo for the review

See #958


instrumentation/http/src/main/java/brave/http/HttpServerHandler.java
```diff
@@ -61,14 +75,36 @@
     );
     this.tracer = httpTracing.tracing().tracer();
     this.sampler = httpTracing.serverSampler();
+    this.defaultExtractor = httpTracing.tracing().propagation().extractor(HttpServerRequest.GETTER);
+    this.defaultHandler = adapter == HttpServerAdapter.DEFAULT
```
adriancole: this is tricky games so we can keep portable with old code, yet use the new `public Span handleReceive(HttpServerRequest request)` method

instrumentation/http/src/main/java/brave/http/HttpServerRequest.java
```diff
@@ -0,0 +1,90 @@
...
+import brave.propagation.Propagation;
+
+/**
+ * Marks an interface for use in {@link HttpServerHandler#handleReceive(HttpServerRequest)}. This
+ * gives a standard type to consider when parsing an incoming context.
+ *
+ * @since 5.7
+ */
+public abstract class HttpServerRequest {
+  static final Propagation.Getter<HttpServerRequest, String> GETTER =
```
adriancole: Now, we pretty much only need one GETTER for http (we'll still need specific types for RPC like gRPC and dubbo, as well GETTERs for messaging until the abstraction is done)

instrumentation/netty-codec-http/src/main/java/brave/netty/http/WrappedHttpRequest.java
```diff
@@ -0,0 +1,61 @@
...
+    if (remoteAddress.getAddress() == null) return false;
+    return span.remoteIpAndPort(Platform.get().getHostString(remoteAddress),
+      remoteAddress.getPort());
+  }
+
+  @Override public String method() {
+    return request.method().name();
+  }
+
+  @Override public String path() {
```
adriancole: PS these TODO comments are old, just moved from another file

instrumentation/servlet/src/main/java/brave/servlet/WrappedHttpServletRequest.java
```diff
@@ -0,0 +1,62 @@
...
+ */
+package brave.servlet;
+
+import brave.Span;
+import brave.http.HttpServerAdapter;
+import brave.http.HttpServerRequest;
+import javax.servlet.http.HttpServletRequest;
+
+/** @since 5.7 */
+// public for others like sparkjava to use
```
devinsba: Instead of public maybe we define a `ServletTracing` class that has factory methods. I was thinking the Runtime class maybe but that isn't public and has a lot of stuff that shouldn't be public

adriancole: actually I don't think there is huge value in sharing this type at all anymore. especially now sparkjava has a route type and we no longer support direct instrumentation with mvc.

I'll hide it.

instrumentation/http/src/main/java/brave/http/HttpServerHandler.java
anuraaga: Not sure I understand the game but should adapter be passed in on the line below? Confused why we can ignore the adapter if it's not default

instrumentation/http/src/main/java/brave/http/HttpServerRequest.java
```diff
@@ -0,0 +1,65 @@
...
+      return carrier.header(key);
+    }
+
+    @Override public String toString() {
+      return "HttpServerRequest::header";
+    }
+  };
+
+  /** Returns the underlying http request object. */
+  public abstract Object unwrap();
```
anuraaga: Can consider `<T> T` so users don't have to cast explicitly

adriancole: if we do that, we can't covariant override, which I think we do somewhere. Plus `<T> T` is a little dodgy as it is less explicit that the type could be not what you think. instanceof is definitely safer right?

adriancole:
> Not sure I understand the game but should adapter be passed in on the line below? Confused why we can ignore the adapter if it's not default

I'll put a comment in here. main thing this does is remain compatible with old signatures and without copy/pasting a lot of code for the default path

adriancole:
@anuraaga thanks for the review, I added some docs which hopefully explain both the internals of the constructor and the rationale of unwrap.

instrumentation/http/src/main/java/brave/http/HttpServerRequest.java
adriancole: rewrote the javadoc.

adriancole:
PS some rationale about defensiveness in terms of preserving compat and against logic duplication:

A breaking api change as it makes too much work on the ecosystem, hence Brave 6 probably won't happen for a long time, even if we did have hands to do it. While we are changing code here, it is important to remember a lot of 3rd party projects use this layer. Some of them commercial (like confluence and alfresco). By using the same call paths as before, we avoid making fixes only to the new path and silently breaking others.

Also on the `Object unwrap()` I know that in ratpack and probably also armeria, too, we've had to refactor internals dramatically on occasion. Being really defensive is being honest about this. When layers need to refactor, they should.. especially sometimes the logic is literally broken and you don't know until later. There should be little need for `unwrap()` over time as we implement some of the backlog such as request and response length. This is just a stop-gap.

instrumentation/http/src/main/java/brave/http/HttpServerAdapter.java
```diff
@@ -14,9 +14,57 @@
...
 import brave.Span;
-import brave.propagation.TraceContext;
 
 public abstract class HttpServerAdapter<Req, Resp> extends HttpAdapter<Req, Resp> {
+  /**
+   * Intentionally not exposed as call-sites shouldn't need this directly.
+   *
+   * @since 5.7
+   */
+  static final HttpServerAdapter<HttpServerRequest, HttpServerResponse> DEFAULT =
```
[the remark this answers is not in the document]
adriancole: excellent point.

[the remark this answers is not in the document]
adriancole: added a commit. thx

adriancole:
https://github.com/openzipkin/brave/pull/988 is for the client side

PR #987, opened 2019-09-03
Enables maven-failsafe-plugin and fixes Dubbo tests

adriancole:
Before, we weren't actually running tests prefixed IT. Both of the Dubbo
implementations had test setup problems that weren't noticed.

cc @glmapper @icom2014


adriancole:
dubbo is currently impossible to make pass in travis I give up

PR #988, opened 2019-09-04
Adds HttpClientRequest and HttpClientResponse

adriancole:
Same as #986, but for clients


adriancole:
I think this is the last one we should do prior to minor release

PR #989, opened 2019-09-04
Supports external finish timestamp in HTTP client and server handling

adriancole:
Armeria supports `RequestLog.responseEndTimeNanos()` for more precise
timestamps, which line up with metrics. This punches a hole so that it
is possible to use this without copy/pasting most of the handler code.

cc @trustin @kojilin


adriancole:
reworked I like this better

PR #990, opened 2019-09-05
Uses unwrapped type for http parsers and samplers

adriancole:
While working on Armeria, I noticed that we accidentally missed types
when moving to the standard HTTP client and server responses. What I
mean is that signatures that formerly got an HttpServletRequest now get
an HttpServerRequest that wraps it. This could break people's parsers
without fair warning.

This makes sure that unwrapped types are used in existing call sites
such as HTTP parsers and samplers. It does so by regearing the new types
as adapters of what they wrap.

See https://github.com/line/armeria/pull/2038#discussion_r321133202


adriancole:
this is what this change prevents

```diff
-                                   if (res instanceof HttpClientResponse) {
-                                       Object unwrapped = ((HttpClientResponse) res).unwrap();
-                                       if (unwrapped instanceof RequestLog) {
-                                           customizer.tag("http.url", SpanTags.generateUrl((RequestLog) unwrapped));
-                                       }
+                                   if (res instanceof RequestLog) {
+                                       customizer.tag("http.url", SpanTags.generateUrl((RequestLog) res));
                                    }
```

adriancole:
ps an alternative to retrofitting is allocating an adapter wrapper at each call site. ex a wrapper when processing the request and another when parsing the response. the current approach in this PR will not allocate any more objects. however it is possible that an adapter wrapper would run quicker due to the reduced calls to unwrap()

adriancole:
I will go with the alternative as it is the cleanest for users to see. Having slept on it, I think this is the best balance and will update the RATIONALE.md accordingly

adriancole:
As this doesn't change public api and is fixing a glitch, I'm gonna merge this as it is blocking release.

PR #994, opened 2019-09-25
Adds HttpRequestSampler and backports HttpSampler to use it

adriancole:
We recently introduced the `HttpClientRequest` and `HttpServerRequest` 
types. These are used during propagation extraction, for example as a 
parameter to `Extractor.extract(request)`. This was to facilitate
secondary sampling, which occurs just prior to the primary sampling as
performed by `HttpSampler`.

One feedback was that it is intuitive to use types like 
`HttpRuleSampler` uniformly across primary and secondary sampling.

This change introduces `HttpRequestSampler` which aligns types so that
rules can be defined with similar code regardless of primary or 
secondary decision. This also takes care to not break api. In order to
reduce complexity around starting a span with old or new types, this
pushes some logic to adapters. Rationale is adapters are straightforward
to test and easier to reason with than other logic.

See https://github.com/openzipkin-contrib/zipkin-secondary-sampling/issues/8

### DIFF
--- a/instrumentation/http/src/main/java/brave/http/HttpClientHandler.java
+++ b/instrumentation/http/src/main/java/brave/http/HttpClientHandler.java
@@ -69,7 +69,7 @@ public final class HttpClientHandler<Req, Resp> extends HttpHandler {
   final Tracer tracer;
   @Nullable final HttpClientAdapter<Req, Resp> adapter; // null when using default types
   final Sampler sampler;
-  final HttpSampler httpSampler;
+  final HttpRequestSampler httpSampler;
   @Nullable final Tracer samplingTracer;
   @Nullable final String serverName;
   final Injector<HttpClientRequest> defaultInjector;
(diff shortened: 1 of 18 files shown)


adriancole:
This will need some more work, a lot of it done locally. ignore for now.

adriancole:
the code is done here, just it needs to be polished (test coverage itself). I'll ping when I have a corresponding pull request in the secondary sampling project that uses it.

adriancole:
This all works now, https://github.com/openzipkin-contrib/zipkin-secondary-sampling/pull/18


FYI: @marcingrzejszczak and @trustin @geminiKim @minwoox @ikhoon, I re-used terminology from sleuth as "rate" has caused confusion in the past when applied to a probability. This didn't break api or anything it just impacted naming of method parameters.

adriancole:
NEXT STEP: backfill unit tests

trustin:
Thabks for the heads up. Let us rename rate to probability.

adriancole:
@marcingrzejszczak and @trustin @geminiKim @minwoox @ikhoon FYI: I moved the probability vs rate change to https://github.com/openzipkin/brave/pull/995 in order to reduce the heft of this PR

adriancole:
eventhough the other code was extracted, there's still some coverage issues here due to the increase of adapters.

NEXT STEP: backfill coverage so that it is as good or better than before

PR #996, opened 2019-09-29
Adds putAllRules to parameterized sampling to allow composition

adriancole:
In order to ease api, this adds `HttpRuleSampler.Builder.putAllRules()`, which
allows you to compose rules without knowing what the former ones were.

This also adds `removeRule` and renames the prefix `addRule` to
`putRule` in order to emphasize replacement based on the match
expression.


adriancole:
toBuilder is unneeded when the builder can accept an existing instance to seed rules from: putAllRules

PR #997, opened 2019-09-29
Switches HttpRuleSampler to directly use matchers

adriancole:
This switches the rule sampler to use more flexible matchers instead of pre-canned method signatures:

```java
httpTracingBuilder.serverSampler(HttpRuleSampler.newBuilder()
  .putRule(pathStartsWith("/favicon"), Sampler.NEVER_SAMPLE)
  .putRule(pathStartsWith("/foo"), RateLimitingSampler.create(100))
  .putRule(and(methodIsEqualTo("POST"), pathStartsWith("/bar")), RateLimitingSampler.create(10))
  .build());
```

This allows one use case @narayaruna asked for, which is a query-based rule.

Ex. 100 requests per second that match /play&country=US

```java
    Matcher<HttpRequest> playInTheUSA = request -> {
      if (!"/play".equals(request.path())) return false;
      String url = request.url();
      if (url == null) return false;
      String query = URI.create(url).getQuery();
      return query != null && query.contains("country=US");
    };

    HttpRuleSampler sampler = HttpRuleSampler.newBuilder()
      .putRule(playInTheUSA, RateLimitingSampler.create(100))
      .build();
```

### DIFF
--- a/brave/src/main/java/brave/sampler/Matchers.java
+++ b/brave/src/main/java/brave/sampler/Matchers.java
@@ -1,0 +1,101 @@
+/*
+ * Copyright 2013-2019 The OpenZipkin Authors
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License"); you may not use this file except
+ * in compliance with the License. You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software distributed under the License
+ * is distributed on an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express
+ * or implied. See the License for the specific language governing permissions and limitations under
+ * the License.
+ */
+package brave.sampler;
+
+import java.util.Arrays;
+
+/**
+ * Convenience functions to compose matchers for {@link ParameterizedSampler}.
+ *
+ * @see Matcher
+ * @see ParameterizedSampler
+ * @since 5.8
+ */
+public final class Matchers {
+
+  /** @since 5.8 */
+  public static <P> Matcher<P> alwaysMatch() {
+    return (Matcher<P>) Constants.ALWAYS_MATCH;
+  }
+
+  /** @since 5.8 */
+  public static <P> Matcher<P> neverMatch() {
+    return (Matcher<P>) Constants.NEVER_MATCH;
+  }
+
+  enum Constants implements Matcher<Object> {
+    ALWAYS_MATCH {
+      @Override public boolean matches(Object parameters) {
+        return true;
+      }
+
+      @Override public String toString() {
+        return "matchAll()";
+      }
+    },
+    NEVER_MATCH {
+      @Override public boolean matches(Object parameters) {
+        return false;
+      }
+
+      @Override public String toString() {
+        return "neverMatch()";
+      }
+    }
+  }
+
+  /** @since 5.8 */
+  public static <P> Matcher<P> and(Matcher<P>... matchers) {
+    return composite(matchers, true);
+  }
+
+  /** @since 5.8 */
+  public static <P> Matcher<P> or(Matcher<P>... matchers) {
+    return composite(matchers, false);
+  }
+
+  static <P> Matcher<P> composite(Matcher<P>[] matchers, boolean and) {
+    if (matchers == null) throw new NullPointerException("matchers == null");
+    if (matchers.length == 0) throw new NullPointerException("matchers is empty");
+    for (int i = 0; i < matchers.length; i++) {
+      if (matchers[i] == null) throw new NullPointerException("matchers[" + i + "] == null");
+    }
+    if (matchers.length == 1) return matchers[0];
+    return new Composite<>(matchers, and);
+  }
+
+  static final class Composite<P> implements Matcher<P> {
+    final Matcher<P>[] matchers; // Array ensures no iterators are created at runtime
+    final boolean and;
+
+    Composite(Matcher<P>[] matchers, boolean and) {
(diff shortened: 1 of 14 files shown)


brave/src/main/java/brave/sampler/Matchers.java
```diff
@@ -0,0 +1,101 @@
...
+      this.matchers = Arrays.copyOf(matchers, matchers.length);
+      this.and = and;
+    }
+
+    @Override public boolean matches(P parameters) {
+      boolean or = false;
+      for (Matcher<P> matcher : matchers) {
+        boolean next = matcher.matches(parameters);
+        if (!next && and) return false;
+        if (next) or = true;
```
anuraaga: This being set once and never set back is hard to follow for me (there's a weird relationship between this `or` variable and the `and`).

It's not that much code so I'd just add two implementations, one for or and one for and

brave/src/main/java/brave/sampler/Matchers.java
```diff
@@ -0,0 +1,101 @@
...
+      }
+
+      @Override public String toString() {
+        return "neverMatch()";
+      }
+    }
+  }
+
+  /** @since 5.8 */
+  public static <P> Matcher<P> and(Matcher<P>... matchers) {
```
anuraaga: I don't know if there is a style in Brave, but if not, would recommend providing a version that takes `Iterable` too, converting an `Iterable` to array in Java is annoying enough that it's nice if libraries do it.

adriancole:
thanks @anuraaga I'll look at your comments now. Sorry I was backfilling tests. You can refresh and look again as I believe parts are all in now.

brave/src/main/java/brave/sampler/Matchers.java
adriancole: no problem.

instrumentation/http/src/test/java/brave/http/HttpRequestMatchersTest.java
```diff
@@ -0,0 +1,67 @@
...
+import static brave.http.HttpRequestMatchers.pathStartsWith;
+import static org.assertj.core.api.Assertions.assertThat;
+import static org.mockito.Mockito.when;
+
+@RunWith(MockitoJUnitRunner.class)
+public class HttpRequestMatchersTest {
+  @Mock HttpRequest httpRequest;
+
+  @Test public void methodEquals_matched() {
+    when(httpRequest.method()).thenReturn("GET");
```
anuraaga: Might want to add a test that matches against lowercase, to explicitly document it doesn't match (or alternatively change the implementation to case insensitive).

adriancole: interesting.. haven't run across mixed case method names. I'll document as such

anuraaga: I'm generally nervous on whether the framework I'm using happens to return all caps or lowercase for a method that returns http method. I think Armeria tags (or tagged? not sure of the new parser changed it) lowercase so I might actually assume lowercase here.

brave/src/main/java/brave/sampler/Matchers.java
adriancole: done

instrumentation/http/src/test/java/brave/http/HttpRequestMatchersTest.java
adriancole: tags sure but this is the actual http request type, and I've literally never seen that despite an inordinate amount of libraries. Never had any issues raised in a decade over mixed case http method. I'm not even sure http spec would allow it..

are you sure we should do mixed case? it would only be here as I suspect all over the internet http methods are presumed caps..

adriancole: yeah it is case sensitive. https://tools.ietf.org/html/rfc7231#section-4.1

I think you probably were thinking about tag converters which do often downcase things.

adriancole: @anuraaga this is not a function of tags, it is a function of a request object. Also this logic is just carried over, as since the beginning of the http abstraction it has been uppercase. That means however many 10 libraries haven't reported an issue in years.

I will mark HttpRequest explicitly that it must follow RFC, so if for some weird reason armeria decides to lowercase something it should uppercase it in the adapter. It hasn't so far coerced to downcase else our tests would have failed anyway. Remember we have ITs on conversion of http.method.

anuraaga: I'm good if it's documented - for me it's just that since I've seen both `GET` and `get` in different places, I'd always need either a doc or code to be absolutely sure which one I'm getting from a library.

adriancole: cool I added a commit to clarify. cheers

adriancole:
@anuraaga I ended up needing two more matchers. PTAL

SecondarySampler sits below the HTTP abstraction. It needs to use rules
that accept multiple types, such as RPC requests or Messaging requests.

This adds `Matchers.cast()` and `Matchers.ifInstanceOf` to allow 
navigation, for example, to use an HTTP sampler underneath a sampler 
whose dispatch type is Object.

adriancole:
actually the cast on is unimportant. I'll delete that.

adriancole:
nm I can't get rid of it.. though it can be deferred :P

adriancole:
moved it to the secondary sampling project

adriancole:
here's the change that uses this https://github.com/openzipkin-contrib/zipkin-secondary-sampling/pull/19

PR #998, opened 2019-09-30
Consolidates samplers under the interface SamplerFunction

adriancole:
This allows more re-use by backporting all single-arity samplers to a
new `SamplerFunction` interface. For example, this consolidates
constants and allows base-layer tools such as `DeclarativeSampler` to
be used in any call site.

This deprecates `Tracer.withSampler` in favor of lighter methods that
accomplish the same:

* `Tracer.startScopedSpan(name, sampler, param)`
* `Tracer.nextSpan(sampler, param)`

With this foundation, we can next implement an RpcRequest type and use
it with less copy/paste of documentation and constants.


brave/README.md
```diff
@@ -288,23 +284,21 @@ is. For example, you might not want to trace requests to static resources
 such as images, or you might want to trace all requests to a new api.
 
 Most users will use a framework interceptor which automates this sort of
-policy. Here's how they might work internally.
+policy.
 
+Here's a simplified version of how this might work internally.
 ```java
-Sampler fallback = tracing.sampler();
+SamplerFunction<Request> parametersSampler = (request) -> {
```
adriancole:
```suggestion
SamplerFunction<Request> requestBased = (request) -> {
```

brave/src/main/java/brave/Tracer.java
```diff
@@ -495,6 +492,46 @@ public ScopedSpan startScopedSpan(String name) {
...
+   * Like {@link #nextSpan()} except when there is no trace in process, the sampler {@link
+   * SamplerFunction#trySample(Object) triggers} against the supplied argument.
+   *
+   * @param samplerFunction invoked if there's no {@link CurrentTraceContext#get() current trace}
+   * @param arg parameter to {@link SamplerFunction#trySample(Object)}
+   * @see #startScopedSpan(String, SamplerFunction, Object)
+   * @see #nextSpan(TraceContextOrSamplingFlags)
+   * @since 5.8
+   */
+  public <T> Span nextSpan(SamplerFunction<T> samplerFunction, T arg) {
```
adriancole: This method is used directly by the http client implementation now. It could easily be used by RPC also except we have no RPC samplers yet :P

brave/src/main/java/brave/Tracer.java
```diff
@@ -495,6 +492,46 @@ public ScopedSpan startScopedSpan(String name) {
...
+   * Like {@link #startScopedSpan(String)} except when there is no trace in process, the sampler
+   * {@link SamplerFunction#trySample(Object) triggers} against the supplied argument.
+   *
+   * @param name the {@link Span#name(String) span name}
+   * @param samplerFunction invoked if there's no {@link CurrentTraceContext#get() current trace}
+   * @param arg parameter to {@link SamplerFunction#trySample(Object)}
+   * @see #nextSpan(SamplerFunction, Object)
+   * @since 5.8
+   */
+  public <T> ScopedSpan startScopedSpan(String name, SamplerFunction<T> samplerFunction, T arg) {
```
anuraaga: Cool, this looks much nicer than `withSampler`

brave/src/main/java/brave/Tracer.java
```diff
@@ -495,6 +492,46 @@ public ScopedSpan startScopedSpan(String name) {
...
+    if (name == null) throw new NullPointerException("name == null");
```
anuraaga: This is the same as old code so probably ok, but just wondering should we also check if it's empty?

brave/src/main/java/brave/Tracer.java
```diff
@@ -495,6 +492,46 @@ public ScopedSpan startScopedSpan(String name) {
...
+   * @see #nextSpan(TraceContextOrSamplingFlags)
+   * @since 5.8
+   */
+  public <T> Span nextSpan(SamplerFunction<T> samplerFunction, T arg) {
+    return _toSpan(nextContext(samplerFunction, arg));
+  }
+
+  <T> TraceContext nextContext(SamplerFunction<T> samplerFunction, T arg) {
+    if (samplerFunction == null) throw new NullPointerException("samplerFunction == null");
+    if (arg == null) throw new NullPointerException("arg == null");
```
anuraaga: If `SamplerFunction` is intended for use by users, it might make sense to let them accept `null` if they want it. Some pattern like a function of an attribute, with arg being an extracted attribute that could be `null`, seems like a reasonable pattern

brave/src/main/java/brave/Tracer.java
adriancole: I noticed some old examples actually pass empty.. we couldn't error, but we could eagerly coerce to null..

brave/src/main/java/brave/Tracer.java
adriancole: that's ok with me.

adriancole: added code PTAL

anuraaga: I think the new tests verify the default matchers accept null but don't we still need to remove this check so users can call this method with null? How would instrumentation pass null to the matcher?

adriancole: sorry I forgot to push most of the commit :P

anuraaga:
LGTM

PR #914, opened 2019-05-29
Messaging instrumentation

jeqo:
Follow up of https://github.com/apache/incubator-zipkin-brave/pull/904 with the right source branch.

PR #665, opened 2018-03-26
Adds RxJava v2 CurrentTraceContextAssemblyTracking

adriancole:
`CurrentTraceContextAssemblyTracking` prevents traces from breaking
during RxJava operations by scoping them with trace context.

The design of this library borrows heavily from https://github.com/akaita/RxJava2Debug

To set this up, create `CurrentTraceContextAssemblyTracking` using the
current trace context provided by your `Tracing` component.

```java
assemblyTracking = CurrentTraceContextAssemblyTracking.create(
  tracing.currentTraceContext()
);
```

After your application-specific changes to `RxJavaPlugins`, enable trace
context tracking like so:

```java
assemblyTracking.enable();
```

Fixes #664

PR #999, opened 2019-09-30
Adds parsers and handlers for Rpc abstraction and ports Dubbo and gRPC

adriancole:
This adds basic infrastructure for uniform parsing of RPC libraries

By default, the following are added to both RPC client and server spans:
* Span.name is the RPC service/method. Ex. "zipkin.proto3.SpanService/Report"
  * If the service is absent, the method is the name and visa versa.
* Tags:
  * "rpc.method", eg "Report"
  * "rpc.service", eg "zipkin.proto3.SpanService"
  * "rpc.error_code", eg "CANCELLED"
  * "error" the RPC error code if there is no exception
* Remote IP and port information

As we know propagation fields are not always headers, "getters and setter" methods used for propagation are protected, not public methods, and not named headers.

### DIFF
--- a/instrumentation/dubbo-rpc/src/main/java/brave/dubbo/rpc/DubboClientResponse.java
+++ b/instrumentation/dubbo-rpc/src/main/java/brave/dubbo/rpc/DubboClientResponse.java
@@ -13,13 +13,12 @@
  */
 package brave.dubbo.rpc;
 
-import brave.Response;
-import brave.Span;
 import brave.internal.Nullable;
+import brave.rpc.RpcClientResponse;
 import com.alibaba.dubbo.rpc.Result;
 import com.alibaba.dubbo.rpc.RpcException;
 
-final class DubboClientResponse extends Response implements DubboResponse {
+final class DubboClientResponse extends RpcClientResponse implements DubboResponse {
   final DubboClientRequest request;
   @Nullable final Result result;
   @Nullable final Throwable error;
@@ -40,10 +39,6 @@ final class DubboClientResponse extends Response implements DubboResponse {
     return result;
   }
 
-  @Override public Span.Kind spanKind() {
-    return Span.Kind.CLIENT;
-  }
-
   @Override public DubboClientRequest request() {
     return request;
   }
@@ -53,7 +48,7 @@ final class DubboClientResponse extends Response implements DubboResponse {
   }
 
   /** Returns the string form of the {@link RpcException#getCode()} */
-  String errorCode() {
+  @Override public String errorCode() {
     return DubboParser.errorCode(error);
   }
 }
--- a/instrumentation/dubbo-rpc/src/main/java/brave/dubbo/rpc/DubboParser.java
+++ b/instrumentation/dubbo-rpc/src/main/java/brave/dubbo/rpc/DubboParser.java
@@ -45,13 +45,11 @@ final class DubboParser {
   }
 
   /**
-   * Returns the {@link URL#getServiceInterface() service interface} of the invocation.
+   * Returns the {@link URL#getServiceInterface() service interface} of the invoker.
    *
    * <p>This was chosen as the {@link URL#getServiceName() service name} is deprecated for it.
    */
-  static @Nullable String service(Invocation invocation) {
-    Invoker<?> invoker = invocation.getInvoker();
-    if (invoker == null) return null;
+  @Nullable static String service(Invoker<?> invoker) {
     URL url = invoker.getUrl();
     if (url == null) return null;
     String service = url.getServiceInterface();
@@ -63,8 +61,8 @@ final class DubboParser {
     InetSocketAddress remoteAddress = rpcContext.getRemoteAddress();
     if (remoteAddress == null) return false;
     return span.remoteIpAndPort(
-      Platform.get().getHostString(remoteAddress),
-      remoteAddress.getPort()
+        Platform.get().getHostString(remoteAddress),
+        remoteAddress.getPort()
     );
   }
 
--- a/instrumentation/dubbo-rpc/src/main/java/brave/dubbo/rpc/FinishSpan.java
+++ b/instrumentation/dubbo-rpc/src/main/java/brave/dubbo/rpc/FinishSpan.java
@@ -1,0 +1,89 @@
+/*
+ * Copyright 2013-2019 The OpenZipkin Authors
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License"); you may not use this file except
+ * in compliance with the License. You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software distributed under the License
+ * is distributed on an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express
+ * or implied. See the License for the specific language governing permissions and limitations under
+ * the License.
+ */
+package brave.dubbo.rpc;
+
+import brave.Span;
+import brave.internal.Nullable;
+import brave.rpc.RpcClientHandler;
+import brave.rpc.RpcClientRequest;
+import brave.rpc.RpcServerHandler;
+import com.alibaba.dubbo.rpc.Result;
+
+// Intentionally the same as Apache Dubbo, even though we don't use the first arg.
+// When the signature is the same, it reduces work porting bug fixes or tests
+abstract class FinishSpan { // implements BiConsumer<Object, Throwable> except Java 6
+  static void finish(TracingFilter filter,
+    DubboRequest request, @Nullable Result result, @Nullable Throwable error, Span span) {
+    if (request instanceof RpcClientRequest) {
+      filter.clientHandler.handleReceive(
+        new DubboClientResponse((DubboClientRequest) request, result, error), span);
+    } else {
+      filter.serverHandler.handleSend(
+        new DubboServerResponse((DubboServerRequest) request, result, error), span);
+    }
+  }
+
+  static FinishSpan create(TracingFilter filter, DubboRequest request, Result result, Span span) {
+    if (request instanceof DubboClientRequest) {
+      return new FinishClientSpan(
+        span, result, filter.clientHandler, (DubboClientRequest) request);
+    }
+    return new FinishServerSpan(span, result, filter.serverHandler, (DubboServerRequest) request);
+  }
+
+  final Span span;
+  final Result result;
+
+  FinishSpan(Span span, Result result) {
+    if (span == null) throw new NullPointerException("span == null");
+    if (result == null) throw new NullPointerException("result == null");
+    this.span = span;
+    this.result = result;
+  }
+
+  /** One, but not both parameters can be {@code null}. */
+  public abstract void accept(@Nullable Object unused, @Nullable Throwable error);
+
+  static final class FinishClientSpan extends FinishSpan {
+    final RpcClientHandler clientHandler;
+    final DubboClientRequest request;
+
+    FinishClientSpan(
+      Span span, Result result, RpcClientHandler clientHandler, DubboClientRequest request) {
+      super(span, result);
+      this.clientHandler = clientHandler;
+      this.request = request;
+    }
+
+    @Override public void accept(@Nullable Object unused, @Nullable Throwable error) {
+      clientHandler.handleReceive(new DubboClientResponse(request, result, error), span);
+    }
+  }
+
+  static final class FinishServerSpan extends FinishSpan {
+    final RpcServerHandler serverHandler;
+    final DubboServerRequest request;
+
+    FinishServerSpan(
+      Span span, Result result, RpcServerHandler serverHandler, DubboServerRequest request) {
+      super(span, result);
+      this.serverHandler = serverHandler;
+      this.request = request;
+    }
+
+    @Override public void accept(@Nullable Object unused, @Nullable Throwable error) {
+      serverHandler.handleSend(new DubboServerResponse(request, result, error), span);
+    }
+  }
+}
--- a/instrumentation/dubbo-rpc/src/main/java/brave/dubbo/rpc/TracingFilter.java
+++ b/instrumentation/dubbo-rpc/src/main/java/brave/dubbo/rpc/TracingFilter.java
@@ -74,9 +74,9 @@ public final class TracingFilter implements Filter {
   @Deprecated public void setTracing(Tracing tracing) {
     if (tracing == null) throw new NullPointerException("rpcTracing == null");
     setRpcTracing(RpcTracing.newBuilder(tracing)
-      .clientResponseParser(LEGACY_RESPONSE_PARSER)
-      .serverResponseParser(LEGACY_RESPONSE_PARSER)
-      .build());
+        .clientResponseParser(LEGACY_RESPONSE_PARSER)
+        .serverResponseParser(LEGACY_RESPONSE_PARSER)
+        .build());
   }
 
   /**
@@ -121,7 +121,7 @@ public final class TracingFilter implements Filter {
       span = serverHandler.handleReceive(serverRequest);
     }
 
-    boolean deferFinish = false;
+    boolean isSynchronous = true;
     Scope scope = currentTraceContext.newScope(span.context());
     Result result = null;
     Throwable error = null;
@@ -129,13 +129,17 @@ public final class TracingFilter implements Filter {
       result = invoker.invoke(invocation);
       error = result.getException();
       Future<Object> future = rpcContext.getFuture(); // the case on async client invocation
-      if (future instanceof FutureAdapter) {
-        deferFinish = true;
+      if (future != null) {
+        if (!(future instanceof FutureAdapter)) {
+          assert false : "we can't defer the span finish unless we can access the ResponseFuture";
+          return result;
+        }
+        isSynchronous = false;
         ResponseFuture original = ((FutureAdapter<Object>) future).getFuture();
         // See instrumentation/RATIONALE.md for why the below response callbacks are invocation context
         TraceContext callbackContext = kind == Kind.CLIENT ? invocationContext : span.context();
         ResponseFuture wrapped =
-          new FinishSpanResponseFuture(original, this, request, result, span, callbackContext);
+            new FinishSpanResponseFuture(original, this, request, result, span, callbackContext);
         RpcContext.getContext().setFuture(new FutureAdapter<>(wrapped));
       }
       return result;
@@ -144,7 +148,7 @@ public final class TracingFilter implements Filter {
       error = e;
       throw e;
     } finally {
-      if (!deferFinish) FinishSpan.finish(this, request, result, error, span);
+      if (isSynchronous) FinishSpan.finish(this, request, result, error, span);
       scope.close();
     }
   }
--- a/instrumentation/dubbo-rpc/src/main/java/brave/dubbo/rpc/TracingFilter.java
+++ b/instrumentation/dubbo-rpc/src/main/java/brave/dubbo/rpc/TracingFilter.java
@@ -15,18 +15,17 @@ package brave.dubbo.rpc;
 
 import brave.Span;
 import brave.Span.Kind;
+import brave.SpanCustomizer;
 import brave.Tag;
-import brave.Tracer;
 import brave.Tracing;
 import brave.propagation.CurrentTraceContext;
 import brave.propagation.CurrentTraceContext.Scope;
 import brave.propagation.TraceContext;
-import brave.propagation.TraceContext.Extractor;
-import brave.propagation.TraceContext.Injector;
-import brave.propagation.TraceContextOrSamplingFlags;
-import brave.rpc.RpcRequest;
+import brave.rpc.RpcClientHandler;
+import brave.rpc.RpcResponse;
+import brave.rpc.RpcResponseParser;
+import brave.rpc.RpcServerHandler;
 import brave.rpc.RpcTracing;
-import brave.sampler.SamplerFunction;
 import com.alibaba.dubbo.common.Constants;
 import com.alibaba.dubbo.common.extension.Activate;
 import com.alibaba.dubbo.common.extension.ExtensionLoader;
@@ -42,10 +41,7 @@ import com.alibaba.dubbo.rpc.protocol.dubbo.FutureAdapter;
 import java.util.Map;
 import java.util.concurrent.Future;
 
-import static brave.dubbo.rpc.DubboClientRequest.SETTER;
-import static brave.dubbo.rpc.DubboServerRequest.GETTER;
 import static brave.internal.Throwables.propagateIfFatal;
-import static brave.sampler.SamplerFunctions.deferDecision;
 
 @Activate(group = {Constants.PROVIDER, Constants.CONSUMER}, value = "tracing")
 // http://dubbo.apache.org/en-us/docs/dev/impls/filter.html
@@ -57,12 +53,15 @@ public final class TracingFilter implements Filter {
       return String.valueOf(((RpcException) input).getCode());
     }
   };
+  static final RpcResponseParser LEGACY_RESPONSE_PARSER = new RpcResponseParser() {
+    @Override public void parse(RpcResponse response, TraceContext context, SpanCustomizer span) {
+      DUBBO_ERROR_CODE.tag(response.error(), span);
+    }
+  };
 
   CurrentTraceContext currentTraceContext;
-  Tracer tracer;
-  Extractor<DubboServerRequest> extractor;
-  Injector<DubboClientRequest> injector;
-  SamplerFunction<RpcRequest> clientSampler = deferDecision(), serverSampler = deferDecision();
+  RpcClientHandler clientHandler;
+  RpcServerHandler serverHandler;
   volatile boolean isInit = false;
 
   /**
@@ -74,23 +73,28 @@ public final class TracingFilter implements Filter {
    */
   @Deprecated public void setTracing(Tracing tracing) {
     if (tracing == null) throw new NullPointerException("rpcTracing == null");
-    setRpcTracing(RpcTracing.create(tracing));
+    setRpcTracing(RpcTracing.newBuilder(tracing)
+      .clientResponseParser(LEGACY_RESPONSE_PARSER)
+      .serverResponseParser(LEGACY_RESPONSE_PARSER)
+      .build());
   }
 
   /**
    * {@link ExtensionLoader} supplies the tracing implementation which must be named "rpcTracing".
    * For example, if using the {@link SpringExtensionFactory}, only a bean named "rpcTracing" will
    * be injected.
+   *
+   * <h3>Custom parsing</h3>
+   * Custom parsers, such as {@link RpcTracing#clientRequestParser()}, can use Dubbo-specific types
+   * {@link DubboRequest} and {@link DubboResponse} to get access such as the Java invocation or
+   * result.
    */
   public void setRpcTracing(RpcTracing rpcTracing) {
     if (rpcTracing == null) throw new NullPointerException("rpcTracing == null");
     // we don't guard on init because we intentionally want to overwrite any call to setTracing
     currentTraceContext = rpcTracing.tracing().currentTraceContext();
-    extractor = rpcTracing.tracing().propagation().extractor(GETTER);
-    injector = rpcTracing.tracing().propagation().injector(SETTER);
-    tracer = rpcTracing.tracing().tracer();
-    clientSampler = rpcTracing.clientSampler();
-    serverSampler = rpcTracing.serverSampler();
+    clientHandler = RpcClientHandler.create(rpcTracing);
+    serverHandler = RpcServerHandler.create(rpcTracing);
     isInit = true;
   }
 
@@ -101,6 +105,7 @@ public final class TracingFilter implements Filter {
     RpcContext rpcContext = RpcContext.getContext();
     Kind kind = rpcContext.isProviderSide() ? Kind.SERVER : Kind.CLIENT;
     Span span;
+    DubboRequest request;
     if (kind.equals(Kind.CLIENT)) {
       // When A service invoke B service, then B service then invoke C service, the parentId of the
       // C service span is A when read from invocation.getAttachments(). This is because
@@ -108,28 +113,20 @@ public final class TracingFilter implements Filter {
       // See com.alibaba.dubbo.rpc.protocol.AbstractInvoker(line 138) from v2.6.7
       Map<String, String> attachments = RpcContext.getContext().getAttachments();
       DubboClientRequest clientRequest = new DubboClientRequest(invoker, invocation, attachments);
-      span = tracer.nextSpan(clientSampler, clientRequest);
-      injector.inject(span.context(), clientRequest);
+      request = clientRequest;
+      span = clientHandler.handleSendWithParent(clientRequest, invocationContext);
     } else {
       DubboServerRequest serverRequest = new DubboServerRequest(invoker, invocation);
-      TraceContextOrSamplingFlags extracted = extractor.extract(serverRequest);
-      span = nextSpan(extracted, serverRequest);
-    }
-
-    if (!span.isNoop()) {
-      span.kind(kind);
-      String service = DubboParser.service(invocation);
-      String method = DubboParser.method(invocation);
-      span.name(service + "/" + method);
-      DubboParser.parseRemoteIpAndPort(span);
-      span.start();
+      request = serverRequest;
+      span = serverHandler.handleReceive(serverRequest);
     }
 
     boolean deferFinish = false;
     Scope scope = currentTraceContext.newScope(span.context());
+    Result result = null;
     Throwable error = null;
     try {
-      Result result = invoker.invoke(invocation);
+      result = invoker.invoke(invocation);
       error = result.getException();
       Future<Object> future = rpcContext.getFuture(); // the case on async client invocation
       if (future instanceof FutureAdapter) {
@@ -138,7 +135,7 @@ public final class TracingFilter implements Filter {
         // See instrumentation/RATIONALE.md for why the below response callbacks are invocation context
         TraceContext callbackContext = kind == Kind.CLIENT ? invocationContext : span.context();
         ResponseFuture wrapped =
-          new FinishSpanResponseFuture(original, this, span, callbackContext);
+          new FinishSpanResponseFuture(original, this, request, result, span, callbackContext);
         RpcContext.getContext().setFuture(new FutureAdapter<>(wrapped));
       }
       return result;
@@ -147,28 +144,8 @@ public final class TracingFilter implements Filter {
       error = e;
       throw e;
     } finally {
-      if (error != null) onError(error, span);
-      if (!deferFinish) span.finish();
+      if (!deferFinish) FinishSpan.finish(this, request, result, error, span);
       scope.close();
     }
-  }
-
-  /** Creates a potentially noop span representing this request */
-  // This is the same code as HttpServerHandler.nextSpan
-  // TODO: pull this into RpcServerHandler when stable https://github.com/openzipkin/brave/pull/999
-  Span nextSpan(TraceContextOrSamplingFlags extracted, DubboServerRequest request) {
-    Boolean sampled = extracted.sampled();
-    // only recreate the context if the sampler made a decision
-    if (sampled == null && (sampled = serverSampler.trySample(request)) != null) {
-      extracted = extracted.sampled(sampled.booleanValue());
-    }
-    return extracted.context() != null
-      ? tracer.joinSpan(extracted.context())
-      : tracer.nextSpan(extracted);
-  }
-
-  static void onError(Throwable error, Span span) {
-    span.error(error);
-    DUBBO_ERROR_CODE.tag(error, span);
   }
 }
--- a/instrumentation/dubbo/src/main/java/brave/dubbo/DubboClientRequest.java
+++ b/instrumentation/dubbo/src/main/java/brave/dubbo/DubboClientRequest.java
@@ -14,26 +14,13 @@
 package brave.dubbo;
 
 import brave.Span;
-import brave.propagation.Propagation;
-import brave.propagation.Propagation.Setter;
 import brave.rpc.RpcClientRequest;
 import java.util.Map;
 import org.apache.dubbo.common.URL;
 import org.apache.dubbo.rpc.Invocation;
 import org.apache.dubbo.rpc.Invoker;
 
-// intentionally not yet public until we add tag parsing functionality
 final class DubboClientRequest extends RpcClientRequest implements DubboRequest {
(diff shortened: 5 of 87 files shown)


adriancole:
https://github.com/openzipkin/brave/pull/1000 pulls the most important parts out into a separate change

adriancole:
@jeqo @anuraaga do you have some modeling thoughts? should we just add the following to the response types?

```
@Nullable String message()
@Nullable Boolean isError()
```

instrumentation/rpc/README.md
```diff
@@ -4,8 +4,58 @@ Most instrumentation are based on RPC communication. For this reason,
...
 are configured with `RpcTracing`.
 
-The `RpcTracing` class holds a reference to a tracing component and
-sampling policy.
+The `RpcTracing` class holds a reference to a tracing component,
+instructions on what to put into rpc spans, and sampling policy.
+
+## Span data policy
+By default, the following are added to both rpc client and server spans:
+* Span.name is the rpc method in lowercase: ex "zipkin.proto3.spanservice/report"
```
adriancole: NOTE: ignore this it is copy/paste and needs to be re-written

jeqo:
maybe `@Nullable String errorMessage()` or similar would be better, as just `message` could be confusing

adriancole:
so we need a message (status) in case of success or fail. probably we need
examples... the main issue is we wont know which message is an error.

otoh maybe you are right we can just have two strings one for status and
one for error (which might be status also or empty string if we know
nothing except it is error)

adriancole:
ok I added error parsing. next steps on this are to integrate gRPC and dubbo.

After this is merged, I'd recommend pulling out the brave.Response type and backporting http to implement it. After that, try again on the messaging abstraction!

instrumentation/rpc/src/main/java/brave/rpc/RpcHandler.java
```diff
@@ -0,0 +1,91 @@
...
+        parseResponse(response, error, span);
+      } finally {
+        ws.close(); // close the scope before finishing the span
+      }
+    } finally {
+      finishInNullScope(span, finishTimestamp);
+    }
+  }
+
+  /** Clears the scope to prevent remote reporters from accidentally tracing */
```
devinsba: This is interesting, I think I would need a longer comment here to understand the need. I'm maybe just wondering why the if/else can't be in the finally above

instrumentation/rpc/src/test/java/brave/rpc/ITRpcTracingClassLoader.java
```diff
@@ -20,7 +20,9 @@
 import static brave.test.util.ClassLoaders.assertRunIsUnloadable;
 import static org.assertj.core.api.Assertions.assertThat;
 
-public class RpcTracingClassLoaderTest {
+// Mockito taints the classloader, through its log initialization. While this isn't an IT, this
+// works around the problem by forcing a different classloader.
+public class ITRpcTracingClassLoader {
```
devinsba: I wonder if instead forcing the mockito using test classes into their own class loader would be a nicer work around.

This person has a solution to force the mockito classes into their own loader, maybe that could be extended to the things that they load https://stackoverflow.com/a/28657203/2232476

adriancole: I'll change the comment. the problem is that it is initializes logging. Our current classpath test thing cannot unload the log manager.

instrumentation/rpc/src/main/java/brave/rpc/RpcHandler.java
adriancole: This happened when [optimizing for RxJava](https://github.com/openzipkin/brave/pull/665/files#diff-71435fd20f1401571a16dc9d9cd6373bR285) [https://github.com/openzipkin/brave/pull/669], and was pre-mature and short-sighted imho. Thanks for pointing out a long term issue. This wouldn't really stop anything unless client sampler was set to disabled. Anyway our servers are rate-limited now anyway.

I'll revert from http and here.

instrumentation/rpc/src/test/java/brave/rpc/ITRpcTracingClassLoader.java
adriancole: I'll leave a note about your comment, as I don't have unlimited time to dig into these things and what I did is a pragmatic workaround. Someone later can run into the rabbit hole, and hopefully will, but there's really no problem with how it is working now.

instrumentation/rpc/src/main/java/brave/rpc/RpcHandler.java
adriancole: https://github.com/openzipkin/brave/pull/1026

adriancole:
PS I plan to stop on this PR last friday :) I will hard stop on it tomorrow morning, so hopefully we can have a volunteer to carry forward after that.

adriancole:
ok formally setting this down for someone else to pick up or delete.

NEXT STEPS:
* implement dubbo, dubbo-rpc, grpc

adriancole:
looks like this stopped, which is also blocking https://github.com/openzipkin/brave/pull/914, which is probably blocking some other things..

adriancole:
so this is from otel (possibly just lifting grpc stuff) https://github.com/open-telemetry/opentelemetry-java/blob/c24e4a963669cfd044d7478a801e8f170faba4fe/api/src/main/java/io/opentelemetry/trace/Status.java#L38

I think we do need a message, regardless of whether it is success or failure. I'm less sure about "code" as not sure that's entirely portable.

@minwoox do you know between the various frameworks armeria supports (grpc, thrift etc) which properties are actually portable?
@trytocatch I looked at this for dubbo.. seems mostly to define stats, not generic properties like status message or code. do you have any comments about properties most required for stats/tracing? https://github.com/apache/dubbo/blob/master/dubbo-rpc/dubbo-rpc-api/src/main/java/org/apache/dubbo/rpc/RpcStatus.java

trytocatch:
> so this is from otel (possibly just lifting grpc stuff) https://github.com/open-telemetry/opentelemetry-java/blob/c24e4a963669cfd044d7478a801e8f170faba4fe/api/src/main/java/io/opentelemetry/trace/Status.java#L38
> 
> I think we do need a message, regardless of whether it is success or failure. I'm less sure about "code" as not sure that's entirely portable.
> 
> @minwoox do you know between the various frameworks armeria supports (grpc, thrift etc) which properties are actually portable?
> @trytocatch I looked at this for dubbo.. seems mostly to define stats, not generic properties like status message or code. do you have any comments about properties most required for stats/tracing? https://github.com/apache/dubbo/blob/master/dubbo-rpc/dubbo-rpc-api/src/main/java/org/apache/dubbo/rpc/RpcStatus.java

sorry, I just fixed a small concurrency problems there

adriancole:
thanks for the reply @trytocatch @beiwei30 do you have any commentary on RPC status/message portability from Dubbo?

minwoox:
> do you know between the various frameworks armeria supports (grpc, thrift etc) which properties are actually portable?

For the gRPC, @anuraaga would know much better and for the Thrift, I need to check it.

adriancole:
@reta does CXF have any abstraction of RPC which might have some prior art to consider here?
@WillemJiang same question for camel

reta:
@adriancole I am not aware of any RPC abstractions in CXF

adriancole:
I took a look at Avro here are some findings.

As usual, common errors such as method not existing on the other side are possible here. This hints at the need for a normal String error as opposed to relying on Throwable. That said, we can always synthesize a list of Throwables to keep things simple (by trimming the stack trace)

Like some other RPC, there is no defined error types actually, just a bit "isError" which neatly maps to "error" -> "" as we do now, when we know nothing else.

Like some other RPC a response is not required (ex. one-way sends nothing back not even isError). This is similar to both dubbo and thrift.

In fact there is no defined result code anyway, you just know it if it might have errored or not.

For example, one-way, you can receive feedback on failed handshake, missing IPC on the other side etc, but nothing back on success (or error at app level)

In two-way, you get a bit on error, but there's no generic response message nor is there any coding of messages.

## My takeaway

It is foolish to expect a common result type across all RPC or even a code. For this reason, we should stick with what I think @jeqo suggested earlier, which is something like...

* `rpc.message` (optional for those who have a code or something)
* `error` (our standard tag which can be empty if the implementation is a single bit)

thoughts?

adriancole:
while it is probably more schema in nature, the only thing I could suggest as another common tag (whether or not it is added by default is another question) is `rpc.is_one_way` I say this because at least 3 RPC systems use this: libthrift, dubbo and avro, and it helps you interpret the end of the span..

wu-sheng:
Right now, SkyWalking handles the error of RPC only by using status(true/false), error_code(http) and log(thread stack in java)

adriancole:
TL'DR' I do not think we should inherit OpenTelemetry's status specification, especially in its current form.

I'm having a very hard time finding a single RPC library that has a code series that includes success, besides grpc. The only 2 example mappings in https://github.com/open-telemetry/opentelemetry-specification are flimsy. In other words, I don't believe OpenTelemetry's status is generic enough to try to map things across. For example, we typically do not create an abstraction at all unless there are 3 different things.

A good example of Status being a leaky abstraction is their attempt to reverse map HTTP to it https://github.com/open-telemetry/opentelemetry-specification/blob/bfb060b23113ba9af492f8c63dd89ecfc500810b/specification/trace/semantic_conventions/http.md#status

The above Status mapping is even bad for errors. For example, alerting in HTTP is far more likely to group by things like HTTP 400 (not even defined as a status in otel) vs a mismatched except grpc thing like Deadline (redirect loop == deadline exceeded looks more like arbitrary attempt to map than anything else).

There's only one other mapping in their spec to it that I can find, messaging, which says "Ok" "Ok" "Ok" https://github.com/open-telemetry/opentelemetry-specification/blob/c04041e16c9571b7ee5cfd5f102fc15e94ffacd7/specification/trace/semantic_conventions/messaging.md#topic-with-multiple-consumers

I've told @jkwatson about this, warning this is probably an oversight aka mistake. We shouldn't inherit mistakes into our model, and we shouldn't inherit non-diverse models as normalized tags. For both reasons, I think we should let this sit, and just advise people to drop that open telemetry status tag, except its only valid mapping: grpc. That is until something changes and they fix it (by renaming it to grpc perhaps)

adriancole:
I looked further and it is easy enough to find coding of errors, though code is usually a string not a number.

For example, we have people tracing SOAP services "faultcode" and "faultstring" are both strings (which could map to an "error" tag or individual tags within the SOAP abstraction).

This hints that while mapping success codes into RPC response seems futile at the moment, there are various types of faults that usually end up in buckets of "short" and "not too long" we can define, and map one of them by default (likely the short one as cardinality is assumed less) to "error". Shortest being boolean true -> "error", "" (empty)

adriancole:
PS in case there's a question about why not tag by default status of OK? you can't assume that anyway. one-way RPC give no response back. in the odd case of mapping messaging to this it is evenn more a lie as almost always it just means it moved to a different non-remote buffer.

Also as we learned in opentracing, people make mistakes in instrumentation routinely which caused the catastrophy of the "error"="false" tag. Rather than repeat these mistakes, we can just note it in the rationale doc in attempts to warn others, even outside projects to not repeat these modeling mistakes.

In tools that have a success code tag, we just add those tags, or let people opt in. For example, grpc has a success code, so call that a grpc tag (as we do now) vs pretending all rpc have a valid answer to this.

adriancole:
Another clarification easy to miss. In our model we do not intend to conflate messaging or http as RPC. This was a point mentioned by @jeqo earlier and also matches Armeria

RPC is its own abstraction, that HTTP has RPC-like characteristics is incidental and there's also no need to re-abstract HTTP. That underlying drivers like redis have RPC like characteristics is often moot and not something to optimize for as users use redis at its storage abstraction, not the implementation detail of how it communicates underneath.

RPC is therefore how you associate RPC abstraction, like thrift, avro, protocol buffers services, gRPC, XML-RPC. These abstractions, which have more business impact as their message/message/function names are more likely to reflect  business code and by that help users identify what's up vs incidental details such as how a MySQL driver is implemented. The lowest RPC tiers are interesting to driver developers and advanced troubleshooting, but not something our model should optimize towards.

adriancole:
offline chat with @jkwatson reminded another hazard we should never forget. The reason we allow already folks to choose what to put as the "error" tag is that even things that seem like errors aren't.

Some may recall users complaining about junk service diagrams because 404 was classified as an error. HTTP GET returning 404 is sometimes treated like Map.containsKey. You can imagine how aggregate service error stats become with this in mind. The main point is that all abstractions we make should have a default mapping, yet allow users to change the primary "error" status if they know better. This is in fact the point of this PR (to add parsers that do exactly that)

adriancole:
Here's a pitch before bringing it to the non-java group.

@jeqo suggested "rpc.error_message" which I think is good, but I've noticed there's two cases when looking at libraries.

I propose we start with "rpc.error_code"

"rpc.error_code" - the shortest human readable code (ex prefer codename not code number). Required when span.tags['error'] is present
  * Ex "grpc.status_code" maps directly to this when not OK
  * Ex "true" or "" when we only have an error bit

optionally also we can define "rpc.error_message" as it is somewhat usual to have a longer message not necessarily in an exception. My tendencies go towards "when in doubt, leave it out", but I admit a longer message, we can find probably 2 or 3 libraries that do this.

@openzipkin/core thumbs up if like, comment if you also want "rpc.error_message" or do not want to map "rpc.error_code" as described. If it seems like we have a mostly consensus, I'll open an issue specing this on https://github.com/openzipkin/zipkin-api while we practice here in Brave.

adriancole:
ps on "rpc.error_message" it is somewhat conjecture at this point as it would be the exception message in SOAP. If folks want a deeper look at that, I can try to find an example where it isn't already the exception (ex brave's mutablespan.error is the exception so any time it is already the exception, this is redundant)

jeqo:
@adriancole agree. I'm +1 on adding "rpc.error_code" and expect `mutablespan.error` already cover "rpc.error_message" purpose.

adriancole:
ok I think we pondered the angles well enough. I'm finishing this now.

jkwatson:
ok, I'll make the change and we can see how it works. Once it's merged, I can point a test service at New Relic's zipkin span ingest and see how it works pretty easily.

adriancole:
@jkwatson wrong issue I think :D

jkwatson:
> @jkwatson wrong issue I think :D

nothing to see here, move along

adriancole:
ok rebased with error code. will proceed to implement dubbo and grpc next. review welcome

adriancole:
Apache Dubbo is done I think. I'll work on Alibaba Dubbo and gRPC tomorrow. Feedback welcome as we near the final implementation.

adriancole:
one other note: unlike http, response parsing doesn't have two args for response or error. That's because errors can sometimes embed response information. We should actually consider changing http abstraction to do the same, as we know HTTP status codes are often smuggled in exception messages (cc @trustin as I know this is somewhat familiar to a recent chat)!

brave/src/main/java/brave/Tag.java
```diff
@@ -87,38 +87,41 @@ public final void tag(I input, Span span) {
   /**
    * Tags the value parsed from the {@code input}.
    *
+   * @return the tag value if one was parsed.
    * @since 5.11
    */
   // ex void parse(HttpRequest request, TraceContext context, SpanCustomizer span);
-  public final void tag(I input, @Nullable TraceContext context, SpanCustomizer span) {
+  @Nullable public final String tag(I input, @Nullable TraceContext context, SpanCustomizer span) {
```
adriancole, line 94: this is compatible but we can't change this again :)

instrumentation/dubbo/src/main/java/brave/dubbo/DubboClientRequest.java
```diff
@@ -13,47 +13,62 @@
...
+    if (invoker == null) throw new NullPointerException("invoker == null");
     if (invocation == null) throw new NullPointerException("invocation == null");
+    this.invoker = invoker;
     this.invocation = invocation;
-    if (attachments == null) throw new NullPointerException("attachments == null");
-    this.attachments = attachments;
+    // When A service invoke B service, then B service then invoke C service, the parentId of the
+    // C service span is A when read from invocation.getAttachments(). This is because
+    // AbstractInvoker adds attachments via RpcContext.getContext(), not the invocation.
+    // See org.apache.dubbo.rpc.protocol.AbstractInvoker(line 141) from v2.7.3
```
adriancole, line 38: this is crazy, but it is moved from the old code. ideally there was a bug fixed at some point!

instrumentation/rpc/src/main/java/brave/rpc/RpcHandler.java
```diff
@@ -0,0 +1,82 @@
...
+    span.kind(request.spanKind());
+    request.parseRemoteIpAndPort(span);
+    requestParser.parse(request, span.context(), span.customizer());
+  }
+
+  void parseResponse(Resp response, Span span) {
+    responseParser.parse(response, span.context(), span.customizer());
+  }
+
+  void handleFinish(Resp response, Span span) {
```
adriancole, line 60: ex here.. we don't have two arguments. because for example in dubbo error_code is smuggled as an exception subtype. This is also routine for HTTP status in framework-specific exception types.

instrumentation/rpc/src/main/java/brave/rpc/RpcResponseParser.java
```diff
@@ -0,0 +1,73 @@
...
+import brave.Tags;
+import brave.Tracing;
+import brave.handler.FinishedSpanHandler;
+import brave.propagation.TraceContext;
+
+/**
+ * Use this to control the response data recorded for an {@link TraceContext#sampledLocal() sampled
+ * RPC client or server span}.
+ *
+ * <p>Here's an example that adds the "rpc.error_code" even though "error" contains it.
```
adriancole, line 27: TODO: change this example as it doesnt' actually change anything as we already set "rpc.error_code"

instrumentation/rpc/src/main/java/brave/rpc/RpcServerHandler.java
```diff
@@ -0,0 +1,119 @@
...
+ * Span span = handler.handleReceive(wrapper); // 1.
+ * Result result = null;
+ * Throwable error = null;
+ * try (Scope ws = currentTraceContext.newScope(span.context())) { // 2.
+ *   return result = process(request); // 3.
+ * } catch (RuntimeException | Error e) {
+ *   error = e; // 4.
+ *   throw e;
+ * } finally {
+ *   RpcServerResponseWrapper response = result != null
```
adriancole, line 49: TODO: change this example and also client to not dance around null result (response should allow null result when there is an exception or return a constant if there is neither a result nor exception)

instrumentation/rpc/src/main/java/brave/rpc/RpcTracing.java
```diff
@@ -45,6 +46,47 @@ public Tracing tracing() {
...
+   * about the request before the server processes it.
+   *
+   * @since 5.12
+   */
+  public RpcRequestParser serverRequestParser() {
+    return serverRequestParser;
+  }
+
+  /**
+   * Used by {@link RpcServerHandler#handleSend(Object, Throwable, Span)} to add tags about the
```
adriancole, line 81: TODO: javadoc links drifted. doublecheck all before merge

instrumentation/rpc/RATIONALE.md
```diff
@@ -87,3 +119,30 @@ parameterized sampling. This happens before a span name is chosen. Moreover, a
 future change will allow parsing into a span name from the same type. In other
 words, `RpcRequest` is an intermediate model that must evaluate properties
 before a span exists.
+
+## Why "rpc.error_code"?
+
+## Why not "rpc.error_message"?
```
jcchavezs, line 125: worth to mention that error_code is more likely to be searchable vs error_message?

instrumentation/rpc/README.md
```diff
@@ -1,32 +1,86 @@
 # brave-instrumentation-rpc
 
-This is a helper for RPC libraries such as gRPC and Dubbo. Specifically, this
-includes samplers for clients and servers, configured with `RpcTracing`.
+Most instrumentation are based on rpc communication. For this reason,
```
jcchavezs, line 3: not sure why RPC was `lowercase`d but maybe good to `uppercase` it as they don't represent a word?

instrumentation/rpc/src/main/java/brave/rpc/RpcClientHandler.java
```diff
@@ -0,0 +1,137 @@
...
+ * portable customizations via {@link RpcRequestParser} and {@link RpcResponseParser}.
+ *
+ * <p>Synchronous interception is the most straight forward instrumentation.
+ *
+ * <p>You generally need to:
+ * <ol>
+ *   <li>Start the span and add trace headers to the request</li>
+ *   <li>Put the span in scope so things like log integration works</li>
+ *   <li>Invoke the request</li>
+ *   <li>Catch any errors</li>
```
jcchavezs, line 35: I'd say `Catch any errors, tag them and throw them up`

instrumentation/rpc/README.md
adriancole: I'll sweep good idea

instrumentation/rpc/RATIONALE.md
adriancole: seems I died in the middle of writing `Why "rpc.error_code"?` good catch :)

instrumentation/rpc/src/main/java/brave/rpc/RpcClientHandler.java
adriancole: catch is maybe imprecise in hindsight as sometimes they are just passed to us. Ack the example catch and rethrows in sync code, but probably we shouldn't focus on the catching part.

`If there was a Throwable, add it to the span`

PS better english for what you said would be

"Catch, tag, and rethrow any errors" because "throw them up" is saying more like vomit, which well.. isn't wrong :)

instrumentation/rpc/src/main/java/brave/rpc/RpcHandler.java
adriancole: added to rationale

instrumentation/rpc/RATIONALE.md
adriancole: done!

adriancole:
As we know gRPC users already tag server response values, I added an example of how to do that in Dubbo. I think now, the feature is complete so it is a better time to start porting it to the other dubbo (alibaba) and gRPC

instrumentation/dubbo/src/test/java/brave/dubbo/ITTracingFilter_Consumer.java
```diff
@@ -240,6 +247,52 @@
...
+      .getExtension("tracing"))
+      .setTracing(tracing);
+
+    assertThatThrownBy(() -> wrongClient.get().sayHello("jorge"))
+      .isInstanceOf(RpcException.class);
+
+    Span span =
+      reporter.takeRemoteSpanWithError(Span.Kind.CLIENT, ".*Not found exported service.*");
+    assertThat(span.tags())
+      .containsEntry("dubbo.error_code", "1");
```
adriancole, line 269: FYI, we hadn't released the change to the legacy dubbo error code format. I put it back when RpcTracing is not in use, and deprecated the Tracing setter (cc @anuraaga)

instrumentation/dubbo/src/test/java/brave/dubbo/ITTracingFilter_Consumer.java
```diff
@@ -240,6 +247,52 @@
...
+          javaValue.tag((DubboResponse) res, span);
+        }
+      }
+    };
+    rpcTracing = RpcTracing.newBuilder(tracing)
+      .clientResponseParser(customResponseParser)
+      .build();
+    init();
+
+    String javaResult = client.get().sayHello("jorge");
```
adriancole, line 293: @jeqo in case you didn't notice how famous you are in tests :D

instrumentation/dubbo/src/main/java/brave/dubbo/TracingFilter.java
```diff
@@ -90,73 +109,86 @@ public Result invoke(Invoker<?> invoker, Invocation invocation) throws RpcExcept
...
     } catch (Throwable e) {
       propagateIfFatal(e);
       error = e;
       throw e;
     } finally {
-      if (error != null) onError(error, span);
-      if (!deferFinish) span.finish();
+      if (!deferFinish) {
+        if (clientRequest != null) {
+          clientHandler.handleReceive(new DubboClientResponse(clientRequest, result, error), span);
```
adriancole, line 148: this tension is ok as it is somewhat uncommon for client and server to use the same handler/filter and types.

adriancole:
grpc is final lap here. backporting the old parsers is a bit interesting as grpc has streaming requests and responses which we already support.

brave/src/main/java/brave/Tag.java
anuraaga: It's compatible with source compilation, but I'm pretty sure it's ABI incompatible, meaning if two libraries target different Brave and use this method, one gets `NoSuchMethodError`.

anuraaga: Found this tool that can allow the void bytecode to be outputted as well. Seems quite hacky though

https://github.com/infradna/bridge-method-injector

adriancole:
> It's compatible with source compilation, but I'm pretty sure it's ABI incompatible, meaning if two libraries target different Brave and use this method, one gets NoSuchMethodError

cool I'll revert this as it isn't important enough not to. hindsight's 20/20 as they say...

anuraaga:
Phew - was worried that would snowball into a hundreds of lines of diff or something ;)

instrumentation/dubbo-rpc/src/main/java/brave/dubbo/rpc/DubboClientResponse.java
```diff
@@ -0,0 +1,54 @@
...
+ * the License.
+ */
+package brave.dubbo.rpc;
+
+import brave.internal.Nullable;
+import brave.rpc.RpcClientResponse;
+import com.alibaba.dubbo.rpc.Result;
+import com.alibaba.dubbo.rpc.RpcException;
+
+class DubboClientResponse extends RpcClientResponse implements DubboResponse { // not final for mock
```
anuraaga, line 21: May consider using mockito's support for mocking finals. I've been using it happily for a while. No idea how it works ;)

https://github.com/mockito/mockito/wiki/What%27s-new-in-Mockito-2#unmockable

instrumentation/dubbo-rpc/src/main/java/brave/dubbo/rpc/DubboClientResponse.java
```diff
@@ -0,0 +1,54 @@
...
+  final DubboClientRequest request;
+  @Nullable final Result result;
+  @Nullable final Throwable error;
+
+  DubboClientResponse(DubboClientRequest request, Result result, @Nullable Throwable error) {
```
anuraaga, line 26: Think `result` is missing `@Nullable` for all the similar types in the PR

instrumentation/dubbo-rpc/src/main/java/brave/dubbo/rpc/DubboParser.java
```diff
@@ -33,7 +33,7 @@
    */
   static @Nullable String method(Invocation invocation) {
     String methodName = invocation.getMethodName();
-    if ("$invoke".equals(methodName)) {
+    if ("$invoke" .equals(methodName)) {
```
anuraaga, line 36: Revert

instrumentation/dubbo-rpc/src/main/java/brave/dubbo/rpc/FinishSpan.java
```diff
@@ -0,0 +1,87 @@
...
+
+    FinishClientSpan(
+      Span span, Result result, RpcClientHandler clientHandler, DubboClientRequest request) {
+      super(span, result);
+      this.clientHandler = clientHandler;
+      this.request = request;
+    }
+
+    @Override void finish(Object resultValue, @Nullable Throwable error) {
+      clientHandler.handleReceive(new DubboClientResponse(request, result, error), span);
```
anuraaga, line 68: Should this be `resultValue`, not `result`? Otherwise `resultValue` is unused and can be removed from the interface.

instrumentation/dubbo-rpc/src/main/java/brave/dubbo/rpc/FinishSpan.java
```diff
@@ -0,0 +1,87 @@
...
+    final DubboClientRequest request;
+
+    FinishClientSpan(
+      Span span, Result result, RpcClientHandler clientHandler, DubboClientRequest request) {
+      super(span, result);
+      this.clientHandler = clientHandler;
+      this.request = request;
+    }
+
+    @Override void finish(Object resultValue, @Nullable Throwable error) {
```
anuraaga, line 67: `@Nullable resultValue`

instrumentation/dubbo-rpc/src/main/java/brave/dubbo/rpc/DubboClientResponse.java
adriancole: great sorcery. I prefer to have oddness in src/test vs src/main any time!

instrumentation/dubbo-rpc/src/main/java/brave/dubbo/rpc/DubboClientResponse.java
adriancole: you caught me avoiding a line wrap :D I'll be more honest next time.

instrumentation/dubbo-rpc/src/main/java/brave/dubbo/rpc/DubboParser.java
adriancole: I'll try to figure out why my IDE is now just doing that. I recently updated but didn't change the formatter..

instrumentation/dubbo-rpc/src/main/java/brave/dubbo/rpc/FinishSpan.java
adriancole: I put this here so that there's parity on the callback. ex in apache dubbo it is 2 args. Maybe I'll rename the callback to represent the same thing for apache dubbo

instrumentation/dubbo-rpc/src/main/java/brave/dubbo/rpc/FinishSpan.java
adriancole: this one was an accident :)

instrumentation/dubbo-rpc/src/main/java/brave/dubbo/rpc/FinishSpan.java
adriancole: I made a comment to why there's parity. mainly to make maintenance easy between the dubbo's

adriancole:
thanks for the feedback @anuraaga heading to grpc now

instrumentation/dubbo-rpc/src/main/java/brave/dubbo/rpc/FinishSpan.java
anuraaga: Ah got it - how about renaming `resultValue` to `unused` for ours? Don't think it' affects maintenance much but makes that explicit.

adriancole: sgtm will do

adriancole:
I noticed quite a few issues in the gRPC instrumentation, so addressed them independently here #1175 Once merged, finishing this will be mechanical

adriancole:
ok this is back in business. I'm happy with the code I think and hope you are. @anuraaga has done a lot of work moving my hands. The commit log will betray this, so please thank him.

tomorrow, I will polish up the readmes for shipit!

instrumentation/grpc/RATIONALE.md
```diff
@@ -25,3 +25,14 @@ record the application exception.
 In short, we choose to not mask the gRPC status with a potential application
 exception on close. In worst case, if there is no instrumentation for the layer
 that throws, its trace and span ID could be known via log correlation.
+
+## Why don't we use RpcClientHandler or RpcServerHandler for unexpected errors?
```
adriancole, line 29: @anuraaga since we now have a more compelling story about parsing, I agree that we should keep it nicer on the happy path vs the edge cases. This follows up your points on the other pull request I hope. Otherwise, let me know how to change it.

instrumentation/grpc/src/main/java/brave/grpc/GrpcParser.java
```diff
@@ -51,8 +75,8 @@ protected ErrorParser errorParser() {
 
   /**
    * Override to change what data from the status or trailers are parsed into the span modeling it.
-   * By default, this tags "grpc.status_code" when it is not OK, and "error" if there was no {@link
-   * Status#getCause()}.
+   * <<<<<<< HEAD By default, this tags "grpc.status_code" when it is not OK, and "error" if there
```
adriancole, line 78: TODO: bad merge

instrumentation/grpc/src/test/java/brave/grpc/BaseITTracingClientInterceptor.java
```diff
@@ -509,6 +489,59 @@ void initMessageTaggingClient() {
     assertThat(reporter.takeRemoteSpan(Span.Kind.CLIENT).parentId()).isNull();
   }
```
adriancole, line 491: NOTE: all ITs still pass, across dubbo and grpc. This means the data policy is exactly the same unless, someone initializes with RpcTracing, which will set the new policy from the rpc abstraction (ex "rpc.error_code")

adriancole:
rebased over a jax-rs test flake fix.

proceeding to add README content to finish this

adriancole:
OK this is done now, imho

instrumentation/dubbo/src/main/java/brave/dubbo/TracingFilter.java
```diff
@@ -139,39 +133,18 @@ public Result invoke(Invoker<?> invoker, Invocation invocation) throws RpcExcept
...
       }
       return result;
     } catch (Throwable e) {
       propagateIfFatal(e);
       error = e;
       throw e;
     } finally {
-      if (error != null) onError(error, span);
-      if (!deferFinish) span.finish();
+      if (!deferFinish) FinishSpan.finish(this, request, result, error, span);
```
anuraaga, line 146: Not really something for this PR, but just wondering, does this mean the timestamp of the span will be bogus?

instrumentation/grpc/src/main/java/brave/grpc/GrpcClientResponse.java
```diff
@@ -13,72 +13,84 @@
...
   }
 
   /**
-   * Returns the status passed to {@link ClientCall.Listener#onClose(Status, Metadata)} or {@code
-   * null} on {@link #error()}.
+   * Returns a copy of headers passed to {@link ClientCall.Listener#onHeaders(Metadata)}.
+   *
+   * @since 5.12
+   */
+  @Override public Metadata headers() {
```
anuraaga, line 75: Thanks!

instrumentation/grpc/src/main/java/brave/grpc/GrpcRequest.java
```diff
@@ -0,0 +1,54 @@
...
+ * @see GrpcResponse
+ * @see GrpcClientRequest
+ * @see GrpcServerRequest
+ * @see RpcTracing#clientRequestParser()
+ * @see RpcTracing#serverRequestParser()
+ * @since 5.12
+ */
+// NOTE: gRPC is Java 1.7+, so we cannot add methods to this later
+public interface GrpcRequest {
+  MethodDescriptor<?, ?> methodDescriptor();
```
anuraaga, line 51: Suggest just `method()`

instrumentation/grpc/src/main/java/brave/grpc/TracingClientInterceptor.java
```diff
@@ -117,9 +90,14 @@ void finishWithError(@Nullable Span span, Throwable error) {
...
         super.start(responseListener, headers);
       } catch (Throwable e) {
+        propagateIfFatal(e);
+
         // Another interceptor may throw an exception during start, in which case no other
         // callbacks are called, so go ahead and close the span here.
-        finishWithError(spanRef.getAndSet(null), e);
+        //
+        // See instrumentation/grpc/RATIONALE.md for why we don't use the handler here
+        spanRef.getAndSet(null);
```
anuraaga, line 99: Either just `set` or pass the return value to the next line.

instrumentation/dubbo/src/main/java/brave/dubbo/TracingFilter.java
adriancole: we usually have synchronous methods finish once they return, whether that's a synchronous exception or result.

instrumentation/grpc/src/main/java/brave/grpc/GrpcRequest.java
adriancole: agree. that's a lot better and now's the only time to change it!

instrumentation/grpc/src/main/java/brave/grpc/TracingClientInterceptor.java
adriancole: set is better yeah. thanks

instrumentation/grpc/src/main/java/brave/grpc/GrpcRequest.java
adriancole: ergh. now I remember. we define method() in RpcRequest already. I'll make a comment

instrumentation/dubbo/src/main/java/brave/dubbo/TracingFilter.java
adriancole: rewrote the code around it a little.

instrumentation/dubbo-rpc/src/main/java/brave/dubbo/rpc/TracingFilter.java
```diff
@@ -121,21 +121,25 @@ public void setRpcTracing(RpcTracing rpcTracing) {
...
-    boolean deferFinish = false;
+    boolean isSynchronous = true;
     Scope scope = currentTraceContext.newScope(span.context());
     Result result = null;
     Throwable error = null;
     try {
       result = invoker.invoke(invocation);
       error = result.getException();
       Future<Object> future = rpcContext.getFuture(); // the case on async client invocation
-      if (future instanceof FutureAdapter) {
```
adriancole, line 135: before, this instanceof was also checking null. In practice, this is always a (dubbo) FutureAdapter

adriancole:
@anuraaga you're the hero of this change! thanks for the weeks of review on this (and all the ancillary changes around it)

PR #1000, opened 2019-10-01
Adds basic infrastructure for RPC request sampling

adriancole:
This adds an abstraction layer for RPC, based on non-deprecated code in
brave-instrumentation-http. It will be a work in progress until at least
Dubbo and gRPC are integrated (ideally also Armeria even though in
another repo.

This is limited to sampling logic to make it easy to integrate and also
as that's the more in-demand functionality. Most notably, we only need
two properties: RPC service and method, to satisfy the first use cases
of RPC sampling.

Ex. Here's a sampler that traces 100 "GetUserToken" requests per second. This
doesn't start new traces for requests to the health check service. Other
requests will use a global rate provided by the tracing component.

```java
import static brave.rpc.RpcRequestMatchers.*;

rpcTracing = rpcTracingBuilder.serverSampler(RpcRuleSampler.newBuilder()
  .putRule(serviceEquals("grpc.health.v1.Health"), Sampler.NEVER_SAMPLE)
  .putRule(methodEquals("GetUserToken"), RateLimitingSampler.create(100))
  .build()).build();

grpcTracing = GrpcTracing.create(rpcTracing);
```

Thanks @trustin for helping me think through namings and such

### DIFF
--- a/instrumentation/grpc/src/main/java/brave/grpc/GrpcClientRequest.java
+++ b/instrumentation/grpc/src/main/java/brave/grpc/GrpcClientRequest.java
@@ -1,0 +1,42 @@
+/*
+ * Copyright 2013-2019 The OpenZipkin Authors
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License"); you may not use this file except
+ * in compliance with the License. You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software distributed under the License
+ * is distributed on an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express
+ * or implied. See the License for the specific language governing permissions and limitations under
+ * the License.
+ */
+package brave.grpc;
+
+import brave.rpc.RpcClientRequest;
+
+// intentionally not yet public until we add tag parsing functionality
+final class GrpcClientRequest extends RpcClientRequest {
+  final String fullMethodName;
+
+  GrpcClientRequest(String fullMethodName) {
+    if (fullMethodName == null) throw new NullPointerException("fullMethodName == null");
+    this.fullMethodName = fullMethodName;
+  }
+
+  @Override public Object unwrap() {
+    return this;
+  }
+
+  @Override public String method() {
+    int index = fullMethodName.lastIndexOf('/');
+    if (index == -1) return null;
+    return fullMethodName.substring(index);
+  }
+
+  @Override public String service() {
+    int index = fullMethodName.lastIndexOf('/');
+    if (index == -1) return null;
+    return fullMethodName.substring(0, index);
+  }
+}
--- a/instrumentation/grpc/src/main/java/brave/grpc/GrpcClientRequest.java
+++ b/instrumentation/grpc/src/main/java/brave/grpc/GrpcClientRequest.java
@@ -25,7 +25,7 @@ final class GrpcClientRequest extends RpcClientRequest {
   static final Setter<GrpcClientRequest, Metadata.Key<String>> SETTER =
     new Setter<GrpcClientRequest, Metadata.Key<String>>() { // retrolambda no like
       @Override public void put(GrpcClientRequest request, Metadata.Key<String> key, String value) {
-        request.metadata(key, value);
+        request.setMetadata(key, value);
       }
 
       @Override public String toString() {
@@ -59,7 +59,7 @@ final class GrpcClientRequest extends RpcClientRequest {
     return this;
   }
 
-  <T> void metadata(Metadata.Key<T> key, T value) {
+  <T> void setMetadata(Metadata.Key<T> key, T value) {
     Metadata metadata = this.metadata;
     if (metadata == null) {
       assert false : "This code should never be called when null";
--- a/instrumentation/grpc/src/main/java/brave/grpc/GrpcClientRequest.java
+++ b/instrumentation/grpc/src/main/java/brave/grpc/GrpcClientRequest.java
@@ -13,15 +13,32 @@
  */
 package brave.grpc;
 
+import brave.internal.Nullable;
+import brave.propagation.Propagation.Setter;
 import brave.rpc.RpcClientRequest;
+import io.grpc.ClientCall;
+import io.grpc.Metadata;
+import io.grpc.MethodDescriptor;
 
 // intentionally not yet public until we add tag parsing functionality
 final class GrpcClientRequest extends RpcClientRequest {
-  final String fullMethodName;
+  static final Setter<GrpcClientRequest, Metadata.Key<String>> SETTER =
+    new Setter<GrpcClientRequest, Metadata.Key<String>>() { // retrolambda no like
+      @Override public void put(GrpcClientRequest request, Metadata.Key<String> key, String value) {
+        request.metadata(key, value);
+      }
 
-  GrpcClientRequest(String fullMethodName) {
-    if (fullMethodName == null) throw new NullPointerException("fullMethodName == null");
-    this.fullMethodName = fullMethodName;
+      @Override public String toString() {
+        return "GrpcClientRequest::metadata";
+      }
+    };
+
+  final String fullMethodName;
+  @Nullable volatile Metadata metadata; // nullable due to lifecycle of gRPC request
+
+  GrpcClientRequest(MethodDescriptor<?, ?> methodDescriptor) {
+    if (methodDescriptor == null) throw new NullPointerException("methodDescriptor == null");
+    this.fullMethodName = methodDescriptor.getFullMethodName();
   }
 
   @Override public Object unwrap() {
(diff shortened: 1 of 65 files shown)


instrumentation/rpc/src/main/java/brave/rpc/RpcRequestMatchers.java
```diff
@@ -0,0 +1,105 @@
...
+      return method.hashCode();
+    }
+
+    @Override public String toString() {
+      return "RpcMethodEquals(" + method + ")";
+    }
+  }
+
+  /**
+   * Matcher for case-sensitive RPC service names, such as "zipkin.proto3.SpanService/Report" or
```
adriancole:
```suggestion
   * Matcher for case-sensitive RPC service names, such as "zipkin.proto3.SpanService" or
```

adriancole:
added gRPC impl

instrumentation/grpc/src/main/java/brave/grpc/TracingClientInterceptor.java
```diff
@@ -62,7 +67,7 @@
   public <ReqT, RespT> ClientCall<ReqT, RespT> interceptCall(
     final MethodDescriptor<ReqT, RespT> method, final CallOptions callOptions,
     final Channel next) {
-    Span span = tracer.nextSpan();
+    Span span = tracer.nextSpan(sampler, new GrpcClientRequest(method.getFullMethodName()));
```
adriancole: sweet right? that's all it needed.

instrumentation/grpc/src/test/java/brave/grpc/ITTracingServerInterceptor.java
```diff
@@ -333,6 +337,23 @@ public void serverParserTest() throws Exception {
     assertThat(span.tags()).hasSize(10);
   }
 
+  @Test public void customSampler() throws Exception {
+    RpcTracing rpcTracing = RpcTracing.newBuilder(tracing).clientSampler(RpcRuleSampler.newBuilder()
+      .putRule(methodEquals("SayHelloWithManyReplies"), Sampler.NEVER_SAMPLE)
```
adriancole: @narayaruna this bud's for you

instrumentation/rpc/RATIONALE.md
```diff
@@ -0,0 +1,61 @@
...
+* gRPC - full method "zipkin.proto3.SpanService/Report" returns "Report"
+* Apache Thrift - full method "scribe.Log" returns "Log"
+* Redis - "EXISTS" command returns "EXISTS"
+
+`rpc.service` - The fully-qualified, case-sensitive service path as defined in IDL.
+
+Examples
+* gRPC - full method "zipkin.proto3.SpanService/Report" returns "zipkin.proto3.SpanService"
+* Apache Thrift - full method "scribe.Log" returns "scribe"
+* Redis - "EXISTS" command returns null
```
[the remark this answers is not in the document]
adriancole: Fair enough I'll use DNS as an example instead.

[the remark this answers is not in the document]
adriancole: actually dubbo since we have it here anyway

instrumentation/grpc/src/main/java/brave/grpc/GrpcClientRequest.java
```diff
@@ -0,0 +1,71 @@
...
+    return GrpcParser.service(fullMethodName);
+  }
+
+  /** Call on {@link ClientCall#start(ClientCall.Listener, Metadata)} */
+  GrpcClientRequest metadata(Metadata metadata) {
+    this.metadata = metadata;
+    return this;
+  }
+
+  <T> void metadata(Metadata.Key<T> key, T value) {
```
anuraaga: Maybe should have `set` in the method name since we remove existing values for the key

instrumentation/grpc/src/test/java/brave/grpc/GrpcClientRequestTest.java
```diff
@@ -0,0 +1,48 @@
...
+    metadata.put(b3Key, "0");
+
+    request.metadata(metadata).metadata(b3Key, "1");
+
+    assertThat(request.metadata.get(b3Key))
+      .isEqualTo("1");
+  }
+
+  @Test public void metadata_null() {
+    assertThatThrownBy(() -> request.metadata(b3Key, "1")) // doesn't NPE
```
anuraaga: Instead of comment is it better to be `isNotInstanceOf(NullPointerException.class`?

instrumentation/grpc/src/test/java/brave/grpc/ITTracingClientInterceptor.java
```diff
@@ -344,6 +347,27 @@ public void clientParserTestStreamingResponse() throws Exception {
...
+    RpcTracing rpcTracing = RpcTracing.newBuilder(tracing).clientSampler(RpcRuleSampler.newBuilder()
+      .putRule(methodEquals("SayHelloWithManyReplies"), Sampler.NEVER_SAMPLE)
+      .build()).build();
+
+    grpcTracing = GrpcTracing.create(rpcTracing);
+    client = newClient();
+
+    // unsampled
+    assertThat(GreeterGrpc.newBlockingStub(client).sayHelloWithManyReplies(HELLO_REQUEST))
+      .hasNext(); // Request is lazy, so you must invoke the iterator
```
anuraaga: Since this isn't actually what we're testing, maybe we don't need `assertThat` and can just call `hasNext` directly.

instrumentation/rpc/RATIONALE.md
```diff
@@ -0,0 +1,61 @@
...
+
+### Wait.. not everything has a `service` namespace?
+
+Not all RPC implementation has a service/namespace. For example, Redis commands
+have a flat namespace. In our model, not applicable properties are represented
+as null. The important part is decoupling `service` from `method` allows one
+property, `method`, to be extractable more frequently vs if we forced
+concatenation through a `full method name` as exists in gRPC.
+
+### But what if I want a `full method name` as exists in gRPC?
```
anuraaga: This is a little confusing, since we do actually have the full method name as the name of the span - only difference is it's lowercase, but then this would become "I want a `case-sensitive full method name`" which would anyone ever ask? ;)

instrumentation/rpc/RATIONALE.md
```diff
@@ -0,0 +1,61 @@
...
+
+We can have the talk about what an idiomatic term could be for a fully
+qualified method name. Only gRPC uses the term `full method name`, so it isn't
+as clean to lift that term up, vs other names such as `service` than exist in
```
anuraaga:
```suggestion
as clean to lift that term up, vs other names such as `service` that exist in
```

instrumentation/rpc/RATIONALE.md
anuraaga: Though after reading the `Matchers` code now I guess this is referring more to matching ability rather than data model, and we'd match by the property not the span name.

instrumentation/rpc/RATIONALE.md
```diff
@@ -0,0 +1,61 @@
...
+multiple tools such as Apache Thrift. For now, concat on your own.
```
anuraaga: can provide some sample code like `using and(serviceEquals(...), methodEquals(...))`

instrumentation/rpc/RATIONALE.md
adriancole: yep you got it right. this happens before what the span name is chosen as. Also, this is an abstract model.

adriancole: I'll add something about this topic to the rationale.

instrumentation/rpc/RATIONALE.md
adriancole: yeah I also was thinking about that

instrumentation/grpc/src/main/java/brave/grpc/GrpcClientRequest.java
adriancole: ok

instrumentation/grpc/src/test/java/brave/grpc/GrpcClientRequestTest.java
adriancole: agree!

instrumentation/grpc/src/test/java/brave/grpc/ITTracingClientInterceptor.java
adriancole: good point

adriancole:
@anuraaga PTAL again at grpc as I'm switching to dubbo meanwhile. thanks for the review!

adriancole:
@anuraaga yeah the idea is to do parsing and agreed it is more straightforward in google grpc which hides the transport. Due to issues like you mentioned, I wanted to keep the simple part (sampling) separate.. this keeps the latter change concentrated on these bits (including if http request should have a nullable getter for rpcRequest or not)

adriancole:
suprisingly I didn't mention this earlier, but #999 will be rebased over this, and has scope for parsing etc.

adriancole:
so one thing I will clear up in the text is that the method name is not required to be different than the java method. There are java based RPC, of course, like Dubbo. It is just preferred to use IDL when something is backed by IDL. We cannot enforce that there is always going to be IDL in other words.

adriancole:
I added apache dubbo. Last thing is to convert that code to alibaba dubbo

adriancole:
Thanks for the help!

PR #1001, opened 2019-10-02
Reverts permittance of duplicate handlers

adriancole:
It is a worse bug to allow duplicate handlers than it is to allow
duplicate configuration. Duplicate handlers translate directly into
runtime overhead, particularly as there is special casing for one
handler.

Throwing an exception was suggested here #961, but that is not a good
choice as it would almost certainly lead to someone having to do an
emergency release to undo it. This changes to logging on dupe, so that
we don't propagate this problem into future work such as collector
handlers.

See https://github.com/openzipkin/zipkin/pull/2807#discussion_r330288383

### DIFF
--- a/brave/src/main/java/brave/Tracing.java
+++ b/brave/src/main/java/brave/Tracing.java
@@ -29,7 +29,9 @@ import brave.sampler.Sampler;
 import brave.sampler.SamplerFunction;
 import java.io.Closeable;
 import java.util.ArrayList;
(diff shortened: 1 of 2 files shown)


adriancole:
cc @anuraaga @jeqo

adriancole:
build has been red since I hooked up dubbo. karmatic :P https://github.com/openzipkin/brave/pull/1003

PR #1005, opened 2019-10-10
doc: fix missed package name

ueokande:
Fix midded package name of [SpanCustomizingHandlerInterceptor](https://github.com/openzipkin/brave/blob/master/instrumentation/spring-webmvc/src/main/java/brave/spring/webmvc/SpanCustomizingHandlerInterceptor.java) in the README.md


adriancole:
thanks!

PR #1006, opened 2019-10-21
Adds `RpcTracing` and `HttpTracing.current()` and fixes class loading

adriancole:
When porting Cassandra to `RpcTracing`, I noticed we have no means to
statically look up the tracing instance, such as we do in `Tracing`.
This backfills the same functionality as `brave.Tracing`.

While implementing classloader tests, I notice we load logging types
even if they will be never used. This takes caution to guard usage,
which allows us to test class unloading properly.

To implement this, we had to move the implicit log hook for orphan
tracking to a new option `Tracing.Builder.trackOrphans()`.


adriancole:
@jeqo I'm mainly raising this now so that when we copy/paste for Messaging, we copy/paste the right things :)

brave/src/main/java/brave/Tracing.java
```diff
@@ -97,16 +97,25 @@ public final Clock clock(TraceContext context) {
 
   abstract public ErrorParser errorParser();
 
-  // volatile for visibility on get. writes guarded by Tracing.class
-  static volatile Tracing current = null;
+  // volatile for get visibility. writes guarded by Tracing.class. Object to ensure unloadable
```
adriancole: strictly speaking, this was unloadable, but anything that used it was not. I made this Object to be consistent across our "current" tests.

brave/src/main/java/brave/Tracing.java
```diff
@@ -97,16 +97,25 @@ public final Clock clock(TraceContext context) {
...
+  }
 
   /**
    * Returns the most recently created tracer if its component hasn't been closed. null otherwise.
    *
    * <p>This object should not be cached.
    */
   @Nullable public static Tracer currentTracer() {
-    Tracing tracing = current;
+    Tracing tracing = (Tracing) current;
```
adriancole: in case forgot the magic java rules like I did.. no, cast won't NPE. Also, we have tests too.

brave/src/main/java/brave/internal/Platform.java
```diff
@@ -34,8 +34,9 @@
  */
 public abstract class Platform {
   private static final Platform PLATFORM = findPlatform();
-  private static final Logger LOG = Logger.getLogger(brave.Tracer.class.getName());
 
+  // Lazy init as a constant eagerly initializes hooks like log4j2 even when the tracer is no-op
```
anuraaga: I think we can avoid the volatile read / missing out on potential inlining optimizations by using a nested class

```java
// Use nested class to ensure logger isn't initialized unless it is accessed once.
private static class LoggerHolder {
  private static final Logger LOG = Logger.getLogger(brave.Tracer.class.getName());
}

public void log() {
  LoggerHolder.LOG.log
}
```

anuraaga: The same doesn't have to apply to `linkLocalIp` since that's only accessed once when constructing `Tracing` which should be infrequent - maybe it doesn't even need to be memoized because of that actually.

brave/src/main/java/brave/internal/Platform.java
```diff
@@ -76,18 +77,27 @@ public static Platform get() {
...
     lr.setParameters(params);
     if (thrown != null) lr.setThrown(thrown);
-    LOG.log(lr);
+    logger.log(lr);
+  }
+
+  Logger logger() {
+    Logger result = this.logger;
+    if (result != null) return result;
+    // synchronization not needed as worst case we overwrite the field twice
```
anuraaga: I think worst case is overwriting it 1000 times with 1000 threads (very unlikely of course). But anyways, if not switching to nested class pattern, I'd make the field non-volatile since we're doing best-effort only to minimize logger instantiation anyways.

brave/src/main/java/brave/internal/recorder/PendingSpans.java
```diff
@@ -140,10 +139,12 @@ void reportOrphanedSpans() {
       );
 
       if (caller != null) {
+        // Verification of this is done manually because log4j2 in our test classpath makes it too
```
anuraaga: This comes out of nowhere in business logic and is a little confusing. I'd just remove it - it's common knowledge that verifying log messages is hard and it's very rare to see anyone do it.

instrumentation/http/src/main/java/brave/http/HttpTracing.java
```diff
@@ -243,4 +252,32 @@ public HttpTracing build() {
...
+   * Returns the most recently created tracing component iff it hasn't been closed. null otherwise.
+   *
+   * <p>This object should not be cached.
+   */
+  @Nullable public static HttpTracing current() {
+    return (HttpTracing) current;
+  }
+
+  private void maybeSetCurrent() {
+    if (current != null) return;
```
anuraaga: Is it simpler to use `AtomicReference`? Pretty sure performance of `current` is the same (inlines the volatile read in the reference), and performance of the other methods (not that it matters) is better with less code.

```java
current.compareAndSet(null, this);
```

```java
current.compareAndSet(this, null);
```

brave/src/main/java/brave/internal/recorder/PendingSpans.java
adriancole: 👍

adriancole:
all good suggestions, @anuraaga, will weave them in!

brave/src/main/java/brave/internal/Platform.java
adriancole: awesome idea on the holder thing!

there's a pattern where people create multiple instances of brave, for better or for worse.. I'll leave the linkedlocal ip thing alone for now.

instrumentation/http/src/main/java/brave/http/HttpTracing.java
adriancole: wonderful

PR #1008, opened 2019-10-22
Log all exceptions when re-setting message props

rmichela:
When receiving JMS messages from Amazon SQS, the message sometimes
has headers like JMSXGroupID and JMS_SQS_DeduplicationId added by
SQS with empty string values.

When brave-instrumentation-jms is wrapped around an SQSConnectionFactory,
PropertyFilter.filterProperties() attempts to re-set these properties
on the received SQSMessage object. Doing so fails becaus SQSMessage
throws an IllegalArgumentException if either the property name or
value are empty. (Even though the properties came from SQSMessage
to begin with.) The IllegalArgumentException escapes the
try/catch block only looking for JMSException, and results in the
message silently failing to process.

This change widens the scope of the catch block so that any failure
when attempting to re-set a message property is caught, preventing
an unexpected exception type from blowing up the entire message.

### DIFF
--- a/instrumentation/jms/src/main/java/brave/jms/PropertyFilter.java
+++ b/instrumentation/jms/src/main/java/brave/jms/PropertyFilter.java
@@ -14,6 +14,8 @@
 package brave.jms;
 
 import brave.internal.Platform;
+import zipkin2.Call;
+
 import java.util.ArrayList;
 import java.util.Enumeration;
 import java.util.List;
@@ -76,7 +78,8 @@ final class PropertyFilter {
       String name = out.get(i).toString();
       try {
         message.setObjectProperty(name, out.get(i + 1));
-      } catch (JMSException e) {
+      } catch (Throwable e) {
+        Call.propagateIfFatal(e);
         log(e, "error setting property {0} on message {1}", name, message);
         // continue on error when re-setting properties as it is better than not.
       }
--- a/instrumentation/jms/src/test/java/brave/jms/PropertyFilterTest.java
+++ b/instrumentation/jms/src/test/java/brave/jms/PropertyFilterTest.java
@@ -22,8 +22,14 @@ import org.junit.After;
 import org.junit.Test;
 
 import static brave.test.util.ClassLoaders.assertRunIsUnloadable;
-import static org.assertj.core.api.Assertions.*;
-import static org.mockito.Mockito.*;
+import static org.assertj.core.api.Assertions.assertThat;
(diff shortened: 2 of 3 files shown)


rmichela:
Here is where `SQSMessage` throws the unexpected `IllegalArgumentException`. 

https://github.com/awslabs/amazon-sqs-java-messaging-lib/blob/master/src/main/java/com/amazon/sqs/javamessaging/message/SQSMessage.java#L904-L909

instrumentation/jms/src/main/java/brave/jms/PropertyFilter.java
```diff
@@ -76,7 +76,7 @@ static void filterProperties(Message message, Set<String> namesToClear, List<Obj
       String name = out.get(i).toString();
       try {
         message.setObjectProperty(name, out.get(i + 1));
-      } catch (JMSException e) {
+      } catch (Throwable e) {
```
adriancole: we tend to use Call.propagatedIfFatal when catching throwable.

adriancole:
Thanks!

PropertyFilterTest probably could get a new test with an example IAE with the link you posted about SQS. mind adding that?

instrumentation/jms/src/main/java/brave/jms/PropertyFilter.java
rmichela: Fixed in next line.

rmichela:
Added tests and addressed feedback.

instrumentation/jms/src/test/java/brave/jms/PropertyFilterTest.java
```diff
@@ -15,13 +15,15 @@
...
 import javax.jms.JMSException;
+import javax.jms.Message;
 import javax.jms.TextMessage;
 import org.apache.activemq.command.ActiveMQTextMessage;
 import org.junit.After;
 import org.junit.Test;
 
 import static brave.test.util.ClassLoaders.assertRunIsUnloadable;
-import static org.assertj.core.api.Assertions.assertThat;
+import static org.assertj.core.api.Assertions.*;
```
adriancole: do you mind undoing wildcard imports? we don't use wildcard anywhere in our projects.

rmichela: Fixed. (shakes fist at IDEs)

rmichela:
It's a testament to Brave's quality. Years of use and this is the first bug we've found. 😁

PR #1009, opened 2019-10-24
Adds OSGi manifests to all non-test jars

### LINKED ISSUE #266: Can we package zipkin dependencies as OSGi bundles
It would be great if we can package zipkin dependencies as OSGi bundles.

adriancole:
This uses file-based detection to ensure projects with bnd.bnd end up with both OSGi and Java Modules manifests

Fixes #266

### DIFF
--- a/brave/bnd.bnd
+++ b/brave/bnd.bnd
@@ -1,8 +1,8 @@
 Import-Package: \
   *
 Export-Package: \
-	brave,\
+  brave,\
   brave.handler,\
-	brave.propagation,\
-	brave.sampler,\
-	brave.internal;braveinternal=true;mandatory:=braveinternal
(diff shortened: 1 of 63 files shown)


adriancole:
I think this PR is not even half-way, so the refactoring is to prevent this from becoming 1k lines of cruft.

adriancole:
fyi @anuraaga @jeqo I've managed to figure out how to remove the redundancy here. Will finish this up in a bit.

adriancole:
https://github.com/openzipkin/brave-karaf/pull/44 works with this change (passes iTests)

adriancole:
I will do same in zipkin-reporter so it is all consistent and stuff :P

PR #1011, opened 2019-10-24
Adds messaging sampler

adriancole:
this is pulling out #914 like we did for RPC

- [x] kafka-clients
- [x] jms
- [x] spring-rabbit

### DIFF
--- a/instrumentation/kafka-clients/src/main/java/brave/kafka/clients/KafkaTracing.java
+++ b/instrumentation/kafka-clients/src/main/java/brave/kafka/clients/KafkaTracing.java
@@ -16,6 +16,7 @@ package brave.kafka.clients;
 import brave.Span;
 import brave.SpanCustomizer;
 import brave.Tracing;
+import brave.messaging.MessagingTracing;
 import brave.propagation.B3SingleFormat;
 import brave.propagation.Propagation;
 import brave.propagation.TraceContext;
@@ -37,23 +38,32 @@ import static brave.kafka.clients.KafkaPropagation.TEST_CONTEXT;
 
 /** Use this class to decorate your Kafka consumer / producer and enable Tracing. */
 public final class KafkaTracing {
-
   public static KafkaTracing create(Tracing tracing) {
-    return new Builder(tracing).build();
+    return newBuilder(tracing).build();
+  }
+
+  /** @since 5.9 */
+  public static KafkaTracing create(MessagingTracing messagingTracing) {
+    return newBuilder(messagingTracing).build();
   }
 
   public static Builder newBuilder(Tracing tracing) {
-    return new Builder(tracing);
+    return newBuilder(MessagingTracing.create(tracing));
+  }
+
+  /** @since 5.9 */
+  public static Builder newBuilder(MessagingTracing messagingTracing) {
+    return new Builder(messagingTracing);
   }
 
   public static final class Builder {
-    final Tracing tracing;
+    final MessagingTracing messagingTracing;
     String remoteServiceName = "kafka";
     boolean writeB3SingleFormat;
 
-    Builder(Tracing tracing) {
-      if (tracing == null) throw new NullPointerException("tracing == null");
-      this.tracing = tracing;
+    Builder(MessagingTracing messagingTracing) {
+      if (messagingTracing == null) throw new NullPointerException("messagingTracing == null");
+      this.messagingTracing = messagingTracing;
     }
 
     /**
@@ -88,9 +98,9 @@ public final class KafkaTracing {
   final String remoteServiceName;
 
   KafkaTracing(Builder builder) { // intentionally hidden constructor
-    this.tracing = builder.tracing;
+    this.tracing = builder.messagingTracing.tracing();
     this.extractor = tracing.propagation().extractor(KafkaPropagation.GETTER);
-    List<String> keyList = builder.tracing.propagation().keys();
+    List<String> keyList = tracing.propagation().keys();
     // Use a more efficient injector if we are only propagating a single header
     if (builder.writeB3SingleFormat || keyList.equals(Propagation.B3_SINGLE_STRING.keys())) {
       TraceContext testExtraction = extractor.extract(B3_SINGLE_TEST_HEADERS).context();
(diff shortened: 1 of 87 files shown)


instrumentation/messaging/src/main/java/brave/messaging/MessagingRequest.java
```diff
@@ -0,0 +1,89 @@
...
+   */
+  public abstract Object unwrap();
+
+  /**
+   * The unqualified, case-sensitive semantic message operation name. The currently defined names
+   * are "send", "receive", "bulk-send" and "bulk-receive".
+   *
+   * <p>Examples:
+   * <pre><ul>
+   *   <li>Amazon SQS - {@code AmazonSQS.sendMessageBatch()} returns "bulk-send" as it can send many messages</li>
```
adriancole: ps I'm still on the fence about bulk-X vs X-bulk :P

devinsba: I'm thinking `send-bulk`

jorgheymans:
stepping back a bit, why conveying bulk/batch notion at all in the
operation name? It's at most a hint that a batch send or retrieve could
have happened. Trying to see if i would care enough to see this level of
detail in a trace, and if it could not lead to false assumptions or
reasoning ie 'receive-batch' in the trace but the kafka consumer is
actually configured with max.poll.records=1 ...

adriancole:
> stepping back a bit, why conveying bulk/batch notion at all in the
operation name? It's at most a hint that a batch send or retrieve could
have happened. Trying to see if i would care enough to see this level of
detail in a trace, and if it could not lead to false assumptions or
reasoning ie 'receive-batch' in the trace but the kafka consumer is
actually configured with max.poll.records=1 ...

For now, I agree. I'll move your comments into RATIONALE.md and we can change the impl easier to include batch later vs remove it. If we did that, we'd update RATIONALE with why

adriancole:
updated with rationale

adriancole:
@jeqo I added a rationale we kindof always forget about clearing headers. Notably, I've something in here about how to handle JMS which I think is more elegant than before.

adriancole:
added kafka sampling impl finally. took a while to figure out what the best way to do this is..

adriancole:
OK I integrated the header work from prior pull requests into kafka. With a look, I'll proceed to JMS

instrumentation/kafka-clients/src/main/java/brave/kafka/clients/KafkaTracing.java
```diff
@@ -110,8 +109,9 @@ public KafkaTracing build() {
     this.remoteServiceName = builder.remoteServiceName;
   }
 
-  public Tracing tracing() {
-    return tracing;
+  /** @since 5.9 exposed for Kafka Streams tracing. */
```
adriancole: @simondean fyi

instrumentation/kafka-streams/src/main/java/brave/kafka/streams/KafkaStreamsTracing.java
```diff
@@ -44,25 +49,27 @@
...
-    return new KafkaStreamsTracing.Builder(KafkaTracing.create(tracing)).build();
+    return create(KafkaTracing.create(tracing));
   }
 
+  /** @since 5.9 */
   public static KafkaStreamsTracing create(KafkaTracing kafkaTracing) {
-    return new KafkaStreamsTracing.Builder(kafkaTracing).build();
-  }
-
-  public KafkaTracing kafkaTracing() {
```
adriancole: no use of this so I removed this. In general, we expose things public upon use case, not before.

instrumentation/kafka-clients/src/main/java/brave/kafka/clients/KafkaTracing.java
```diff
@@ -134,27 +140,44 @@ public Tracing tracing() {
...
+  /** Creates a potentially noop remote span representing this request */
+  Span nextMessagingSpan(
+    SamplerFunction<MessagingRequest> sampler,
+    MessagingRequest request,
+    TraceContextOrSamplingFlags extracted
+  ) {
+    Boolean sampled = extracted.sampled();
+    // only recreate the context if the messaging sampler made a decision
+    if (sampled == null && (sampled = sampler.trySample(request)) != null) {
+      extracted = extracted.sampled(sampled.booleanValue());
```
jeqo: sweet! so this is the place where sampling will be overwritten if `samplerFunction` is provided, effectively solving the issue I post a few days ago in gitter :)

instrumentation/kafka-clients/src/main/java/brave/kafka/clients/KafkaTracing.java
```diff
@@ -81,33 +85,35 @@ public KafkaTracing build() {
...
+  final SamplerFunction<MessagingRequest> producerSampler, consumerSampler;
   final Set<String> propagationKeys;
   final String remoteServiceName;
 
   KafkaTracing(Builder builder) { // intentionally hidden constructor
-    this.tracing = builder.tracing;
-    this.extractor = tracing.propagation().extractor(KafkaPropagation.GETTER);
-    List<String> keyList = builder.tracing.propagation().keys();
-    // Use a more efficient injector if we are only propagating a single header
-    if (builder.writeB3SingleFormat || keyList.equals(Propagation.B3_SINGLE_STRING.keys())) {
```
jeqo: Really happy this is gone :)

instrumentation/kafka-streams/src/main/java/brave/kafka/streams/TracingFilter.java
```diff
@@ -32,7 +32,7 @@
   TracingFilter(KafkaStreamsTracing tracing, String spanName,
     Predicate<K, V> delegatePredicate, boolean filterNot) {
     this.kafkaStreamsTracing = tracing;
```
jeqo: nit: should we align variable names?

instrumentation/kafka-clients/src/main/java/brave/kafka/clients/KafkaTracing.java
adriancole: yep though primary sampling is only at the start of a trace

instrumentation/kafka-streams/src/main/java/brave/kafka/streams/TracingFilter.java
adriancole: sure I will sweep on that

PR #1012, opened 2019-10-28
Removes accidental zipkin lib dependency and fixes throwable catching

adriancole:
In #1008, we only fixed one place where semi-implementation of JMS could
raise unexpected throwables. This fixes the rest and also removes the
accidental compile dep on zipkin2.


adriancole:
landing this prior to the osgi thing

PR #1014, opened 2019-10-29
Fix: keep context on Producer callback

### LINKED ISSUE #1007: TraceContext missing on TracingProducer's callback when not sampling
## Describe the Bug
The TracingProducer returned from KafkaTracing wraps the underlying Producer's `send(ProducerRecord, Callback)` method. When this method is invoked with a span in context that has sampling enabled, a new child span is appropriately created and scoped when the Callback is asynchronously invoked by the underlying Producer. However, if the span in context does not have the sampling flag set, there is no TraceContext when the callback is invoked. This makes it impossible to do log correlation or extra field propagation for unsampled spans.

For background - `TracingCallback.create(Callback, Span, CurrentTraceContext)` returns the delegate back if `Span.isNoop()` is true. However, it does not appear to be the overall design of Brave to not continue a TraceContext for NoopSpan - the other common instrumentation doesn't do this. Indeed, the javadoc on `Span.isNoop()` indicates that no recording or reporting is done, but that the span should still be injected downstream - the code in TracingCallback makes that impossible, as the span is never put into scope or otherwise made available to consuming code.

I discussed this with @devinsba on gitter, who suggested I file this bug. There was some discussion there about the nature of a NoopSpan - upon investigation, a NoopSpan is created by `Tracer.nextSpan()` whenever `FLAG_SAMPLED` is not set (by way of `Tracer._toSpan(TraceContext)`

There is an existing test that confirms this behavior - I believe the test is wrong and the callback should be invoked with a span in scope regardless of sampling: https://github.com/openzipkin/brave/blob/master/instrumentation/kafka-clients/src/test/java/brave/kafka/clients/TracingCallbackTest.java#L28

## Steps to Reproduce
1. Use KafkaTracing.producer(Producer) to instrument a Kafka Producer
2. Set a span in scope that has sampling disabled
3. Call Producer.send(ProducerRecord, Callback) on the instrumented producer
4. Observe that, when Callback is invoked asynchronously on a clean thread, CurrentTraceContext.get() is not set.

## Expected Behaviour
When the callback is invoked, CurrentTraceContext.get() should contain a TraceContext with the same tracing information, extras, etc as existed when Producer.send() was invoked.

jeqo:
Fix #1007

### DIFF
--- a/instrumentation/kafka-clients/src/test/java/brave/kafka/clients/TracingCallbackTest.java
+++ b/instrumentation/kafka-clients/src/test/java/brave/kafka/clients/TracingCallbackTest.java
@@ -18,20 +18,25 @@ import brave.propagation.TraceContextOrSamplingFlags;
 import org.apache.kafka.clients.producer.Callback;
 import org.apache.kafka.clients.producer.RecordMetadata;
 import org.apache.kafka.common.TopicPartition;
+import org.junit.Ignore;
 import org.junit.Test;
 
 import static org.assertj.core.api.Assertions.assertThat;
+import static org.mockito.ArgumentMatchers.any;
 import static org.mockito.Mockito.mock;
 import static org.mockito.Mockito.verify;
+import static org.mockito.Mockito.when;
 
 public class TracingCallbackTest extends BaseTracingTest {
   @Test public void create_returns_input_on_noop() {
     Span span = tracing.tracer().nextSpan(TraceContextOrSamplingFlags.NOT_SAMPLED);
 
-    Callback delegate = mock(Callback.class);
+    Callback delegate =
+        (metadata, exception) -> assertThat(tracing.tracer().currentSpan()).isEqualTo(span);
     Callback tracingCallback = TracingCallback.create(delegate, span, current);
 
-    assertThat(tracingCallback).isSameAs(delegate);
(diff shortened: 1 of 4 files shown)


jeqo:
cc @beowulfe here is an attempt to solve this bug. Let me know if it works on your side.

instrumentation/kafka-clients/src/test/java/brave/kafka/clients/TracingCallbackTest.java
```diff
@@ -18,20 +18,25 @@
...
     Span span = tracing.tracer().nextSpan(TraceContextOrSamplingFlags.NOT_SAMPLED);
 
-    Callback delegate = mock(Callback.class);
+    Callback delegate =
+        (metadata, exception) -> assertThat(tracing.tracer().currentSpan()).isEqualTo(span);
     Callback tracingCallback = TracingCallback.create(delegate, span, current);
 
-    assertThat(tracingCallback).isSameAs(delegate);
+    tracingCallback.onCompletion(null, null);
+    //assertThat(tracingCallback).isSameAs(delegate);
```
adriancole: accidental comment

beowulfe:
@jeqo - works perfectly, thank you for the fast solution!

PR #1015, opened 2019-10-29
Fixes javadoc settings for internal packages

adriancole:


pom.xml
```diff
@@ -682,9 +682,10 @@
 
   <profiles>
     <profile>
+      <!-- don't JDK 13 until https://github.com/google/error-prone/issues/1106 -->
       <id>error-prone-9+</id>
       <activation>
-        <jdk>[9,)</jdk>
+        <jdk>[9,12)</jdk>
```
anuraaga: Is this supposed to be `12]` to allow JDK 12? Currently it's not the same as the comment

adriancole: in my mind I skip 12 :P 12 is also busted but better to go to 13 right?

PR #1017, opened 2019-10-30
trim mysql statement name.

qinhang3:
The mysql span name like "select" will be "select\n"  if you are using mybatis or ibatis.
A sql generated by mybatis looks like 
`select\n          item_ids\n        from item_recommand\n        where\n          delete_mark = 0 and user_id = ?`
Then we get a span name "select\n"
A TRIM will fix it .


adriancole:
maybe we can look for index of new line before space?

PR #1018, opened 2019-10-30
remove " " and "\n" in end of mysql statement name.

qinhang3:
As your recommend, I look for index of new line and space both .
then we can get a span name without "\n" and " " in the end.

### DIFF
--- a/instrumentation/mysql/src/main/java/brave/mysql/TracingStatementInterceptor.java
+++ b/instrumentation/mysql/src/main/java/brave/mysql/TracingStatementInterceptor.java
@@ -50,7 +50,7 @@ public class TracingStatementInterceptor implements StatementInterceptorV2 {
     if (interceptedStatement instanceof PreparedStatement) {
       sql = ((PreparedStatement) interceptedStatement).getPreparedSql();
     }
-    int spaceIndex = sql.indexOf(' '); // Allow span names of single-word statements like COMMIT
+    int spaceIndex = Math.min(sql.indexOf(' '), sql.indexOf('\n')); // Allow span names of single-word statements like COMMIT
     span.kind(Span.Kind.CLIENT).name(spaceIndex == -1 ? sql : sql.substring(0, spaceIndex));
     span.tag("sql.query", sql);
     parseServerIpAndPort(connection, span);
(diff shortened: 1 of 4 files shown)


instrumentation/mysql/src/main/java/brave/mysql/TracingStatementInterceptor.java
```diff
@@ -50,7 +50,7 @@ public ResultSetInternalMethods preProcess(String sql, Statement interceptedStat
     if (interceptedStatement instanceof PreparedStatement) {
       sql = ((PreparedStatement) interceptedStatement).getPreparedSql();
     }
-    int spaceIndex = sql.indexOf(' '); // Allow span names of single-word statements like COMMIT
+    int spaceIndex = Math.min(sql.indexOf(' '), sql.indexOf('\n')); // Allow span names of single-word statements like COMMIT
```
adriancole: slick

adriancole:
can you please add p6spy as this prevents same issue there also

instrumentation/mysql/src/test/java/brave/mysql/ITTracingStatementInterceptor.java
```diff
@@ -34,6 +34,7 @@
 
 public class ITTracingStatementInterceptor {
   static final String QUERY = "select 'hello world'";
```
adriancole: in a unit test I think you want the old condition. I'm worried that Math.min won't work because if space is present and \n is not, won't it return -1 instead of the index of space?

instrumentation/mysql/src/main/java/brave/mysql/TracingStatementInterceptor.java
anuraaga: This is pretty inefficient since we need to iterate the string twice (and many queries won't have newline so it's a full scan of potentially very long ones).

Let's instead loop through the characters and return the first that matches

```
int findWhitespace(String str) {
for (int i = 0; i < str.length(); i++) {
  char c = str.charAt(i);
  switch (c) {
    case ' ':
    case '\n':
    case '\r':
    case '\t':
      return I;
    default:
      // Skip
 }
}
return -1;
}
```

Not sure how official it is but I found this grammar that lists those four as whitespace possibilities

https://github.com/antlr/grammars-v4/blob/master/mysql/Positive-Technologies/MySqlLexer.g4

adriancole: sgtm good point on long queries. I forgot literally these are full query expressions

adriancole:
@qinhang3 ps you can add commits to your branch instead of raising new pull requests. Adding new pull requests pollutes the issues list.

PR #1020, opened 2019-10-30
Retrofits B3Propagation to consider request kind

### LINKED ISSUE #1019: Enhance B3Propagation to implement both formats on inject
Right now, users choose `B3Propagation` or `B3SinglePropagation` for all trace instrumentation. Because messaging can have problems with multi-format, JMS ignores this and only writes b3 single, and Kafka has an option to override for the same reason. This is a proposal to make  `B3Propagation` smart enough to default to the correct basic format for the kind of span.

## What this changes
`B3Propagation` will write b3 multi for client spans and b3 single for producer/consumer spans by default. `B3SinglePropagation` will be deprecated.

## What this improves
This removes special casing for messaging, which allows propagation add-ons such as secondary sampling to work there. It also allows someone to write both formats on client spans, something some proxies do when they don't know their downstream.

### Other formats
There are other trace formats including `grpc-trace-bin`, `x-cloud-trace-context`, `x-amzn-trace-id`, and the currently developing `traceparent` formats. Solving for B3 doesn't solve for them, and some similar concerns will exist there.

Solving generically will result in a larger effort that may take more time than people have to offer now, and it isn't aligned with user demand. Right now, if we don't solve for B3, messaging just won't work with secondary sampling at all. Solving now for B3 doesn't stop us from making a generic solution later, ideally one that separates the primary trace format from secondary data such as extra fields or secondary sampling.

### Runtime data
Looking solely at the kind of the span won't solve where there are framework specifics. For example, we have special-case code in grpc for their private format `grpc-trace-bin`. This solution can't read the type of request, and even if we had a type of request it would be unlikely that in the brave core abstraction, we would define gRPC.

adriancole:
This allows us to undo messaging hacks that ignore the propagation
component in order to use B3 single propagation. It does this by top
leveling a `brave.Request` type and looking at this in `B3Propagation`.

This change obviates the `B3SinglePropagation` class.

Fixes #1019

#1011 is blocked on this


brave/src/main/java/brave/Request.java
```diff
@@ -0,0 +1,59 @@
...
+  public abstract Span.Kind kind();
+
+  /**
+   * Returns the underlying request object or {@code this} if there is none. Here are some request
+   * objects: {@code org.apache.http.HttpRequest}, {@code org.apache.dubbo.rpc.Invocation}, {@code
+   * org.apache.kafka.clients.consumer.ConsumerRecord}.
+   *
+   * <p>Note: Some implementations are composed of multiple types, such as a request and a socket
+   * address of the client. Moreover, an implementation may change the type returned due to
+   * refactoring. Unless you control the implementation, cast carefully (ex using {@code instance
```
jorgheymans: `instanceof`

adriancole:
I think I want to refactor this a little. I don't like the code in B3, though the request type makes sense.

brave/src/main/java/brave/Request.java
adriancole: thx

brave/src/main/java/brave/propagation/B3Propagation.java
```diff
@@ -13,31 +13,74 @@
...
+    return new FactoryBuilder();
+  }
+
+  /**
+   * Defaults to {@link Format#MULTI} for client/server spans and {@link Format#SINGLE_NO_PARENT}
+   * for messaging. Non-request spans default to {@link Format#MULTI}.
+   */
+  public static final class FactoryBuilder {
+    Format injectFormat = Format.MULTI;
+    final Map<Span.Kind, Format> kindToInjectFormat = new LinkedHashMap<>();
```
anuraaga: `EnumMap`

adriancole: thx will do now

adriancole:
ok I changed the implementation a bit as the BOTH thing bothered me. The current impl bothers less, but might bother someone else..

brave/src/main/java/brave/propagation/B3Propagation.java
```diff
@@ -94,6 +153,29 @@
     }
 
     @Override public void inject(TraceContext traceContext, C carrier) {
+      Format[] formats = propagation.injectFormats;
+      if (carrier instanceof Request) {
+        Span.Kind kind = ((Request) carrier).kind();
+        formats = propagation.kindToInjectFormats.get(kind);
+      }
+
+      for (Format format : formats) {
```
adriancole: ps aware of EnumSet, but using array doesn't allocate anything

brave/src/main/java/brave/propagation/B3Propagation.java
```diff
@@ -13,31 +13,86 @@
...
     }
-  };
+
+    /**
+     * Like {@link #injectFormat}, but writes multiple formats.
+     *
+     * For example, you can set {@link Span.Kind#CLIENT} spans to inject both {@link Format#MULTI}
+     * and {@link Format#SINGLE}, for transition use cases.
+     */
+    public FactoryBuilder injectFormats(Span.Kind kind, Format format1, Format format2) {
```
anuraaga: Should this check that both `format1` and `format2` are different and not `SINGLE` and `SINGLE_NO_PARENT`?

brave/src/main/java/brave/propagation/B3Propagation.java
```diff
@@ -13,31 +13,86 @@
...
+    public FactoryBuilder injectFormat(Span.Kind kind, Format format) {
+      if (kind == null) throw new NullPointerException("kind == null");
+      if (format == null) throw new NullPointerException("format == null");
+      kindToInjectFormats.put(kind, new Format[] {format});
+      return this;
     }
-  };
+
+    /**
+     * Like {@link #injectFormat}, but writes multiple formats.
```
anuraaga: `two formats`

adriancole:
running benchmarks to see how much overhead this flexibility adds

brave/src/main/java/brave/propagation/B3Propagation.java
```diff
@@ -73,6 +130,8 @@
     this.fields = Collections.unmodifiableList(
```
anuraaga: Does this need to be updated?

brave/src/main/java/brave/propagation/B3SinglePropagation.java
```diff
@@ -16,8 +16,12 @@
...
 import java.util.List;
 
-/** Implements the propagation format described in {@link B3SingleFormat}. */
-public final class B3SinglePropagation<K> implements Propagation<K> {
+/**
+ * Implements the propagation format described in {@link B3SingleFormat}.
+ *
+ * @deprecated Since 5.9, use {@link B3Propagation#newFactoryBuilder()} to control inject formats.
+ */
+@Deprecated public final class B3SinglePropagation<K> implements Propagation<K> {
```
anuraaga: Could remove the code here and delegate to the new stuff. The unit tests for this class still passing would give good confidence :)

adriancole:
inject benchmarks are the same before and after this for the default. I verified the allocation profile is identical before and after.

brave/src/main/java/brave/propagation/B3Propagation.java
adriancole: sgtm

brave/src/main/java/brave/propagation/B3SinglePropagation.java
adriancole: good idea

adriancole:
added this about next version defaulting kafka to b3 single https://github.com/openzipkin/brave/issues/1021

brave/src/main/java/brave/propagation/B3Propagation.java
adriancole: yes. good eye.. though this change invalidates this slightly as almost always you will have all fields possible (if any injector has multi..)

adriancole: adds a lot of code to be precise here, heh

brave/src/main/java/brave/propagation/B3SinglePropagation.java
adriancole: the impl was package private so I just ripped out everything except the public ctor

adriancole: PS worthwhile trade of code!

adriancole:
Excellent feedback, @anuraaga! I like it much better now.

adriancole:
this is a change to make non-request default to b3 single as they end up used in things like spring messaging even though  "local". The header count explodes so.. https://github.com/openzipkin/brave/pull/1204

PR #1023, opened 2019-10-30
Switches non-request based header default to B3 single format

adriancole:
This might be contentious. This switches the default when not using http
messaging or rpc abstractions to b3 single. In doing so, this implicitly
lowers overhead of things like message processors (which are neither
producers nor consumers). The overhead is lower because one header would
be used instead of many (for each stage in Kafka Streams, for example).

On the other hand, not everything uses our http, messaging or rpc
abstractions. What would happen is if someone manually created an http
instrumentation, they would silently switch to B3 single by default.
This risk should be considered.


adriancole:
I think this is too risky right now. We should revisit once the http and messaging abstractions are out longer.

Issue #1022, opened 2019-10-30
Allow setting remoteServiceName with TracingKafkaClientSupplier


msmsimondean:
I'm trying to set the remoteServiceName when creating an instance of [TracingKafkaClientSupplier](https://github.com/openzipkin/brave/blob/master/instrumentation/kafka-streams/src/main/java/brave/kafka/streams/TracingKafkaClientSupplier.java).  At the moment, there is no way to do this.  

**Feature:**
Add an override to [KafkaStreamsTracing.kafkaClientSupplier](https://github.com/openzipkin/brave/blob/master/instrumentation/kafka-streams/src/main/java/brave/kafka/streams/KafkaStreamsTracing.java#L67) that takes [KafkaTracing](https://github.com/openzipkin/brave/blob/master/instrumentation/kafka-clients/src/main/java/brave/kafka/clients/KafkaTracing.java) as an argument.

**Rational**
Enables people to use a customised KakfaTracing obect when creating a KafkaClientSupplier.  

This will likely be of use in one or both of the following Spring Cloud Sleuth issues: 
*  https://github.com/spring-cloud/spring-cloud-sleuth/issues/1480
* https://github.com/spring-cloud/spring-cloud-sleuth/issues/1481

adriancole:
I think line 55  is the better option. the field saved should be
KafkaTracing not Tracing. so add a second static factory method iotw.

msmsimondean:
@adriancole sounds good

adriancole:
wanna raise the PR?

PR #1024, opened 2019-10-30
Enable custom KafkaTracing to be passed to KafkaStreamsTracing.kafkaClientSupplier

simondean:
This is a fix for https://github.com/openzipkin/brave/issues/1022

### DIFF
--- a/instrumentation/kafka-streams/src/main/java/brave/kafka/streams/KafkaStreamsTracing.java
+++ b/instrumentation/kafka-streams/src/main/java/brave/kafka/streams/KafkaStreamsTracing.java
@@ -42,18 +42,26 @@ import org.apache.kafka.streams.processor.ProcessorSupplier;
 /** Use this class to decorate Kafka Stream Topologies and enable Tracing. */
 public final class KafkaStreamsTracing {
 
-  final Tracing tracing;
+  final KafkaTracing kafkaTracing;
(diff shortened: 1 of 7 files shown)


instrumentation/kafka-streams/src/main/java/brave/kafka/streams/KafkaStreamsTracing.java
```diff
@@ -42,18 +42,26 @@
...
   public static KafkaStreamsTracing create(Tracing tracing) {
-    return new KafkaStreamsTracing.Builder(tracing).build();
+    return new KafkaStreamsTracing.Builder(KafkaTracing.create(tracing)).build();
+  }
+
+  public static KafkaStreamsTracing create(KafkaTracing kafkaTracing) {
+    return new KafkaStreamsTracing.Builder(kafkaTracing).build();
+  }
+
+  public KafkaTracing kafkaTracing() {
```
adriancole: this doesn't need to be exposed, but I'll scrub it out later. We should document when things are exposed. The only reason we expose in KafkaTracing is for this component. If we expose something public, there should be a use case for it.

adriancole: edf2b1b1952233bfad5363fd2d93be8b835a567d

PR #1026, opened 2019-11-03
Makes all instrumentation call Span.finish in scope

adriancole:
See https://github.com/openzipkin/brave/pull/999#discussion_r341813329


instrumentation/http-tests/src/main/java/brave/test/http/ITHttp.java
```diff
@@ -154,23 +148,7 @@ protected Tracer tracer() {
 
   protected Tracing.Builder tracingBuilder(Sampler sampler) {
     return Tracing.newBuilder()
-      .spanReporter(s -> {
```
adriancole: NOTE: one sad side-effect is this removes our ability to do span leak detection this way.. someone clever can figure another one.

adriancole: actually this only helps with third party impls anyway as it proved nothing when using our middleware HttpHandler

PR #1030, opened 2019-11-09
catch Throwable instead of RuntimeException | Error

### LINKED ISSUE #1029: Tracing stream components catch RuntimeException | Error
Lombok's `@SneakyThrows` defeats this because it allows you to throw checked exceptions. Maybe `Throwable` would be better here ?

jorgheymans:
fixes #1029


instrumentation/kafka-streams/src/main/java/brave/kafka/streams/TracingTransformer.java
```diff
@@ -53,7 +53,7 @@ public R transform(K k, V v) {
     Throwable error = null;
     try {
       return delegateTransformer.transform(k, v);
-    } catch (RuntimeException | Error e) {
+    } catch (Throwable e) {
```
anuraaga: Should call propogateIfFatal when catching Throwable.

instrumentation/kafka-streams/pom.xml
```diff
@@ -64,6 +64,14 @@
       <scope>test</scope>
     </dependency>
 
+    <dependency>
+      <!-- used to test checked exceptions from transformers -->
```
anuraaga: Can consider defining it yourself instead of adding lombok which is surprising to see in a project not using its codege , just a few lines.

https://github.com/line/armeria/blob/master/core/src/main/java/com/linecorp/armeria/common/util/Exceptions.java#L197

jorgheymans:
@anuraaga thanks for the feedback !

instrumentation/kafka-streams/src/test/java/brave/kafka/streams/ITKafkaStreamsTracing.java
```diff
@@ -726,6 +728,119 @@ public void close() {
...
+    assertThat(spanInput.traceId()).isEqualTo(spanProcessor.traceId());
+    assertThat(spanProcessor.tags()).containsEntry("error", "file-not-found");
+
+    assertThat(!streams.state().isRunning());
+
+    streams.close();
+    streams.cleanUp();
+  }
+
+  // Armeria Black Magic copied from
```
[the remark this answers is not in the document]
anuraaga: Both same magic but with less surface :)

instrumentation/kafka-streams/src/main/java/brave/kafka/streams/TracingProcessor.java
```diff
@@ -55,6 +56,7 @@ public void process(K k, V v) {
       delegateProcessor.process(k, v);
     } catch (RuntimeException | Error e) {
       error = e;
+      Call.propagateIfFatal(e);
```
anuraaga: Nit might be better to use static import - seeing `Call` out of nowhere could trigger some confusion. Better would be if we moved the method to a separate utility at some point though :)

jorgheymans:
@jeqo @anuraaga thanks for looking at this
- `catch (Throwable e)` now applied everywhere
- zipkin2.Call is now a static import
- removed lombok dark magic comment

PR #1033, opened 2019-11-13
add flag to not share trace on consumption

### LINKED ISSUE #1032: Kafka Consumer/Streams flag to do not group traces when no incoming context
**Feature:**
Add a flag on Kafka instrumentation to affect current Kafka Consumer behavior, and KafkaStreams, where we tell instrumentation to do not create a parent trace by partition when no incoming context is available. This will allow consumers to decide when to create a trace per record consumed. 

**Rational**
Kafka Consumers, different to other messaging libraries, consume batches instead of individual messages. This batches could be small (10s of records) or could be very large depending on the load. Given than consumers, as Kafka Streams applications, will potentially write records to new topics, leading to even larger traces, a batch trace could potentially very big (>1MB).

**Example Scenario**
There are scenarios where records produced on a Kafka topic are created by an application that is not instrumented (e.g. Kafka Connectors without interceptors. integration platforms, etc). Then these records are been processed by an instrumented pipeline composed by Kafka Streams applications. Then it will be preferable to have a trace by record instead of a trace by batch.

**Prior Art**
* Currently, to implement this, you'll add an extra field to propagation, forcing KafkaConsumer to do not batch. Even though this is supported, it would be easier for users to define this based on a flag.

jeqo:
fix #1032 

**Context:**

Kafka Consumers obtain records from a Kafka topics by polling from an offset. By design, poll operation return an array of records, to be processed usually in a loop.

This `poll` operation is traced and create an initial span on the consumer application. Right before creating this span, incoming record is checked to see if it has a trace context or not. If it does, `poll` span will be created to continue a record's trace. If it doesn't, it will create a `poll` span to be shared with other records that doesn't have a trace context in its headers.

 This PR is aimed to allow users to opt-out this mechanism of creating a shared span for records without a trace context. Instead, a `poll` span will be created by record.

I'd like to propose the following defaults:

- Kafka Consumer: `shareTraceOnConsumption=true` to don't break current behavior on consumers; as plain Kafka Consumers are usually used to sink data somewhere.
- Kafka Streams: `shareTraceOnConsumption=false` as Kafka Streams usually propagate context to other topics, leading to even larger traces that if batched, parent trace will be too big.

### DIFF
--- a/instrumentation/kafka-clients/README.md
+++ b/instrumentation/kafka-clients/README.md
@@ -8,7 +8,6 @@ Add decorators for Kafka producer and consumer to enable tracing.
 First, setup the generic Kafka component like this:
 ```java
 kafkaTracing = KafkaTracing.newBuilder(messagingTracing)
-                           .writeB3SingleFormat(true) // for more efficient propagation
                            .remoteServiceName("my-broker")
                            .build();
 ```
--- a/instrumentation/kafka-clients/README.md
+++ b/instrumentation/kafka-clients/README.md
@@ -10,7 +10,6 @@ First, setup the generic Kafka component like this:
 kafkaTracing = KafkaTracing.newBuilder(messagingTracing)
                            .writeB3SingleFormat(true) // for more efficient propagation
                            .remoteServiceName("my-broker")
-                           .singleRootSpanOnReceiveBatch(true) // to share a single root span when starting a trace from consumer
                            .build();
 ```
 
@@ -103,6 +102,14 @@ to trace manually or you can do similar via automatic instrumentation like Aspec
   }
 }
 ```
+
+## Single Root Span on Consumer
+
+When a Tracing Kafka Consumer is processing records that do not have trace-context (i.e. Producer is not tracing)
+it will reuse the same root span `poll` to group all processing of records returned.
+
+If this is not the desired behavior, users can customize it by setting `singleRootSpanOnReceiveBatch` to `false`. 
+This will create a root span `poll` for each record received. 
 
 ## Notes
 * This tracer is only compatible with Kafka versions including headers support ( > 0.11.0).
--- a/instrumentation/kafka-clients/README.md
+++ b/instrumentation/kafka-clients/README.md
@@ -10,7 +10,7 @@ First, setup the generic Kafka component like this:
 kafkaTracing = KafkaTracing.newBuilder(messagingTracing)
                            .writeB3SingleFormat(true) // for more efficient propagation
                            .remoteServiceName("my-broker")
-                           .shareTraceOnConsumption(true) // to create a trace with all records consumed in a batch
+                           .singleRootSpanOnReceiveBatch(true) // to share a single root span when starting a trace from consumer
                            .build();
 ```
 
--- a/instrumentation/kafka-clients/README.md
+++ b/instrumentation/kafka-clients/README.md
@@ -108,8 +108,31 @@ to trace manually or you can do similar via automatic instrumentation like Aspec
 When a Tracing Kafka Consumer is processing records that do not have trace-context (i.e. Producer is not tracing)
 it will reuse the same root span `poll` to group all processing of records returned.
 
+```
+trace 1:
+poll
+|- processing1
+|- processing2
+...
++- processing N
+```
+
 If this is not the desired behavior, users can customize it by setting `singleRootSpanOnReceiveBatch` to `false`. 
 This will create a root span `poll` for each record received. 
+
+```
+trace 1:
+poll
++- processing1
+
+trace 2:
+poll
++- processing2
+...
+trace N:
+poll
++- processing N
+```
 
 ## Notes
 * This tracer is only compatible with Kafka versions including headers support ( > 0.11.0).
--- a/instrumentation/kafka-clients/README.md
+++ b/instrumentation/kafka-clients/README.md
@@ -10,6 +10,7 @@ First, setup the generic Kafka component like this:
 kafkaTracing = KafkaTracing.newBuilder(messagingTracing)
                            .writeB3SingleFormat(true) // for more efficient propagation
                            .remoteServiceName("my-broker")
+                           .shareTraceOnConsumption(true) // to create a trace with all records consumed in a batch
                            .build();
 ```
 
--- a/instrumentation/kafka-clients/src/main/java/brave/kafka/clients/KafkaTracing.java
+++ b/instrumentation/kafka-clients/src/main/java/brave/kafka/clients/KafkaTracing.java
@@ -85,10 +85,11 @@ public final class KafkaTracing {
     }
 
     /**
-     * Controls the sharing of a poll span for incoming spans with no trace context.
+     * Controls the sharing of a {@code poll} span for incoming spans with no trace context.
      *
-     * <b/>If true, all the spans received in a poll batch that do not have trace-context will be added
-     * to a single new poll root span. Otherwise, a poll span will be created for each such message.
+     * <b/>If true, all the spans received in a {@code poll} batch that do not have trace-context
+     * will be added to a single new poll root span. Otherwise, a {@code poll} span will be created
+     * for each such message.
      *
      * @since 5.10
      */
--- a/instrumentation/kafka-clients/src/main/java/brave/kafka/clients/KafkaTracing.java
+++ b/instrumentation/kafka-clients/src/main/java/brave/kafka/clients/KafkaTracing.java
@@ -13,6 +13,16 @@
  */
 package brave.kafka.clients;
 
+import java.util.Iterator;
+import java.util.LinkedHashSet;
+import java.util.Set;
+
+import org.apache.kafka.clients.consumer.Consumer;
+import org.apache.kafka.clients.consumer.ConsumerRecord;
+import org.apache.kafka.clients.producer.Producer;
+import org.apache.kafka.common.header.Header;
+import org.apache.kafka.common.header.Headers;
+
 import brave.Span;
 import brave.SpanCustomizer;
 import brave.Tracer;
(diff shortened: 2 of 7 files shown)


jeqo:
cc @jorgheymans

instrumentation/kafka-clients/src/main/java/brave/kafka/clients/KafkaTracing.java
```diff
@@ -72,6 +84,21 @@ public Builder remoteServiceName(String remoteServiceName) {
       return this;
     }
 
+    /**
+     * When no trace context or extra fields have been added to an incoming consumed message propagation,
+     * records from the same partition are been traced under the same parent id, sharing the same trace.
+     *
+     * <b/>If true, this will share a trace for messages received as part of
+     * a batch with no trace context, and create an individual trace by message; leading to 1 trace
```
anuraaga: I'm not sure what you mean by `and create an individual trace by message`.

instrumentation/kafka-clients/src/main/java/brave/kafka/clients/KafkaTracing.java
```diff
@@ -72,6 +84,21 @@ public Builder remoteServiceName(String remoteServiceName) {
...
+     * per poll operation.
+     *
+     * @since 5.10
+     */
+    public Builder shareTraceOnConsumption(boolean shareTraceOnConsumption) {
```
anuraaga: I think it'd be nice to indicate that this applies to new traces in the name itself. `shareNewTraceOnConsumption`? 

Though for me it's easier to grok with the opposite - `splitNewTraceOnConsumption`. What do you think?

instrumentation/kafka-clients/src/main/java/brave/kafka/clients/KafkaTracing.java
```diff
@@ -72,6 +84,21 @@ public Builder remoteServiceName(String remoteServiceName) {
       return this;
     }
 
+    /**
+     * When no trace context or extra fields have been added to an incoming consumed message propagation,
```
anuraaga: If an incoming message contains no trace context or extra fields, we create a new trace context. Normally, we add all the spans for a topic received in a batch into a single new trace. This flag can be used to instead force each message in the batch into its own trace.

instrumentation/kafka-clients/src/main/java/brave/kafka/clients/KafkaTracing.java
anuraaga: Also I know the class is called consumer but I think receive might be easier to understand.

jeqo: Thanks! I'm not convinced about naming either hehe. I'll rephrase my explanation to have a better background for naming.

jeqo:
@anuraaga, I've rephrased API comments, and added context on the PR. Let me know if its clearer now to focus on flag name.

instrumentation/kafka-clients/src/main/java/brave/kafka/clients/KafkaTracing.java
jorgheymans: tend to agree, since `shareTraceOnConsumption` is the current behaviour it might stand out more if the flag refers to enabling the opposite behaviour . `splitTraceOnConsumption` , `spanPerMessage`, `noSpanPerBatch`, 

Above names are still slightly misleading as they could make ppl (like me) think that it allows you to break a span, whereas the flag only applies when no parent trace is around. `spanPerMessageWhenNoParentTraceFound` ?

jeqo: would `shareRootSpanOnReceive` or `rootSpanOnReceiveBatch` be a better name? (I think `span...WhenNoParentTrace` ~= `rootSpan`)

jeqo: btw @jorgheymans I think we should have a similar flag to do not continue a trace on the consumer side (e.g. instead that removing headers on the producer), that would be related with this one. for instance: if producer is instrumented and injecting trace context, on the consumer side we can decide to do not continue the same trace context, and just create a new root span. Then, this flag can help to opt-out of sharing that root span between consumed messages in a batch.

Will tackle this on an additional PR as it will impact other messaging instrumentation.

jorgheymans: @jeqo so maybe this should be bundled together as one feature then, where the `separate-span-if-no-existing-root` and `force-start-new-trace` are just different strategies of this feature ? This 'something' would allow one to relatively easily define new processing like 'if the message header contains value X we want to start a new trace, otherwise join existing'. Or is that a bridge too far ?

jeqo: I'd say that (1) starting a new trace on consume is independent from (2) reusing/sharing rootSpan when consuming a batch. Where (1) is shared across messaging instrumentations, probably should be added at that level of abstraction; and (2) is specific to Kafka so far as we don't have other consumer instrumentation polling batches, but could eventually be moved to `messaging` as well. So, I'd take these as 2 diff features.

About `if (header=x) then start new trace`, I haven't thought about this as initial use-case. My first answer will be to do this once you have consumed a message, as part of your code. But we can discuss this further here https://github.com/openzipkin/brave/issues/1039

instrumentation/kafka-clients/src/main/java/brave/kafka/clients/KafkaTracing.java
```diff
@@ -72,6 +84,22 @@ public Builder remoteServiceName(String remoteServiceName) {
       return this;
     }
 
+    /**
+     * Opt-out of sharing poll span when no trace-context or extra propagation field is available on
+     * incoming messages.
+     *
+     * <b/>If true, a poll root span will be created for all messages without trace-context, and all
```
anuraaga: `If true, all the spans received in a poll batch that do not have trace-context will be added to a single new poll root span. Otherwise, a poll span will be created for each such message.`

instrumentation/kafka-clients/src/main/java/brave/kafka/clients/KafkaTracing.java
```diff
@@ -72,6 +84,22 @@ public Builder remoteServiceName(String remoteServiceName) {
       return this;
     }
 
+    /**
+     * Opt-out of sharing poll span when no trace-context or extra propagation field is available on
```
anuraaga: Since the flag defaults to true, this isn't really an opt-out. Can just say `Controls the sharing of a poll span for incoming spans with no trace context`.

instrumentation/kafka-clients/src/main/java/brave/kafka/clients/KafkaTracing.java
```diff
@@ -72,6 +84,22 @@ public Builder remoteServiceName(String remoteServiceName) {
...
+     * incoming messages.
+     *
+     * <b/>If true, a poll root span will be created for all messages without trace-context, and all
+     * following processing spans will be under the same trace.
+     *
+     * <b/>If false, a poll span will be created by message received.
+     *
+     * @since 5.10
+     */
+    public Builder rootSpanOnReceiveBatch(boolean rootSpanOnReceiveBatch) {
```
anuraaga: `singleRootSpanOnReceiveBatch`?

jeqo: like it!

instrumentation/kafka-clients/src/main/java/brave/kafka/clients/KafkaTracing.java
```diff
@@ -25,14 +35,6 @@
 import brave.propagation.TraceContext.Injector;
 import brave.propagation.TraceContextOrSamplingFlags;
 import brave.sampler.SamplerFunction;
-import java.util.Iterator;
-import java.util.LinkedHashSet;
-import java.util.Set;
-import org.apache.kafka.clients.consumer.Consumer;
```
anuraaga: Can you revert this import order change before merge?

instrumentation/kafka-clients/README.md
```diff
@@ -10,6 +10,7 @@ First, setup the generic Kafka component like this:
 kafkaTracing = KafkaTracing.newBuilder(messagingTracing)
                            .writeB3SingleFormat(true) // for more efficient propagation
                            .remoteServiceName("my-broker")
+                           .singleRootSpanOnReceiveBatch(true) // to share a single root span when starting a trace from consumer
```
jorgheymans: Since `singleRootSpanOnReceiveBatch` is the default you may omit it there IMO, it's not needed to setup kafkatracing. We may explain the feature better in a dedicated paragraph of README.md to make sure people understand why this feature was added, its impact and when it comes into play. The rationale you wrote in #1039 is a good start.

(not part of this PR but consider then removing the line `writeB3SingleFormat(true)` as well since it's the default and deprecated anyway)

jeqo:
@anuraaga @jorgheymans , I've applied latest changes based on your feedback. Let me know if it's ready to merge :)

instrumentation/kafka-clients/README.md
```diff
@@ -103,6 +102,37 @@ to trace manually or you can do similar via automatic instrumentation like Aspec
...
+...
++- processing N
+```
+
+If this is not the desired behavior, users can customize it by setting `singleRootSpanOnReceiveBatch` to `false`. 
+This will create a root span `poll` for each record received. 
+
+```
+trace 1:
+poll
```
jorgheymans: should this not be `poll1` , `poll2` etc ?

anuraaga: IMO it's fine as is, the span name is always `poll` (at least I think so ;) ).

instrumentation/kafka-clients/src/main/java/brave/kafka/clients/KafkaTracing.java
```diff
@@ -72,6 +84,19 @@ public Builder remoteServiceName(String remoteServiceName) {
       return this;
     }
 
+    /**
+     * Controls the sharing of a poll span for incoming spans with no trace context.
+     *
+     * <b/>If true, all the spans received in a poll batch that do not have trace-context will be added
```
anuraaga: Might be good to use `{@code poll}` to highlight it as a keyword, not an English word.

instrumentation/kafka-clients/README.md
jeqo: yeap it is, that's why :)

jeqo:
thanks for your feedback! I'll go ahead and merge this one :+1:

PR #1036, opened 2019-11-19
Fix path parsing of HttpClient to return the whole path, not only las…

### LINKED ISSUE #1035: Why is everything before the last / removed when getting the path and how to include the full path.
I just updated my project from using `brave-instrumentation-http` 5.6.5 -> 5.8.0. We noticed that the zipkin span names for our client requests removed the elements of the url before the last slash and we were wondering what the purpose was behind this and if there was a way we could include the entire url 

current behavior
 "/v1/digest" -> "/digest"

desired behavior
 "/v1/digest" -> "v1/digest"

Thank you very much for your help and for maintaining openzipkin!

anuraaga:
…t part.

Parsers are supposed to return the whole path, not just last part. I copied the implementation from https://github.com/openzipkin/brave/blob/master/instrumentation/httpasyncclient/src/main/java/brave/httpasyncclient/TracingHttpAsyncClientBuilder.java#L269

Fixes #1035

### DIFF
--- a/instrumentation/httpclient/src/main/java/brave/httpclient/TracingProtocolExec.java
+++ b/instrumentation/httpclient/src/main/java/brave/httpclient/TracingProtocolExec.java
@@ -81,6 +81,9 @@ final class TracingProtocolExec implements ClientExecChain {
     }
 
     @Override public String path() {
+      if (delegate instanceof HttpRequestWrapper) {
+        return ((HttpRequestWrapper) delegate).getURI().getPath();
+      }
       String result = delegate.getRequestLine().getUri();
       int queryIndex = result.indexOf('?');
       return queryIndex == -1 ? result : result.substring(0, queryIndex);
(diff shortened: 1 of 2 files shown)


anuraaga:
Found another test that failed which may have resulted in the logic :) I think special-casing `HttpRequestWrapper` is fine, similar to the other method, but I'm not too sure what this class is since I've never really used apache client ;)

instrumentation/httpclient/src/main/java/brave/httpclient/TracingProtocolExec.java
```diff
@@ -81,10 +81,12 @@
     }
 
     @Override public String path() {
+      if (delegate instanceof HttpRequestWrapper) {
```
devinsba: Instead of special casing how we get this value, could we instead unwrap using `getOriginal()` on the wrapper class then follow the flow we already have. I see the benefit here given that URI already will contain the parsed path. Just thinking out loud about maintaining parity

anuraaga: Thanks for the idea, I like it, but unfortunately it seems that `getOriginal().getRequestLine()` still contains the host, so we'd need some special casing or more sophisticated URL parsing logic anyways. Best probably to take advantage of the parsing that already happened.

PR #1040, opened 2019-11-20
Flag to start a new trace on Messaging Consumers

### LINKED ISSUE #1039: Messaging consumers flag to force starting a new trace instead of child span
**Feature:**
Adding a messaging consumer flag to opt-out from continuing a trace, and create a new one.

**Rational**
Messaging pipelines, where a message producer is fan-out to (potentially) multiple consumers, we should allow users to tell which of these consumers should form part of the same trace, and which should start a new trace (probably linking these via tags).

**Example Scenario**
We have `producer1->topic1` and consumers `topic1->consumer1`, `topic1->consumer2` and `topic1->consumer3`; where consumer 1 and 2 are part of the same "pipeline" as producer 1, but `consumer3->topic2` where other consumers subscribe to.

In this scenario, if we have a messaging consumer flag, we could say on consumer 3:

```
MessagingTracing.create(...).forceNewTraceOnReceive(true)
```

Leading to 2 traces for each message:

trace 1: `producer1:send ... consumer1:receive .. consumer2:receive ...`

trace 2: `consumer3:receive consumer3:send ... consumerN:receive ...`

Where trace 2 `consumer:receive` span can be tagged with a "parentTraceId" linking to trace 1.

jeqo:
fix #1039

### DIFF
--- a/instrumentation/messaging/src/main/java/brave/messaging/MessagingTracing.java
+++ b/instrumentation/messaging/src/main/java/brave/messaging/MessagingTracing.java
@@ -78,6 +78,16 @@ public class MessagingTracing implements Closeable {
     return consumerSampler;
   }
 
+  /**
+   * Returns flag to create a new trace when consuming a message with an existing trace context, or
+   * to continue.
+   *
+   * @since 5.10 (potentially)
+   */
+  public boolean newTraceOnReceive() {
+    return newTraceOnReceive;
+  }
+
   public Builder toBuilder() {
     return new Builder(this);
   }
@@ -86,10 +96,13 @@ public class MessagingTracing implements Closeable {
   final SamplerFunction<MessagingRequest> producerSampler;
   final SamplerFunction<MessagingRequest> consumerSampler;
 
+  final boolean newTraceOnReceive;
+
   MessagingTracing(Builder builder) {
     this.tracing = builder.tracing;
     this.producerSampler = builder.producerSampler;
     this.consumerSampler = builder.consumerSampler;
+    this.newTraceOnReceive = builder.newTraceOnReceive;
     // assign current IFF there's no instance already current
     CURRENT.compareAndSet(null, this);
   }
@@ -98,6 +111,7 @@ public class MessagingTracing implements Closeable {
     Tracing tracing;
     SamplerFunction<MessagingRequest> producerSampler;
     SamplerFunction<MessagingRequest> consumerSampler;
+    boolean newTraceOnReceive = false;
 
     Builder(Tracing tracing) {
       if (tracing == null) throw new NullPointerException("tracing == null");
@@ -110,6 +124,7 @@ public class MessagingTracing implements Closeable {
       this.tracing = source.tracing;
       this.producerSampler = source.producerSampler;
       this.consumerSampler = source.consumerSampler;
+      this.newTraceOnReceive = source.newTraceOnReceive;
     }
 
     /** @see MessagingTracing#tracing() */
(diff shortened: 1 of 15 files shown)


jcchavezs:
My 2c on this:
1. Span tags are snake case so I'd rather use parent.trace_id instead of parent.traceId.
2. Is parent.trace_id good naming? It feels like it is but would like to get some input from people like @basvanbeek or @shakuzen.
3. Is it worth adding the parent.span_id also? Otherwise how can you know what was the operation triggering this?

jeqo:
@jcchavezs thanks! I agree with you. I'll fix casing, and add span_id. Looking forward to more feedback to check if naming is right.

adriancole:
I feel like this could be done in propagation. for example, in secondary sampling we satisfy the same use case (for the second trace).. does this jog any more flexible ideas? https://github.com/openzipkin-contrib/zipkin-secondary-sampling

instrumentation/messaging/src/main/java/brave/messaging/MessagingTracing.java
```diff
@@ -126,13 +141,23 @@ public Builder producerSampler(SamplerFunction<MessagingRequest> producerSampler
...
     public Builder consumerSampler(SamplerFunction<MessagingRequest> consumerSampler) {
       if (consumerSampler == null) throw new NullPointerException("consumerSampler == null");
       this.consumerSampler = consumerSampler;
       return this;
     }
 
+    /**
+     * @see MessagingTracing#newTraceOnReceive()
+     */
+    public Builder newTraceOnReceive(boolean newTraceOnReceive) {
```
adriancole, line 156: I think this might need to be an expression, similar to a sampling function?

jeqo: Yes! sampler function is a very similar interface. We can add a similar design here.

Would make sense to discuss something as https://cloud.google.com/trace/docs/reference/v2/rest/v2/Links supporting this links as a Zipkin/Brave type? Or tags would be enough (e.g. `parent.trace.id=... parent.span.id=...`)?

adriancole: I think the links thing is being inherited into otel. I haven't followed that part, but in opencensus it was hardly defined. Ex not sure it is for this case or something else. Would appreciate if you could summarize later on otel, but for now we can chat about stackdriver.

At any rate, we have some choices in Brave as we don't have to map 1-1 with zipkin model. For example, `MutableSpan.error()` In other words, let's get the api right first then consider normalizing a model.

I suspect if the semantics match what links are supposed to be used for, we can consider updating zipkin-gcp accordingly ( I think at some point that code should implement a SpanHandler anyway). Ex, the direction is important. I think in the stackdriver they discuss a batchhandler who is in an existing trace, and notices inbound trace IDs from messages. This is not the same as what I believe we are discussing here. Here we only have one inbound and one outbound trace, We are choosing to sever it in order to limit the tree (this use case needs to be fleshed out actually). Maybe you can add a RATIONALE file with all the use cases you've seen?

@saturnism do you know if the intent of links are for the case where you intentionally break a trace from an inbound message in order for "processor down" to not cause the upstream trace to be massive?

janainamendonca:
@jeqo is this going ahead? I have a use case where I'd like to apply this

codefromthecrypt:
this is out-of-date and over 2.5 years old. This doesn't mean it isn't valid, but please re-open if necessary today.

PR #1042, opened 2019-11-21
fix re-setting properties on BytesMessage

me3x:
Sleuth JMS instrumentation cleared all properties on ActiveMQBytesMessage and was unable to put them back. 

fix MessageNotWriteableException - Message body is read-only while re-setting properties on BytesMessage

please squash commits with merge

### DIFF
--- a/instrumentation/jms/src/main/java/brave/jms/PropertyFilter.java
+++ b/instrumentation/jms/src/main/java/brave/jms/PropertyFilter.java
@@ -52,7 +52,7 @@ final class PropertyFilter {
       return;
     }
     
-    if (disjoint(namesToClear, names)) {
+    if (disjoint(namesToClear, names) && message instanceof BytesMessage) {
       return;
     }
 
@@ -81,10 +81,10 @@ final class PropertyFilter {
     }
 
     // workaround for ActiveMQBytesMessage
-    if ( message instanceof BytesMessage ) {
-      resetBytesMessageProperties( message, out );
+    if (message instanceof BytesMessage) {
+      resetBytesMessageProperties(message, out);
     } else {
-      resetProperties( message, out );
+      resetProperties(message, out);
     }
   }
 
--- a/instrumentation/jms/src/main/java/brave/jms/PropertyFilter.java
+++ b/instrumentation/jms/src/main/java/brave/jms/PropertyFilter.java
@@ -14,7 +14,6 @@
 package brave.jms;
 
 import java.util.ArrayList;
-import java.util.Enumeration;
 import java.util.List;
 import java.util.Set;
 import javax.jms.BytesMessage;
@@ -23,6 +22,8 @@ import javax.jms.Message;
 
 import static brave.internal.Throwables.propagateIfFatal;
 import static brave.jms.JmsTracing.log;
+import static java.util.Collections.disjoint;
+import static java.util.Collections.list;
 
 // Similar to https://github.com/apache/camel/blob/b9a3117f19dd19abd2ea8b789c42c3e86fe4c488/components/camel-jms/src/main/java/org/apache/camel/component/jms/JmsMessageHelper.java
 final class PropertyFilter {
@@ -42,17 +43,20 @@ final class PropertyFilter {
   }
 
   static void filterProperties(Message message, Set<String> namesToClear, List<Object> out) {
-    Enumeration<?> names;
+    List<String> names;
     try {
-      names = message.getPropertyNames();
+      names = list(message.getPropertyNames());
     } catch (Throwable t) {
       propagateIfFatal(t);
       log(t, "error getting property names from {0}", message, null);
       return;
     }
+    
+    if (disjoint(namesToClear, names)) {
+      return;
+    }
 
-    while (names.hasMoreElements()) {
-      String name = (String) names.nextElement();
+    names.forEach( name -> {
       Object value;
       try {
         value = message.getObjectProperty(name);
@@ -65,7 +69,7 @@ final class PropertyFilter {
         out.add(name);
         out.add(value);
       }
-    }
+    });
 
     // redo the properties to keep
     try {
@@ -89,18 +93,18 @@ final class PropertyFilter {
       BytesMessage bytesMessage = (BytesMessage) message;
       byte[] body = new byte[(int) bytesMessage.getBodyLength()];
       bytesMessage.reset();
-      bytesMessage.readBytes( body );
+      bytesMessage.readBytes(body);
       message.clearBody();
-      resetProperties( message, out );
-      bytesMessage.writeBytes( body );
+      resetProperties(message, out);
+      bytesMessage.writeBytes(body);
       bytesMessage.reset();
-    } catch ( JMSException e ) {
+    } catch (JMSException e) {
       propagateIfFatal(e);
-      log( e, "unable to reset BytesMessage body {0}", message, e );
+      log(e, "unable to reset BytesMessage body {0}", message, e);
     }
   }
 
-  private static void resetProperties( Message message, List<Object> out ) {
+  private static void resetProperties(Message message, List<Object> out) {
     for (int i = 0, length = out.size(); i < length; i += 2) {
       String name = out.get(i).toString();
       try {
--- a/instrumentation/jms/src/main/java/brave/jms/PropertyFilter.java
+++ b/instrumentation/jms/src/main/java/brave/jms/PropertyFilter.java
@@ -13,88 +13,112 @@
  */
 package brave.jms;
 
+import static brave.jms.JmsTracing.log;
+import javax.jms.BytesMessage;
+import javax.jms.JMSException;
+import javax.jms.Message;
 import java.util.ArrayList;
 import java.util.Enumeration;
 import java.util.List;
 import java.util.Set;
-import javax.jms.Message;
-
-import static brave.internal.Throwables.propagateIfFatal;
-import static brave.jms.JmsTracing.log;
+import brave.internal.Platform;
 
 // Similar to https://github.com/apache/camel/blob/b9a3117f19dd19abd2ea8b789c42c3e86fe4c488/components/camel-jms/src/main/java/org/apache/camel/component/jms/JmsMessageHelper.java
 final class PropertyFilter {
-  /**
-   * This implies copying properties because the JMS spec says you can't write properties upon
-   * receipt until {@link Message#clearProperties()} has been called.
-   *
-   * <p> See https://docs.oracle.com/javaee/6/api/javax/jms/Message.html
-   */
-  static void filterProperties(Message message, Set<String> namesToClear) {
-    ArrayList<Object> retainedProperties = messagePropertiesBuffer();
-    try {
-      filterProperties(message, namesToClear, retainedProperties);
-    } finally {
-      retainedProperties.clear(); // ensure no object references are held due to any exception
-    }
-  }
 
-  static void filterProperties(Message message, Set<String> namesToClear, List<Object> out) {
-    Enumeration<?> names;
-    try {
-      names = message.getPropertyNames();
-    } catch (Throwable t) {
-      propagateIfFatal(t);
-      log(t, "error getting property names from {0}", message, null);
-      return;
+    /**
+     * This implies copying properties because the JMS spec says you can't write properties upon
+     * receipt until {@link Message#clearProperties()} has been called.
+     * <p> See https://docs.oracle.com/javaee/6/api/javax/jms/Message.html
+     */
+    static void filterProperties( Message message, Set<String> namesToClear ) {
+        ArrayList<Object> retainedProperties = messagePropertiesBuffer();
+        try {
+            filterProperties( message, namesToClear, retainedProperties );
+        } finally {
+            retainedProperties.clear(); // ensure no object references are held due to any exception
+        }
     }
 
-    while (names.hasMoreElements()) {
-      String name = (String) names.nextElement();
-      Object value;
-      try {
-        value = message.getObjectProperty(name);
-      } catch (Throwable t) {
-        propagateIfFatal(t);
-        log(t, "error getting property {0} from message {1}", name, message);
-        return;
-      }
-      if (!namesToClear.contains(name) && value != null) {
-        out.add(name);
-        out.add(value);
-      }
+    static void filterProperties( Message message, Set<String> namesToClear, List<Object> out ) {
+        Enumeration<?> names;
+        try {
+            names = message.getPropertyNames();
+        } catch ( JMSException e ) {
+            Platform.get().log( "error getting property names from {0}", message, e );
+            return;
+        }
+
+        while ( names.hasMoreElements() ) {
+            String name = (String) names.nextElement();
+            Object value;
+            try {
+                value = message.getObjectProperty( name );
+            } catch ( JMSException e ) {
+                log( e, "error getting property {0} from message {1}", name, message );
+                return;
+            }
+            if ( !namesToClear.contains( name ) && value != null ) {
+                out.add( name );
+                out.add( value );
+            }
+        }
+
+        // redo the properties to keep
+        try {
+            message.clearProperties();
+        } catch ( JMSException e ) {
+            Platform.get().log( "error clearing properties of {0}", message, e );
+            return;
+        }
+
+        // workaround for ActiveMQBytesMessage
+        if ( message instanceof BytesMessage ) {
+            resetBytesMessageProperties( message, out );
+        } else {
+            resetProperties( message, out );
+        }
+
     }
 
-    // redo the properties to keep
-    try {
-      message.clearProperties();
-    } catch (Throwable t) {
-      propagateIfFatal(t);
-      log(t, "error clearing properties of {0}", message, null);
-      return;
+    private static void resetBytesMessageProperties( Message message, List<Object> out ) {
+        try {
+            BytesMessage bytesMessage = (BytesMessage) message;
+            byte[] body = new byte[(int) bytesMessage.getBodyLength()];
+            bytesMessage.reset();
+            bytesMessage.readBytes( body );
+            message.clearBody();
+            resetProperties( message, out );
+            bytesMessage.writeBytes( body );
+            bytesMessage.reset();
+        } catch ( JMSException e ) {
+            log( e, "unable to reset BytesMessage body {0}", message, e );
+        }
     }
 
-    for (int i = 0, length = out.size(); i < length; i += 2) {
-      String name = out.get(i).toString();
-      try {
-        message.setObjectProperty(name, out.get(i + 1));
-      } catch (Throwable t) {
-        propagateIfFatal(t);
-        log(t, "error setting property {0} on message {1}", name, message);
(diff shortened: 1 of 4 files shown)


instrumentation/jms/src/main/java/brave/jms/PropertyFilter.java
```diff
@@ -40,17 +43,20 @@ static void filterProperties(Message message, Set<String> namesToClear) {
...
     try {
-      names = message.getPropertyNames();
+      names = list(message.getPropertyNames());
     } catch (Throwable t) {
       propagateIfFatal(t);
       log(t, "error getting property names from {0}", message, null);
       return;
     }
+    
+    if (disjoint(namesToClear, names) && message instanceof BytesMessage) {
```
jcchavezs: This is maybe a naive comment but could we a dd a comment on the reasoning for `message instanceof BytesMessage` ?

instrumentation/jms/src/main/java/brave/jms/PropertyFilter.java
```diff
@@ -74,6 +80,31 @@ static void filterProperties(Message message, Set<String> namesToClear, List<Obj
       return;
     }
 
+    // workaround for ActiveMQBytesMessage
+    if (message instanceof BytesMessage) {
```
jcchavezs: I see this check being done here too. Worth to put into a private method that selfdescribes the why?

instrumentation/jms/src/main/java/brave/jms/PropertyFilter.java
me3x: This line is to avoid unnecessary clearing properties when there's nothing to clear. I would like to remove the condition. Could you please help me fix https://github.com/openzipkin/brave/blob/master/instrumentation/jms/src/test/java/brave/jms/ITJms_1_1_TracingMessageConsumer.java 

It fails on line 300 receive_customSampler.

instrumentation/jms/src/main/java/brave/jms/PropertyFilter.java
me3x: Please help me to get rid of the other usage. I believe instanceof is simple enough thought.

jcchavezs: Yeah, it is. so maybe a comment explaining the why is enough.

instrumentation/jms/src/main/java/brave/jms/PropertyFilter.java
jcchavezs: I don't think I have the right knowledge to help you on this matter. @jeqo or @anuraaga might help. My comment is more for keeping it readable even for someone like me who has very little context.

instrumentation/jms/src/main/java/brave/jms/PropertyFilter.java
devinsba: Lets add that comment to this method with why there is a special case for `BytesMessage`

me3x: I added a comment to resetBytesMessageProperties

jcchavezs: I don't wanna be a pain but I'd really prefer a comment in the `instanceof` check. It is way easier to read rather than having go to to the function to understand the check.

instrumentation/jms/src/main/java/brave/jms/PropertyFilter.java
jeqo: @me3x if you check `disjoint` implementation is actually doing the same as we are doing on the loop later:

```
for (Object e : iterate) {
            if (contains.contains(e)) {
               // Found a common element. Collections are not disjoint.
                return false;
            }
        }
```

I'd like to propose to refactor this logic. Here is one way: we can remove this check and go directly to the `for` loop. Loop logic can be restructured to first do `if (!namesToClear.contains(name))` and if true, then extract value and check if not null, add to `out`.

instrumentation/jms/src/main/java/brave/jms/PropertyFilter.java
```diff
@@ -74,6 +80,32 @@ static void filterProperties(Message message, Set<String> namesToClear, List<Obj
...
```
jeqo:
```suggestion
    if (message instanceof BytesMessage) { // BytesMessage requires clearing the body before reset properties otherwise it fails with MessageNotWriteableException
```

instrumentation/jms/src/main/java/brave/jms/PropertyFilter.java
me3x: I refactored as you proposed. Please check.

instrumentation/jms/src/main/java/brave/jms/PropertyFilter.java
me3x: fixed

me3x:
I only tested with ActiveMQ. Please check.

instrumentation/jms/src/main/java/brave/jms/PropertyFilter.java
```diff
@@ -40,31 +42,39 @@ static void filterProperties(Message message, Set<String> namesToClear) {
...
-        log(t, "error getting property {0} from message {1}", name, message);
-        return;
-      }
-      if (!namesToClear.contains(name) && value != null) {
-        out.add(name);
-        out.add(value);
+    
+    boolean disjoint = true;
+    for (String name: names) {
+      if (!namesToClear.contains(name)) {
```
jcchavezs: small nit: use early termination for easier readability: 
```java
for (String name: names) {
      if (namesToClear.contains(name)) {
          disjoint = false;
          continue;
      }
// all the rest

**Comment by me3x** [comment]
@jeqo could you please help me to fix a test? I believe it shouldn't be failing. It probably relies on clearing properties even if there's no need to clear.

 ITJmsTracingMessageConsumer>ITJms_1_1_TracingMessageConsumer.receive_customSampler:300

File: instrumentation/jms/src/main/java/brave/jms/PropertyFilter.java (hunk 4 above)
Comment (me3x) [reply instrumentation/jms/src/main/java/brave/jms/PropertyFilter.java]: I usually use it in methods, but I wanted to avoid "continue".

Comment (me3x) [reply instrumentation/jms/src/main/java/brave/jms/PropertyFilter.java]: refactored as you proposed

**Comment by jeqo** [comment]
Hey @me3x thanks for the fixes!

I've test it further and actually that test case is preventing to break functionality here. Let me explain:
- That case is a non-traced producer sending a message, and a traced consumer with a sampler based on queue name.
- traced consumer should be able to extract context from producer (null in this case) and inject context for further processing.
- as in the current implementation we are returning when disjoint before reset properties, properties are kept as read-only, failing to inject context later. Check by debugging a breaking point on `MessagePropagation:45`  

Based on this I recommend to remove `disjoint` flag as we need to reset properties anyway, simplifying loop logic.

**Comment by jeqo** [comment]
btw, I've test it with Oracle WebLogic JMS and works before (doesn't have a similar bug as ActiveMQ) and works after. So, I think we are good to merge on that side.

**Comment by jeqo** [comment]
@me3x , could you re-enable this IT that was validating this issue: `ITJms_1_1_TracingMessageConsumer:253`

If that one goes green we should be good to merge.

**Comment by me3x** [comment]
I removed disjoint check. It's a little bit weird design to rely on properties being writable for this scenario, but let's merge without it.

**Comment by me3x** [comment]
@jeqo Should we have similar test with BytesMessage. It failed only without instance of condition. In case of ActiveMQBytesMessage, setObjectProperty will not work after I call bytesMessage.reset();

File: instrumentation/jms/src/main/java/brave/jms/PropertyFilter.java [hunk 5]
```diff
@@ -40,17 +42,19 @@ static void filterProperties(Message message, Set<String> namesToClear) {
...
       log(t, "error getting property names from {0}", message, null);
       return;
     }
-
-    while (names.hasMoreElements()) {
-      String name = (String) names.nextElement();
+    
+    for (String name: names) {
+      if (namesToClear.contains(name)) {
+        continue;
```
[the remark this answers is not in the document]
me3x: fixed

instrumentation/jms/src/main/java/brave/jms/PropertyFilter.java
```diff
@@ -74,6 +78,35 @@ static void filterProperties(Message message, Set<String> namesToClear, List<Obj
...
+    // workaround for ActiveMQBytesMessage
+    if (message instanceof BytesMessage) { // BytesMessage requires clearing the body before reset properties otherwise it fails with MessageNotWriteableException
+      resetBytesMessageProperties(message, out);
+    } else {
+      resetProperties(message, out);
+    }
+  }
+
+  private static void resetBytesMessageProperties( Message message, List<Object> out ) {
+    if (out.isEmpty()) {
```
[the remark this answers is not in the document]
me3x: fixed

me3x:
@jeqo do you have any idea why it started to fail?  [ERROR]   ITJmsTracingMessageConsumer>ITJms_1_1_TracingMessageConsumer.receive_resumesTrace_bytes:255->ITJms_1_1_TracingMessageConsumer.receive_resumesTrace:275 expected:<"463ac35c9f6413ad-8fdeb02e21ce5744-1"> but was:<null>

jeqo:
@me3x take a look to the last `bytesMessage.reset()` on `PropertyFilter#resetBytesMessageProperties`, it is closing the message again; triggering failure later when injecting context at the end:

```
  void handleReceive(Message message) {
    if (message == null || tracing.isNoop()) return;
    MessageConsumerRequest request = new MessageConsumerRequest(message, destination(message));

    TraceContextOrSamplingFlags extracted =
      jmsTracing.extractAndClearProperties(extractor, request, message);
    Span span = jmsTracing.nextMessagingSpan(sampler, request, extracted);

    if (!span.isNoop()) {
      span.name("receive").kind(Span.Kind.CONSUMER);
      Destination destination = destination(message);
      if (destination != null) jmsTracing.tagQueueOrTopic(request, span);
      if (remoteServiceName != null) span.remoteServiceName(remoteServiceName);

      // incur timestamp overhead only once
      long timestamp = tracing.clock(span.context()).currentTimeMicroseconds();
      span.start(timestamp).finish(timestamp);
    }
    injector.inject(span.context(), request);
  }
```

I think we can remove this `reset` and let JMS to close the message before sending it.

me3x:
I tried and it didn't work without reset. It failed on JmsListener ... unable to read, because body was in write mode.

instrumentation/jms/src/main/java/brave/jms/TracingConsumer.java
```diff
@@ -62,6 +63,9 @@ void handleReceive(Message message) {
       span.start(timestamp).finish(timestamp);
     }
     injector.inject(span.context(), request);
+    if (message instanceof BytesMessage) {
+      (BytesMessage message).reset();
```
jeqo: @me3x sorry if I wasn't clear enough. This part is not compiling, and actually not needed. I believe jms will take care of the message itself; no need to reset ourselves. I've give a try removing this and verifying ITs locally and all's green :)

me3x: Instrumentation should left the message in the same state it started. I tried to read body in JmsListener and it failed.

jeqo:
Could you add a test to show this? As you can see we have this issues based
on jms design choices and bug with activemq bytes message, and we are
trying to work around it. Maybe we are not considering the scenario of
readOnly msgs enough, which can lead to a deeper discussion (potentially
other Issue/PR). But we are currently supporting jms clients with this
implementation and we are fixing specific issue with activemq messages, so
I don’t think closing message for these should be added.

me3x:
I don't have a test, but here is a stack trace: 
```
Caused by: javax.jms.MessageNotReadableException: Message body is write-only
	at org.apache.activemq.command.ActiveMQBytesMessage.checkWriteOnlyBody(ActiveMQBytesMessage.java:867) ~[activemq-client-5.15.10.jar:5.15.10]
	at org.apache.activemq.command.ActiveMQBytesMessage.initializeReading(ActiveMQBytesMessage.java:872) ~[activemq-client-5.15.10.jar:5.15.10]
	at org.apache.activemq.command.ActiveMQBytesMessage.getBodyLength(ActiveMQBytesMessage.java:203) ~[activemq-client-5.15.10.jar:5.15.10]
	at com.cs.ifs.oxygen.common.JmsUtils.getBytesMessagePayloadAsString(JmsUtils.java:27) ~[classes/:na]
	... proxies and transactions ...

	at com.cs.ifs.oxygen.am.CoreRefDataListener.onMessage(CoreRefDataListener.java:60) ~[classes/:na]
	at sun.reflect.NativeMethodAccessorImpl.invoke0(Native Method) ~[na:1.8.0_171-1-redhat]
	at sun.reflect.NativeMethodAccessorImpl.invoke(NativeMethodAccessorImpl.java:62) ~[na:1.8.0_171-1-redhat]
	at sun.reflect.DelegatingMethodAccessorImpl.invoke(DelegatingMethodAccessorImpl.java:43) ~[na:1.8.0_171-1-redhat]
	at java.lang.reflect.Method.invoke(Method.java:498) ~[na:1.8.0_171-1-redhat]
```

jeqo:
@me3x I won't be able to help much if we don't have a test-case to check this scenario, I'm sorry. 

If we want to cover this case where [this change](https://github.com/openzipkin/brave/pull/1042/files#diff-fcac8ed18c129e381a445860fea0b7a5R69-R75) is needed, we need test cases first to prove that works and doesn't affect other scenarios.

If we need to review how we are handling JMS messages and if we need to reset them or not at the end, I'd prefer to take this on another PR and only solve issue with ActiveMQ Bytes Message here first.

me3x:
from javax.jms.BytesMessage:
  * <P>If {@code clearBody} is called on a message in read-only mode, 
  * the message body is cleared and the message is in write-only mode.
  *
  * <P>If a client attempts to read a message in write-only mode, a 
  * {@code MessageNotReadableException} is thrown.

me3x:
I removed body reset. I'll create another PR

jeqo:
@me3x I've second thoughts about my previous comments. I realize now that we are only "touching" message content on this PR and not in the existing implementation where we deal with properties only, then we have to solve the `read-only` issue here. Sorry for the confusion.

To have temporal `bytesMessage` handling on the same place, we can move changes upper to `TracingConsumer` instead of `PropertyFilter`. Here is a draft with some comments:

```java
  void handleReceive(Message message) {
    if (message == null || tracing.isNoop()) return;
    MessageConsumerRequest request = new MessageConsumerRequest(message, destination(message));

    byte[] body = null;
    if (message instanceof BytesMessage) {
      try {
        BytesMessage bytesMessage = (BytesMessage) message;
        body = new byte[(int) bytesMessage.getBodyLength()];
        //bytesMessage.reset(); // do we need this reset() before reading?
        bytesMessage.readBytes(body);
        bytesMessage.clearBody(); // any exception after this can lead to loose message
      } catch (Throwable t) {
        propagateIfFatal(t);
        //log(t, "error getting property {0} from message {1}", name, message);
        return;
      }
    }

    TraceContextOrSamplingFlags extracted =
      jmsTracing.extractAndClearProperties(extractor, request, message);
    // ....current code
    injector.inject(span.context(), request);

    if (message instanceof BytesMessage) {
      try {
        BytesMessage bytesMessage = (BytesMessage) message;
        bytesMessage.writeBytes(body);
        bytesMessage.reset();
      } catch (Throwable t) {
        propagateIfFatal(t);
        //log(t, "error getting property {0} from message {1}", name, message);
        //return;
      }
    }
  }

```

This will also mean we can rollback some of the changes on PropertyFilter back: mainly property names enumeration was enough (we are not using disjoint anymore).

Lastly, I'm worried about the impact of this changes as we are carrying the technical debt of an ActiveMQ bug, as well as touching message content. I'm asking other contributors (@openzipkin/brave) to give feedback here.

There are different options from my point of view: (1) add this workaround and maintain it until activemq client is fixed. (2) move this workaround to a new module (e.g. `activemq`) which will only impact/help ActiveMQ users (3) have this workaround on another repo (e.g. inside `openzipkin-contrib`) until bug is fixed.

me3x:
We can write body back immediately. We don't have to wait. Clearing it is the only way to put the message to write mode, but it will remain writable until reset.

jeqo:
@me3x right, so we can do the write back right after the `clearBody` isn't it? This is much better. What do you think on moving things to `TracingConsumer`?

We can test if possible reduce the filter even further by adding `(message instanceof BytesMessage && message.getName().equals("org.apache.activemq.command.ActiveMQBytesMessage"))`, so we can go for option 1 and keep the fix as part of the current module. I'll test this today.

me3x:
@jeqo It's a good idea. I found that it's used also in class JmsTracing and TracingMessageListener, so there are 3 places. Do we have ActiveMQBytesMessage class available? If not, maybe it could be part of some recovery after exception?

me3x:
@jeqo I mean current behavior is the worst possible, because we loose user's properties completely.

me3x:
we could rethrow MessageNotWriteableException and handle it with clearing the body

instrumentation/jms/src/main/java/brave/jms/PropertyFilter.java
```diff
@@ -40,28 +42,29 @@ static void filterProperties(Message message, Set<String> namesToClear) {
...
     try {
-      names = message.getPropertyNames();
+      names = list(message.getPropertyNames());
     } catch (Throwable t) {
       propagateIfFatal(t);
       log(t, "error getting property names from {0}", message, null);
       return;
     }
-
-    while (names.hasMoreElements()) {
```
adriancole: why is this logic changed? was there a problem? Using collections.list will add allocate an array, and changing this logic makes it hard to tell what change is actually a problem or in focus.

instrumentation/jms/src/main/java/brave/jms/PropertyFilter.java
```diff
@@ -74,6 +77,31 @@ static void filterProperties(Message message, Set<String> namesToClear, List<Obj
...
 
+    // workaround for ActiveMQBytesMessage
+    if (message instanceof BytesMessage) { // BytesMessage requires clearing the body before reset properties otherwise it fails with MessageNotWriteableException
+      resetBytesMessageProperties(message, out);
+    } else {
+      resetProperties(message, out);
+    }
+  }
+
+  private static void resetBytesMessageProperties( Message message, List<Object> out ) {
```
adriancole: this will cause a lot of overhead as we are subject to OOM. as this is a bug in activemq, we need to be more specific than this.

adriancole:
@jeqo if I recall correctly, there's a fix to this upstream now?

jeqo:
@adriancole still waiting for the next ActiveMQ client release to be available. 

I will create a PR with the changes and notes about target versions on README, and close this one.

adriancole:
we should close this now with ActiveMQ release fixing this internally correct? meanwhile it is nearing a year old :D

codefromthecrypt:
closing as out-of-date and seems possible addressed upstream. If not the case, please re-open

PR #1050, opened 2019-12-13
Remove scope on httpasyncclient interceptor fail

andylintner:
Previous to this change, if a user supplied interceptor failed after
executing the HandlerSend interceptor, the scope created there was
never removed. The scope is now removed in the error handler if the
RemoveScope interceptor was not executed.

### DIFF
--- a/instrumentation/httpasyncclient/src/main/java/brave/httpasyncclient/TracingHttpAsyncClientBuilder.java
+++ b/instrumentation/httpasyncclient/src/main/java/brave/httpasyncclient/TracingHttpAsyncClientBuilder.java
@@ -91,16 +91,16 @@ public final class TracingHttpAsyncClientBuilder extends HttpAsyncClientBuilder 
 
   static final class RemoveScope implements HttpRequestInterceptor {
     @Override public void process(HttpRequest request, HttpContext context) {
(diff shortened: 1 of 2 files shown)


instrumentation/httpasyncclient/src/main/java/brave/httpasyncclient/TracingHttpAsyncClientBuilder.java
```diff
@@ -91,6 +91,10 @@ public static HttpAsyncClientBuilder create(HttpTracing httpTracing) {
 
   static final class RemoveScope implements HttpRequestInterceptor {
     @Override public void process(HttpRequest request, HttpContext context) {
+      process(context);
+    }
+
+    static void process(HttpContext context) {
```
adriancole: small change.. pull up as static method removeScope. then its use inline doesn't need qualification and is a bit more readable.

andylintner: Makes sense - I've updated the PR

adriancole:
thanks, makes total sense. we appreciate the help!

PR #1051, opened 2019-12-15
TracingAdvice should come first when decorating listener by SpringRabbitTracing

kubamarchwicki:
When decorating an existing `SimpleRabbitListenerContainerFactory` in `spring-rabbit`, the tracingAdvice should come first, in front of existing advice, not last. 

**Rationale**
`adviceChain` is - for example - used to manage retrials via `RetryOperationInterceptors`. When `tracingAdvice` comes first, it sets MDC headers, so that consecutive interceptors log with property set trace and span ids. 

**Example scenario**
`StatelessRetryOperationsInterceptorFactoryBean` warns on dropping message due to absent recoverer (https://github.com/spring-projects/spring-amqp/blob/master/spring-rabbit/src/main/java/org/springframework/amqp/rabbit/config/StatelessRetryOperationsInterceptorFactoryBean.java#L71). With the current way listener container is decorated, tracing advice will be the last one and the log message will lack the tracing information. 
With the chain reversed, the MDC context is set up as expected.

### DIFF
--- a/instrumentation/spring-rabbit/src/test/java/brave/spring/rabbit/SpringRabbitTracingTest.java
+++ b/instrumentation/spring-rabbit/src/test/java/brave/spring/rabbit/SpringRabbitTracingTest.java
@@ -15,14 +15,6 @@ package brave.spring.rabbit;
 
 import brave.Tracing;
 import brave.propagation.ThreadLocalCurrentTraceContext;
-
-import java.util.Arrays;
-import java.util.Collection;
-import java.util.List;
-
-import org.aopalliance.aop.Advice;
-import org.assertj.core.api.Condition;
-import org.assertj.core.data.Index;
 import org.junit.After;
 import org.junit.Test;
 import org.springframework.amqp.rabbit.config.SimpleRabbitListenerContainerFactory;
@@ -31,7 +23,10 @@ import org.springframework.amqp.support.postprocessor.UnzipPostProcessor;
 import org.springframework.cache.interceptor.CacheInterceptor;
 import zipkin2.reporter.Reporter;
 
-import static java.util.Arrays.*;
+import java.util.Collection;
+import java.util.List;
+
+import static java.util.Arrays.asList;
 import static org.assertj.core.api.Assertions.assertThat;
 
 public class SpringRabbitTracingTest {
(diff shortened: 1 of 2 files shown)


adriancole:
sounds smart. maybe this intended behavior can be proven with a unit test? cc @jonathan-lo

kubamarchwicki:
Totally. Test ammended - I missed that in the first place. Sorry

instrumentation/spring-rabbit/src/test/java/brave/spring/rabbit/SpringRabbitTracingTest.java
```diff
@@ -82,11 +83,12 @@
...
 
-  @Test public void decorateSimpleRabbitListenerContainerFactory_appends_when_absent() {
+  @Test public void decorateSimpleRabbitListenerContainerFactory_appends_as_first_when_absent() {
     SimpleRabbitListenerContainerFactory factory = new SimpleRabbitListenerContainerFactory();
     factory.setAdviceChain(new CacheInterceptor());
 
     assertThat(rabbitTracing.decorateSimpleRabbitListenerContainerFactory(factory).getAdviceChain())
-      .anyMatch(advice -> advice instanceof TracingRabbitListenerAdvice);
+      .hasSize(2)
+      .matches(adviceArray -> asList(adviceArray).get(0) instanceof TracingRabbitListenerAdvice);
```
adriancole: would this work? (also removing the import)

```suggestion
      .matches(adviceArray -> adviceArray[0] instanceof TracingRabbitListenerAdvice);
```

instrumentation/spring-rabbit/src/test/java/brave/spring/rabbit/SpringRabbitTracingTest.java
```diff
@@ -82,11 +83,12 @@
       .hasSize(1);
   }
 
-  @Test public void decorateSimpleRabbitListenerContainerFactory_appends_when_absent() {
+  @Test public void decorateSimpleRabbitListenerContainerFactory_appends_as_first_when_absent() {
```
adriancole: maybe add a comment saying this is needed so that downstream interceptors can see the trace context

kubamarchwicki:
Thank you for your assistance!

PR #1055, opened 2020-01-15
Propagate current span to http client callback thread

simontoens:


instrumentation/httpasyncclient/src/main/java/brave/httpasyncclient/TracingHttpAsyncClientBuilder.java
```diff
@@ -131,7 +134,7 @@ static void parseTargetAddress(HttpHost target, Span span) {
         new TracingAsyncRequestProducer(requestProducer, context),
         new TracingAsyncResponseConsumer<>(responseConsumer, context),
         context,
-        callback
+        new FutureCallbackWithCurrentSpan(tracer.currentSpan(), callback)
```
adriancole: the tracecontext here in currentSpan should be the same as what's above set in context.setAttribute right? as there's some cost to fetching it, I think we should remember or reuse the context value.

instrumentation/httpasyncclient/src/main/java/brave/httpasyncclient/TracingHttpAsyncClientBuilder.java
```diff
@@ -148,6 +151,37 @@ static void parseTargetAddress(HttpHost target, Span span) {
     }
   }
 
+  final class FutureCallbackWithCurrentSpan<T> implements FutureCallback<T> {
```
adriancole: I'd reduce the cost of this by making it a static final, passing the CurrentTraceContext, TraceContext, and the callback. CurrentTraceContext scope commands are a bit cheaper than tracer based ones.

simontoens:
@adriancole thanks for the review.  I have updated the code to use the current trace ctx instead of the current span.  Please let me know what you want me to do regarding the test.

anuraaga:
Just for reference, here's a long discussion on the issue of automatic propagation for Armeria

https://github.com/line/armeria/issues/2347

In the end, we decided not to do it, since it's very easy to set up callbacks in a way that loses the context (just as simple as `CompletableFuture.allOf`, so instead of something that will easily break, it seems better for users to understand context and propagate it at a higher level.

So instead of propagation we're adding examples of good patterns for framework-based propagation - https://github.com/line/armeria/pull/2363

Not exactly sure if the same applies here but figured it's good food for thought.

adriancole:
@anuraaga I think that you can sort this in the framework is one of the advantages of using a client/server framework which has some compatible notions of context. Given that this is about apache http client, with no server counterpart, not sure we can convey that here. It is an interesting thought about generic'ness of the feature. there's probably a RATIONALE we can extract about rules when in a relatively homogenous (and literally brave integrated) framework vs piecemeal code. wdyt?

adriancole:
case in point, I think armeria has a context decorator which will work across client and server, and that's more consistent and understandable vs doing something in apache for the client side, and something else for an incoming kafka message.

Probably in the hodge-podge area, implicitly wrapping context can still make sense. Likely we do need to mention the warnings.. For example, when we manage the context (driven by a framework or otherwise) it also drives things attached to it, like MDC sync etc. When users are in control, they would have to have explicit code or aspects or wrapped things to instrument those on their behalf, for example, decorated executors.. then there's the notion of which context makes sense to place on the callback. for example, here we are talking about the invocation context being replaced on the callback.. it is uncertain which context would be placed on with automatic executors.. probably why you are mentioning the dragons. Finally, a concrete test which breaks, even for the sake of being honest about limitations, is likely a good thing. regardless of policy like this.

instrumentation/httpasyncclient/src/test/java/brave/httpasyncclient/ITTracingHttpAsyncClientBuilder.java
```diff
@@ -92,4 +98,54 @@ protected void post(CloseableHttpAsyncClient client, String pathIncludingQuery,
     
     takeSpan();
   }
+
+  @Test public void currentSpanIsVisibleInCallbackThread() throws Exception {
```
adriancole: I would go a bit farther, because you are not only setting a current span, but specifically the invocation context. Ex if you change spanIsVisible to something that catches the current tracecontext, you can compare it against the one in line 124

anuraaga:
I think there are two issues

- What trace context should be propagated. Indeed, with the concept of server / client in e.g., Armeria it's easy to have a best practice of always propagate the server context. Maybe it'd be good for brave to have a similar concept of root context (the first one created) and a root context executor? 

- Ease of losing context with standard APIs

This code would lose context even with this PR

```java
// Traced
asyncClient.execute(request, future1);
asyncClient.execute(request, future2);

CompletableFutures.allOf(future1, future2)
  .thenCompose(() -> {
   // Not traced
  asyncClient.execute(request, future3);

});
```

So you would need to do

```java

.thenCompose(CurrentTraceContext.wrap(() -> {
   // Traced
  asyncClient.execute(request, future3);

}));
```

Given how easy it is to lose context with futures, I personally don't know if it helps much to automatically propagate. And it might be better as an option, since if the user does wrap context themselves (usually because they know which one they want), it's wasteful to propagate automatically.

I guess it's a choice of showing an example of how to propagate yourself, or showing an example of when automatic propagation won't work, and how to propagate yourself to work around. Former is a bit simpler? :-) Just my 2 yen

adriancole:
I agree with your points except that I would treat the future composition
thing separately. for example instrumenting executors is relatively normal
and still allows a DI approach vs literally embedding tracing code.

where callbacks are instrumented (with little overhead actually as there is
that maybe thing.. remember this isnt rx) I think this also solves the DI
vs locking into tracing code in your logic.

wdyt?

adriancole:
another thing about this PR is that this isnt JDK8 futures either. this is
literally a callback model even if some may try to fit this into JDK 8
model. iotw for the era it could make more sense to wrap in context
callbacks here as we do in all the other similar callback libraries
https://hc.apache.org/httpcomponents-core-ga/httpcore/apidocs/org/apache/http/concurrent/FutureCallback.html

anuraaga:
Yeah it definitely makes more sense here since while I got a little tricked by the futures in the PR, it is just callbacks. Wrapping callbacks makes sense.

simontoens:
@adriancole do you want me to make any other changes here?

instrumentation/httpasyncclient/src/test/java/brave/httpasyncclient/ITTracingHttpAsyncClientBuilder.java
```diff
@@ -92,4 +99,52 @@ protected void post(CloseableHttpAsyncClient client, String pathIncludingQuery,
...
+    AtomicBoolean callbackHasTraceCtx = new AtomicBoolean(false);
+    AtomicBoolean callbackCompleted = new AtomicBoolean(false);
+    server.enqueue(new MockResponse());
+    closeClient(client);
+    client = TracingHttpAsyncClientBuilder.create(httpTracing).build();
+    client.start();
+    ScopedSpan span = tracer.startScopedSpan("test");
+    TraceContext expectedTraceCtx = span.context();
+
+    try {
```
adriancole: I think what's left here (besides nit on formatting) is to simplify the code and remove the wait loop. For example, if we use a countdownlatch instead of AtomicBoolean for callbackCompleted. You can search for references to CountdownLatch in brave for similar examples.

next, the assertion in 117 may actually occur off the main thread, which is why you are saving a boolean there. I think it is a bit cleaner to have an atomic reference of TraceContext, then you can do assertThat(callbackTraceContext.get()).isSameAs(expected) right after your countdown latch is done.

instrumentation/httpasyncclient/src/test/java/brave/httpasyncclient/ITTracingHttpAsyncClientBuilder.java
```diff
@@ -92,4 +99,52 @@ protected void post(CloseableHttpAsyncClient client, String pathIncludingQuery,
     
     takeSpan();
   }
+
+  @Test public void currentTraceContextIsVisibleInCallbackThread() throws Exception {
+    Tracer tracer = httpTracing.tracing().tracer();
+    AtomicBoolean callbackHasTraceCtx = new AtomicBoolean(false);
+    AtomicBoolean callbackCompleted = new AtomicBoolean(false);
+    server.enqueue(new MockResponse());
+    closeClient(client);
```
adriancole: is restarting the client needed here? what happens if you don't close it? it looks like it would be the same as what you're doing. maybe try without lines 108-110 or add a comment above why we have to recreate it.

simontoens:
@adriancole thanks I have made those changes.  

Sorry about the formatting.  Which code formatter should I use? I didn't see anything documented in HACKING.md.

adriancole:
hmm yeah I guess it isnt intuitive to have two similarly named files that
dont ref eachother. sorry about that
https://github.com/openzipkin/brave/blob/master/.github/CONTRIBUTING.md

adriancole:
The code formatting here looks good. We do need to follow-up on some more elaborate RATIONALE about when to propagate what and with what impact, as this stuff is tricky.

adriancole:
good work @simontoens! thanks for the review @anuraaga

simontoens:
Thank you @adriancole and @anuraaga!

PR #1056, opened 2020-01-16
Corrects attribution of the fragment of guava we modified

adriancole:
This corrects the LICENSE by moving the copyright citation to NOTICE instead.
This also adds LICENSE and NOTICE to each jar and sources jar.

See https://www.apache.org/dev/licensing-howto.html#mod-notice

### DIFF
--- a/LICENSE
+++ b/LICENSE
@@ -199,8 +199,3 @@
    WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
    See the License for the specific language governing permissions and
    limitations under the License.
-
-This product contains a modified part of Guava, distributed by Google:
-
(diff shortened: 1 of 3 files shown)


adriancole:
@anuraaga PTAL I added a resources section and verified these actually end up in the jars.

adriancole:
ps the travis fails aren't reproducable locally (the grpc thing..) I can try to nuke my local maven cache, but for example, the travis cache doesn't exist and it is still hosed

adriancole:
build fail unrelated https://github.com/openzipkin/brave/pull/1057

adriancole:
rebased over what I believe to be a fixed travis

PR #1060, opened 2020-01-17
Add composite propagation support.

csabakos:
This adds a `CompositePropagation` class, which allows one to use multiple propagations.

The purpose is to allow people to more easily transition between various propagations by being able to extract all of them and to emit either one of them or all of them.

Concrete example: `CompositePropagation` in conjunction with `B3Propagation` and a vendor-specific `VendorPropagation` allows one to migrate a distributed system from `VendorPropagation` to `B3Propagation` in a piece by piece fashion while maintaining interoperability.

### DIFF
--- a/brave/src/main/java/brave/propagation/CompositePropagation.java
+++ b/brave/src/main/java/brave/propagation/CompositePropagation.java
@@ -40,13 +40,13 @@ public class CompositePropagation<K> implements Propagation<K> {
     boolean injectAll = true;
     final List<Propagation.Factory> propagationFactories = new ArrayList<>();
 
-    public FactoryBuilder addPropagationFactory(final Propagation.Factory propagationFactory) {
+    public FactoryBuilder addPropagationFactory(Propagation.Factory propagationFactory) {
       if (propagationFactory == null) throw new NullPointerException("propagationFactory == null");
       propagationFactories.add(propagationFactory);
       return this;
     }
 
-    public FactoryBuilder addAllPropagationFactories(final Collection<Propagation.Factory> propagationFactories) {
+    public FactoryBuilder addAllPropagationFactories(Collection<Propagation.Factory> propagationFactories) {
       if (propagationFactories == null) throw new NullPointerException("propagationFactories == null");
       this.propagationFactories.addAll(propagationFactories);
       return this;
@@ -57,7 +57,7 @@ public class CompositePropagation<K> implements Propagation<K> {
       return this;
     }
 
-    public FactoryBuilder injectAll(final boolean injectAll) {
+    public FactoryBuilder injectAll(boolean injectAll) {
       this.injectAll = injectAll;
       return this;
     }
@@ -67,15 +67,15 @@ public class CompositePropagation<K> implements Propagation<K> {
     }
   }
 
-  private final List<Propagation<K>> propagations;
-  private final List<K> keys;
-  private final boolean injectAll;
+  final List<Propagation<K>> propagations;
+  final List<K> keys;
+  final boolean injectAll;
 
-  CompositePropagation(final List<Propagation<K>> propagations, final boolean injectAll) {
+  CompositePropagation(List<Propagation<K>> propagations, boolean injectAll) {
     this.propagations = propagations;
     this.injectAll = injectAll;
-    final Set<K> keySet = new LinkedHashSet<>();
-    for (final Propagation<K> propagation : propagations) {
+    Set<K> keySet = new LinkedHashSet<>();
+    for (Propagation<K> propagation : propagations) {
       keySet.addAll(propagation.keys());
     }
     keys = new ArrayList<>(keySet);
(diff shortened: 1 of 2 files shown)


adriancole:
understood that this could be helpful, though I think there's only 2 copy/pastes so far. sitting on this a bit and meanwhile feel free to copy/paste.

reason is that there was a propagation plugin design thing that I can't fetch up right now which would be more efficient and this late in the game adding a base class to deprecate later might not be wise.

brave/src/main/java/brave/propagation/CompositePropagation.java
```diff
@@ -0,0 +1,144 @@
...
+    };
+  }
+
+  @Override public <C> TraceContext.Extractor<C> extractor(final Getter<C, K> getter) {
+    return new TraceContext.Extractor<C>() {
+      @Override
+      public TraceContextOrSamplingFlags extract(final C carrier) {
+        for (final Propagation<K> propagation : propagations) {
+          final TraceContextOrSamplingFlags result = propagation.extractor(getter).extract(carrier);
+          if (SamplingFlags.EMPTY != result.samplingFlags()) {
```
adriancole, line 108: sorry this isn't quite right, as when extra field/baggage is in use it won't necessarily equal the constant.

adriancole:
related example https://github.com/openzipkin/brave/pull/1144

adriancole:
Something like this or otherwise that helps multiple "primary" formats should ideally be released before Brave 6.

Particularly w3c trace-context will cause more mixed environments than before

brave/src/main/java/brave/propagation/CompositePropagation.java
csabakos: I made an update that I think fixes this issue

adriancole:
to make concrete.. what I mean by B3 clone is that the format needed for the composite doesn't require putting anything in TraceContext.extra() in order to work. The way I'm thinking about to simplify this is just allow this to be unsealed (possibly adding an extract side of it)

https://github.com/openzipkin/brave/blob/master/brave/src/main/java/brave/propagation/B3Propagation.java#L38

For example, b3 single is in fact a different format than b3 multi, and we already have code that has some sophisticated features, but does compositing also. When we get to Brave 6 (soon) we'll deprecate some of the Propagation type (notably non-string types), so basically want to help solve this in a way that makes the least code, highest performance and simplest experience.

You can imagine a future where wrapping is limited mostly to things like baggage + pure trace context, this will give lots of leverage to people, and I think we finally have the tools (now that Brave 5.12 is out) to make that a reality

csabakos:
> In your PR you mention abstractly needing to combine formats, which is true, but can you also be more specific?

My use-case is switching from the LightStep `ot-tracer-(traceid|spanid|samples)` and `ot-baggage-.*` headers to B3 headers (which are also supported by LightStep).

While we are switching, my app needs to be able to both consume and emit both sets of headers to maintain compatibility with any upstream/downstream apps.

adriancole:
@csabakos how are you achieving the .* baggage now? considering the propagation component is a whitelist.. Are you using BaggagePropagation (formerly ExtraFieldPropagation) to map them (similar to what sleuth does) or are you custom handling them in your component. If you are doing custom, are you just passing the values across or are you integrating them somehow?

If you have a code link it would be helpful. thanks tons for the feedback so far.

adriancole:
related: composite was deleted out of -gcp as it wasn't actually used. in all cases so far, there's only primary/secondary https://github.com/openzipkin/zipkin-gcp/pull/174

csabakos:
> @csabakos how are you achieving the .* baggage now? considering the propagation component is a whitelist.. Are you using BaggagePropagation (formerly ExtraFieldPropagation) to map them (similar to what sleuth does) or are you custom handling them in your component. If you are doing custom, are you just passing the values across or are you integrating them somehow?
> 
> If you have a code link it would be helpful. thanks tons for the feedback so far.

@adriancole I haven't really thought about how to handle baggage because nobody at my company has requested that feature so far. This is the code that I'm using:

* [LightStepPropagation](https://github.com/csabakos/brave-lightstep/blob/master/src/main/java/brave/propagation/LightStepPropagation.java)
* Initialization:
```
CompositePropagation.newFactoryBuilder()
        .addPropagationFactory(B3Propagation.FACTORY)
        .addPropagationFactory(LightStepPropagation.FACTORY)
        .build()
```

adriancole:
@csabakos thanks for the link.. so what you've got there is basically B3 clone (no use of extra) which is what I hinted at for https://github.com/openzipkin/brave/pull/1060#issuecomment-628997291

meaning right now, B3Propagation has different inject formats, some are in-process only. I don't think you'd want to use lightstep internally. Basically, if we open up the ability to add a format like what you have, which is the same thing as stackdriver, jaeger and also x-ray.. people can choose which B3'ish formats they want for which transport.

This helps avoid washing away the features that already exist now, like remote detection etc, and calls a spade a spade. IOTW, if it is just a B3'ish format, no need to even wrap a factory, no need to go to every config tool and teach it compositing, just allow B3Propagation.FactoryBuilder to specify an additional compatible format.

adriancole:
anyway I think I'll move to "show not tell" as some of the nuance, the trouble of triple wrapping etc, performance and also feature detection wise, aren't easy to see.

codefromthecrypt:
this is out-of-date and 3.5 years old. This doesn't mean it isn't valid, but please re-open if necessary today.

PR #1062, opened 2020-01-22
Updates dependency versions, notably gRPC which has some changes

adriancole:
* gRPC 1.22 census integration stopped propagating "method", so caused test drift
* gRPC also removed the parameter to withPlainText
* Dubbo 2.7.5 removed some constants, but tests fail. I removed use of the constants.


instrumentation/dubbo/src/main/java/brave/dubbo/DubboParser.java
```diff
@@ -19,9 +19,6 @@
 import org.apache.dubbo.rpc.Invoker;
 import org.apache.dubbo.rpc.support.RpcUtils;
 
-import static org.apache.dubbo.rpc.Constants.$INVOKE;
```
adriancole: a separate change would need to make the next version of dubbo not crash tests. this just avoids referencing a constant they later deleted

instrumentation/grpc/src/test/java/brave/grpc/BaseITTracingClientInterceptor.java
```diff
@@ -0,0 +1,395 @@
...
+import static brave.rpc.RpcRequestMatchers.serviceEquals;
+import static brave.sampler.Sampler.ALWAYS_SAMPLE;
+import static brave.sampler.Sampler.NEVER_SAMPLE;
+import static org.assertj.core.api.Assertions.assertThat;
+import static org.assertj.core.api.Assertions.catchThrowableOfType;
+import static org.assertj.core.api.Assertions.entry;
+import static org.assertj.core.api.Assertions.failBecauseExceptionWasNotThrown;
+import static org.junit.Assume.assumeTrue;
+
+public abstract class BaseITTracingClientInterceptor {
```
adriancole: diff didn't notice this file was just renamed with one method added.

PR #1065, opened 2020-01-23
Handle null MessageListeners

### LINKED ISSUE #1061: JMS Support does not handle removal of MessageListener from TracingSession
If the MessageListener is removed from a JMS TracingSession by calling TracingSession#setMessageListener(null), an "empty" TracingMessageListener is set in the session instead.

E.g. JmsPoolSession does this here when closing:
https://github.com/messaginghub/pooled-jms/blob/9e56282b25ddd3c89124dce9f4d306013c2af55e/pooled-jms/src/main/java/org/messaginghub/pooled/jms/JmsPoolSession.java#L91
So the session is then returned to the pool in a rather questionable state. Particularly, after the session is obtained again, doing synchronous receive() calls an ActiveMQMessageReceiver are rejected because the MessageListener is not null.

I would expect that the MessageListener is actually set to null in the underlying session instance.

ohr:
fixes #1061

### DIFF
--- a/instrumentation/jms/src/main/java/brave/jms/TracingJMSConsumer.java
+++ b/instrumentation/jms/src/main/java/brave/jms/TracingJMSConsumer.java
@@ -40,7 +40,7 @@ import javax.jms.MessageListener;
   }
 
   @Override public void setMessageListener(MessageListener listener) throws JMSRuntimeException {
-    delegate.setMessageListener(listener != null ? TracingMessageListener.create(listener, jmsTracing) : null);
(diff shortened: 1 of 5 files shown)


adriancole:
thanks for the help. to test this might be trickier.. I might instead recommend making TracingMessageListener.create() return null on null which is easier to test in TracingMessageListenerTest wdyt?

ohr:
That's what I first thought, too, but then I didn't want to change the semantics of the create() function (that it may return null now). If you prefer this solution, I'll change the PR.

adriancole:
thanks, in this case it was a pragmatic choice. appreciated!

PR #1067, opened 2020-01-25
Ensure http client callbacks are on invocation context

adriancole:
This is a generalization of @simontoens work in #1055 towards the other
http client impls. This ensures http client callbacks are made with invocation context as the current trace context.

- [x] httpasyncclient #1055
- [x] jaxrs2
- [x] okhttp3 (users need to configure the dispatcher's executor service)
- [x] spring-web

### DIFF
--- a/instrumentation/http-tests/src/main/java/brave/test/http/ITHttpAsyncClient.java
+++ b/instrumentation/http-tests/src/main/java/brave/test/http/ITHttpAsyncClient.java
@@ -1,5 +1,5 @@
 /*
- * Copyright 2013-2019 The OpenZipkin Authors
+ * Copyright 2013-2020 The OpenZipkin Authors
  *
  * Licensed under the Apache License, Version 2.0 (the "License"); you may not use this file except
  * in compliance with the License. You may obtain a copy of the License at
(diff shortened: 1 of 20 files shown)


adriancole:
for this to work in spring web, we'd have to make a forwarding version of spring's ListenableFuture, or return a proxy that hooks into the two addCallback methods

Note: if we make a forwarding variant, we have to be careful about the recently added default method https://docs.spring.io/spring/docs/current/javadoc-api/org/springframework/util/concurrent/ListenableFuture.html#completable--

adriancole:
okhttp3 is similar. we can return a proxy or a ForwardingCall. The Call type has been stable throughout, with one small incompatibility in v3.12 (timeout) https://github.com/square/okhttp/commit/5373160d20aa3f6036ff5ebc91137888c4abffbe

If doing a forwarding type, we can solve by not adding the Override annotation on that method that doesn't exist before 3.12 and dancing on that similar to how we do in JMS for methods added later.

adriancole:
note okhttp3 is fine at the moment for those using interceptors, the gap is limited to those using the CallFactory, which happens to be the recommended way heh.

adriancole:
So, the things not needing change are jax-rs, apache (done previously), okhttp3 interceptor.
Things needing change are spring-web, okhttp3 call.factory

adriancole:
I noticed that the two okhttp impls differed in test setup. With a configured executorservice, you don't need to make any code changes.

adriancole:
will proceed the last step after https://github.com/openzipkin/brave/pull/1070

adriancole:
thinking supporting okhttp without mucking with dispatcher might be good

adriancole:
alright should be good to go

PR #1068, opened 2020-01-27
chore(license): copyright fixed

petermetz:
Template string needed expanding.

### DIFF
--- a/LICENSE
+++ b/LICENSE
@@ -186,7 +186,7 @@
       same "printed page" as the copyright notice for easier
       identification within third-party archives.
 
-   Copyright {yyyy} {name of copyright owner}
+   Copyright 2020 The OpenZipkin Authors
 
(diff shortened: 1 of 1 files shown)


adriancole:
I can see why you might do this. there are a few variants of the original ASL file. I think they are all templated. Let's opt out of the change as probably some are checksumming to see if the text matches expectations?

https://www.apache.org/licenses/LICENSE-2.0.txt

petermetz:
> I can see why you might do this. there are a few variants of the original ASL file. I think they are all templated. Let's opt out of the change as probably some are checksumming to see if the text matches expectations?
> 
> https://www.apache.org/licenses/LICENSE-2.0.txt

Sure thing, I'm fine either way, no skin in the game.
Feel free to re-open/merge if the consensus changes.

PR #1074, opened 2020-02-04
Fixes startScopedSpanWithParent null

adriancole:
javadocs lied!

### DIFF
--- a/brave/src/main/java/brave/Tracer.java
+++ b/brave/src/main/java/brave/Tracer.java
@@ -514,7 +514,7 @@ public class Tracer {
    * }</pre>
    */
   public ScopedSpan startScopedSpan(String name) {
-    return startScopedSpanWithParent(name, null);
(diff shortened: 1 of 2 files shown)


brave/src/main/java/brave/Tracer.java
```diff
@@ -566,7 +566,6 @@ public ScopedSpan startScopedSpan(String name) {
   // this api is needed to make tools such as executors which need to carry the invocation context
   public ScopedSpan startScopedSpanWithParent(String name, @Nullable TraceContext parent) {
     if (name == null) throw new NullPointerException("name == null");
-    if (parent == null) parent = currentTraceContext.get();
```
adriancole: this subverted the whole point. the javadocs suggest you can force a new span by passing null. we didn't have coverage for this, eventhough coverage tools couldn't see it.

PR #1076, opened 2020-02-06
Ensures integration tests won't hang

adriancole:
This adds a global timeout of 5s/method for HTTP integration tests.
It also carries over the 3s/poll timeout from taking a span for other
queue based operations. This is an effort to reduce the impact of
testing async libraries with bugs in their callback management.

### DIFF
--- a/brave/src/main/java/brave/propagation/B3Propagation.java
+++ b/brave/src/main/java/brave/propagation/B3Propagation.java
@@ -1,5 +1,5 @@
 /*
- * Copyright 2013-2019 The OpenZipkin Authors
+ * Copyright 2013-2020 The OpenZipkin Authors
  *
  * Licensed under the Apache License, Version 2.0 (the "License"); you may not use this file except
  * in compliance with the License. You may obtain a copy of the License at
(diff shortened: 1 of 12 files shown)


adriancole:
specifically, I'm trying to test the reactor netty client in sleuth, and it is hanging and this would allow me to know which tests don't

adriancole:
ack but that would only apply to this project. the downstream would have to
copy that in maven or gradle etc. maybe I can add a comment clarifying why
here

anuraaga:
Ah yeah that's right - a comment could be a good reminder.

Also just a random thought for future reference that is probably too much work, but junit5 does support a global timeout we could propagate downstream using a properties file. Maybe someday ;)

adriancole:
I will help in fallout if you feel up to converting the tests to junit5. we
could change it for the next minor (not next patch)

adriancole:
thx for the advice @anuraaga

adriancole:
While a timeout to pull a request or span off the queue makes sense. GlobalTimeout should be reverted as it interferes with debugging. You can imagine trying to step through a debugger in less than 5 seconds :)

anuraaga:
Ah, can wrap with `DisableOnDebug` to solve that.

https://junit.org/junit4/javadoc/4.12/org/junit/rules/DisableOnDebug.html

adriancole:
wow cool!

PR #1080, opened 2020-02-11
Preserve case of local and remote service names

csabakos:
`MutableSpan` should preserve the case of the service name string in the `localServiceName(String)` and `remoteServiceName(String)` setters.

This will allow one to report service names in a case-preserving manner to back-ends that support it. (The Zipkin back-end will continue to see lowercase service names since `zipkin2.Endpoint` enforces lowercase.)

See [this Gitter discussion](https://gitter.im/openzipkin/zipkin?at=5e3db3f855b6b04bf69b3c6d) for more reasoning.

### DIFF
--- a/brave/src/main/java/brave/Tracing.java
+++ b/brave/src/main/java/brave/Tracing.java
@@ -1,5 +1,5 @@
 /*
- * Copyright 2013-2019 The OpenZipkin Authors
+ * Copyright 2013-2020 The OpenZipkin Authors
  *
  * Licensed under the Apache License, Version 2.0 (the "License"); you may not use this file except
  * in compliance with the License. You may obtain a copy of the License at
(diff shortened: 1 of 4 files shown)


csabakos:
> looks good. you might want to add a test to tracer as well, just so it doesn't become "fixed" :P

I thought about adding a test for `Tracing` but `localServiceName` is only used in conjunction with `ZipkinFinishedSpanHandler` and the service name ends up getting lowercased in that code path due to `zipkin2.Endpoint`.

I could add a getter for `localServiceName` to `Tracing` and then use that in the test. How does that sound?

adriancole:
we dont add getters for tests. you can use assertj like so

https://github.com/openzipkin/brave/blob/master/spring-beans/src/test/java/brave/spring/beans/TracingFactoryBeanTest.java#L63

adriancole:
thanks much!

adriancole:
this is out as 5.9.5 once maven sync is complete. Thanks for the help and patience!

PR #1083, opened 2020-02-14
Allows client instrumentation to avoid thread local scoping

### LINKED ISSUE #1082: Make tracing more efficient when explicit propagation is in use
When creating a client or message producer span, we can use `Tracer.nextSpan(clientRequestSampler(), request)` which handles sampling correctly. However, this relies on the `CurrentTraceContext` component, namely the `.get()` method to look at the current span.

When instrumentation happens via callbacks, this implies having to wrap the whole thing like so:
```java
try (Scope ws = currentTraceContext().maybeScope(ref.parent)){
	Span clientSpan = tracer().nextSpan(clientRequestSampler(), request);
}
```

The scoping operation is expensive, especially in sleuth as it associates logging context. For example, we'd be doing that work only for the single line of code to make and sample the span.

To avoid this overhead, we should allow a third parameter of the parent context.

adriancole:
This allows async instrumentation to avoid any use of thread-local current trace context between callbacks.

The motivation for this was reactor-netty `WebClient` which needs to carry the subscription trace context to the request hook. Before this change, the code implied a threadlocal access (sometimes some other scope access) which was unnecessary and expensive.

It may not be obvious why this saves overhead. Mainly, folks hang things like log context synchronization on top of the scoping api. The more times a context is synced in tracing, the more times anything that hangs off it is synced. The total overhead is a sum of all those things, not just the threadlocal part.

Here's the jist of what this allows:

```java
class MyFilter extends Filter {

  public void onSchedule(Attributes attributes) {
    // Stash the invoking trace context as this will be the correct parent of
    // the request.
    attributes.put(TraceContext.class, currentTraceContext.get());
  }

  public void onStart(Request request, Attributes attributes) {
    // Retrieve the parent, if any, and start the span.
    TraceContext parent = attributes.get(TraceContext.class);
    Span span = tracer.nextSpanWithParent(samplerFunction, request, parent);
```

This tidies up okhttp and apache async client.. lowering overhead a bit.

Fixes #1082

### DIFF
--- a/brave/src/main/java/brave/Tracer.java
+++ b/brave/src/main/java/brave/Tracer.java
@@ -529,7 +529,7 @@ public class Tracer {
    */
   public <T> ScopedSpan startScopedSpan(String name, SamplerFunction<T> samplerFunction, T arg) {
     if (name == null) throw new NullPointerException("name == null");
-    return newScopedSpan(name, nextContext(samplerFunction, arg));
+    return newScopedSpan(name, nextContext(samplerFunction, arg, currentTraceContext.get()));
   }
 
   /**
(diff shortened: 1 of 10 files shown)


instrumentation/okhttp3/src/main/java/brave/okhttp3/TracingInterceptor.java
```diff
@@ -52,7 +55,14 @@ public static Interceptor create(HttpTracing httpTracing) {
   @Override public Response intercept(Chain chain) throws IOException {
     HttpClientRequest request = new HttpClientRequest(chain.request());
 
-    Span span = handler.handleSend(request);
+    Span span;
+    TraceContext parent = chain.request().tag(TraceContext.class);
+    if (parent != null) {
+      span = handler.handleSend(request, parent != NULL_SENTINEL ? parent : null);
+    } else {
+      span = handler.handleSend(request);
```
adriancole: TODO: comment that we are using interceptor only in this case

instrumentation/okhttp3/src/main/java/brave/okhttp3/TracingCallFactory.java
```diff
@@ -33,6 +29,12 @@
 // NOTE: this is not an interceptor because the current span can get lost when there's a backlog.
 // This will be completely different after https://github.com/square/okhttp/issues/270
 public final class TracingCallFactory implements Call.Factory {
+  /**
+   * To save overhead, we use a null sentinel when there's no parent. This helps avoid looking at
+   * the current trace context when there was no span in scope at invocation time.
+   */
+  static final TraceContext NULL_SENTINEL = TraceContext.newBuilder()
```
adriancole: note: we only need this because the interceptor can be used two different ways and we can't distinguish based on null alone. This is a conscious alternative to allocating another wrapper.

brave/src/main/java/brave/Tracer.java
```diff
@@ -543,13 +543,28 @@ public ScopedSpan startScopedSpan(String name) {
...
    */
   public <T> Span nextSpan(SamplerFunction<T> samplerFunction, T arg) {
-    return _toSpan(nextContext(samplerFunction, arg));
+    return _toSpan(nextContext(samplerFunction, arg, currentTraceContext.get()));
   }
 
-  <T> TraceContext nextContext(SamplerFunction<T> samplerFunction, T arg) {
+  /**
+   * Like {@link #nextSpan(SamplerFunction, Object)} except this controls the parent context
+   * explicitly. This is useful when an invocation context is propagated manually, such as commonly
```
anuraaga: Remove `such as`

brave/src/main/java/brave/Tracer.java
```diff
@@ -543,13 +543,28 @@ public ScopedSpan startScopedSpan(String name) {
...
+   * the case with asynchronous client frameworks.
+   *
+   * @param samplerFunction invoked if there's no {@link CurrentTraceContext#get() current trace}
+   * @param arg parameter to {@link SamplerFunction#trySample(Object)}
+   * @param parent the potentially null parent to use for this span
```
anuraaga: `parent of the new span, or {@code null} if it should have no parent`

adriancole:
thanks @anuraaga!

PR #1084, opened 2020-02-16
Introduces HttpRequestParser and HttpResponseParser

adriancole:
This completes the transition of HTTP instrumentation to the new HttpRequest
and HttpResponse types by implementing parsing. Notably, the new parsing
functions can be implemented by lambdas and pass a TraceContext vs scoping.

The main change is implementing parsers more simply.

Ex before:
```java
    .clientParser(new HttpClientParser() {
      @Override
      public <Req> void request(HttpAdapter<Req, ?> adapter, Req req, SpanCustomizer span) {
        span.name(adapter.method(req)); // default span name
        span.tag("http.url", adapter.url(req)); // the whole url, not just the path
      }
    })
    .build();
```

Now
```java
    .clientRequestParser((req, context, span) -> {
      span.name(req.method()); // default span name
      span.tag("http.url", req.url()); // the whole url, not just the path
    })
    .build();
```

This increased performance and allows easier access to extra fields. For
example, a commonly request was to add extra fields as tags. This can now
be done in the parser instead of the FinishedSpanHandler, if desired.

Ex.
```java
    .clientRequestParser((req, context, span) -> {
      HttpClientRequestParser.DEFAULT.parse(req, context, span);
      String userName = ExtraFieldPropagation.get(context, "user-name");
      if (userName != null) span.tag("user-name", userName);
    })
    .build();
```

In order to not break api, all the other logic was kept in
`HttpParserAdapter`, which we can't remove until the next major release.


instrumentation/http/src/main/java/brave/http/HttpClientParserAdapter.java
```diff
@@ -0,0 +1,69 @@
...
+    SpanCustomizer customizer) {
+    HttpResponse response;
+    if (res instanceof HttpClientResponse) {
+      response = new HttpClientResponse.FromHttpAdapter((HttpClientAdapter) adapter, res);
+    } else if (adapter instanceof HttpClientResponse.ToHttpAdapter) {
+      response = ((HttpClientResponse.ToHttpAdapter) adapter).delegate;
+    } else {
+      throw new AssertionError("programming bug");
+    }
+    errorParser.error(error, customizer);
```
anuraaga: Just wondering is there any ordering requirement between these two lines?

instrumentation/http/src/main/java/brave/http/HttpParserAdapter.java
```diff
@@ -0,0 +1,111 @@
...
+    this.currentTraceContext = currentTraceContext;
+    this.parser = parser;
+  }
+
+  @Override public void parse(HttpRequest request, TraceContext context, SpanCustomizer span) {
+    HttpAdapter<Object, ?> adapter;
+    Object req;
+
+    // The order matters here because the deprecated HttpClientHandler.create(httpTracing, adapter)
+    // can be used even with the new HttpRequest type. This means we have to check the HttpResponse
```
anuraaga: HttpRequest type

instrumentation/http/src/main/java/brave/http/HttpParserAdapter.java
```diff
@@ -0,0 +1,111 @@
...
+    // type after checking to see if we are in an adapter scenario.
+    if (response instanceof HttpClientResponse.FromHttpAdapter) {
+      HttpClientResponse.FromHttpAdapter wrapped = (HttpClientResponse.FromHttpAdapter) response;
+      adapter = wrapped.adapter;
+      res = wrapped.response;
+    } else if (response instanceof HttpServerResponse.FromHttpAdapter) {
+      HttpServerResponse.FromHttpAdapter wrapped = (HttpServerResponse.FromHttpAdapter) response;
+      adapter = wrapped.adapter;
+      res = wrapped.response;
+    } else if (response instanceof HttpClientResponse) {
```
anuraaga: Is it ok to have the `else`? (we don't have it above)

instrumentation/http/src/main/java/brave/http/HttpParserAdapter.java
```diff
@@ -0,0 +1,111 @@
...
+
+import brave.Span;
+import brave.SpanCustomizer;
+import brave.internal.Nullable;
+import brave.propagation.CurrentTraceContext;
+import brave.propagation.CurrentTraceContext.Scope;
+import brave.propagation.TraceContext;
+
+/** Adapts {@link HttpParser} to the new request and response objects. */
+@Deprecated final class HttpParserAdapter implements HttpRequestParser, HttpResponseParser {
```
anuraaga: This class is mostly impossible to grok ;) But we can leave it at that since it's deprecated, I wonder if there isn't a small readability improvement possible though by splitting client and server.

instrumentation/http/src/main/java/brave/http/HttpResponse.java
```diff
@@ -0,0 +1,90 @@
...
+import brave.propagation.TraceContext;
+
+/**
+ * Abstract response type used for parsing and sampling of http clients and servers.
+ *
+ * @see HttpClientResponse
+ * @see HttpServerResponse
+ * @since 5.10
+ */
+public abstract class HttpResponse extends Response {
```
anuraaga: Seal to only allow subclasses in brave?

instrumentation/http/src/main/java/brave/http/HttpRequestParser.java
```diff
@@ -0,0 +1,86 @@
...
+   * @see Default
+   */
+  void parse(HttpRequest request, TraceContext context, SpanCustomizer span);
+
+  /**
+   * The default data policy sets the span name to the HTTP method and adds the "http.method" and
+   * "http.path" tags.
+   */
+  // Eventhough the default span name is the method, we have no way of knowing that a user hasn't
+  // overwritten the name to something else. If that occurs during response parsing, it is too late
```
anuraaga: I think it's good to add the method as a tag either way, but for this statement, can't we pass the request to the response parser?

instrumentation/http/src/main/java/brave/http/HttpResponse.java
```diff
@@ -0,0 +1,90 @@
...
+  /**
+   * Like {@link HttpRequest#method()} except used in response parsing.
```
anuraaga: Related to above comment, isn't it easier to reason about (and possibly more powerful) if we pass the request during response parsing instead of duplicating fields into the response?

instrumentation/http/src/main/java/brave/http/HttpResponseParser.java
```diff
@@ -0,0 +1,153 @@
...
+ * <pre>{@code
+ * httpTracing = httpTracing.toBuilder()
+ *   .clientResponseParser((response, error, context, span) -> {
+ *     HttpResponseParser.DEFAULT.parse(response, error, context, span);
+ *     int statusCode = response != null ? response.statusCode() : 0;
+ *     if (statusCode > 0) span.tag("http.status_code", String.valueOf(statusCode));
+ *   }).build();
+ * }</pre>
+ *
+ * <h3>Relationship to {@link HttpParser}</h3>
```
anuraaga: To keep the javadoc of modern, non-deprecated types simpler for users, maybe this sort of reasoning should be in the javadoc of the deprecated type instead.

instrumentation/http/src/main/java/brave/http/HttpRequest.java
```diff
@@ -24,21 +26,63 @@
...
+   * The HTTP method, or verb, such as "GET" or "POST" or null if unreadable.
+   *
+   * <p>Conventionally associated with the key "http.method"
+   *
+   * <h3>Note</h3>
+   * <p>It is part of the <a href="https://tools.ietf.org/html/rfc7231#section-4.1">HTTP RFC</a>
+   * that an HTTP method is case-sensitive. Do not downcase results. If you do, not only will
+   * integration tests fail, but you will surprise any consumers who expect compliant results.
+   */
   @Nullable public abstract String method();
```
anuraaga: It's probably too late now due to backwards compatibility, but I guess we didn't need to use `@Nullable` on attributes that are directly part of the HTTP spec. e.g., an `HttpRequest` without a `method` never exists and doesn't make sense. I find the business logic hard to follow because of the null checks that end up scattered around.

instrumentation/http/src/main/java/brave/http/HttpRequestParser.java
```diff
@@ -0,0 +1,86 @@
...
+  class Default implements HttpRequestParser {
+    /**
+     * This sets the span name to the HTTP method and adds the "http.method" and "http.path" tags.
+     *
+     * <p>If you only want to change the span name, subclass and override {@link
+     * #spanName(HttpRequest, TraceContext)}.
+     */
+    @Override public void parse(HttpRequest req, TraceContext context, SpanCustomizer span) {
+      String name = spanName(req, context);
+      if (name != null) span.name(name);
```
anuraaga: For example here, the fact that method is allowed to be null makes this single line of code a lot more complex (what are the implications of not setting name, for example)

instrumentation/http/src/main/java/brave/http/HttpResponseParser.java
```diff
@@ -0,0 +1,153 @@
...
+      return null;
+    }
+
+    static String spanNameFromRoute(HttpResponse response, int statusCode) {
+      String method = response.method();
+      if (method == null) return null; // don't undo a valid name elsewhere
+      String route = response.route();
+      if (route == null) return null; // don't undo a valid name elsewhere
+      if (!"".equals(route)) return method + " " + route;
+      if (statusCode / 100 == 3) return method + " redirected";
```
anuraaga: I guess most 3xx are redirects, but not 304 Not Modified.

instrumentation/http/src/main/java/brave/http/HttpResponseParser.java
```diff
@@ -0,0 +1,153 @@
...
+      // library being unable to get the http status and a bad status (0). We don't classify zero as
+      // error in case instrumentation cannot read the status. This prevents tagging every response as
+      // error.
+      if (httpStatus == 0) return;
+
+      // Unlike success path tagging, we only want to indicate something as error if it is not in a
+      // success range. 1xx-3xx are not errors. It is endpoint-specific if client codes like 404 are
+      // in fact errors. That's why this is overridable.
+
+      // 1xx, 2xx, and 3xx codes are not all valid, but the math is good enough vs drift and opinion
```
anuraaga: It's not brave's job to validate HTTP, so this comment is a bit overkill

instrumentation/http/src/main/java/brave/http/HttpServerHandler.java
```diff
@@ -148,22 +151,12 @@ Span nextSpan(TraceContextOrSamplingFlags extracted, HttpServerRequest request)
    * @since 4.3
    */
   public void handleSend(@Nullable Resp response, @Nullable Throwable error, Span span) {
-    if (response == null && error == null) {
```
anuraaga: Keep this check?

adriancole:
have a lot of code to push (battery dead).


one thing we could consider I guess is requiring all frameworks to pass
http request to the the response callback. I'm not sure how viable this is
though as many times the callback is async and has no request at all.. in
other words the request would have to be nullable if we did that and
similar to how method is checked null at response time now..it would have
to be double checked.. once for the request and again for the method.

the thing that generally complicates things is being portable

adriancole:
another thing we can do is hold back 5.10 a bit.. ex merge once this part
is stable.. then revisit about request in the response callback as the code
for latter will be less blinding vs doing it now in this pr which will be
thousand plus lines as it is. assuming we even can.

anuraaga:
Would be cool to see an example of where the request may not be available. A request always happens before a response, so if an async framework is able to keep a span / context through to the callback that initiates response parsing, intuitively seems like it'd generally be able to do similar for the request.

anuraaga:
Yeah mostly just food for thought, doesn't have to be in this PR

adriancole:
I'm actually happy to try this and happy for the idea. just knowing the
code I think best to isolate it.. maybe we can open a tracking issue if a
few things pile up. main thing is getting the burnt effort done as that no
one will want to do or manage drift on heh.

adriancole:
@anuraaga thanks for the review! I pushed commits but they were ships passing in the night.. in other words, I didn't see your review when writing this stuff. I'll go through all your comments regardless.

the current status should be that everything works and it is all wired up except spring-beans which I hadn't done, yet. Some of the code supporting deprecated methods still need some unit test coverage and refactoring.

adriancole:
@anuraaga your comments got me thinking.. the throwable arg to the http
parsers predated span.error

I think it might be simplest to just invoke span.error and remove the
throwable arg to the parser. I would also remove the client and server
specific error parsers too.


this could reduce confusion and buy us back a parameter to use for the
request when we revisit that.

wdyt?

anuraaga:
IIUC, that means span.error contains all the logic for an error span. Otherwise, there would be a chance for a parser to handle it specially 

If that sounds right, I'm good with it - I tend to prefer error cases to have extra complexity (FinishedSpanHandler in this case in the chance a user wants something special?) to simplify the standard one.

adriancole:
> IIUC, that means span.error contains all the logic for an error span.
> Otherwise, there would be a chance for a parser to handle it specially
>
indeed.. it also allows finished span handlers to see the exception which
currently they dont because we just parse it. this is important in sending
to other clouds which can render a stack trace.

yes there is an edge case where you get both and error status and also an
exception. in this case we would mark the error tag as "500" which could
mask an exception with a better message. during completion of the span on
the zipkin reporting side, by default an exception is parsed into an error
tag only if there is no existing "error" tag.

so for example if an http server library bubbles an exception yet is
returning an http response 500 .. we wouldnt see the exception error
message by default. not unless we make a new tag to do that.

If that sounds right, I'm good with it - I tend to prefer error cases to
> have extra complexity (FinishedSpanHandler in this case in the chance a
> user wants something special?) to simplify the standard one.
>
yep they can handle the edge case for sure an do something like always add
a tag "exception.message" in a finishedspanhandler.

the code simplifies for the normal case and also helps us similar code
based on this discussion ex. messaging and RPC abstractions.. there we
could also make similar decisions though mostly it would apply to RPC. ex
you could get an RPC error message and also an exception.

instrumentation/http/src/main/java/brave/http/HttpSampler.java
```diff
@@ -75,9 +75,9 @@
     if (request == null) return null;
     HttpAdapter<Object, Void> adapter;
     if (request instanceof HttpClientRequest) {
-      adapter = new HttpClientRequest.ToHttpAdapter((HttpClientRequest) request);
+      adapter = new HttpClientAdapters.ToRequestAdapter((HttpClientRequest) request);
```
jeqo: Should L35 be updates from:

```
 * @deprecated Since 5.8, use {@code RequestSampler<HttpRequest>}.
```

to 

```
 * @deprecated Since 5.8, use {@code SamplerFunction<HttpRequest>}.
```

instrumentation/http/src/main/java/brave/http/HttpHandler.java
```diff
@@ -13,68 +13,66 @@
...
     }
     return span;
   }
 
-  /** parses remote IP:port and tags while the span is in scope (for logging for example) */
-  abstract <Req> void parseRequest(HttpAdapter<Req, ?> adapter, Req request, Span span);
+  abstract void parseRequest(HttpRequest request, Span span);
 
+  void handleFinish(@Nullable HttpResponse response, @Nullable Throwable error, Span span) {
+    if (response == null && error == null) {
```
jeqo: @anuraaga this is the check that was moved from `HttpClientHandler#handleReceive`, but I'm also not sure if this check should be moved back to the beginning of `handleReceive`

brave/src/main/java/brave/Response.java
```diff
@@ -0,0 +1,53 @@
...
+ *
+ * @see Request
+ * @since 5.10
+ */
+public abstract class Response {
+  /** The remote {@link Span.Kind} describing the direction and type of the response. */
+  public abstract Span.Kind spanKind();
+
+  /**
+   * Returns the underlying response object or {@code this} if there is none. Here are some response
```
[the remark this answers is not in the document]
adriancole: users are told to use instanceof, so we can switch this to null as instanceof will fail on null gracefully. I'll open a PR on the other side (Request) and update this.

instrumentation/http/src/main/java/brave/http/HttpRequest.java
adriancole: agree on this.. throwing when method is null would be new, but we could remove the nullable

instrumentation/http/src/main/java/brave/http/HttpRequestParser.java
adriancole: spanName can be overridden, so it can return null anyway (because it allowed this in the past). For example, you can intentionally return null to only set the name at response time. Anyway, I get your point about method though.

instrumentation/http/src/main/java/brave/http/HttpResponseParser.java
adriancole: you are reviewing old code, so I guess it is a good thing to move things around occasionally :)

adriancole:
massive rebase needed  now :)

adriancole:
rebased. next steps:

* pull adapter refactoring/test backfilling into a different PR as it is blinding but mechanical
* remove the errorParser stuff here
* address feedback mentioned so far

adriancole:
this takes the heft of adapter refactor out of this PR https://github.com/openzipkin/brave/pull/1088

adriancole:
TODO: weave in feedback. remove the throwable parameter from the parsers

instrumentation/http/src/main/java/brave/http/HttpClientHandler.java
```diff
@@ -196,22 +207,12 @@ public Span handleSend(HttpClientRequest request, Span span) {
    * @since 4.3
    */
   public void handleReceive(@Nullable Resp response, @Nullable Throwable error, Span span) {
-    if (response == null && error == null) {
```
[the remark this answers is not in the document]
adriancole: moved to the parent class

instrumentation/http/src/main/java/brave/http/HttpHandler.java
adriancole: less work is happening in clientHandler.handleReceive, serverHandler.handleSend now... this allows us to avoid copy/paste

instrumentation/http/src/main/java/brave/http/HttpParserAdapter.java
adriancole: because HttpClientAdapters.FromResponseAdapter is an instanceof HttpClientResponse I think we wouldn't intentionally break the if/else chain. as that would cause us to re-enter. I put stronger comments above this which can be improved if you feel need be.

instrumentation/http/src/main/java/brave/http/HttpServerHandler.java
adriancole: same answer as other copy/paste. we defer the check to the super

instrumentation/http/src/main/java/brave/http/HttpRequestParser.java
adriancole: https://github.com/openzipkin/brave/issues/1086

instrumentation/http/src/main/java/brave/http/HttpRequestParser.java
adriancole: https://github.com/openzipkin/brave/issues/1089

instrumentation/http/src/main/java/brave/http/HttpClientParserAdapter.java
adriancole: obviated now. thanks!

instrumentation/http/src/main/java/brave/http/HttpParserAdapter.java
adriancole: agreed

instrumentation/http/src/main/java/brave/http/HttpResponseParser.java
adriancole: agreed. fixed

instrumentation/http/src/main/java/brave/http/HttpAdapter.java
```diff
@@ -13,113 +13,57 @@
  */
 package brave.http;
 
-import brave.Clock;
-import brave.Span;
 import brave.internal.Nullable;
-import brave.propagation.TraceContext;
 import java.net.URI;
 
 public abstract class HttpAdapter<Req, Resp> {
```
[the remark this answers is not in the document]
adriancole: yep

instrumentation/http/src/main/java/brave/http/HttpRequest.java
adriancole: https://github.com/openzipkin/brave/issues/1089

instrumentation/http/src/main/java/brave/http/HttpSampler.java
adriancole: yep!

adriancole:
adjusted to the much nicer signature now that `Response.error()` exists.

adriancole:
This is ready for review while I backfill unit tests.

adriancole:
ok did my best to reduce the amount of if statements when we are converting existing HttpParser into request/response parsers. The cruft is a reality of keeping portability and I'm kindof done I think on this :)

adriancole:
Epic thanks @anuraaga for all the feedback, especially the critical eye. this is much better. not just code here, but code elsewhere you noticed that is now in the project.

@jeqo thanks also for having a look. This should help when we think about the other unfinished abstractions.

Issue #851, opened 2018-12-18
investigate if we need to improve asynchronous handling of servlet


adriancole:
I was pinged on https://github.com/elastic/apm-agent-java/pull/385 by @felixbarny maybe we have something to adjust. cc also @nicmunroe my other servlet friend ;)

felixbarny:
The check for the completed flag in AsyncListener can also be racy as multiple threads could concurrently enter the callbacks but the flag is only set to true after the method has been executed. Could be fixed with `AtomicBoolean.compareAndSet(false, true)`.

nicmunroe:
Yeah this makes sense. Thanks for the heads-up.

Also @adriancole not sure if you saw this: https://github.com/Nike-Inc/wingtips/pull/82#discussion_r228643265 - a subtlety that causes the wrong HTTP response status code to be set on the span when `onTimeout()` or `onError()` are called.

Also (also!) yet another one more thing - if you're using undertow as your servlet container it completely blows away request attributes before `onComplete()` is called: https://issues.jboss.org/browse/UNDERTOW-1437, screwing up lots of stuff you'd want to do when handling spans for async servlet calls.

nicmunroe:
@eyalkoren I've been able to call `event.getThrowable()` in `onComplete()` and get the associated throwable in the containers I've tested. ... Which might just be one container, I can't remember for sure how much I tested different containers. :) At the very least I did a sanity check just now using some tests from wingtips it works in Jetty 9.3.21. 🤷‍♂️ 

What container(s) does it not work with?

(It's frustrating that we get such wildly different behavior on different servlet containers.)

eyalkoren:
@nicmunroe 
> It's frustrating that we get such wildly different behavior on different servlet containers

Totally agree! It's especially so with the async APIs - see [this comment in the corresponding PR](https://github.com/elastic/apm-agent-java/pull/399/files#diff-0ee5792bbcf396ce2afa01954c8e1216R46) as to what I tested and how different behaviour was.

I tested two error scenarios - one where `AsyncConstext.dispatch` is used and the dispatched servlet ends with throwing `ServletException` and the other where `AsyncContext.start` is used with a `Runnable` that ends throwing a `RuntimeException`.

[...]

PR #1085, opened 2020-02-18
Introduces `Response.error()` and uses it in HTTP abstraction

adriancole:
This introduces `Response.error()`, present when there was an exception
invoking a client request, or processing a server request. This also
ensures that `FinishedSpanHandler` can read it independently of the
"error" tag.

This implied retro-fitting this in the HTTP abstraction and backfilling
tests that tell the difference between HTTP status and a parsed
exception. This identified a gap in Servlet libraries which formerly
swallowed application level exceptions. To bridge that gap, we now
look at the request attribute "error" when ending a server span.

### DIFF
--- a/instrumentation/http-tests/src/main/java/brave/test/http/ITHttpClient.java
+++ b/instrumentation/http-tests/src/main/java/brave/test/http/ITHttpClient.java
@@ -16,20 +16,23 @@ package brave.test.http;
 import brave.ScopedSpan;
 import brave.SpanCustomizer;
 import brave.Tracer;
+import brave.handler.FinishedSpanHandler;
+import brave.handler.MutableSpan;
 import brave.http.HttpAdapter;
 import brave.http.HttpClientParser;
 import brave.http.HttpRuleSampler;
 import brave.http.HttpTracing;
 import brave.propagation.ExtraFieldPropagation;
+import brave.propagation.TraceContext;
 import brave.sampler.Sampler;
 import brave.sampler.SamplerFunctions;
 import java.util.Arrays;
 import java.util.concurrent.TimeUnit;
+import java.util.concurrent.atomic.AtomicReference;
 import okhttp3.mockwebserver.MockResponse;
 import okhttp3.mockwebserver.MockWebServer;
 import okhttp3.mockwebserver.RecordedRequest;
 import okhttp3.mockwebserver.SocketPolicy;
-import org.jetbrains.annotations.NotNull;
 import org.junit.After;
 import org.junit.Before;
 import org.junit.Rule;
@@ -298,6 +301,27 @@ public abstract class ITHttpClient<C> extends ITHttp {
     checkReportsSpanOnTransportException();
   }
 
+  /**
+   * This ensures custom finished span handlers can see the actual exception thrown, not just the
+   * "error" tag value.
+   */
+  @Test public void finishedSpanHandlerSeesException() throws Exception {
+    AtomicReference<Throwable> caughtThrowable = new AtomicReference<>();
+    close();
+    httpTracing = HttpTracing.create(tracingBuilder(Sampler.ALWAYS_SAMPLE)
+      .addFinishedSpanHandler(new FinishedSpanHandler() {
+        @Override public boolean handle(TraceContext context, MutableSpan span) {
+          caughtThrowable.set(span.error());
+          return true;
+        }
+      })
+      .build());
+    client = newClient(server.getPort());
+
+    checkReportsSpanOnTransportException();
+    assertThat(caughtThrowable.get()).isNotNull();
+  }
+
   protected Span checkReportsSpanOnTransportException() throws InterruptedException {
     server.enqueue(new MockResponse().setSocketPolicy(SocketPolicy.DISCONNECT_AT_START));
 
@@ -310,10 +334,15 @@ public abstract class ITHttpClient<C> extends ITHttp {
     return takeSpan();
   }
 
-  @Test public void addsErrorTagOnTransportException() throws Exception {
+  @Test public void errorTag_onTransportException() throws Exception {
     Span span = checkReportsSpanOnTransportException();
-    assertThat(span.tags())
-      .containsKey("error");
+    assertThat(span.tags()).containsKey("error");
+  }
+
+  @Test public void errorTag_exceptionOverridesHttpStatus() throws Exception {
+    Span span = checkReportsSpanOnTransportException();
+    assertThat(span.tags().get("error"))
+      .isNotEqualTo(span.tags().get("http.status_code"));
   }
 
   @Test public void httpPathTagExcludesQueryParams() throws Exception {
--- a/instrumentation/http-tests/src/main/java/brave/test/http/ITHttpClient.java
+++ b/instrumentation/http-tests/src/main/java/brave/test/http/ITHttpClient.java
@@ -27,6 +27,7 @@ import brave.propagation.TraceContext;
 import brave.sampler.Sampler;
 import brave.sampler.SamplerFunctions;
 import java.util.Arrays;
+import java.util.concurrent.Callable;
 import java.util.concurrent.TimeUnit;
 import java.util.concurrent.atomic.AtomicReference;
 import okhttp3.mockwebserver.MockResponse;
@@ -297,15 +298,42 @@ public abstract class ITHttpClient<C> extends ITHttp {
       .isEqualTo("post");
   }
 
+  @Test public void httpPathTagExcludesQueryParams() throws Exception {
+    String path = "/foo?z=2&yAA=1";
+
+    server.enqueue(new MockResponse());
+    get(client, path);
+
+    Span span = takeSpan();
+    assertThat(span.tags())
+      .containsEntry("http.path", "/foo");
+  }
+
+  @Test public void finishedSpanHandlerSeesException() throws Exception {
+    finishedSpanHandlerSeesException(get());
+  }
+
   @Test public void reportsSpanOnTransportException() throws Exception {
-    checkReportsSpanOnTransportException();
+    checkReportsSpanOnTransportException(get());
+  }
+
+  @Test public void errorTag_onTransportException() throws Exception {
+    Span span = checkReportsSpanOnTransportException(get());
+    assertThat(span.tags()).containsKey("error");
+  }
+
+  Callable<Void> get() {
+    return () -> {
+      get(client, "/foo");
+      return null;
+    };
   }
 
   /**
    * This ensures custom finished span handlers can see the actual exception thrown, not just the
    * "error" tag value.
    */
-  @Test public void finishedSpanHandlerSeesException() throws Exception {
+  void finishedSpanHandlerSeesException(Callable<Void> get) throws Exception {
     AtomicReference<Throwable> caughtThrowable = new AtomicReference<>();
     close();
     httpTracing = HttpTracing.create(tracingBuilder(Sampler.ALWAYS_SAMPLE)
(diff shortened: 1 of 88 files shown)


instrumentation/http/src/main/java/brave/http/HttpResponse.java
```diff
@@ -0,0 +1,90 @@
...
+ */
+public abstract class HttpResponse extends Response {
+  /**
+   * Like {@link HttpRequest#method()} except used in response parsing.
+   *
+   * <p>Notably, this is used to create a route-based span name.
+   *
+   * @since 5.10
+   */
+  @Nullable public String method() {
```
anuraaga: Just to confirm, we're adding this but may remove it later if we find it works well to pass the request into the response parser?

instrumentation/http/src/main/java/brave/http/HttpServerRequest.java
```diff
@@ -39,13 +39,34 @@
   }
 
   /**
-   * Override and return true when it is possible to parse the {@link Span#remoteIpAndPort(String,
-   * int) remote IP and port} from the {@link #unwrap() delegate}. Defaults to false.
+   * Used by {@link HttpServerHandler#handleReceive(HttpServerRequest)} to add remote socket
+   * information about the client from the {@link #unwrap() delegate}.
```
anuraaga: `linkplain` when rendering a code link as a normal English word.

instrumentation/http/src/main/java/brave/http/HttpServerRequest.java
```diff
@@ -39,13 +39,34 @@
...
    *
-   * @see HttpServerAdapter#parseClientIpAndPort(Object, Span)
+   * <p>By default, this tries to parse the {@link #parseClientIpFromXForwardedFor(Span) forwarded
```
anuraaga: Ditto

instrumentation/http/src/main/java/brave/http/HttpServerRequest.java
```diff
@@ -39,13 +39,34 @@
...
+   * IP}. Override to add client socket information when forwarded info is not available.
+   *
+   * <p>Aside: the ability to parse socket information on server request objects is likely even if
```
anuraaga: Not sure I parse this sentence well. Is it the same as It is more likely a server request object will be able to parse socket information as opposed to a client object.

adriancole:
While almost all things are dealing with exceptions now in a testable way, there are a couple frameworks not addressing user exceptions properly. In these cases, they can only see the status_code as the "error" tag, not the actual exception.

* jax-rs: nothing we can do about this because the spec had no error handler
* sparkjava over servlet: needs new code to stop the default from eating the exception. Note: sparkjava not over servlet is fine..
* netty http: needs to implement the exception hook, but I tried, and seem to be in an ordering snafu.

instrumentation/servlet/src/main/java/brave/servlet/HttpServletRequestWrapper.java
```diff
@@ -0,0 +1,73 @@
...
+    return delegate.getMethod();
+  }
+
+  @Override public final String path() {
+    return delegate.getRequestURI();
+  }
+
+  // not final as some implementations may be able to do this more efficiently
+  @Override public String url() {
+    StringBuffer url = delegate.getRequestURL();
```
anuraaga: Almost commented this should be `StringBuilder` but realized it's not our API...

instrumentation/http/src/main/java/brave/http/HttpResponse.java
adriancole: good point.. I'll put a TODO!

instrumentation/http/src/main/java/brave/http/HttpServerRequest.java
adriancole: thanks for the help!

adriancole:
I almost regret doing asserts on our "error" tag... there were so many bugs in that!

Anyway the heft of the code is done as now all the integration tests pass. I need to backfill some more unit tests and will likely merge tomorrow unless someone screams. reviews are welcome.

cc @tylerbenson who was curious about what I was up to

instrumentation/http-tests/src/main/java/brave/test/http/ITHttpClient.java
```diff
@@ -310,10 +334,15 @@ protected Span checkReportsSpanOnTransportException() throws InterruptedExceptio
     return takeSpan();
   }
 
-  @Test public void addsErrorTagOnTransportException() throws Exception {
+  @Test public void errorTag_onTransportException() throws Exception {
```
adriancole: TODO: make this test as hard as the server ones (eventhough it will hurt). Right now, this only looks for existance of the "error" tag, which in the case of server, hid bugs.

instrumentation/http-tests/src/main/java/brave/test/http/ITHttpServer.java
```diff
@@ -386,30 +390,82 @@ public void addsStatusCode_badRequest() throws Exception {
...
   }
 
-  Span reportsSpanOnException(String path) throws Exception {
+  void errorTag_exceptionOverridesHttpStatus(String path) throws Exception {
     get(path);
 
-    return takeSpan();
+    Span span = takeSpan();
+    assertThat(span.tags().get("error"))
+      .contains("not ready"); // some controllers format the exception
```
adriancole: when I made the assertion actually look at content, I found most of our code didn't return what we thought it would.. rather it just set the http status.

instrumentation/http-tests/src/main/java/brave/test/http/ITServlet25Container.java
```diff
@@ -71,9 +77,13 @@
...
 
   static class ExceptionServlet extends HttpServlet {
-    @Override
-    protected void doGet(HttpServletRequest req, HttpServletResponse resp) throws IOException {
-      throw new IOException(); // null exception message!
+    @Override protected void doGet(HttpServletRequest req, HttpServletResponse resp)
+      throws UnavailableException {
+      // Change the status from 500 to 503
+      req.setAttribute(RequestDispatcher.ERROR_STATUS_CODE, 503);
+      // TODO: org.eclipse.jetty.server.HttpChannelState.onError() clobbers ^^ and hard-codes
```
adriancole: this was a lot of time lost trying to figure out where it clobbers the code, only in sync errors, not async ones! :P

instrumentation/http-tests/src/main/java/brave/test/http/ITServlet3Container.java
```diff
@@ -30,59 +34,88 @@
...
+  }
+
   static class ExceptionAsyncServlet extends HttpServlet {
-    @Override
-    protected void doGet(HttpServletRequest req, HttpServletResponse resp) {
+    @Override protected void doGet(HttpServletRequest req, HttpServletResponse resp) {
+      if (DispatcherType.ERROR.equals(req.getDispatcherType())) return; // don't loop
+
+      AsyncContext async = req.startAsync();
+      // unless we add a listener, the onError hook will never occur
```
adriancole: I have seen some tests out there that say they are testing onError when they are really testing timeout. beware!

below, I prove onError works by switching the http status

instrumentation/http-tests/src/main/java/brave/test/http/Jetty9ServerController.java
```diff
@@ -0,0 +1,34 @@
...
+ * the License.
+ */
+package brave.test.http;
+
+import brave.test.http.ServletContainer.ServerController;
+import org.eclipse.jetty.server.Connector;
+import org.eclipse.jetty.server.Server;
+import org.eclipse.jetty.server.ServerConnector;
+
+public final class Jetty9ServerController implements ServerController {
```
adriancole: jetty 7 and 9 don't like eachother.

instrumentation/servlet/src/main/java/brave/servlet/HttpServletResponseWrapper.java
```diff
@@ -0,0 +1,98 @@
...
+
+  HttpServletResponseWrapper(HttpServletResponse delegate, @Nullable Throwable caught) {
+    if (delegate == null) throw new NullPointerException("delegate == null");
+    this.delegate = delegate;
+    this.caught = caught;
+  }
+
+  @Override public int statusCode() {
+    int result = ServletRuntime.get().status(delegate);
+    if (caught != null && result == 200) { // We may have a potentially bad status due to defaults
```
adriancole: this magically became needed when I upgraded from jetty 8 to 9.

instrumentation/servlet/src/main/java/brave/servlet/internal/ServletRuntime.java
```diff
@@ -91,43 +92,54 @@ private static ServletRuntime findServletRuntime() {
...
-        complete = true;
+        if (!complete.compareAndSet(false, true)) return; // already completed
+        HttpServletRequest req = (HttpServletRequest) e.getSuppliedRequest();
+        HttpServletResponse res = (HttpServletResponse) e.getSuppliedResponse();
+        HttpServerResponse response =
+          brave.servlet.HttpServletResponseWrapper.create(req, res, e.getThrowable());
+        handler.handleSend(response, response.error(), span);
       }
 
+      // Per Servlet 3 section 2.3.3.3, we can't see the final HTTP status, yet. defer to onComplete
```
adriancole: This stuff addresses #851 thanks @nicmunroe @felixbarny @eyalkoren

instrumentation/servlet/src/main/java/brave/servlet/internal/ServletRuntime.java
```diff
@@ -91,43 +92,54 @@ private static ServletRuntime findServletRuntime() {
...
+      // https://download.oracle.com/otndocs/jcp/servlet-3.0-mrel-eval-oth-JSpec/
       @Override public void onTimeout(AsyncEvent e) {
-        if (complete) return;
-        span.tag("error", String.format("Timed out after %sms", e.getAsyncContext().getTimeout()));
-        handler.handleSend(httpServerResponse(e), null, span);
-        complete = true;
+        if (!timeout.compareAndSet(false, true)) return; // already timed out
```
adriancole: I noticed timeout hitting many times in the same request, but I might double-check this before merge.

adriancole: we weren't and what we are doing in this callback isn't worth guarding

instrumentation/servlet/src/main/java/brave/servlet/internal/ServletRuntime.java
```diff
@@ -91,43 +92,55 @@ private static ServletRuntime findServletRuntime() {
...
         this.handler = handler;
         this.span = span;
       }
 
       @Override public void onComplete(AsyncEvent e) {
-        if (complete) return;
-        handler.handleSend(httpServerResponse(e), null, span);
-        complete = true;
+        if (!complete.compareAndSet(false, true)) {
+          // TODO: None of our tests reach this condition. Make a concrete case that re-enters the
```
adriancole: @nicmunroe @felixbarny PS I did something special to execute the entire suite and fail if this ever was the case.. it never was. Right now, I know I've been told this could be re-entrant, but I'd like to see what test we are missing. Or if any of you have a test that proves it is re-entrant?

nicmunroe: @adriancole I can't remember if I've ever seen it be reentrant in practice. But it's servlet, so my cynical side says it's always possible. Even if you find something in the servlet spec that says impls shouldn't do it, I bet there's some framework that does either due to a bug or because they didn't realize they shouldn't or they feel like they have a use case that warrants it. That's all FUD - I don't have any concrete examples. But I wouldn't be surprised in the least. The question is whether it's worth the extra bulletproofing if we can't find a way to reproduce it. I don't have a good answer on that one.

So I don't really have any advice on this one, sorry. I think there are defensible arguments for going either way.

adriancole: Thanks for the feedback Nic! in our case, the only impact to doing this twice is overhead, because this is on the finish hook not the start hook (eg it won't leak a span). In this case, and as I can't reproduce it normally, I'm inclined to remove the bullet-proofing, maybe mentioning that there's no impact to calling twice (and a test to prove it)

PR #1088, opened 2020-02-21
Fixes null on request or response unwrap and deprecates adapters

adriancole:
Our adapters would never expect the raw request to be null, so this uses
a dummy value to prevent NPEs.

This also backfills our adapters to 100% coverage on paths that are
called.


instrumentation/http/src/main/java/brave/http/HttpClientHandler.java
```diff
@@ -125,7 +128,12 @@ public Span handleSend(HttpClientRequest request, Span span) {
     if (request == null) throw new NullPointerException("request == null");
     if (span == null) throw new NullPointerException("span == null");
     defaultInjector.inject(span.context(), request);
-    return handleStart(new HttpClientRequest.ToHttpAdapter(request), request.unwrap(), span);
+
+    Object unwrapped = request.unwrap();
+    if (unwrapped == null) unwrapped = NULL_SENTINEL; // Handle bad implementation
```
anuraaga: Worth doing a log-once for these sort of bad implementation handlers? I always wonder how a bad implementation will become a good one if we just handle it silently.

adriancole:
good point. logOnce will be trickier as we dont have a type based de-duper
I think.. unless we log once regardless of type..

adriancole:
also we dont actually define this as bad anymore :) the comment was old

anuraaga:
Ah right guess we wouldn't need to log here then. But just in general, I'd say if we had to choose between no logging, logging once globally, or logging once per type with a lot of complexity, logging once globally gives a lot of bang for the buck :)

Issue #1086, opened 2020-02-18
Consider `httpResponse.request()` and deprecating `httpResponse.method()` to 


adriancole:
In Brave 5.7,  `HttpClientRequest`, `HttpClientResponse`, `HttpServerRequest` and `HttpServerResponse`. These are wrappers for underlying types like HttpServletResponse.
Existing code needed to carry-over a use case of reading the HTTP method from a response callback. This was used to do things like make a route-based span name.

#1085  normalizes this to `HttpClientResponse` and `HttpServerResponse` commonality into `HttpResponse`. @anuraaga had an idea that we add `HttpResponse.request()` which could navigate to the corresponding method. `response.request()` would allow more flexible after-the-fact naming vs `response.method()`. For example, someone could fallback to a path based name. Rag may have other ideas, also.

There are a few things we should consider when doing this:

* `HttpClientResponse.method()` and `HttpServerResponse.method()` are already public, even if normalized, so we can't remove them.. however we could make them dispatch and deprecate.
* Sometimes the request reference changes between the start and finish stages of the span. This is particularly possible with some requests being immutable. We'd have to decide to always pass the same reference we parsed, or document that the request returned in `HttpResponse.request().unwrap()` may not be the same reference as the one we started with.
* There may be cases where we can't conveniently pass the original request directly to the response handler, and events may not have a full reference to the request. (thinking async servlet). This may imply response.request() is nullable.
* Holding a reference to a string for method is safer than holding a request object until a potentially async response/error callback. If we think this danger exists, we need `HttpResponse.request()` nullable to permit us flexibility to undo an unforeseen disaster.

Mostly, I'm concerned about the last point, and maybe @felixbarny @nicmunroe @raphw @eyalkoren know of some horror stories about holding references to requests during response callbacks.. I'll admit maybe I'm too nervous.

nicmunroe:
Yeah there are certainly cases where you can get into trouble hanging onto request references, but it's not super common in my personal experience (YMMV depending on the services you tend you support). So I'm not sure how worried you should be...

One example I can think of - undertow will start cleaning up the request before it calls the async listeners for async requests: https://issues.redhat.com/browse/UNDERTOW-1437. Or at least it used to, I'm not sure if newer versions ever cleared that up. And that was for request attributes, not request method, but it's indicative of the kind of nastiness you can run into depending on the framework.

[...]

PR #1091, opened 2020-02-23
Makes HTTP request method mandatory

### LINKED ISSUE #1089: Escalate `HttpRequest.method()` as not nullable
While all other properties, possibly even path, could result in exceptions coerced to null, the HTTP method of the request should never end up null.

While there is a slight NPE risk in some implementations, I think it should be safe to enforce HTTP method (`HttpRequest.method()`) as not nullable. This allows one less null check on an extremely common path.

adriancole:
While all other properties can be absent or invalid, this makes the HTTP
method mandatory. This will help when we create a path to the request
from the response, as it allows chaining to the method without a null
check.

Note: Eventhough we've seen method incorrectly set in the past,
we've never seen it missing.

PS: While it seems path would also always be available even recently,
it isn't valid for all HTTP methods (ex OPTIONS and CONNECT). Also,
it can be null due to bad parsing (ex some things don't natively return
the path, only the URL). Also, it has been found missing before, most
recently due to a [request availablility bug](https://github.com/reactor/reactor-netty/pull/998) which is still not released! If we started enforcing path be present, it
would force a known set of scenarios to NPE causing people who can't
change their implementations to become revlocked until they can. This
makes forcing path to be not nullable, so late in the game, to be a cure
worse than the disease.

Fixes #1089

### DIFF
--- a/instrumentation/http/src/main/java/brave/http/HttpAdapter.java
+++ b/instrumentation/http/src/main/java/brave/http/HttpAdapter.java
@@ -19,7 +19,7 @@ import java.net.URI;
 /** @deprecated Since 5.10, use {@link HttpRequest} and {@link HttpResponse} */
 @Deprecated public abstract class HttpAdapter<Req, Resp> {
   /** @see HttpRequest#method() */
-  @Nullable public abstract String method(Req request);
+  public abstract String method(Req request);
 
   /** @see HttpRequest#path() */
   @Nullable public String path(Req request) {
--- a/instrumentation/http/src/main/java/brave/http/HttpRequestParser.java
+++ b/instrumentation/http/src/main/java/brave/http/HttpRequestParser.java
@@ -14,7 +14,7 @@
 package brave.http;
 
 import brave.SpanCustomizer;
-import brave.propagation.ExtraFieldPropagation;
+import brave.internal.Nullable;
 import brave.propagation.TraceContext;
 
 /**
@@ -67,8 +67,11 @@ public interface HttpRequestParser {
       if (path != null) span.tag("http.path", path);
     }
 
-    /** Returns the span name of the request. Defaults to the http method. */
-    protected String spanName(HttpRequest req, TraceContext context) {
+    /**
+     * Returns the span name of the request or null if the data needed is unavailable. Defaults to
+     * the http method.
+     */
+    @Nullable protected String spanName(HttpRequest req, TraceContext context) {
       return req.method();
     }
   }
--- a/instrumentation/http/src/main/java/brave/http/HttpRequestParser.java
+++ b/instrumentation/http/src/main/java/brave/http/HttpRequestParser.java
@@ -62,8 +62,7 @@ public interface HttpRequestParser {
     @Override public void parse(HttpRequest req, TraceContext context, SpanCustomizer span) {
       String name = spanName(req, context);
       if (name != null) span.name(name);
-      String method = req.method();
-      if (method != null) span.tag("http.method", method);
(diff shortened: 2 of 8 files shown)


instrumentation/http/src/main/java/brave/http/HttpAdapter.java
```diff
@@ -19,7 +19,7 @@
 /** @deprecated Since 5.10, use {@link HttpRequest} and {@link HttpResponse} */
 @Deprecated public abstract class HttpAdapter<Req, Resp> {
   /** @see HttpRequest#method() */
-  @Nullable public abstract String method(Req request);
+  public abstract String method(Req request);
 
   /** @see HttpRequest#path() */
   @Nullable public String path(Req request) {
```
anuraaga: Agree that HTTP spec doesn't actually require a path (OPTIONS commonly has * which I don't know if we call a path, CONNECT has only authority). So nullable seems fine.

But not so sympathetic with the availability bug issue - seems like bad http library as opposed to bad instrumentation, not sure if that specifically should fall into our defensive requirements.

instrumentation/http/src/main/java/brave/http/HttpRequestParser.java
```diff
@@ -62,8 +62,7 @@
     @Override public void parse(HttpRequest req, TraceContext context, SpanCustomizer span) {
       String name = spanName(req, context);
       if (name != null) span.name(name);
```
anuraaga: Think we can remove this null check (it's the one I most wanted to remove since it's the span name :) ).

But if you think we need it I guess we are missing `@Nullable` on the protected `spanName` method.

instrumentation/http/src/main/java/brave/http/HttpAdapter.java
adriancole: This was one example of path missing, something from last week, not the only time.

I think it is wise to be cautious as instrumentation should do no harm. Users can't always control versions of things and for example, even in armeria there are availability checks to prevent a known class of bugs. I don't expect you to agree with me but I disagree that we should just trust path will exist and anything that causes it not to be should crash requests.

instrumentation/http/src/main/java/brave/http/HttpRequestParser.java
adriancole: I can fix the lack of nullable

instrumentation/http/src/main/java/brave/http/HttpAdapter.java
adriancole: I reworded the javadoc towards the method rationale without suggesting that's the only reason why. The other reasons could be in the class of parse failures, or availability drift outside our control, but IMHO shouldn't cause requests to crash. hope the words balance!

adriancole: concretely, if we "fixed path" now, we'd definitely blow up former versions of instrumentation, causing revlock.

adriancole: PS I'd be fine with forcing path to be present on the methods relevant in a major release. We'd just know this would invalidate third party stuff and they'd get revlocked to the old one, causing the old version to need patches in worst case. Not signing up for that work personally, but agree in a green field or known break scenario stricter rules could apply.

adriancole:
added rationale because I think many people have not seen the horrors I have and wouldn't want anyone walking into a disaster by accident of less experience cc @openzipkin/core

adriancole:
merging as #1086 rests on this assumption

PR #1092, opened 2020-02-23
Adds HttpResponse.request()

### LINKED ISSUE #1086: Consider `httpResponse.request()` and deprecating `httpResponse.method()` to

adriancole:
This adds nullable `HttpResponse.request()` per discussion in #1086

This doesn't deprecate method or route as there are a couple places you
can't see the request (unless we propagate it ourselves). Also, we have
a scenario in vert.x where the data holding the route isn't on the
request.

Fixes #1086

### DIFF
--- a/instrumentation/http/src/main/java/brave/http/HttpResponse.java
+++ b/instrumentation/http/src/main/java/brave/http/HttpResponse.java
@@ -45,7 +45,8 @@ public abstract class HttpResponse extends Response {
    * Returns the {@linkplain HttpRequest#method() HTTP method} of the request that caused this
    * response or {@code null} if unreadable.
    *
-   * <p>Note: This value may be present even when {@link #request()} is {@code null}.
+   * <p>Note: This value may be present even when {@link #request()} is {@code null}, but
+   * implementations should implement {@link #request()} when possible.
    *
    * @see HttpRequest#method()
    * @since 5.10
@@ -59,7 +60,8 @@ public abstract class HttpResponse extends Response {
    * Returns the {@linkplain HttpRequest#route() HTTP route} of the request that caused this
    * response or {@code null} if unreadable.
    *
-   * <p>Note: This value may be present even when {@link #request()} is {@code null}.
+   * <p>Note: This value may be present even when {@link #request()} is {@code null}, but
+   * implementations should implement {@link #request()} when possible.
    *
    * @see HttpRequest#route()
    * @since 5.10
--- a/instrumentation/http/src/main/java/brave/http/HttpResponse.java
+++ b/instrumentation/http/src/main/java/brave/http/HttpResponse.java
@@ -28,31 +28,45 @@ import brave.propagation.TraceContext;
  */
 public abstract class HttpResponse extends Response {
   /**
-   * Like {@link HttpRequest#method()} except used in response parsing.
+   * The request that initiated this HTTP response or {@code null} if unknown.
    *
-   * <p>Notably, this is used to create a route-based span name.
+   * <p>Implementations should return the last wire-level request that caused this response or
+   * error. HTTP properties like {@linkplain HttpRequest#path() path} and {@linkplain
+   * HttpRequest#header(String) headers} might be different, due to redirects or authentication.
+   * Some properties might not be visible until response processing, notably {@link #route()}.
    *
    * @since 5.10
    */
-  // TODO: Consider `httpResponse.request()` and deprecating `httpResponse.method()` #1086
-  @Nullable public String method() {
+  @Nullable public HttpRequest request() {
     return null;
   }
 
   /**
-   * Returns an expression such as "/items/:itemId" representing an application endpoint,
-   * conventionally associated with the tag key "http.route". If no route matched, "" (empty string)
-   * is returned. Null indicates this instrumentation doesn't understand http routes.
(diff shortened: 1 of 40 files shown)


adriancole:
before I start doing unit tests and porting code to use this.. @nicmunroe @anuraaga anything fire your nerves so far?

instrumentation/http/src/main/java/brave/http/HttpResponse.java
```diff
@@ -28,31 +28,45 @@
...
+  @Nullable public HttpRequest request() {
     return null;
   }
 
   /**
-   * Returns an expression such as "/items/:itemId" representing an application endpoint,
-   * conventionally associated with the tag key "http.route". If no route matched, "" (empty string)
-   * is returned. Null indicates this instrumentation doesn't understand http routes.
+   * Returns the {@linkplain HttpRequest#method() HTTP method} of the request that caused this
+   * response or {@code null} if unreadable.
```
anuraaga: Maybe it's worth adding Where possible, prefer to populate `{@link #request()} to provide both the method and other information that may be useful during response parsing.

instrumentation/servlet/src/main/java/brave/servlet/HttpServletResponseWrapper.java
```diff
@@ -31,29 +30,37 @@
...
+    this.response = response;
     this.caught = caught;
   }
 
+  @Override public HttpRequest request() {
+    return request;
+  }
+
+  @Override public Throwable error() {
+    if (caught != null || request == null) return caught;
```
anuraaga: I had to do a doubletake here, it's an extra line but this could help readers

```java
if (caught != null) return caught;
if (request == null) return null;
return request.maybeError();
```

instrumentation/servlet/src/main/java/brave/servlet/HttpServletResponseWrapper.java
```diff
@@ -65,34 +72,7 @@ public static HttpServerResponse create(@Nullable HttpServletRequest req,
...
 
-  @Override public Throwable error() {
-    return caught;
-  }
-
   @Override public final Object unwrap() {
-    return delegate;
-  }
-
-  static final class WithRequestProperties extends HttpServletResponseWrapper {
```
anuraaga: Fun to see this class get deleted

instrumentation/http/src/main/java/brave/http/HttpResponse.java
adriancole: sure

instrumentation/http/src/main/java/brave/http/HttpResponse.java
```diff
@@ -28,31 +28,45 @@
...
+   * <p>Implementations should return the last wire-level request that caused this response or
+   * error. HTTP properties like {@linkplain HttpRequest#path() path} and {@linkplain
+   * HttpRequest#header(String) headers} might be different, due to redirects or authentication.
+   * Some properties might not be visible until response processing, notably {@link #route()}.
    *
    * @since 5.10
    */
-  // TODO: Consider `httpResponse.request()` and deprecating `httpResponse.method()` #1086
-  @Nullable public String method() {
+  @Nullable public HttpRequest request() {
```
adriancole: @anuraaga I feel like this should be on Response, not HttpResponse. We added Response in 5.10 (this pending version). If we pull it up afterwards (for Rpc) it will look awkward I think..

anuraaga: Ah yeah that makes sense to me

adriancole:
PTAL if you can

adriancole:
Thanks for all the review help @anuraaga!

PR #1095, opened 2020-02-24
Ensures gRPC client response callbacks happen in invocation context

### LINKED ISSUE #1094: ui display wrong hierarchy between rpc, `PRODUCER` and `CONSUMER` spans
## Describe the Bug

zipkin ui (both old ui and lens) display wrong hierarchy between rpc, `PRODUCER` and `CONSUMER` spans.

rpc span:

rpc

produce span(should be child of rpc):

produce

## Steps to Reproduce

Not sure

## Expected Behaviour

the produce span should be child of the rpc span in the ui.

adriancole:
Same as #1067 (HTTP), except for gRPC.

Fixes  #1094

### DIFF
--- a/instrumentation/grpc/src/main/java/brave/grpc/TracingClientInterceptor.java
+++ b/instrumentation/grpc/src/main/java/brave/grpc/TracingClientInterceptor.java
@@ -1,5 +1,5 @@
 /*
- * Copyright 2013-2019 The OpenZipkin Authors
+ * Copyright 2013-2020 The OpenZipkin Authors
  *
  * Licensed under the Apache License, Version 2.0 (the "License"); you may not use this file except
  * in compliance with the License. You may obtain a copy of the License at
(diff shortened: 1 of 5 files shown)


adriancole:
cc @kojilin as eventhough a year since our pow-wow in Fukuoka where we discuss who should be parent during response parsing.. well, we still have some slight glitches due to lack of test

instrumentation/grpc/src/main/java/brave/grpc/TracingClientInterceptor.java
```diff
@@ -93,10 +111,62 @@
...
+      @Nullable TraceContext invocationContext
+    ) {
+      super(delegate);
+      this.currentTraceContext = currentTraceContext;
+      this.invocationContext = invocationContext;
+    }
+
+    @Override public void onReady() {
+      Scope scope = currentTraceContext.maybeScope(invocationContext);
+      try {
```
anuraaga: Not sure if gRPC being Java 8 means we would prefer try/resources here but just in case

adriancole: I didn't move over the retrolambda comment. we still support gRPC 1.2 :P

adriancole: I will make a comment in the pom.xml in next change which is that when we bring the floor version up to v1.16.0, we can bump to Java 1.7 and use try/resources naturally (unless retrolambda still complains)

adriancole: wait that's already there.. I'll just add the comment :)

PR #1096, opened 2020-02-24
Updates to latest gRPC and stops writing the "method" tag

adriancole:
This completes the promise made in Brave 5.9.2's release notes:
> Our gRPC instrumentation was written to be compatible with the Census project. Starting with gRPC v1.22, the Census integration stopped propagating the tag "method". While we still propagate this, the next minor version of Brave will also stop, and save needless in-process and wire overhead. As gRPC dropped Census as a core dependency in v1.27, we will similarly also de-prioritize interop with Census.


adriancole:
forgot some code.. a couple mins please

instrumentation/grpc/src/main/java/brave/grpc/GrpcPropagation.java
```diff
@@ -39,14 +38,6 @@
   static final Metadata.Key<Map<String, String>> GRPC_TAGS_BIN =
     Metadata.Key.of("grpc-tags-bin", new TagContextBinaryMarshaller());
 
-  /**
-   * The census tag key corresponding to the {@link MethodDescriptor#getFullMethodName()}.
-   *
-   * <p>Note: this was removed in gRPC 1.22. We currently continue to propagate it for interop with
-   * earlier versions of gRPC.
-   */
-  static final String RPC_METHOD = "method";
```
adriancole: note: we still propagate it now, just we don't refresh it.

PR #1099, opened 2020-02-24
Handles special case when a destination is both queue and topic

### LINKED ISSUE #1098: Writing a message into an OracleAQJms Topic throws NPE if a trace context already exists
I get a NPE (java.lang.NullPointerException: value of jms.queue == null) when trying to enqueue a message to an Oracle AQJms Topic. It works with an Oracle Queue.
The Exception atually is thrown in MutableSpan.java:266 when we try to tag a span with the destination name.

The issue is caused when trying to read the channelName from the Destination in [MessageParser.java](https://github.com/openzipkin/brave/blob/0f03bcbdb419ac92c51761c377a57e09a0b63d73/instrumentation/jms/src/main/java/brave/jms/MessageParser.java#L46)

[AQjmsDestination implements Queue and Topic](https://docs.oracle.com/cd/B19306_01/server.102/b14291/oracle/jms/AQjmsDestination.html) so in this method we always treat any destination as queue. Which results in null when reading the queueName if it's actually a Topic.

adriancole:
Fixes #1098

### DIFF
--- a/instrumentation/jms/src/main/java/brave/jms/MessageParser.java
+++ b/instrumentation/jms/src/main/java/brave/jms/MessageParser.java
@@ -1,5 +1,5 @@
 /*
- * Copyright 2013-2019 The OpenZipkin Authors
+ * Copyright 2013-2020 The OpenZipkin Authors
  *
  * Licensed under the Apache License, Version 2.0 (the "License"); you may not use this file except
  * in compliance with the License. You may obtain a copy of the License at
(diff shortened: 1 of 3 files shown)


instrumentation/jms/src/main/java/brave/jms/MessageParser.java
```diff
@@ -36,16 +36,35 @@
...
-    if (destination instanceof Queue) return "queue";
-    if (destination instanceof Topic) return "topic";
-    return null;
+    if (destination == null) return null;
+    return isQueue(destination) ? "queue" : "topic";
+  }
+
+  /**
+   * Handles special case of a destination being both a Queue and a Topic by checking if the
+   * {@linkplain Queue#getQueueName() queue name} is readable.
```
anuraaga: Perhaps need to link to the issue here or provide a bit more background. The JMS APIs are generally sketchy, but the javadoc doesn't seem to indicate this name would ever return null.

adriancole: we check null anyway, especially as everything can throw JMSException and we coerce those to null. Our bug was one place where we didn't: this line is more about fallback behavior than null.

This era of library design didn't have clarifying annotations or docs on null as a matter of course. This matter is compounded by TCKs not being open source either.. I don't think it would help to assume non-null on JMS properties with or without this change.

I could add a comment about the example of something implementing both, but I'd hesitate to use this as a why something can return null as generally they can become null and don't want people mistakenly thinking null is an edge case.

wdyt?

instrumentation/jms/src/main/java/brave/jms/MessageParser.java
```diff
@@ -36,16 +36,35 @@
...
+   */
+  static boolean isQueue(@Nullable Destination destination) {
+    boolean isQueue = destination instanceof Queue;
+    boolean isTopic = destination instanceof Topic;
+    if (isQueue && isTopic) {
+      try {
+        isQueue = ((Queue) destination).getQueueName() != null;
+      } catch (Throwable t) {
+        propagateIfFatal(t);
+        log(t, "error getting destination name from {0}", destination, null);
```
adriancole: @anuraaga ex we catch this. we had to recently add these catches everywhere because implementations are free to throw for dumb reasons.

instrumentation/jms/src/main/java/brave/jms/MessageParser.java
adriancole: https://github.com/openzipkin/brave/pull/1100 hoping this topic doesn't eat our afternoon

PR #1100, opened 2020-02-25
Adds really long explanation about JMS being horrible

adriancole:
This is mainly to answer a request, the same rationale would apply to
just about anything as it is common for implementations to implement
multiple interfaces (ex connection factory) and also everything in old
apis to return null.

See https://github.com/openzipkin/brave/pull/1099#discussion_r383629865

### DIFF
--- a/instrumentation/jms/src/main/java/brave/jms/MessageParser.java
+++ b/instrumentation/jms/src/main/java/brave/jms/MessageParser.java
@@ -43,13 +43,17 @@ final class MessageParser {
 
   /**
    * Handles special case of a destination being both a Queue and a Topic by checking if the
-   * {@linkplain Queue#getQueueName() queue name} is readable.
+   * {@linkplain Queue#getQueueName() queue name} is readable and not {@code null}.
    */
   static boolean isQueue(@Nullable Destination destination) {
     boolean isQueue = destination instanceof Queue;
     boolean isTopic = destination instanceof Topic;
     if (isQueue && isTopic) {
       try {
+        // The JMS 1.1 specification does not define the result of queue name or topic name,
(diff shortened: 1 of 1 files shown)


adriancole:
cc @jeqo as this is important for the pending messaging abstraction. while it is unlikely that the channel name could be non-null, we may have to provide some awful dummy value if we assume it must not be null.

An awful dummy value is what we used to do in span names "unknown" which cluttered things, leading to us later killing that practice for just being truly null. It was a massive amount of work to undo "unknown" span name, in other words. I don't suggest we do anything in the future that relies on channel name being present.. If we think it is truly an edge case, we could do the "unknown" hack, just knowing that we've had to remove tech debt around this in the past and that was very labor intensive.

adriancole:
incidentally our abstraction already says things can be null, the above defensiveness was just to pray no-one undoes that, or at least doesn't do that and run away from the mountain of bugs doing so would create https://github.com/openzipkin/brave/blob/master/instrumentation/messaging/src/main/java/brave/messaging/MessagingRequest.java#L67

instrumentation/jms/src/main/java/brave/jms/MessageParser.java
```diff
@@ -58,6 +59,16 @@ static boolean isQueue(@Nullable Destination destination) {
     return isQueue;
   }
 
+  /**
+   * Similar to other properties, {@code null} should be expected even if it seems unintuitive.
```
anuraaga: I probably should have just given a suggestion than request ;) I didn't mean channel name, just a line or so in `isQueue` above the `getQueueName` null check saying while the spec doesn't seem to define it, in practice implementations that implement both interfaces return only one of the names for which they are active.

PR #1101, opened 2020-02-25
Enforces sampled field must be well formed

adriancole:
Currently, we restart the trace if the identity fields: traceId and
spanId are malformed. This restarts on malformed sampled flag. Formerly,
we treated malformed the same as an explicit sampled=false decision.

### DIFF
--- a/brave/src/main/java/brave/propagation/B3Propagation.java
+++ b/brave/src/main/java/brave/propagation/B3Propagation.java
@@ -117,7 +117,7 @@ public final class B3Propagation<K> implements Propagation<K> {
    */
   static final String SAMPLED_NAME = "X-B3-Sampled";
   static final String SAMPLED_MALFORMED =
-    "Invalid input: expected 0 or 1 for " + SAMPLED_NAME + ": {0}";
+    "Invalid input: expected 0 or 1 for " + SAMPLED_NAME + ", but found '{0}'";
 
   /**
    * "1" implies sampled and is a request to override collection-tier sampling policy.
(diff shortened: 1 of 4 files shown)


brave/src/main/java/brave/propagation/B3Propagation.java
```diff
@@ -115,10 +116,14 @@ public FactoryBuilder injectFormats(Span.Kind kind, Format format1, Format forma
    * decision to the receiver of this header).
    */
   static final String SAMPLED_NAME = "X-B3-Sampled";
+  static final String SAMPLED_MALFORMED =
+    "Invalid input: expected 0 or 1 for " + SAMPLED_NAME + ": {0}";
```
anuraaga: Perhaps `: '{0}'` with quotes - IIUC empty string will also log so if it just logs as empty it's tricky to read the log. But either's fine.

adriancole: good idea.. in this case I'll put the whole thing in single quotes as was aiming for header format.

Ex. `X-B3-Sampled: ` in the case of empty string

adriancole: but then again, that would still have the same problem.. yeah single quoting it

brave/src/test/java/brave/propagation/B3PropagationTest.java
```diff
@@ -179,9 +214,62 @@ void header(String key, String value) {
...
+
+    assertThat(extract(headers).sampled()).isFalse();
+  }
+
+  @Test public void extract_sampledCorrupt() {
+    Map<String, String> headers = new LinkedHashMap<>();
+    headers.put("X-B3-TraceId", traceId);
+    headers.put("X-B3-SpanId", spanId);
+
+    Stream.of("", "d", "💩", "hello").forEach(sampled -> {
```
anuraaga: Nice

PR #1103, opened 2020-02-27
Slightly adjusts Javadoc on HttpClientRequest

adriancole:
Before, we said calls during parsing were invalid, but we didn't say
they could be ignored. This clarifies the point to avoid a temptation
to throw an exception which could crash things.

### DIFF
--- a/instrumentation/http/src/main/java/brave/http/HttpClientRequest.java
+++ b/instrumentation/http/src/main/java/brave/http/HttpClientRequest.java
@@ -44,7 +44,7 @@ public abstract class HttpClientRequest extends HttpRequest {
    *
    * <p>This is only used when {@link TraceContext.Injector#inject(TraceContext, Object) injecting}
    * a trace context as internally implemented by {link HttpClientHandler}. Calls during sampling or
-   * parsing are invalid and may be ignored.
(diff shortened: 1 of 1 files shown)


instrumentation/http/src/main/java/brave/http/HttpClientRequest.java
```diff
@@ -40,11 +40,11 @@
...
   /**
-   * Sets a request header with the indicated name. Null values are unsupported.
+   * Sets a request header with the indicated name. {@code null} values are unsupported.
    *
-   * This is only used when {@link TraceContext.Injector#inject(TraceContext, Object) injecting} a
-   * trace context as internally implemented by {link HttpClientHandler}. Calls during sampling or
-   * parsing are invalid.
+   * <p>This is only used when {@link TraceContext.Injector#inject(TraceContext, Object) injecting}
+   * a trace context as internally implemented by {link HttpClientHandler}. Calls during sampling or
+   * parsing are invalid and may be ignored.
```
anuraaga: Maybe can add `may be ignored by the instrumentation.`

I first read this as a user ignoring something and got confused.

adriancole: yeah that's better.

Issue #826, opened 2018-11-14
support tunneling correlation entries through one propagation field


adriancole:
We discussed this in dev chat 3-6 Sep with  @shakuzen @zeagord @ivantopo

The scenario is this: Someone uses ExtraFieldPropagation, yet they hit a snag as they enter JMS which has strict naming concerns. Propagation fields are lost across links like this.

We referred to w3c correlation-context which is described in other issues. This includes an comma+equals encoding which is generally understood and used today. However, obviously a name like "correlation-context" would not only break JMS, but also would be lying as the spec isn't even draft https://github.com/w3c/correlation-context/blob/master/correlation_context/HTTP_HEADER_FORMAT.md

Practically speaking, comma/equals joined is useful meanwhile.. we just need to be careful of names and dodge getting into "forbidden character replacement" games.

Over the chat @zeagord identified "b3ext" could be a decent name as it hints at either "extra" or "extension" either of which intentionally not defined in "b3" as this is just a placeholder. Folks can choose their own name, but suffer catastrophes if they add dots or hyphens to it :P

From a loose spec POV folks know beyond 8192bytes would certainly be bad. OTOH, there's no clear best answer either.. probably something smaller would still get data needed and avoid people using extra fields as application data storage. One way is to have an order that is itself a priority. For example, we prioritize the names listed first and they will be truncated or deleted last.

Now, this feature could be used as a normal thing holding propagation properties, or a tunnel for transports with constraints. For example, you can create whitelist of properties allowed to tunnel over this like `packed=user-id,correlation-key,x-ivan,x-diego` which creates a special header named packed, which is understood to be different than "b3ext" if such clarity is needed.

Anyway this was the topic! thumbsup reply or otherwise.

PR #1109, opened 2020-03-06
Allows scope decorators to synchorize "extra fields" to logging context

adriancole:
This pulls up a major feature currently only available in Sleuth:

https://github.com/spring-cloud/spring-cloud-sleuth/blob/289ec4755e6bcda77830a25c958c19fb1f5af7e2/spring-cloud-sleuth-core/src/main/java/org/springframework/cloud/sleuth/log/Slf4jScopeDecorator.java#L111-L119

Ex.
```java
currentTraceContext = ThreadLocalCurrentTraceContext.newBuilder()
  .addScopeDecorator(MDCScopeDecorator.newBuilder()
    .addExtraField("user-id") // this pushes the extra field into log scope
    .build())
  .build();
```


brave/src/main/java/brave/internal/propagation/CorrelationFieldScopeDecorator.java
```diff
@@ -13,85 +13,184 @@
...
-    if (currentSpan != null) {
-      maybeReplaceTraceContext(
-        currentSpan, previousTraceId, previousParentId, previousSpanId, previousSampled);
-    } else {
-      remove("traceId");
-      remove("parentId");
-      remove("spanId");
-      remove("sampled");
+  @Override public Scope decorateScope(@Nullable TraceContext context, Scope scope) {
+    String[] previousValues = new String[updaters.length];
```
adriancole: using a mutable array here is intentional. It allows a later change to implement immediate flushing https://github.com/openzipkin/brave/issues/682

anuraaga: Not completely sure but I thought the immediate flushing would mean there are extra fields added, so doesn't it need to be expandable like a `List`?

brave/src/main/java/brave/internal/propagation/CorrelationFieldScopeDecorator.java
```diff
@@ -13,85 +13,176 @@
...
+    boolean update(@Nullable String previous, String current) {
+      if (!current.equals(previous)) {
+        decorator.put(field, current);
+        return true;
+      }
+      return false;
+    }
+
+    boolean updateNullable(@Nullable String previous, @Nullable String current) {
+      if (current != null) return update(previous, current);
```
anuraaga: Small nit this is probably easier to read if the if/else pattern is consistent, currently this line seems particularly special for some reason.

```java
if (current != null) {
  return update(previous, current);
} else if (previous != null) {
  decorator.remove(field);
  return true;
} else {
  return false;
}
```

Could remove the elses if desired.

brave/src/main/java/brave/internal/propagation/CorrelationFieldScopeDecorator.java
```diff
@@ -13,85 +13,176 @@
...
  * is used as a hint that a span found in logs might be in Zipkin.
  */
 public abstract class CorrelationFieldScopeDecorator implements ScopeDecorator {
+  final Updater[] updaters;
+
+  protected static abstract class Builder<B extends Builder<B>> {
+    final Set<String> extraFields = new LinkedHashSet<>(); // insertion order;
+
+    /**
+     * Adds a field from {@link ExtraFieldPropagation} into the correlation context.
```
anuraaga: Just want to confirm `ExtraFieldPropagation` is the only type of propagation this would ever make sense for?

adriancole: it might not be the only possible thing but it is what's often requested leading to the sleuth thing a couple years ago.

brave/src/main/java/brave/internal/propagation/CorrelationFieldScopeDecorator.java
adriancole: actually the late thing is change in a value not change in the whitelist. so should be fine. this will be separate change anyway

adriancole:
@anuraaga I think I agree with your hints about this maybe flexing as we still have folks wanting [unrestricted key names](https://github.com/openzipkin/brave/issues/826). That said, I'm not sure how exactly that would work.. it may be that we solve this now, or later, but in the mean time we should definitely not expose internal types used here as they are not ready.

Regardless, at least the addExtraField javadoc should change to constrain this directly to a specific act on fixed `ExtraFieldPropagation`.. entry vs a dynamic one. That or move the config to  `ExtraFieldPropagation` somehow..

Next, the performance impact of this design isn't good yet. The lookup of the "extra" object happens once per key, which is not ideal.. whichever internal types used to perform updates should probably be redone to handle multiple keys.

Anyway.. I'm going for more think time on this. Thanks!

brave/src/main/java/brave/internal/propagation/CorrelationFieldScopeDecorator.java
adriancole: yeah this is just late updates keys, set in the builder and immutable after. so the value holder needn't be expandable. If we change later to make this expandable, we would need to update this code, as would other code which assumes the list is fixed.

brave/src/main/java/brave/internal/propagation/CorrelationFieldScopeDecorator.java
adriancole: after thinking about this, yeah I believe it is better to be explicit as we have nothing else and it is hard to predict when we could.

adriancole:
cool those interested in performance (GC overhead mostly) should remove parentId and sampled

(default is only updating a thread local, that's why it is so cheap)
```
CurrentTraceContextBenchmarks.newScope_default:newScope_default·p0.999                         sample             0.295             us/op
CurrentTraceContextBenchmarks.newScope_default:·gc.alloc.rate.norm                             sample      15     0.003 ±   0.001    B/op
CurrentTraceContextBenchmarks.newScope_log4j2:newScope_log4j2·p0.999                           sample            11.552             us/op
CurrentTraceContextBenchmarks.newScope_log4j2:·gc.alloc.rate.norm                              sample      15  1920.210 ±   0.020    B/op
CurrentTraceContextBenchmarks.newScope_log4j2_extra:newScope_log4j2_extra·p0.999               sample            13.712             us/op
CurrentTraceContextBenchmarks.newScope_log4j2_extra:·gc.alloc.rate.norm                        sample      15  2680.283 ±   0.027    B/op
CurrentTraceContextBenchmarks.newScope_log4j2_onlyTraceId:newScope_log4j2_onlyTraceId·p0.999   sample            10.544             us/op
CurrentTraceContextBenchmarks.newScope_log4j2_onlyTraceId:·gc.alloc.rate.norm                  sample      15   592.059 ±   0.007    B/op
```

adriancole:
added special casing of single propagation field (ex just traceId or just an extra field), which saves considerable overhead. Bear in mind that when using reactive assembly hooks, scope happens as bad as each operator!

```
CurrentTraceContextBenchmarks.newScope_default:newScope_default·p0.99                          sample              0.095             us/op
CurrentTraceContextBenchmarks.newScope_default:newScope_default·p0.999                         sample              0.824             us/op
CurrentTraceContextBenchmarks.newScope_default:·gc.alloc.rate.norm                             sample      15      0.003 ±   0.001    B/op
CurrentTraceContextBenchmarks.newScope_log4j2:newScope_log4j2·p0.99                            sample              0.239             us/op
CurrentTraceContextBenchmarks.newScope_log4j2:newScope_log4j2·p0.999                           sample              9.104             us/op
CurrentTraceContextBenchmarks.newScope_log4j2:·gc.alloc.rate.norm                              sample      15     56.022 ±   0.002    B/op
CurrentTraceContextBenchmarks.newScope_log4j2_extra:newScope_log4j2_extra·p0.99                sample              3.620             us/op
CurrentTraceContextBenchmarks.newScope_log4j2_extra:newScope_log4j2_extra·p0.999               sample             16.480             us/op
CurrentTraceContextBenchmarks.newScope_log4j2_extra:·gc.alloc.rate.norm                        sample      15   2248.309 ±   0.061    B/op
CurrentTraceContextBenchmarks.newScope_log4j2_onlyExtra:newScope_log4j2_onlyExtra·p0.99        sample              0.974             us/op
CurrentTraceContextBenchmarks.newScope_log4j2_onlyExtra:newScope_log4j2_onlyExtra·p0.999       sample             10.912             us/op
CurrentTraceContextBenchmarks.newScope_log4j2_onlyExtra:·gc.alloc.rate.norm                    sample      15    768.076 ±   0.013    B/op
CurrentTraceContextBenchmarks.newScope_log4j2_onlyTraceId:newScope_log4j2_onlyTraceId·p0.99    sample              0.129             us/op
CurrentTraceContextBenchmarks.newScope_log4j2_onlyTraceId:newScope_log4j2_onlyTraceId·p0.999   sample              3.393             us/op
CurrentTraceContextBenchmarks.newScope_log4j2_onlyTraceId:·gc.alloc.rate.norm                  sample      15     24.007 ±   0.001    B/op
```

PR #1114, opened 2020-03-13
Fixes bugs in B3 single parsing and completes coverage

adriancole:
This hardens the code, as it formerly conflated missing fields or too many fields with truncation in general. It also formerly checked valid characters in multiple places.

It sporadically mentioned an offset, which is not helpful as the source is not logged and is often a reusable buffer. Instead, this mentions the field that had an error as opposed to the string position.

This is preparation for the trace context format as it is a simplification of the B3 single code.

### DIFF
--- a/brave/src/main/java/brave/internal/HexCodec.java
+++ b/brave/src/main/java/brave/internal/HexCodec.java
@@ -1,5 +1,5 @@
 /*
- * Copyright 2013-2019 The OpenZipkin Authors
+ * Copyright 2013-2020 The OpenZipkin Authors
  *
  * Licensed under the Apache License, Version 2.0 (the "License"); you may not use this file except
  * in compliance with the License. You may obtain a copy of the License at
@@ -33,19 +33,36 @@ public final class HexCodec {
   /**
    * Parses a 16 character lower-hex string with no prefix into an unsigned long, starting at the
    * specified index.
+   *
+   * <p>This reads a trace context a sequence potentially larger than the format. The use-case is
+   * reducing garbage, by re-using the input {@code value} across multiple parse operations.
+   *
+   * @param value the sequence that contains a lower-hex encoded unsigned long.
+   * @param beginIndex the inclusive begin index: {@linkplain CharSequence#charAt(int) index} of the
+   * first lower-hex character representing the unsigned long.
    */
-  public static long lowerHexToUnsignedLong(CharSequence lowerHex, int index) {
-    int endIndex = Math.min(index + 16, lowerHex.length());
-    long result = lenientLowerHexToUnsignedLong(lowerHex, index, endIndex);
-    if (result == 0) throw isntLowerHexLong(lowerHex);
+  public static long lowerHexToUnsignedLong(CharSequence value, int beginIndex) {
+    int endIndex = Math.min(beginIndex + 16, value.length());
+    long result = lenientLowerHexToUnsignedLong(value, beginIndex, endIndex);
+    if (result == 0) throw isntLowerHexLong(value);
     return result;
   }
 
-  /** Like {@link #lowerHexToUnsignedLong(CharSequence, int)}, but returns zero on invalid input */
-  public static long lenientLowerHexToUnsignedLong(CharSequence lowerHex, int index, int endIndex) {
+  /**
+   * Like {@link #lowerHexToUnsignedLong(CharSequence, int)}, but returns zero on invalid input.
+   *
+   * @param value the sequence that contains a lower-hex encoded unsigned long.
+   * @param beginIndex the inclusive begin index: {@linkplain CharSequence#charAt(int) index} of the
+   * first lower-hex character representing the unsigned long.
+   * @param endIndex the exclusive end index: {@linkplain CharSequence#charAt(int) index}
+   * <em>after</em> the last lower-hex character representing the unsigned long.
+   */
+  public static long lenientLowerHexToUnsignedLong(CharSequence value, int beginIndex,
+    int endIndex) {
     long result = 0;
-    while (index < endIndex) {
-      char c = lowerHex.charAt(index++);
+    int pos = beginIndex;
+    while (pos < endIndex) {
+      char c = value.charAt(pos++);
       result <<= 4;
       if (c >= '0' && c <= '9') {
         result |= c - '0';
--- a/brave/src/main/java/brave/propagation/B3SingleFormat.java
+++ b/brave/src/main/java/brave/propagation/B3SingleFormat.java
@@ -135,8 +135,8 @@ public final class B3SingleFormat {
   }
 
   @Nullable
-  public static TraceContextOrSamplingFlags parseB3SingleFormat(CharSequence value) {
-    return parseB3SingleFormat(value, 0, value.length());
+  public static TraceContextOrSamplingFlags parseB3SingleFormat(CharSequence b3) {
+    return parseB3SingleFormat(b3, 0, b3.length());
   }
 
   /**
--- a/brave/src/main/java/brave/propagation/B3SingleFormat.java
+++ b/brave/src/main/java/brave/propagation/B3SingleFormat.java
@@ -68,9 +68,9 @@ public final class B3SingleFormat {
    * with the client.
    */
   public static String writeB3SingleFormatWithoutParentId(TraceContext context) {
-    char[] value = getCharBuffer();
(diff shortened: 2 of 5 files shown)


adriancole:
A bit quicker!

Before: quickest run:
```
B3SinglePropagationBenchmarks.extract:extract·p0.999                       sample            9.312            us/op
```

After: quickest run:

```
B3SinglePropagationBenchmarks.extract:extract·p0.999                       sample            6.036            us/op
```

brave/src/main/java/brave/internal/HexCodec.java
```diff
@@ -33,19 +33,36 @@ public static long lowerHexToUnsignedLong(CharSequence lowerHex) {
...
    * specified index.
+   *
+   * <p>This reads a trace context a sequence potentially larger than the format. The use-case is
+   * reducing garbage, by re-using the input {@code value} across multiple parse operations.
+   *
+   * @param value the sequence that contains a lower-hex encoded unsigned long.
+   * @param beginIndex the inclusive begin index: {@linkplain CharSequence#charAt(int) index} of the
+   * first lower-hex character representing the unsigned long.
    */
-  public static long lowerHexToUnsignedLong(CharSequence lowerHex, int index) {
```
adriancole: the sequence isn't necessarily all lowerHex

brave/src/main/java/brave/propagation/B3SingleFormat.java
```diff
@@ -135,106 +135,114 @@ static int writeB3SingleFormat(TraceContext context, long parentId, char[] resul
...
+   *
+   * @param value the sequence that contains a B3 single formatted trace context
+   * @param beginIndex the inclusive begin index: {@linkplain CharSequence#charAt(int) index} of the
+   * first character in B3 single format.
+   * @param endIndex the exclusive end index: {@linkplain CharSequence#charAt(int) index}
+   * <em>after</em> the last character in B3 single format.
    */
   @Nullable
-  public static TraceContextOrSamplingFlags parseB3SingleFormat(CharSequence b3, int beginIndex,
+  public static TraceContextOrSamplingFlags parseB3SingleFormat(CharSequence value, int beginIndex,
```
adriancole: the sequence isn't necessarily all b3 data

brave/src/main/java/brave/propagation/B3SingleFormat.java
```diff
@@ -140,101 +140,109 @@ public static TraceContextOrSamplingFlags parseB3SingleFormat(CharSequence b3) {
...
+    long spanId = tryParseId("span", value, pos, endIndex);
+    if (spanId == 0) return null;
     pos += 16; // spanid
 
     int flags = 0;
     long parentId = 0L;
     if (endIndex > pos) {
-      if (isLowerHex(b3.charAt(pos))) {
-        Platform.get().log("Invalid input: span ID is too long", null);
+      pos++; // consume the delimiter
```
anuraaga:
```suggestion
      pos++; // consume the hyphen
```

brave/src/main/java/brave/propagation/B3SingleFormat.java
```diff
@@ -140,101 +140,109 @@ public static TraceContextOrSamplingFlags parseB3SingleFormat(CharSequence b3) {
...
+
+      if (endIndex == pos) { // traceid-spanid-
+        Platform.get().log("Truncated after reading span ID", null);
```
anuraaga: If the precision in the error messages isn't important, it might be nice to have these checks in the validation loop above so it's all together.

brave/src/main/java/brave/propagation/B3SingleFormat.java
```diff
@@ -140,101 +140,109 @@ public static TraceContextOrSamplingFlags parseB3SingleFormat(CharSequence b3) {
...
       }
-      if (!checkHyphen(b3, pos++)) return null;
 
-      // If our position is at the end of the string, or another delimiter is one character past our
-      // position, try to read sampled status.
-      boolean afterSampledField = notHexFollowsPos(b3, pos, endIndex);
-      if (endIndex == pos + 1 || afterSampledField) {
-        flags = parseFlags(b3, pos);
+      // At this point, we are either at the sampling field, or first character of the parentId
+      int nextPos = pos + 1;
```
anuraaga: My mind is being blown by `nextChar` and `nextPos` not being the same position here. Is it possible to remove `nextPos` for readability?

brave/src/main/java/brave/propagation/B3SingleFormat.java
```diff
@@ -140,101 +140,109 @@ public static TraceContextOrSamplingFlags parseB3SingleFormat(CharSequence b3) {
...
-        pos++; // consume the sampled status
-        if (afterSampledField && !checkHyphen(b3, pos++)) return null; // consume the delimiter
+        if (hasSampledAndParent && value.charAt(pos++) != '-') {
+          Platform.get().log("Invalid input: sampled is too long", null);
+          return null;
+        }
       }
 
-      if (endIndex > pos || afterSampledField) {
+      if (endIndex > nextPos) {
```
anuraaga: Think this is supposed to be `pos` - if not then minutes of staring still hasn't let my brain fully grok `nextPos`

brave/src/main/java/brave/propagation/B3SingleFormat.java
```diff
@@ -251,69 +259,48 @@ public static TraceContextOrSamplingFlags parseB3SingleFormat(CharSequence b3, i
   }
 
   /** Returns zero if truncated, malformed, or too big after logging */
-  static long tryParseParentId(CharSequence b3, int pos, int endIndex) {
-    if (endIndex < pos + 16) {
-      Platform.get().log("Invalid input: truncated", null);
+  static long tryParseId(String name, CharSequence value, int beginIndex, int endIndex) {
+    int endOfId = beginIndex + 16;
+    if (beginIndex == endIndex || value.charAt(beginIndex) == '-') {
```
anuraaga: Not sure if this method is used elsewhere - but at a glance it seems this duplicates some validation with the validation code in `parseB3SingleFormat`. Maybe fine to be safer than sorry though.

brave/src/main/java/brave/propagation/B3SingleFormat.java
adriancole: sure

brave/src/main/java/brave/propagation/B3SingleFormat.java
adriancole: traceID is a special case, as it can be variable length, and the sampled flag is single-character.. I couldn't re-use this logic for all field permutations, but most of them I could!

brave/src/main/java/brave/propagation/B3SingleFormat.java
adriancole: it isn't pos.. anyway I'll try to rename things

adriancole: tried a different approach. PTAL

brave/src/main/java/brave/propagation/B3SingleFormat.java
adriancole: I loosened one log entry to reduce the logic with little loss in fidelity. PTAL!

brave/src/main/java/brave/propagation/B3SingleFormat.java
```diff
@@ -140,102 +140,96 @@ public static TraceContextOrSamplingFlags parseB3SingleFormat(CharSequence b3) {
...
-        return null;
-      }
-      if (!checkHyphen(b3, pos++)) return null;
-
-      // If our position is at the end of the string, or another delimiter is one character past our
-      // position, try to read sampled status.
-      boolean afterSampledField = notHexFollowsPos(b3, pos, endIndex);
-      if (endIndex == pos + 1 || afterSampledField) {
-        flags = parseFlags(b3, pos);
+      if (hyphenCount == 3) { // we should parse sampled AND parent ID
```
anuraaga: Thanks this is very clear now!

brave/src/main/java/brave/propagation/B3SingleFormat.java
adriancole: think we're good here.. only a small loosening made a lot of things possible. good idea

adriancole:
thanks for the review @anuraaga

PR #1115, opened 2020-03-16
Simplifies B3 single by parsing in a single loop

adriancole:
This consolidates logic and removes duplication via the following:
* it is odd that parent ID is sent without sampled flag, assume it isn't
* exploit being in the same package as `SamplingFlags`

Note: the performance isn't better, it is just less code.

### DIFF
--- a/brave-tests/src/main/java/brave/test/propagation/B3SingleFormatClassLoaderTest.java
+++ b/brave-tests/src/main/java/brave/test/propagation/B3SingleFormatClassLoaderTest.java
@@ -1,0 +1,34 @@
+/*
+ * Copyright 2013-2020 The OpenZipkin Authors
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License"); you may not use this file except
+ * in compliance with the License. You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software distributed under the License
+ * is distributed on an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express
+ * or implied. See the License for the specific language governing permissions and limitations under
+ * the License.
+ */
+package brave.test.propagation;
+
+import brave.propagation.B3SingleFormat;
+import org.junit.Test;
+
+import static brave.test.util.ClassLoaders.assertRunIsUnloadable;
+
+public class B3SingleFormatClassLoaderTest {
+  @Test public void unloadable_afterBasicUsage() {
+    assertRunIsUnloadable(BasicUsage.class, getClass().getClassLoader());
+  }
+
+  static class BasicUsage implements Runnable {
+    @Override public void run() {
+      String expected = "1234567812345678-1234567812345678-1";
+      String written =
+        B3SingleFormat.writeB3SingleFormat(B3SingleFormat.parseB3SingleFormat(expected).context());
+      if (!written.equals(expected)) throw new AssertionError();
+    }
+  }
+}
--- a/brave/src/main/java/brave/propagation/B3SingleFormat.java
+++ b/brave/src/main/java/brave/propagation/B3SingleFormat.java
@@ -207,7 +207,7 @@ public final class B3SingleFormat {
           case FIELD_SPAN_ID:
             spanId = buffer;
 
-            // The handle malformed cases like below, it is easier to assume the next field is
+            // To handle malformed cases like below, it is easier to assume the next field is
             // sampled and revert if later not vs peek to determine if it is sampled or parent ID.
             // 'traceId-spanId--parentSpanId'
             // 'traceId-spanId-'
--- a/brave/src/main/java/brave/propagation/B3SingleFormat.java
+++ b/brave/src/main/java/brave/propagation/B3SingleFormat.java
@@ -54,6 +54,7 @@ public final class B3SingleFormat {
   static final int FORMAT_MAX_LENGTH = 32 + 1 + 16 + 3 + 16; // traceid128-spanid-1-parentid
 
   enum Field {
+    TRACE_ID_HIGH("trace ID", 16, false),
     TRACE_ID("trace ID", 16, true),
     SPAN_ID("span ID", 16, true),
     SAMPLED("sampled", 1, false),
@@ -181,7 +182,8 @@ public final class B3SingleFormat {
     long traceIdHigh = 0L, traceId = 0L, spanId = 0L, parentId = 0L;
     int flags = 0;
 
-    Field currentField = Field.TRACE_ID;
+    // Assume it is a 128-bit trace ID and revise back as necessary
+    Field currentField = Field.TRACE_ID_HIGH;
     int currentFieldLength = 0;
     long buffer = 0L;
 
@@ -198,6 +200,9 @@ public final class B3SingleFormat {
           if (isEof && currentFieldLength > 1) {
             currentField = Field.PARENT_SPAN_ID;
           }
+        } else if (currentField == Field.TRACE_ID_HIGH) {
+          // We reached a hyphen before the 17th hex character. This means it is a 64-bit trace ID.
+          currentField = Field.TRACE_ID;
         }
 
         if (!validateFieldLength(currentField, currentFieldLength)) {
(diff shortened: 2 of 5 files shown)


brave/src/main/java/brave/propagation/B3SingleFormat.java
```diff
@@ -158,79 +171,100 @@ public static TraceContextOrSamplingFlags parseB3SingleFormat(CharSequence value
...
+
+          // The handle malformed cases like below, it is easier to assume the next field is sampled
+          // and revert if definitely not vs try to determine if it is definitely sampled.
+          // 'traceId-spanId--parentSpanId'
+          // 'traceId-spanId-'
+          currentField = Field.SAMPLED;
+        } else if (currentField.equals(Field.SAMPLED)) {
+          SamplingFlags samplingFlags = tryParseSamplingFlags(value.charAt(pos -1));
+          if (samplingFlags == null) return null;
+          flags = samplingFlags.flags;
```
adriancole: forgot we were in the same package before. this removes duplicate logic.

brave/src/main/java/brave/propagation/B3SingleFormat.java
```diff
@@ -158,79 +171,98 @@ public static TraceContextOrSamplingFlags parseB3SingleFormat(CharSequence value
...
-      traceIdHigh = lenientLowerHexToUnsignedLong(value, pos, pos + 16);
-      pos += 16; // upper 64 bits of the trace ID
-    } else if (traceIdLength == 16) {
-      traceIdHigh = 0L;
-    } else {
-      Platform.get().log("Invalid input: expected a 16 or 32 lower hex trace ID", null);
-      return null;
-    }
+    // Instead of pos < endIndex, this uses pos <= endIndex to keep field processing consolidated.
+    // Otherwise, we'd have to process again when outside the loop.
```
adriancole:
```suggestion
    // Otherwise, we'd have to process again when outside the loop to handle dangling data on EOF.
```

brave/src/main/java/brave/propagation/B3SingleFormat.java
```diff
@@ -158,79 +171,98 @@ public static TraceContextOrSamplingFlags parseB3SingleFormat(CharSequence value
...
+
+        // This character is the next hex. If it isn't, the next iteration will throw. Either way,
+        // reset so that we can capture the next 16 characters of the trace ID.
+        buffer = 0;
+        currentFieldLength = 0;
       }
+
+      // The rest of this is normal lower-hex decoding
+      buffer <<= 4;
+      if (c >= '0' && c <= '9') {
```
anuraaga: Realize this is the same as current, but if pursuing more benchmarking, maybe worth looking into either having a switch with sixteen cases or a lookup table.

brave/src/main/java/brave/propagation/B3SingleFormat.java
```diff
@@ -158,79 +171,98 @@ public static TraceContextOrSamplingFlags parseB3SingleFormat(CharSequence value
...
+          if (samplingFlags == null) return null;
+          flags = samplingFlags.flags;
+
+          currentField = Field.PARENT_SPAN_ID;
+          currentFieldLength = 0;
+          continue;
         } else {
-          parentId = tryParseParentId(value, endIndex, pos);
-          if (parentId == 0L) return null;
+          parentId = buffer;
```
anuraaga: Think it's helpful to have one of
```java
assert currentField == Field.PARENT_SPAN_ID;
```

```java
// Parent Span ID
```

adriancole:
latest commit p0.999 slightly quicker.. 9.5us instead of over 10us

I've only been paying attention to p0.999 as p0.99 is fast anyway.. currently 0.37us (0.95 is half that)

brave/src/main/java/brave/propagation/B3SingleFormat.java
```diff
@@ -158,79 +171,98 @@ public static TraceContextOrSamplingFlags parseB3SingleFormat(CharSequence value
...
+          // 'traceId-spanId-'
+          currentField = Field.SAMPLED;
+        } else if (currentField.equals(Field.SAMPLED)) {
+          SamplingFlags samplingFlags = tryParseSamplingFlags(value.charAt(pos - 1));
+          if (samplingFlags == null) return null;
+          flags = samplingFlags.flags;
+
+          currentField = Field.PARENT_SPAN_ID;
+          currentFieldLength = 0;
+          continue;
```
anuraaga: Reasoning this `continue` out is tough since it's buried among much code. IIUC, this just avoids a single `buffer = 0L` so maybe not worth special casing?

brave/src/main/java/brave/propagation/B3SingleFormat.java
```diff
@@ -158,79 +171,98 @@ public static TraceContextOrSamplingFlags parseB3SingleFormat(CharSequence value
...
+          if (!isEof) {
+            Platform.get().log("Invalid input: more than 4 fields exist", null);
+            return null;
+          }
         }
+
+        buffer = 0L;
+        currentFieldLength = 0;
+        continue;
+      } else if (currentField == Field.TRACE_ID && beginIndex + 16 == pos) {
```
anuraaga: The above if block is huge so I'd add a comment

```java
  continue;
}

// Not a hyphen

if (currentField == Field.TRACE_ID...

```

brave/src/main/java/brave/propagation/B3SingleFormat.java
adriancole: ooo I like

brave/src/main/java/brave/propagation/B3SingleFormat.java
adriancole: this was fuzz indeed.. good eye!

brave/src/main/java/brave/propagation/B3SingleFormat.java
adriancole: yeah intellij kept getting mad at me for the redundant else if. this is better though

brave/src/main/java/brave/propagation/B3SingleFormat.java
adriancole: good idea

brave/src/main/java/brave/propagation/B3SingleFormat.java
adriancole: running benchmarks again now. I did all the other tricks I could think of along the way

adriancole:
benchmarking switch vs using normal char subtraction for hex decoding.. the latter wins by a lot in the p0.99 on my host. I'm going to revert to the smaller

adriancole:
@anuraaga PTAL the current code is as fast as I can get it. The traceparent variant can be much faster, but I want to base the main/tests on it, so unless you've any more feedback, let's merge!

adriancole:
not hugely smaller, but this is smaller than master's copy:

```
± |master {7} ✓| → ls -al  ./brave/target/classes/brave/propagation/B3SingleFormat*
-rw-r--r--  1 acole  staff  6726 Mar 16 17:37 ./brave/target/classes/brave/propagation/B3SingleFormat.class
± |simpler-b3 {7} ✓| → ls -al  ./brave/target/classes/brave/propagation/B3SingleFormat*
-rw-r--r--  1 acole  staff  6611 Mar 16 17:38 ./brave/target/classes/brave/propagation/B3SingleFormat.class
```

brave-tests/src/main/java/brave/test/propagation/B3SingleFormatClassLoaderTest.java
```diff
@@ -0,0 +1,34 @@
...
+ * the License.
+ */
+package brave.test.propagation;
+
+import brave.propagation.B3SingleFormat;
+import org.junit.Test;
+
+import static brave.test.util.ClassLoaders.assertRunIsUnloadable;
+
+public class B3SingleFormatClassLoaderTest {
```
adriancole: added this because we usually do when we use thread locals

brave/src/main/java/brave/propagation/B3SingleFormat.java
```diff
@@ -158,81 +161,118 @@ public static TraceContextOrSamplingFlags parseB3SingleFormat(CharSequence value
...
+        switch (currentField) {
+          case FIELD_TRACE_ID:
+            traceId = buffer;
+
+            currentField = FIELD_SPAN_ID;
+            break;
+          case FIELD_SPAN_ID:
+            spanId = buffer;
+
+            // The handle malformed cases like below, it is easier to assume the next field is
```
adriancole:
```suggestion
            // To handle malformed cases like below, it is easier to assume the next field is
```

adriancole:
Thanks for all the help @anuraaga !

PR #1116, opened 2020-03-19
Fixes accidental parse fail when trace ID is padded right

adriancole:
Padding the right side of the trace ID with zeros is bad practice, but
it should be permitted. This changes the code and tests to ensure we
don't accidentally fail.

### DIFF
--- a/brave/src/main/java/brave/propagation/TraceContext.java
+++ b/brave/src/main/java/brave/propagation/TraceContext.java
@@ -1,5 +1,5 @@
 /*
- * Copyright 2013-2019 The OpenZipkin Authors
+ * Copyright 2013-2020 The OpenZipkin Authors
  *
  * Licensed under the Apache License, Version 2.0 (the "License"); you may not use this file except
  * in compliance with the License. You may obtain a copy of the License at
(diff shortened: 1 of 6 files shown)


adriancole:
this is a more robust version of the first commit on #693 (will rebase that after)

brave/src/main/java/brave/propagation/TraceContext.java
```diff
@@ -445,21 +445,30 @@ boolean invalidIdLength(Object key, int length, int max) {
...
-    boolean isNull(Object key, String maybeNull) {
+    static boolean isNull(Object key, String maybeNull) {
       if (maybeNull != null) return false;
       Platform.get().log("{0} was null", key, null);
       return true;
     }
 
-    void maybeLogNotLowerHex(String notLowerHex) {
+    /** Helps differentiate a parse failure from a successful parse of all zeros. */
+    static boolean isAllZeros(String value, int beginIndex, int endIndex) {
```
adriancole: later, we could consider sneaky sharing between b3 single format and here, as the parsing requirements are similar

adriancole: I definitely want to refactor this afterwards as it is annoying. for now, just fixing the bug

PR #1117, opened 2020-03-20
Hardens HTTP integration tests to catch more bugs easier

adriancole:
It was very hard to detect subtle bugs in reactor until I redid the
tests to force us to be conscious of whether a span error is expected or
if we are expecting a remote or local span. Further, this makes sure
these two most difficult to debug things end up with better assertion
failures, including the entire span json. This is super handy when
tests flake due to timing. For example, using a debugger could actually
slow things to the point they pass!

### DIFF
--- a/instrumentation/http-tests/src/main/java/brave/test/http/ITHttp.java
+++ b/instrumentation/http-tests/src/main/java/brave/test/http/ITHttp.java
@@ -49,7 +49,8 @@ import static org.assertj.core.api.Assertions.assertThat;
  *   <li>After tests complete, the queue is strictly checked to catch redundant span reporting</li>
  * </ul></pre>
  *
- * <p>As a blocking queue is used, {@link #takeSpan take a span} to perform assertions on it.
+ * <p>As a blocking queue is used, {@linkplain #takeSpan() take a span} to perform assertions on
+ * it.
  *
  * <pre>{@code
  * Span span = takeSpan();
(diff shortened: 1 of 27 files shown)


adriancole:
PS I ran these test against armeria and pass all w/o any changes cc @minwoox @kojilin @anuraaga @trustin

adriancole:
ironically the test flake is a good reason why logging span json is helpful. Sometimes things only fail in CI and you'd not know why:

```
ITTracingAsyncClientHttpRequestInterceptor>ITHttpClient.clientTimestampAndDurationEnclosedByParent:123->ITHttpClient.takeClientSpan:437->ITHttpClient.assertClientSpan:451 Expected span {"traceId":"e7c168c78748af9a","id":"e7c168c78748af9a","name":"test","timestamp":1584667932681036,"duration":12810,"localEndpoint":{"serviceName":"unknown","ipv4":"172.17.0.1"}} to have kind=CLIENT
```

adriancole:
latest commit still passes with armeria

instrumentation/http-tests/src/main/java/brave/test/http/ITHttp.java
```diff
@@ -157,6 +182,84 @@ protected Tracer tracer() {
...
+  protected Span[] assertSpansReportedKindInAnyOrder(@Nullable Span.Kind kind1,
+    @Nullable Span.Kind kind2) throws InterruptedException {
+    if (Objects.equals(kind1, kind2)) {
+      throw new AssertionError("Expected test to pass different span kinds");
+    }
+
+    // Intentionally pull both spans first to ensure neither are errors.
+    Span span1 = takeSpan(), span2 = takeSpan();
+
+    Stream.of(span1, span2).forEach(span -> assertThat(span.kind())
```
anuraaga: Recommend a normal loop with `Arrays.asList` since it's very similar code but there won't be a lambda in the stack trace if failure

adriancole: great idea

PR #1119, opened 2020-03-21
Decouples instrumentation-http-tests from servlet and okhttp 4

adriancole:
This decouples instrumentation-http-tests from servlet as many libraries
don't require it. This particularly helps removes Jetty classpath clash.

Also, this reduces the runtime version of OkHttp from v4 to v3, at the
expense of using a deprecated method.


adriancole:
The motivation for this is re-using our tests in the sleuth project, which has a somewhat difficult classpath due to the number of integrations involved.

PR #1120, opened 2020-03-23
Extracts ITRemote and re-uses in RPC frameworks

adriancole:
This takes the base functionality of ITHttp into ITRemote so that it can
be used in other remote integration tests such as RPC and MongoDB.

This ports only the setup and assertion code to Dubbo and gRPC. There is
a chance to port tests themselves, but it is too effortful to do in a
single change.


adriancole:
will do follow-up on top, to migrate other tests

PR #1121, opened 2020-03-24
Polishes ITRemote and re-uses for JMS testing

adriancole:
This is a part of an overall deduplication of text fixtures and will
continue across all instrumentation.


brave-tests/src/main/java/brave/test/ITRemote.java
```diff
@@ -239,15 +239,39 @@ private Span doTakeSpan(boolean flushed) throws InterruptedException {
     }
   };
 
+  /**
+   * Blocks until a local span was reported. We define a local span as one with a timestamp and
```
adriancole:
```suggestion
   * Blocks until a local span was reported. We define a local span as one with a timestamp and no
```

instrumentation/jms/src/main/java/brave/jms/TracingCompletionListener.java
```diff
@@ -40,23 +40,19 @@ static CompletionListener create(CompletionListener delegate, Span span,
   }
 
   @Override public void onCompletion(Message message) {
-    Scope ws = current.maybeScope(span.context());
```
adriancole: this is just polishing

brave-tests/src/main/java/brave/test/ITRemote.java
```diff
@@ -239,15 +239,39 @@ private Span doTakeSpan(boolean flushed) throws InterruptedException {
...
+   * failure, use {@link #takeLocalSpanWithError(String)} instead.
+   */
   protected Span takeLocalSpan() throws InterruptedException {
     Span local = takeSpan();
+    assertLocalSpan(local);
+    return local;
+  }
+
+  /** Like {@link #takeLocalSpan()} except an error tag must match the given value. */
+  protected Span takeLocalSpanWithError(String errorTag)
```
anuraaga: I wonder if this results in too many permutations. I'm thinking of adding `BraveAssertions` to allow something like

```java
assertThat(takeSpan()).isLocal().hasError("failed");
```

Not something you have to do :) But wondering what you think of the idea.

adriancole: right now there are two permutations, local and remote. there can't be any other permutations.

the explicit handling of error is really really important because until we made error mandatory we actually had bugs we never noticed.

adriancole: it could be the case that some of the comparison operators could fit in an assertions class though, ex childOf parentOf etc

let's just please not take the base preconditions out, that verify this is or is not a local span and is or is not error. deferring these things would re-introduce the problems the code just only for a day has so far solved.

anuraaga: Ah ok makes sense - remembered that `takeSpan` verifies there is no error which does seem better.

instrumentation/jms/src/test/java/brave/jms/ITJms_1_1_TracingMessageProducer.java
```diff
@@ -191,7 +191,7 @@ void should_record_properties(Map<String, String> producerTags) throws Exception
 
   @Test public void should_record_error() throws Exception {
     tracedSession.close();
-    should_record_error(() -> messageProducer.send(message));
+    should_record_error(() -> messageProducer.send(jms.destination, message));
```
adriancole: @anuraaga for example, the error assertion showed that the earlier tests were bad syntax. the error returned had nothing to do with the session that was caused. it is really important to not undo this as there are still a boatload of hidden bugs to find.

brave-tests/src/main/java/brave/test/ITRemote.java
adriancole: the other thing is accidental permutation. we had tests that thought they were acting on a local span when they were acting on a remote one.. I tried to limit most of this by using scopes instead of parent spans

adriancole:
@anuraaga fyi I just added a commit as there was another permutation which is that annoying one-way span thing (we should never implement this again now that we have messaging spans). It is used in dubbo so less confusing imho if pinned there.

anuraaga:
Thanks!

adriancole:
The other thing I had in mind also, to save you some time in case you hadn't..

is that brave has in the past had problems being dependent on zipkin types. If we are to do assertion helper, it should probably be for MutableSpan. Once all the tests are organized, we can change the base test to be a queue of MutableSpan (fed by a FinishedSpanHandler as opposed to a zipkin reporter). Especially in this case, it would be helpful to have an assertion helper as MutableSpan isn't as nice to use as Span is.

PR #1123, opened 2020-03-25
Ports Kafka tests to ITRemote and fixes propagation bug

adriancole:
Note: there's significant opportunity for deduplication in the kafka tests right now. However, I'd like to conserve energy as some of it would be magically deduplicated if we do that messaging abstraction. It is also possible to do more deduping offline.

cc @jeqo @jorgheymans

### DIFF
--- a/instrumentation/kafka-clients/src/main/java/brave/kafka/clients/KafkaTracing.java
+++ b/instrumentation/kafka-clients/src/main/java/brave/kafka/clients/KafkaTracing.java
@@ -1,5 +1,5 @@
 /*
- * Copyright 2013-2019 The OpenZipkin Authors
+ * Copyright 2013-2020 The OpenZipkin Authors
  *
  * Licensed under the Apache License, Version 2.0 (the "License"); you may not use this file except
  * in compliance with the License. You may obtain a copy of the License at
@@ -25,8 +25,10 @@ import brave.propagation.TraceContext.Extractor;
 import brave.propagation.TraceContext.Injector;
 import brave.propagation.TraceContextOrSamplingFlags;
 import brave.sampler.SamplerFunction;
+import java.util.Collections;
 import java.util.Iterator;
 import java.util.LinkedHashSet;
+import java.util.Map;
 import java.util.Set;
 import org.apache.kafka.clients.consumer.Consumer;
 import org.apache.kafka.clients.consumer.ConsumerRecord;
@@ -118,8 +120,9 @@ public final class KafkaTracing {
   final Extractor<Headers> processorExtractor;
   final Injector<KafkaProducerRequest> producerInjector;
   final Injector<KafkaConsumerRequest> consumerInjector;
-  final SamplerFunction<MessagingRequest> producerSampler, consumerSampler;
   final Set<String> propagationKeys;
+  final TraceContextOrSamplingFlags emptyExtraction;
+  final SamplerFunction<MessagingRequest> producerSampler, consumerSampler;
   final String remoteServiceName;
   final boolean singleRootSpanOnReceiveBatch;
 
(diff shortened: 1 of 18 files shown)


instrumentation/kafka-clients/src/main/java/brave/kafka/clients/KafkaTracing.java
```diff
@@ -132,9 +135,12 @@ public KafkaTracing build() {
     this.processorExtractor = propagation.extractor(KafkaPropagation.GETTER);
     this.producerInjector = propagation.injector(KafkaProducerRequest::setHeader);
     this.consumerInjector = propagation.injector(KafkaConsumerRequest::setHeader);
+    this.propagationKeys = new LinkedHashSet<>(propagation.keys());
+    // When Extra Fields or similar are in use, the result != TraceContextOrSamplingFlags.EMPTY
+    this.emptyExtraction = propagation.<Map<String, String>>extractor(Map::get)
```
adriancole: this was a bug.. I searched for all cases we (like me) did incorrect comparisons against EMPTY

instrumentation/kafka-clients/src/test/java/brave/kafka/clients/ITKafkaTracing.java
```diff
@@ -45,97 +40,57 @@
...
-    .spanReporter(producerSpans::add)
-    .build());
-
+public class ITKafkaTracing extends ITKafka {
   @ClassRule
   public static KafkaJunitRule kafkaRule = new KafkaJunitRule(EphemeralKafkaBroker.create());
-  @Rule
-  public TestName testName = new TestName();
+
+  @Rule public TestSpanReporter producerReporter = new TestSpanReporter();
```
adriancole: @anuraaga this is using multiple reporters.. indeed efforts like this are piecemeal as things come up. Mainly because it is so much work to refactor all the tests across all modules. Once things settle, and things stop changing, we can think hard about cleaner design with the benefit of seeing warts bit by bit.

instrumentation/kafka-clients/src/test/java/brave/kafka/clients/ITKafkaTracing.java
```diff
@@ -331,19 +301,19 @@ void checkB3Unsampled(ConsumerRecords<String, String> records) {
...
+  @Test public void customSampler_consumer() {
     String topic = testName.getMethodName();
 
-    consumerTracing = KafkaTracing.create(MessagingTracing.newBuilder(
-      Tracing.newBuilder().spanReporter(consumerSpans::add).build()
-    ).consumerSampler(MessagingRuleSampler.newBuilder()
-      .putRule(operationEquals("receive"), Sampler.NEVER_SAMPLE)
-      .build()).build());
+    consumerTracing = KafkaTracing.create(
+      MessagingTracing.newBuilder(consumerTracing.messagingTracing.tracing())
```
adriancole: this sort of stuff looks gross now, but when we complete the messaging abstraction, should end up contained.

instrumentation/kafka-clients/src/main/java/brave/kafka/clients/TracingConsumer.java
```diff
@@ -98,7 +101,7 @@
 
         // If we extracted neither a trace context, nor request-scoped data (extra),
         // and sharing trace is enabled make or reuse a span for this topic
-        if (extracted.equals(TraceContextOrSamplingFlags.EMPTY) && singleRootSpanOnReceiveBatch) {
+        if (extracted.equals(emptyExtraction) && singleRootSpanOnReceiveBatch) {
```
adriancole: this is the code that tests now caught as a bug. Basically, when we had an empty incoming context, we'd still split into traces for each message. The reason is that we don't have an abstract notion to say if "extra" are empty or not. Meanwhile, the trick to just make an empty context works fine.

instrumentation/kafka-clients/src/main/java/brave/kafka/clients/KafkaTracing.java
anuraaga: Does this sort of bug only affect Kafka? If the pattern is generic enough, one idea is something like `Propagation.isEmpty(TraceContextOrSamplingFlags)` so we don't need to repeat this logic as much.

adriancole:
> Does this sort of bug only affect Kafka? If the pattern is generic enough, one idea is something like Propagation.isEmpty(TraceContextOrSamplingFlags) so we don't need to repeat this logic as much.

I've looked to verify but only messaging is playing games like this. When the messaging abstraction happens, this will be hidden inside there ideally only once.

If we find this more general, we can add a method publicly. I only hesitate now as Propagation already is showing warts and I don't want to add more weird cases. I think we can revisit this when the trace context and correlation fields are in, as there might be other weird glitches as more things layer together.

adriancole:
thanks for the look @anuraaga! I'm trucking along to the next stuff in order to keep willpower and momentum going.

PR #1125, opened 2020-03-26
Cleans up MongoDB for release

adriancole:
* Removes unused dependency on async
* Removes irrelevant dependency on reactivestreams
* Adds note to vote for async
* Removes toBuilder
* Removes unnecessary null defense on command failed
* Backports tests to ITRemote
* Makes tags checking more strict, notably with cluster ID
* Adjusts formatting
* Fixes pom versions


instrumentation/mongodb/src/main/java/brave/mongodb/TraceMongoCommandListener.java
```diff
@@ -104,8 +104,7 @@
   @Override public void commandFailed(CommandFailedEvent event) {
     Span span = threadLocalSpan.remove();
     if (span == null) return;
-    Throwable throwable = event.getThrowable();
-    span.error(throwable == null ? new MongoException("Command failed but no throwable was reported") : throwable);
```
adriancole: I checked every reference in mongo's driver. My guess is that this guard was added because of mocks. I've noticed some irrelevant null defense elsewhere when mocks are used primarily.

PR #1127, opened 2020-03-26
Makes StrictScopeDecorator stricter without renaming it

adriancole:
This fixes some fuzz in tests, but notably makes `StrictScopeDecorator`
`Closeable`. In doing so, this will expose any instrumentation that
leaked scopes as opposed to before which could only show if something
closed on the wrong thread.

This type was only ever documented for development, so there should be
no harm in updating this in a way that could OOM in extreme leak
circumstances (due to hard references).

### DIFF
--- a/brave-tests/src/main/java/brave/test/ITRemote.java
+++ b/brave-tests/src/main/java/brave/test/ITRemote.java
@@ -15,7 +15,6 @@ package brave.test;
 
 import brave.Tracing;
 import brave.propagation.B3Propagation;
-import brave.propagation.CurrentTraceContext;
 import brave.propagation.ExtraFieldPropagation;
 import brave.propagation.Propagation;
 import brave.propagation.SamplingFlags;
@@ -45,6 +44,8 @@ import static org.assertj.core.api.Assertions.assertThat;
  * </ul></pre>
  */
 public abstract class ITRemote {
+  public static final String EXTRA_KEY = "user-id";
+
   /**
    * We use a global rule instead of surefire config as this could be executed in gradle, sbt, etc.
    * This way, there's visibility on which method hung without asking the end users to edit build
@@ -56,8 +57,6 @@ public abstract class ITRemote {
   @Rule public TestRule globalTimeout = new DisableOnDebug(Timeout.seconds(20)); // max per method
   @Rule public TestSpanReporter reporter = new TestSpanReporter();
   @Rule public TestName testName = new TestName();
-
-  public static final String EXTRA_KEY = "user-id";
 
   /** Returns a trace context for use in propagation tests. */
   protected TraceContext newTraceContext(SamplingFlags flags) {
@@ -104,6 +103,11 @@ public abstract class ITRemote {
   @After public void close() throws Exception {
     Tracing current = Tracing.current();
     if (current != null) current.close();
+    checkForLeakedScopes();
+  }
+
+  /** Override to control scope leak enforcement. */
+  protected void checkForLeakedScopes() {
     strictScopeDecorator.close();
   }
 
--- a/brave-tests/src/main/java/brave/test/ITRemote.java
+++ b/brave-tests/src/main/java/brave/test/ITRemote.java
@@ -24,6 +24,7 @@ import brave.propagation.ThreadLocalCurrentTraceContext;
 import brave.propagation.TraceContext;
 import brave.sampler.Sampler;
 import org.junit.After;
+import org.junit.Before;
 import org.junit.Rule;
 import org.junit.rules.DisableOnDebug;
 import org.junit.rules.TestName;
@@ -73,7 +74,7 @@ public abstract class ITRemote {
   final StrictScopeDecorator strictScopeDecorator = StrictScopeDecorator.create();
 
   // field because this allows subclasses to initialize a field Tracing
-  protected final CurrentTraceContext currentTraceContext =
+  protected final ThreadLocalCurrentTraceContext currentTraceContext =
     ThreadLocalCurrentTraceContext.newBuilder()
       .addScopeDecorator(strictScopeDecorator)
       .build();
@@ -89,6 +90,11 @@ public abstract class ITRemote {
       .propagationFactory(propagationFactory)
       .currentTraceContext(currentTraceContext)
       .sampler(sampler);
+  }
+
+  /** Ensure any leaks present were caused by code invoked in the current method */
+  @Before public void clearThreadLocal() {
+    currentTraceContext.clear();
   }
 
   /**
--- a/brave-tests/src/main/java/brave/test/ITRemote.java
+++ b/brave-tests/src/main/java/brave/test/ITRemote.java
@@ -98,7 +98,20 @@ public abstract class ITRemote {
 
   /**
    * This closes the current instance of tracing, to prevent it from being accidentally visible to
-   * other test classes which call {@link Tracing#current()}. It also checks for scope leaks!
+   * other test classes which call {@link Tracing#current()}.
+   *
+   * <p>This also checks for scope leaks. It is important that you have closed all resources prior
+   * to this method call. Otherwise, in-flight request cleanup may be mistaken for scope leaks. This
+   * may involve blocking on completion, if using executors.
+   *
+   * <p>Ex.
+   * <pre>{@code
+   * @After @Override public void close() throws Exception {
+   *   executorService.shutdown();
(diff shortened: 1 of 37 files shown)


adriancole:
@marcingrzejszczak @trustin @minwoox @ikhoon @kojilin

this will help sniff down code that leaks and causes little bombs in CI (note that in armeria it "shouldn't" be needed as scopes are owned by the incoming server request, and therefore should be cleaned up always)

brave-tests/src/main/java/brave/test/ITRemote.java
```diff
@@ -60,19 +60,22 @@
 
   /** Returns a trace context for use in propagation tests. */
   protected TraceContext newTraceContext(SamplingFlags flags) {
+    long id = System.nanoTime(); // Random enough as tests are run serially anyway
     TraceContext result = TraceContext.newBuilder()
-      .traceIdHigh(1L).traceId(2L).parentId(3L).spanId(1L)
+      .traceIdHigh(id).traceId(id + 1).parentId(id + 2).spanId(id + 3)
```
adriancole: I made a bug somewhere where I called this twice and expected different results :)

brave/src/main/java/brave/propagation/StrictScopeDecorator.java
```diff
@@ -30,38 +35,66 @@
...
+  /**
+   * Identifies problems by throwing {@link IllegalStateException} when a scope is closed on a
+   * different thread.
+   */
+  @Override public Scope decorateScope(@Nullable TraceContext context, Scope scope) {
+    return new StrictScope(scope, new CallerStackTrace(context), currentCallers);
+  }
+
+  /**
+   * @throws AssertionError if any scopes were left unclosed.
```
adriancole:
```suggestion
   * @throws IllegalStateException if any scopes were left unclosed.
```

brave/src/main/java/brave/propagation/StrictScopeDecorator.java
```diff
@@ -30,38 +35,66 @@
  *                  ).build();
  * }</pre>
  */
-public final class StrictScopeDecorator implements ScopeDecorator {
-  public static ScopeDecorator create() {
+// Closeable so things like Spring will automatically execute it on shutdown and expose leaks!
```
adriancole: This is a neat trick.. Let's say some instrumentation you didn't write might be leaking a scope. If you add a bean of this type to an app with sleuth.. when you ctrl-c it *should* expose any leak problem automatically.

adriancole:
Interestingly this caught a flake in the spring web interceptor. However, the build logs don't say the cause.. probably config

```
[ERROR] brave.spring.web.ITTracingAsyncClientHttpRequestInterceptor.propagatesChildOfCurrentSpan  Time elapsed: 0.047 s  <<< ERROR!

java.lang.IllegalStateException: Thread [I/O dispatcher 39] opened scope, but never closed it

Caused by: brave.propagation.StrictScopeDecorator$CallerStackTrace: Thread [I/O dispatcher 39] opened scope for 000000694274a933000000694274a934/000000694274a936 here:
```

brave/src/main/java/brave/propagation/StrictScopeDecorator.java
```diff
@@ -30,38 +35,66 @@
...
       this.caller = caller;
+      this.currentCallers.add(caller);
     }
 
     @Override public void close() {
-      if (Thread.currentThread().getId() != threadId) {
-        throw new IllegalStateException(
-          "scope closed in a different thread: " + Thread.currentThread().getName(),
-          caller);
+      currentCallers.remove(caller);
```
anuraaga: Probably don't care about performance of this but just in case, can consider using a `LinkedBlockingDeque` to just remove the last element instead of iterating to the end.

adriancole: I so wanted to, but that's Java 1.7 plus (animal sniffer found this)

adriancole:
@anuraaga PTAL I redid some code because the caller stack trace wasn't helpful

adriancole:
I manually checked this by leaking a scope in a test controller
```
java.lang.IllegalStateException: Thread [reactor-http-nio-2] leaked a scope of 1809a0789e30d524/63ce355ef96d3d8b here:

	at org.springframework.cloud.sleuth.instrument.web.WebFluxController.foo(WebFluxController.java:72)
	at sun.reflect.NativeMethodAccessorImpl.invoke0(Native Method)
	at sun.reflect.NativeMethodAccessorImpl.invoke(NativeMethodAccessorImpl.java:62)
	at sun.reflect.DelegatingMethodAccessorImpl.invoke(DelegatingMethodAccessorImpl.java:43)
	at java.lang.reflect.Method.invoke(Method.java:498)
	at org.springframework.web.reactive.result.method.InvocableHandlerMethod.lambda$invoke$0(InvocableHandlerMethod.java:147)
	at reactor.core.publisher.MonoFlatMap$FlatMapMain.onNext(MonoFlatMap.java:118)
```

adriancole:
TracingAsyncClientHttpRequestInterceptor is leaky only in CI. I can't see the stack trace, so added 

```xml
          <trimStackTrace>false</trimStackTrace>
```

to surefire and failsafe

brave/src/main/java/brave/propagation/ThreadLocalCurrentTraceContext.java
```diff
@@ -42,13 +42,27 @@ public static CurrentTraceContext create() {
...
 
-  public static CurrentTraceContext.Builder newBuilder() {
+  public static Builder newBuilder() {
     return new Builder();
   }
 
-  static final class Builder extends CurrentTraceContext.Builder {
+  /**
+   * This component is backed by a possibly static shared thread local. Call this to clear the
+   * reference when you are sure any residual state is due to a leak.
```
anuraaga: Is it worth adding something like "This is generally only useful in tests"?

adriancole: sure

adriancole: well it is being done in a hack way in sleuth...

here 
https://github.com/spring-cloud/spring-cloud-sleuth/blob/master/spring-cloud-sleuth-core/src/main/java/org/springframework/cloud/sleuth/instrument/web/TraceWebFilter.java#L123-L126

and also here

https://github.com/spring-cloud/spring-cloud-sleuth/blob/master/spring-cloud-sleuth-core/src/main/java/org/springframework/cloud/sleuth/instrument/scheduling/TraceSchedulingAspect.java#L63-L65

While I can't say either are correct, it might be that some bad instrumentation (ex written by an end user) leaks, and someone wants to correct it.

adriancole: regardless, your statement is correct.. this should (not must) be only used in tests.

adriancole:
finally some good output!

```
[ERROR] brave.spring.web.ITTracingAsyncClientHttpRequestInterceptor.propagatesUnsampledContext  Time elapsed: 0.02 s  <<< FAILURE!

java.lang.AssertionError: Thread [I/O dispatcher 19] leaked a scope of 0000006b7024e01f0000006b7024e020/0000006b7024e022 here:

	at brave.spring.web.TraceContextListenableFuture$TraceContextListenableFutureCallback.onSuccess(TraceContextListenableFuture.java:107)

	at org.springframework.util.concurrent.ListenableFutureCallbackRegistry.notifySuccess(ListenableFutureCallbackRegistry.java:76)

	at org.springframework.util.concurrent.ListenableFutureCallbackRegistry.success(ListenableFutureCallbackRegistry.java:142)

	at org.springframework.http.client.HttpComponentsAsyncClientHttpRequest$HttpResponseFutureCallback.completed(HttpComponentsAsyncClientHttpRequest.java:127)

	at org.springframework.http.client.HttpComponentsAsyncClientHttpRequest$HttpResponseFutureCallback.completed(HttpComponentsAsyncClientHttpRequest.java:102)

	at org.apache.http.concurrent.BasicFuture.completed(BasicFuture.java:122)
```

adriancole:
this is turning into a joy of finding all the implicit dispatchers and making them stop :) 

basically in-flight callback hooks can be mistaken for leaked stuff depending on a race on close()

I'm putting this down for the night but will come back tomorrow.

adriancole:
thanks for the review and suggestions @anuraaga

PR #1128, opened 2020-03-27
Works around possibility to have an empty set of extra fields

adriancole:
We didn't eagerly validate `ExtraFieldPropagation`, that it had any
fields at all. This leads to late bugs where an exception is thrown for
trying to inject nothing.

As they type `ExtraFieldPropagation` has public methods in use, this
doesn't change signatures. Rather, it creates a wrapper in the case
nothing exists.

WIP as some unit tests are due.

### DIFF
--- a/brave/src/main/java/brave/propagation/ExtraFieldPropagation.java
+++ b/brave/src/main/java/brave/propagation/ExtraFieldPropagation.java
@@ -97,13 +97,12 @@ import java.util.Set;
  * String countryCode = ExtraFieldPropagation.get(span.context(), "country-code");
  * }</pre>
  */
-public final class ExtraFieldPropagation<K> implements Propagation<K> {
+public class ExtraFieldPropagation<K> implements Propagation<K> {
   /** Wraps an underlying propagation implementation, pushing one or more fields */
   public static Factory newFactory(Propagation.Factory delegate, String... fieldNames) {
     if (delegate == null) throw new NullPointerException("delegate == null");
     if (fieldNames == null) throw new NullPointerException("fieldNames == null");
-    String[] validated = ensureLowerCase(new LinkedHashSet<>(Arrays.asList(fieldNames)));
-    return new Factory(delegate, validated, validated, new BitSet());
+    return newFactory(delegate, Arrays.asList(fieldNames));
   }
 
   /** Wraps an underlying propagation implementation, pushing one or more fields */
@@ -112,7 +111,8 @@ public final class ExtraFieldPropagation<K> implements Propagation<K> {
     if (delegate == null) throw new NullPointerException("delegate == null");
     if (fieldNames == null) throw new NullPointerException("fieldNames == null");
     String[] validated = ensureLowerCase(new LinkedHashSet<>(fieldNames));
-    return new Factory(delegate, validated, validated, new BitSet());
+    return validated.length == 0
+      ? new Factory(delegate) : new RealFactory(delegate, validated, validated, new BitSet());
   }
 
   public static FactoryBuilder newFactoryBuilder(Propagation.Factory delegate) {
@@ -163,6 +163,7 @@ public final class ExtraFieldPropagation<K> implements Propagation<K> {
       return this;
     }
 
+    /** Returns a wrapper of the delegate if there are no fields to propagate. */
     public Factory build() {
       BitSet redacted = new BitSet();
       List<String> fields = new ArrayList<>(), keys = new ArrayList<>();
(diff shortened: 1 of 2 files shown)


brave/src/main/java/brave/propagation/ExtraFieldPropagation.java
```diff
@@ -272,15 +275,39 @@ public static void set(TraceContext context, String name, String value) {
...
 
-  public static final class Factory extends Propagation.Factory {
+  public static class Factory extends Propagation.Factory {
     final Propagation.Factory delegate;
+
+    Factory(Propagation.Factory delegate) {
+      this.delegate = delegate;
+    }
+
+    @Override public <K> ExtraFieldPropagation<K> create(Propagation.KeyFactory<K> keyFactory) {
```
adriancole: this signature is retained on empty

brave/src/main/java/brave/propagation/ExtraFieldPropagation.java
```diff
@@ -348,7 +360,7 @@ public static void set(TraceContext context, String name, String value) {
   // This is here to support extraction from carriers missing a get field by name function. The only
   // known example is OpenTracing TextMap https://github.com/opentracing/opentracing-java/issues/305
   public List<K> extraKeys() {
```
adriancole: this signature is retained on empty

brave/src/main/java/brave/propagation/ExtraFieldPropagation.java
```diff
@@ -163,6 +163,7 @@ public FactoryBuilder addPrefixedFields(String prefix, Collection<String> fieldN
       return this;
     }
 
+    /** Returns a wrapper of the delegate if there are no fields to propagate. */
     public Factory build() {
```
adriancole: this signature is retained on empty

adriancole:
sorry here was the original stack trace. If you use extract (implicitly via http or messaging), it would throw like below, sadly per-request. Eventhough the config is easy to correct, it is dangerous.

```
Caused by: java.lang.NullPointerException: fieldNames is empty
	at brave.internal.PredefinedPropagationFields.<init>(PredefinedPropagationFields.java:28)
	at brave.propagation.ExtraFieldPropagation$Extra.<init>(ExtraFieldPropagation.java:464)
	at brave.propagation.ExtraFieldPropagation$ExtraFactory.create(ExtraFieldPropagation.java:450)
	at brave.propagation.ExtraFieldPropagation$ExtraFieldExtractor.extract(ExtraFieldPropagation.java:414)
```

brave/src/main/java/brave/propagation/ExtraFieldPropagation.java
```diff
@@ -272,15 +275,39 @@ public static void set(TraceContext context, String name, String value) {
...
+    @Override public boolean requires128BitTraceId() {
+      return delegate.requires128BitTraceId();
+    }
+
+    @Override public TraceContext decorate(TraceContext context) {
+      return delegate.decorate(context);
+    }
+  }
+
+  public static final class RealFactory extends Factory {
```
anuraaga: I believe we can keep this class package-private

brave/src/main/java/brave/propagation/ExtraFieldPropagation.java
```diff
@@ -199,8 +200,10 @@ public Factory build() {
       for (i = 0; i < keyToField.length; i++) {
         keyToField[i] = keyToFieldList.get(i);
       }
-      return new Factory(delegate, fields.toArray(new String[0]), keys.toArray(new String[0]),
-        keyToField, redacted);
+      String[] validated = fields.toArray(new String[0]);
```
anuraaga: Since we're not doing any extra validation here this is confusing to me. How about

```java
if (fields.isEmpty()) return new Factory(delegate);
return new RealFactory(delegate, fields.toArray(...
```

brave/src/main/java/brave/propagation/ExtraFieldPropagation.java
adriancole: good catch!

brave/src/main/java/brave/propagation/ExtraFieldPropagation.java
adriancole: sgtm

adriancole:
thanks for the help @anuraaga

brave/src/test/java/brave/propagation/ExtraFieldPropagationTest.java
```diff
@@ -105,12 +110,58 @@
...
+  // We formerly enforced presence of field names in the factory's factory method
+  @Test public void noFields_newFactory_disallowed() {
+    assertThatThrownBy(() -> ExtraFieldPropagation.newFactory(B3Propagation.FACTORY))
+      .hasMessage("no field names");
+
+    assertThatThrownBy(() -> ExtraFieldPropagation.newFactory(B3Propagation.FACTORY, asList()))
+      .hasMessage("no field names");
+  }
+
+  // We formerly accepted .build() when no fields were present
```
adriancole: @meltsufin  this bud's for you

regardless of if extra fields are added to the builder or not, behavior will be predictable

meltsufin: Thanks! Much appreciated! 😃

PR #1129, opened 2020-03-28
Makes ITHttpAsyncClient accept a normal type

adriancole:


instrumentation/http-tests/src/main/java/brave/test/http/ITHttpAsyncClient.java
```diff
@@ -36,9 +37,11 @@
    *
    * <p>One of success or failure callbacks must be invoked even on unexpected scenarios. For
```
adriancole:
```suggestion
   * <p>Invoke the callback with one of success or failure even in unexpected scenarios. For
```

PR #1132, opened 2020-03-31
Actually don't check SingletonList as comment says.

anuraaga:
Noticed implementation does not reflect the comment "Faster to make a copy than check the type to see if it is already a singleton list". I haven't actually benchmarked this but it seems to make sense. Alternatively I can revert the code changes and remove the comment, let me know.

Additionally optimizes for an empty mutable list parameter.

### DIFF
--- a/brave/src/main/java/brave/internal/Lists.java
+++ b/brave/src/main/java/brave/internal/Lists.java
@@ -1,5 +1,5 @@
 /*
- * Copyright 2013-2019 The OpenZipkin Authors
+ * Copyright 2013-2020 The OpenZipkin Authors
  *
  * Licensed under the Apache License, Version 2.0 (the "License"); you may not use this file except
  * in compliance with the License. You may obtain a copy of the License at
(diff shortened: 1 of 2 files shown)


brave/src/main/java/brave/internal/Lists.java
```diff
@@ -29,19 +29,19 @@
...
+    if (isImmutable(list)) return list;
+    return Collections.unmodifiableList(new ArrayList<>(list));
   }
 
   static boolean isImmutable(List<Object> extra) {
-    if (extra == Collections.EMPTY_LIST) return true;
+    if (extra == Collections.emptyList()) return true;
     // avoid copying datastructure by trusting certain names.
     String simpleName = extra.getClass().getSimpleName();
-    return simpleName.equals("SingletonList")
```
adriancole: that SingletonList isn't here hurts my brain because it is a well known type. that said I understand why you are doing this.. it is under the assumption nothing else is using this method which is fair enough. It might hurt brain less to say intentionally not checking SingletonList due as the only call site wont' have it

anuraaga: Good call, added a comment both for that and because I realized the empty list check was similarly redundant.

adriancole:
thanks!

PR #1133, opened 2020-04-01
Adds Tag, Tags and HttpTags

adriancole:
This adds a long overdue feature to ease support of tagging spans. `Tag`
bakes in all logic needed to add a tag to a `Span`, `ScopedSpan`,
`SpanCustomizer` or `MutableSpan` leaving the user left only to decide
what the key is and how to parse it.

This also takes care of error handling and obviation of parsing into a
no-op span.

Here's an example of a potentially expensive tag:
```java
SUMMARY_TAG = new Tag<Summarizer>("summary") {
  @Override protected String parseValue(Summarizer input, TraceContext context) {
    return input.computeSummary();
  }
}

// This works for any variant of span
SUMMARY_TAG.tag(summarizer, span);
```

We need to solve a common case for `BaggageField` which is to add that
as a span tag. This is the best time to fix this problem.

The closest type we have is `ErrorParser` as that does a similar
dispatch. Externally, the closest is [OpenTracing Tag](https://github.com/opentracing/opentracing-java/blob/master/opentracing-api/src/main/java/io/opentracing/tag/Tag.java).

Our approach is different notably as we seal everything except how to
parse the value. This helps avoid leaking exceptions or accidentally
adding null values.

### DIFF
--- a/brave/src/main/java/brave/Tag.java
+++ b/brave/src/main/java/brave/Tag.java
@@ -1,0 +1,140 @@
+/*
+ * Copyright 2013-2020 The OpenZipkin Authors
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License"); you may not use this file except
+ * in compliance with the License. You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software distributed under the License
+ * is distributed on an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express
+ * or implied. See the License for the specific language governing permissions and limitations under
+ * the License.
+ */
+package brave;
+
+import brave.handler.FinishedSpanHandler;
+import brave.handler.MutableSpan;
+import brave.internal.Nullable;
+import brave.internal.Platform;
+import brave.propagation.TraceContext;
+
+import static brave.internal.Throwables.propagateIfFatal;
+
+/**
+ * This is a centralized type to parse a tag into any variant of a span. This also avoids the
+ * clutter of checking null or guarding on exceptions.
+ *
+ * Here's an example of a potentially expensive tag:
+ * <pre>{@code
+ * SUMMARY_TAG = new Tag<Summarizer>("summary") {
+ *   @Override protected String parseValue(Summarizer input, TraceContext context) {
+ *     return input.computeSummary();
+ *   }
+ * }
+ * SUMMARY_TAG.tag(span);
+ * }</pre>
+ *
+ * @see Tags
+ * @see SpanCustomizer#tag(String, String)
+ * @see ScopedSpan#tag(String, String)
+ * @see MutableSpan#tag(String, String)
+ * @since 5.11
+ */
+public abstract class Tag<I> {
+  public final String key() {
+    return key;
+  }
+
+  /**
+   * Override to change what data from the input are parsed into the span modeling it. Any
+   * exceptions will be logged and ignored.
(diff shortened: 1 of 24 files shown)


brave/src/main/java/brave/Tag.java
```diff
@@ -0,0 +1,154 @@
...
+public abstract class Tag<I> {
+  public final String key() {
+    return key;
+  }
+
+  /**
+   * Override to change what data from the input are parsed into the span modeling it. Any
+   * exceptions will be logged and ignored.
+   *
+   * @return possibly nullable result. Note: empty string is a valid tag value!
```
anuraaga, line 53: I guess it's either `Nullable result` or `Possibly null result`. Though maybe better to phrase something like `The result to add as a span tag. {@code null} means no tag will be added.`

brave/src/main/java/brave/Tag.java
```diff
@@ -0,0 +1,154 @@
...
+   * @since 5.11
+   */
+  @Nullable protected abstract String parseValue(I input, @Nullable TraceContext context);
+
+  /**
+   * {@linkplain ScopedSpan#tag(String, String) Tags} the value parsed from the {@code input} when
```
anuraaga, line 59: Maybe can leave off these docs at `Tags the value parsed from the {@code input}.`. The non-null is really a job of parseValue, and there's no way to tag a no-op span anyways, the key point is we don't waste time parsing when it's no-op but that's an implementation detail.

brave/src/main/java/brave/Tag.java
```diff
@@ -0,0 +1,154 @@
...
+    String value;
+    try {
+      value = parseValue(input, context);
+    } catch (Throwable e) {
+      propagateIfFatal(e);
+      Platform.get().log("Error parsing tag value of input %s", input, e);
+      return;
+    }
+    if (value == null) return;
+    if (span instanceof SpanCustomizer) {
```
anuraaga, line 139: I'm sure it must have been thought no need for detail - any thoughts on a shared interface?

anuraaga:
The test flaked for my Lists PR too btw, nothing was obvious and haven't had a chance to look more.

brave/src/main/java/brave/Tag.java
adriancole: sgtm

brave/src/main/java/brave/Tag.java
adriancole: cool

brave/src/main/java/brave/Tag.java
adriancole: sure. here's the context

Span is indeed instanceof SpanCustomizer
ScopedSpan doesn't allow name overrides
MutableSpan doesn't have an implicit timestamp

adriancole: we could allow ScopedSpan to override its name.. now wouldn't be a bad time to do that..

adriancole: https://github.com/openzipkin/brave/pull/1134

brave/src/main/java/brave/Tag.java
```diff
@@ -0,0 +1,154 @@
...
+   * @since 5.11
+   */
+  @Nullable protected abstract String parseValue(I input, @Nullable TraceContext context);
+
+  /**
+   * Tags the value parsed from the {@code input}.
+   *
+   * @since 5.11
+   */
+  public final void tag(I input, ScopedSpan span) {
```
adriancole, line 63: I left these overloads to help reduce branching to get noop

PR #1136, opened 2020-04-02
Allows correlation fields to not match `BaggageField` names

adriancole:
This helps port code which formerly added different naming conventions
for span IDs.

Ex. If your log property is %X{trace-id}, you can do this:
```java
builder.clear(); // traceId is a default field!
builder.addField(BaggageFields.TRACE_ID, "trace-id");
```

### DIFF
--- a/brave/src/main/java/brave/baggage/BaggagePropagation.java
+++ b/brave/src/main/java/brave/baggage/BaggagePropagation.java
@@ -102,7 +102,7 @@ public class BaggagePropagation<K> implements Propagation<K> {
      * @since 5.11
      */
     public Map<BaggageField, Set<String>> fieldToKeyNames() {
-      return Collections.unmodifiableMap(fieldToKeyNames);
+      return Collections.unmodifiableMap(new LinkedHashMap<>(fieldToKeyNames));
     }
 
     /**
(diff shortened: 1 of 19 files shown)


brave/src/main/java/brave/baggage/BaggagePropagation.java
```diff
@@ -102,7 +102,7 @@ public static FactoryBuilder newFactoryBuilder(Propagation.Factory delegate) {
      * @since 5.11
      */
     public Map<BaggageField, Set<String>> fieldToKeyNames() {
-      return Collections.unmodifiableMap(fieldToKeyNames);
+      return Collections.unmodifiableMap(new LinkedHashMap<>(fieldToKeyNames));
```
adriancole, line 105: bug

brave/src/main/java/brave/baggage/BaggagePropagation.java
```diff
@@ -171,7 +171,12 @@ public FactoryBuilder addRemoteField(BaggageField field, Iterable<String> keyNam
         allKeyNames.add(lcName);
         lcKeyNames.add(lcName);
       }
-      if (lcKeyNames.isEmpty()) lcKeyNames.add(field.lcName);
+
+      if (lcKeyNames.isEmpty()) { // add the default name
```
adriancole, line 175: bug

brave/src/main/java/brave/propagation/CurrentTraceContext.java
```diff
@@ -53,6 +51,7 @@
      */
     public Builder addScopeDecorator(ScopeDecorator scopeDecorator) {
       if (scopeDecorator == null) throw new NullPointerException("scopeDecorator == null");
+      if (scopeDecorator == ScopeDecorator.NOOP) return this;
```
adriancole, line 54: we need this because customizers act on builders.. we don't always know if there will be any configuration at all.

adriancole:
PS I noticed sleuth has an old thing that maps multiple names for trace ID. It is better to overlap scope decorators than handle this internally

PR #1138, opened 2020-04-03
Allows subtypes of ITRemote to override the CurrentTraceContext

adriancole:
We recently added enforcement of scopes as a test post-condition.
However, this broke a subtyping pattern used by Armeria.

### DIFF
--- a/brave-tests/src/main/java/brave/test/ITRemote.java
+++ b/brave-tests/src/main/java/brave/test/ITRemote.java
@@ -14,12 +14,14 @@
 package brave.test;
 
 import brave.Tracing;
-import brave.propagation.B3Propagation;
 import brave.baggage.BaggageField;
 import brave.baggage.BaggagePropagation;
+import brave.propagation.B3Propagation;
+import brave.propagation.CurrentTraceContext;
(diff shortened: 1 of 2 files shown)


adriancole:
Here's the usage:

```java

    private final StrictScopeDecorator strictScopeDecorator = StrictScopeDecorator.create();

    public BraveServiceIntegrationTest() {
        this.currentTraceContext = RequestContextCurrentTraceContext.builder()
            .addScopeDecorator(strictScopeDecorator)
            .build();
        this.tracing = tracingBuilder(Sampler.ALWAYS_SAMPLE).build();
        this.httpTracing = HttpTracing.create(tracing);
    }

    @Override
    protected void checkForLeakedScopes() {
        strictScopeDecorator.close();
    }
```

adriancole:
will change to return a current trace context builder as that is simpler

adriancole:
usage is simplified now

```java
  @Override
  protected CurrentTraceContext.Builder currentTraceContextBuilder() {
    return RequestContextCurrentTraceContext.builder();
  }
```

Users don't have to deal with anything else

PR #1142, opened 2020-04-05
Prepares for dynamic baggage and fixes broken OSGi config

adriancole:
This refactors internal code around baggage so that it can handle
dynamic baggage in a single propagated field. This also removes unused
tag parsing logic in gRPC as we no longer mutate tags.

`ExtraBaggageFields` is the entry in `TraceContext.extra()` that holds all
baggage, whether they are fixed or not. `BaggageHandler` is given
one slot and currently uses one slot per `BaggageField`. However,
`DynamicBaggageHandler` shows we can do multiple fields in one
slot, just we aren't doing that yet as we need to define an encoding.


brave/src/test/java/brave/features/baggage/DynamicBaggageTest.java
```diff
@@ -0,0 +1,44 @@
...
+  }
+
+  @Test public void encodes_arbitrary_fields() {
+    BaggageState baggageState = factory.create();
+
+    baggageState.updateValue(field1, "1");
+    baggageState.updateValue(field2, "2");
+    baggageState.updateValue(field3, "3");
+
+    assertThat(baggageState.encodeState(handler))
```
adriancole, line 38: encodeState is what Setter.set will use. I was bored, so I used property encoding in the test, but we will likely do comma, equals in real life.

This is how we can tunnel many baggage fields over one header, named for example, "baggage" or "b3ext" or whatever someone wants.

instrumentation/grpc/src/main/java/brave/grpc/GrpcPropagation.java
```diff
@@ -35,8 +39,8 @@
     Metadata.Key.of("grpc-trace-bin", Metadata.BINARY_BYTE_MARSHALLER);
 
   /** This stashes the tag context in "extra" so it isn't lost */
```
adriancole, line 41:
```suggestion
  /** This stashes the tag context into ad-hoc baggage state so it isn't lost */
```

instrumentation/grpc/src/main/java/brave/grpc/HexBinaryMarshaller.java
```diff
@@ -0,0 +1,49 @@
...
+    }
+  }
+
+  @Override public String parseBytes(byte[] buf) {
+    if (buf == null || buf.length == 0) return null;
+    try {
+      return ENCODING.encode(buf);
+    } catch (Throwable e) {
+      propagateIfFatal(e);
+      Platform.get().log("error decoding gRPC tags", e);
```
adriancole, line 45:
```suggestion
      Platform.get().log("error encode gRPC tags", e);
```

brave/src/test/java/brave/features/baggage/DynamicBaggageHandler.java
```diff
@@ -0,0 +1,102 @@
...
+    return newState;
+  }
+
+  @Override
+  public Map<BaggageField, String> mergeState(Map<BaggageField, String> state, BaggageField field,
+    String value) {
+    if (equal(value, state.get(field))) return state;
+    if (value == null) {
+      if (!state.containsKey(field)) return state;
+      if (state.size() == 1) return null;
```
anuraaga, line 69: `Collections.emptyMap`? Otherwise the interface needs to be updated to have a `@Nullable` return with what it means, but empty map seems simpler without any overhead.

brave/src/main/java/brave/internal/baggage/BaggageHandler.java
```diff
@@ -0,0 +1,85 @@
...
+   */
+  S newState(BaggageField field, String value);
+
+  /**
+   * Updates a state object to handle a field value change.
+   *
+   * @see BaggageField#updateValue(TraceContext, String)
+   * @see BaggageField#updateValue(TraceContextOrSamplingFlags, String)
+   */
+  S mergeState(S state, BaggageField field, @Nullable String value);
```
anuraaga, line 64: Any reason not to use `updateState` like `updateValue`?

brave/src/test/java/brave/internal/baggage/EntrySplitter.java
```diff
@@ -0,0 +1,92 @@
...
+ */
+package brave.internal.baggage;
+
+import brave.internal.Platform;
+
+/**
+ * Splits a character sequence that's in a delimited string trimming optional whitespace (OWS)
+ * before or after delimiters.
+ */
+public final class EntrySplitter {
```
anuraaga, line 22: Guessing you'll replace the properties load with this later but maybe add a TODO

brave/src/test/java/brave/internal/baggage/EntrySplitter.java
```diff
@@ -0,0 +1,92 @@
...
+    this.keyValueSeparator = keyValueSeparator;
+    this.entrySeparator = entrySeparator;
+    this.missingKeyValueSeparator =
+      "Invalid input: missing '" + keyValueSeparator + "' between key and value";
+  }
+
+  public boolean parse(String buffer, Handler handler, boolean shouldThrow) {
+    int i = 0, length = buffer.length();
+    while (i < length) {
+      if (isOWS(buffer.charAt(i++))) break; // skip whitespace
```
anuraaga, line 50: Think you may have intended to have the below comment here too, or only here since it's the first occurrence

brave/src/test/java/brave/internal/baggage/EntrySplitter.java
```diff
@@ -0,0 +1,92 @@
...
+    if (i == length) return logOrThrow("Invalid input: only whitespace", shouldThrow);
+    if (buffer.charAt(i) == keyValueSeparator) {
+      return logOrThrow("Invalid input: missing key", shouldThrow);
+    }
+
+    int beginName = i, endName = -1, beginValue = -1;
+    while (i < length) {
+      char c = buffer.charAt(i++);
+      // OWS is zero or more spaces or tabs https://httpwg.org/specs/rfc7230.html#rfc.section.3.2
+      if (isOWS(c)) continue; // trim whitespace
```
anuraaga, line 62: I think we need to only do this check when not inside a field or value, maybe not fields but presumably value should allow spaces.

adriancole: sorry this type was not intended yet.. it isn't used.. let me delete and we can do a proper review later?

brave/src/test/java/brave/internal/baggage/EntrySplitter.java
adriancole: accident.. let's review properly when actually used

brave/src/test/java/brave/internal/baggage/EntrySplitter.java
adriancole: 4680f9f27

brave/src/test/java/brave/features/baggage/DynamicBaggageHandler.java
adriancole: punted to #1153.. agree empty map doesn't have more overhead as the copy/constructor of LHM is cheap.  However, it does complicate the state model. let me know what you think..

brave/src/main/java/brave/internal/baggage/BaggageHandler.java
adriancole: good!

PR #1144, opened 2020-04-08
Adds example of custom trace ID parsing

adriancole:
This shows how people constrained to only propagating a trace ID can stash numerical data into
the trace ID. For example, a load balancer can set some bit flags like this.

Rumor has it that Twitter used to do this for embedding device information to ensure a
specific device couldn't absorb the entire random space. It isn't known if the amount of bits
stolen was a nibble or a byte.

See https://github.com/spring-cloud/spring-cloud-sleuth/issues/1106

### DIFF
--- a/brave/src/test/java/brave/features/propagation/HackedTraceIdTest.java
+++ b/brave/src/test/java/brave/features/propagation/HackedTraceIdTest.java
@@ -1,0 +1,111 @@
+/*
+ * Copyright 2013-2020 The OpenZipkin Authors
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License"); you may not use this file except
+ * in compliance with the License. You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software distributed under the License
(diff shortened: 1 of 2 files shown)


brave/src/test/java/brave/features/propagation/HackedTraceIdTest.java
```diff
@@ -0,0 +1,113 @@
...
+  @Test public void testB3MultiWins() {
+    headers.put("X-B3-TraceId", "1111111111111111");
+    headers.put("X-B3-SpanId", "2222222222222222");
+    headers.put(customTraceIdName, "1000000000000000e457b5a2e4d86bd1");
+    assertThat(extractor.extract(headers).context())
+      .satisfies(c -> assertThat(c.traceIdString()).isEqualTo("1111111111111111"))
+      .satisfies(c -> assertThat(c.spanIdString()).isEqualTo("2222222222222222"));
+  }
+
+  @Test public void invalidTraceIdsDontCrash() {
```
anuraaga, line 104: Delete I guess

adriancole: thx!

PR #1146, opened 2020-04-10
Adds an example about counting children

adriancole:
Those who need to process statistics on children can do something
like this, which works for any instrumentation.

Requested by @fletch3555

### DIFF
--- a/brave/src/test/java/brave/features/handler/CountingChildrenTest.java
+++ b/brave/src/test/java/brave/features/handler/CountingChildrenTest.java
@@ -1,0 +1,108 @@
+/*
+ * Copyright 2013-2020 The OpenZipkin Authors
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License"); you may not use this file except
+ * in compliance with the License. You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software distributed under the License
+ * is distributed on an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express
+ * or implied. See the License for the specific language governing permissions and limitations under
+ * the License.
+ */
+package brave.features.handler;
+
+import brave.Tracer;
+import brave.Tracing;
+import brave.handler.FinishedSpanHandler;
+import brave.handler.MutableSpan;
+import java.util.ArrayList;
+import java.util.List;
+import org.junit.After;
+import org.junit.Test;
+import zipkin2.Span;
+
+import static org.assertj.core.api.Assertions.assertThat;
+import static org.assertj.core.api.Assertions.tuple;
+
+/**
+ * This shows how a {@link FinishedSpanHandler} can add data some external formats need, such as
+ * child count.
+ *
+ * <p><em>Note:</em> this currently only works with children enclosed by their parents. If you have
+ * spans that finish after their parent, you'll need a more fancy implementation.
+ */
+public class CountingChildrenTest {
+  static final class TagChildCount extends FinishedChildrenHandler {
+    @Override protected void handle(MutableSpan parent, Iterable<MutableSpan> children) {
+      int count = 0;
+      for (MutableSpan child : children) {
+        count++;
+      }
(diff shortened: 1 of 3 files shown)


brave/src/test/java/brave/features/handler/CountingChildrenTest.java
```diff
@@ -0,0 +1,108 @@
...
+import zipkin2.Span;
+
+import static org.assertj.core.api.Assertions.assertThat;
+import static org.assertj.core.api.Assertions.tuple;
+
+/**
+ * This shows how a {@link FinishedSpanHandler} can add data some external formats need, such as
+ * child count.
+ *
+ * <p><em>Note:</em> this currently only works with children enclosed by their parents. If you have
```
anuraaga, line 33:
```suggestion
 * <p><em>Note:</em> this currently only works with children fully enclosed by their parents. If you have
```

Very small suggestion that I think makes it quite a bit more readable.

brave/src/test/java/brave/features/handler/FinishedChildrenHandler.java
```diff
@@ -0,0 +1,59 @@
...
+    }
+
+    Iterable<MutableSpan> children = parentToChildren.remove(context.spanIdString());
+
+    FinishedChildrenHandler.this.handle(span, children != null ? children : emptyList());
+    return true;
+  }
+
+  static final class ParentToChildren {
+    final Map<String, Set<MutableSpan>> delegate = new WeakHashMap<>();
```
anuraaga, line 49: A bit less noisy, and easier to switch to a possibly more efficient structure, using `Collections.synchronizedMap` instead of synchronizing the methods.

brave/src/test/java/brave/features/handler/FinishedChildrenHandler.java
```diff
@@ -0,0 +1,59 @@
...
+  @Override public boolean supportsOrphans() {
+    return true; // We need to clear the map entry even if the parent doesn't finish normally
+  }
+
+  @Override public boolean handle(TraceContext context, MutableSpan span) {
+    if (!context.isLocalRoot()) { // a child
+      parentToChildren.add(context.parentIdString(), span);
+    }
+
+    Iterable<MutableSpan> children = parentToChildren.remove(context.spanIdString());
```
anuraaga, line 42: Wonder if we should go ahead and copy into another list when passing to handler.

adriancole: I guess I'm confusing things by being inconsistent. I intentionally used synchronized above because we don't have a better collection type at the moment. That isn't not good for performance, so me avoiding array allocation here is probably not intuitive.

In real code I'd probably not do that, as it is possibly going to happen inline with a random call to Span.finish (time is stolen to invoke this hook). I'd probably prefer exposing Iterator instead..

brave/src/test/java/brave/features/handler/FinishedChildrenHandler.java
adriancole: yeah better!

brave/src/test/java/brave/features/handler/FinishedChildrenHandler.java
adriancole: updated with some notes on why.. ack allocation defence it is unimportant with how it is currently implemented though.

brave/src/test/java/brave/features/handler/FinishedChildrenHandler.java
```diff
@@ -0,0 +1,60 @@
...
+
+  @Override public boolean handle(TraceContext context, MutableSpan span) {
+    if (!context.isLocalRoot()) { // a child
+      parentToChildren.add(context.parentIdString(), span);
+    }
+
+    // There could be a lot of children. Instead of copying the list result, expose the iterator.
+    // The main goal is to not add too much overhead as this is invoked on the same thread as
+    // application code which implicitly call Span.finish() through instrumentation.
+    Set<MutableSpan> children = parentToChildren.remove(context.spanIdString());
```
adriancole, line 43: @anuraaga actually in this impl we could probably pass control to the user.. just anyway I want to get the other hooks in before we do that, just in case in a concurrent weak map variant, we can't give such a thing.

PR #1147, opened 2020-04-10
Reverts defensiveness in generic params and shares a buffer

adriancole:
it was a relic of an older JDK, or complete fantasy. This reverts it.

This also fixes where we accidentally didn't share a buffer.

Thanks @felixbarny for mentioning this..

### DIFF
--- a/brave-tests/src/test/java/brave/propagation/ThreadLocalCurrentTraceContextClassLoaderTest.java
+++ b/brave-tests/src/test/java/brave/propagation/ThreadLocalCurrentTraceContextClassLoaderTest.java
@@ -1,5 +1,5 @@
 /*
- * Copyright 2013-2019 The OpenZipkin Authors
+ * Copyright 2013-2020 The OpenZipkin Authors
  *
  * Licensed under the Apache License, Version 2.0 (the "License"); you may not use this file except
  * in compliance with the License. You may obtain a copy of the License at
@@ -40,6 +40,17 @@ public class ThreadLocalCurrentTraceContextClassLoaderTest {
       try (Scope scope = current.newScope(TraceContext.newBuilder().traceId(1).spanId(1).build())) {
 
       }
+    }
+  }
+
+  @Test public void leakedNullScope() {
+    assertRunIsUnloadable(LeakedNullScope.class, getClass().getClassLoader());
+  }
+
+  static class LeakedNullScope implements Runnable {
+    @Override public void run() {
+      CurrentTraceContext current = ThreadLocalCurrentTraceContext.newBuilder().build();
+      current.newScope(null);
     }
   }
 
--- a/brave-tests/src/test/java/brave/propagation/ThreadLocalCurrentTraceContextClassLoaderTest.java
+++ b/brave-tests/src/test/java/brave/propagation/ThreadLocalCurrentTraceContextClassLoaderTest.java
@@ -55,8 +55,12 @@ public class ThreadLocalCurrentTraceContextClassLoaderTest {
   }
 
   /**
-   * TODO: While it is an instrumentation bug to leak a scope, we should be tolerant, for example
-   * considering weak references or similar.
+   * TODO: While it is an instrumentation bug to leak a scope, we should be tolerant.
+   *
+   * <p>The current design problem is we don't know a reference type we can use that clears when
+   * the classloader is unloaded, regardless of GC. For example, having {@link Scope} extend {@link
+   * java.lang.ref.WeakReference} to hold the value to revert. This would only help if GC happened
+   * prior to the classloader unload, which would be an odd thing to rely on.
    */
   @Test(expected = AssertionError.class) public void leakedScope_preventsUnloading() {
     assertRunIsUnloadable(LeakedScope.class, getClass().getClassLoader());
--- a/brave/src/main/java/brave/Tracing.java
+++ b/brave/src/main/java/brave/Tracing.java
@@ -49,8 +49,7 @@ import zipkin2.reporter.Sender;
  * for example via spring or when mocking.
  */
 public abstract class Tracing implements Closeable {
-  // AtomicReference<Object> instead of AtomicReference<Tracing> to ensure unloadable
-  static final AtomicReference<Object> CURRENT = new AtomicReference<>();
+  static final AtomicReference<Tracing> CURRENT = new AtomicReference<>();
 
   public static Builder newBuilder() {
     return new Builder();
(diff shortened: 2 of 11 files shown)


adriancole:
Note: I tried JDK 1.8, though mine is latest version. Honestly, I have no idea why it failed to unload earlier.. we still fail to unload when someone leaks a scope.. this doesn't change that.

brave/src/main/java/brave/propagation/ThreadLocalCurrentTraceContext.java
```diff
@@ -92,12 +92,21 @@ public void clear() {
...
-      @Override public void close() {
-        local.set(previous);
-      }
-    }
-    Scope result = new ThreadLocalScope();
+    Scope result = new ThreadLocalScope(local, previous);
     return decorateScope(currentSpan, result);
   }
+
+  static final class ThreadLocalScope implements Scope {
```
adriancole, line 99: note this is not really much different than before.. I can revert it.. I was just playing around.

brave/src/main/java/brave/Tracing.java
```diff
@@ -49,8 +49,7 @@
  * for example via spring or when mocking.
  */
 public abstract class Tracing implements Closeable {
-  // AtomicReference<Object> instead of AtomicReference<Tracing> to ensure unloadable
-  static final AtomicReference<Object> CURRENT = new AtomicReference<>();
+  static final AtomicReference<Tracing> CURRENT = new AtomicReference<>();
```
anuraaga, line 52: Agree - the JVM has no idea what's in these brackets anyways without reflection

brave/src/main/java/brave/propagation/ThreadLocalCurrentTraceContext.java
anuraaga, line 99: This seems nice - seems like it could fix something?

adriancole: well  I'll update.. with where I was going.. basically revert to null can be a singleton. I still can't get the thing to unload when leaked as the cover type is brave's.

adriancole: ptal

adriancole:
@felixbarny curious do you know if there is any reference that can be reclaimed when its classloader is being unloaded? I'm stumped..

adriancole:
sorry cleared, not reclaimed

brave-tests/src/test/java/brave/propagation/ThreadLocalCurrentTraceContextClassLoaderTest.java
```diff
@@ -43,9 +43,24 @@
...
+  }
+
   /**
-   * TODO: While it is an instrumentation bug to leak a scope, we should be tolerant, for example
-   * considering weak references or similar.
+   * TODO: While it is an instrumentation bug to leak a scope, we should be tolerant.
+   *
+   * <p>The current design problem is we don't know a reference type we can use that clears when
+   * the classloader is unloaded, regardless of GC. For example, having {@link Scope} extend {@link
+   * java.lang.ref.WeakReference} to hold the value to revert. This would only help if GC happened
```
anuraaga, line 62: Doesn't block this PR at all, just to record my thought and maybe @felixbarny has some ideas too.

I haven't worked with unloading scenarios much but wonder whether this is so odd - since most things get cleaned up when GC happens, it seems natural to me for unloaded classes to also wait until GC to be cleaned up.

adriancole:
thanks @anuraaga any other things we can do in another PR if they come up

felixbarny:
> @felixbarny curious do you know if there is any reference that can be reclaimed when its classloader is being unloaded? I'm stumped..

Not sure I understand your question. There’s no explicit operation to unload classes or a classloader. This is just a side effect of the classloader being GC‘d. As long as there’s one object of one class, declared by the class that is referenced strongly, the classloader can’t be collected. That’s because an object references it’s class and a class has a reference to its classloader. 

References among the objects of that classloader are ok, as long as there are no objects in, for example, the bootstrap classloader referencing objects of the classloader that is supposed to be GC‘d. This is a common scenario with ThreadLocals. That is because ThreadLocals are stored within Thread instances (whose classes are loaded by the bootstrap classloader). If the ThreadLocals store non-bootstrap classes, leaving those objects in the ThreadLocal holds the classloader of those objects alive.

adriancole:
@felixbarny thanks, so I guess maybe I had the problem inverted. Appreciate you bringing that forward.. I suppose what this means is.. the only solution to unload things that leak values in a thread local would be to have the threadlocal hold a weak reference right? At least that's solvable even if it seems expensive..

felixbarny:
That’s typically not a good idea because then there’s nothing that prevents the value, a scope, for example, to be collected.
A better way would be to use Rafael‘s DetachedThreadLocal that is basically a ConcurrentHashMap with a weakly referenced Thread as the key. The lookups are a bit more expensive, however as they cause an allocation.

adriancole:
@felixbarny thanks for the hints.. to put a bit concretely though, the reference enclosed by the scope is TraceContext, which would have a hard reference elsewhere (such as the span where it came from). Wouldn't it only be in an error case, then that the reference would become clear when leaked? Maybe I misunderstand, but I would have thought that the WeakReference wouldn't clear if there was another hard reference to the same object...

adriancole:
ex

```java
try (Scope scope1 = currentTraceContext.newScope(context1)) {
  try (Scope scope2 = currentTraceContext.newScope(context2)) {
    // my assumption is that a weak ref of scope2.context which contains context1,
    // wouldn't clear since there's a hard reference to context1 still
  }
}
```

again thanks for the lesson either way!

adriancole:
PS to folks watching... I thought I was crazy, but then was able to find the thing that was the fanciestestest reference manager I've seen in the wild. This was deleted from google/dagger though not everyone agreed that it should be. regardless, you can take a look for fun!

https://github.com/google/dagger/issues/1117
last version of code here https://github.com/google/dagger/commit/a72c3428a3489ff837a1a4d3023214f157bad82d

@raphw @nicmunroe I think you guys like this gnarly stuff, so if you hadn't seen it ^^

felixbarny:
Sure, if there’s another strong reference to a weakly referenced object it will be kept alive.

adriancole:
> Sure, if there’s another strong reference to a weakly referenced object it will be kept alive.

cool, in this case having scope with a weak reference will make sense as it is a bug to leave the scope open when you exit a method [*]

This moves the question into the "should we do this" category.. ex.. is the overhead of extending WeakReference worth it for the advantage of not pinning the classloader? I suppose benchmarks could answer, unless folks think this is not a good idea for another reason..



[*] sometimes people stash scope in an attribute, but they also stash the context directly or indirectly as a member of Span.

adriancole:
it is a little nuanced I guess because depending on the impl of scoping someone could have a parent which isn't stacked somewhere, and in this case, yeah it could end up collected. for example, if a child finishes after the parent and during unlucky time of high GC..

If this happened the parent ref would clear and when the scope reverts it would clobber whatever the parent was with null. That isn't the worst outcome but it would be inaccurate (ex logs would show no trace ID after the scope was closed instead of the former parent. If this thinking is correct, then probably holding the class loader is ok, especially as we have `StrictCurrentTraceContext` which allows people to find leaked scopes which are bugs in the first place..

PR #1148, opened 2020-04-10
Adds SpanHandler

adriancole:
SpanHandler deprecates FinishedSpanHandler and allows tracking data recorded by brave. Notably, this allows aggregations not formerly possible, such as child counting, and data correlation.


brave/src/test/java/brave/features/handler/CountingChildrenTest.java
```diff
@@ -103,7 +104,7 @@
     assertThat(spans)
       .extracting(Span::name, s -> s.tags().get("childCount"))
       .containsExactly(
-        tuple("root1", "0"), tuple("root1child1", "0"), tuple("root1child2", "0")
+        tuple("root1", "2"), tuple("root1child1", "0"), tuple("root1child2", "0")
```
adriancole, line 107: yay we can implement `child_span_count` in the future with a native brave SpanHandler for stackdriver

cc @meltsufin @saturnism

https://cloud.google.com/trace/docs/reference/v2/rpc/google.devtools.cloudtrace.v2#google.devtools.cloudtrace.v2.Span

saturnism: neat!

adriancole:
currently working to add the parent parameter to the handleCreate api. there's some trickiness as `Tracer.toSpan` exists and can result in a different reference of `TraceContext` vs what other callbacks accept. I will try to fix this.

adriancole:
rebased but not ported to use weak maps yet

adriancole:
@jorgheymans @anuraaga @jeqo I ended up going with a listener instead of sharing inheritance with FinishedSpanHandler. The reason is that there are some shortcuts done that mean FinishedSpanHandler can't see all signals.

While ack I haven't documented anything here, I'm wondering if it makes sense to you in general. I'm happy that the child counter example works, and also the orphan tracker works with less (though still a little) hacks..

I don't plan to merge this until I have an example listener that can hold a localRoot until it is all ready or timeout occurred.

brave/src/main/java/brave/handler/SpanListener.java
```diff
@@ -0,0 +1,103 @@
...
+ * the License.
+ */
+package brave.handler;
+
+import brave.Span;
+import brave.internal.Nullable;
+import brave.propagation.TraceContext;
+import java.lang.ref.WeakReference;
+
+/**
```
adriancole, line 21: This is the important file

brave/src/main/java/brave/internal/recorder/OrphanTracker.java
```diff
@@ -0,0 +1,38 @@
...
+ */
+package brave.internal.recorder;
+
+import brave.handler.MutableSpan;
+import brave.handler.SpanListener;
+import brave.internal.InternalHandler;
+import brave.internal.Platform;
+import brave.propagation.TraceContext;
+
+// TODO make a weak map of MutableSpan -> Throwable instead of a hacky field in MutableSpan
```
adriancole, line 22: PS I plan to shade @raphw's https://github.com/raphw/weak-lock-free/blob/master/src/main/java/com/blogspot/mydailyjava/weaklockfree/WeakConcurrentMap.java into the core jar to support orphan tracker as it is much nicer to use a weak map instead of a hacky field. The code is small (won't make the jar too big), has the same bytecode level as us (Java 6) and stable. We can cite it in the NOTICE file.

If you think this is a bad idea, do scream at some point :D

brave/src/main/java/brave/handler/SpanListener.java
```diff
@@ -0,0 +1,106 @@
...
+ * <p>The {@link TraceContext} parameter from {@link #onCreate} will be the same reference for
+ * all callbacks, except {@link #onOrphan}, which has value, but not reference equality.
+ *
+ * <p>The {@link MutableSpan} parameter from {@link #onCreate} will be the same reference for
+ * all callbacks. Do not mutate {@link MutableSpan} between callbacks as it is not thread safe.
+ *
+ * <p>If caching the {@link TraceContext} parameter, consider a {@link WeakReference} to avoid
+ * holding up garbage collection.
+ */
+public class SpanListener {
```
anuraaga, line 39: abstract? Even in NOOP we subclass it :)

brave/src/main/java/brave/handler/SpanListener.java
```diff
@@ -0,0 +1,106 @@
...
+   * <p><em>Note:</em>Abandoned spans should be ignored as they aren't indicative of an error. Some
+   * instrumentation speculatively create a span for possible outcomes such as retry.
+   */
+  public void onAbandon(TraceContext context, MutableSpan span) {
+  }
+
+  /**
+   * Called on {@link Span#flush()}.
+   *
+   * <p>Even though the span here will is incomplete (missing {@link MutableSpan#finishTimestamp()},
```
anuraaga, line 82: is incomplete

brave/src/main/java/brave/handler/SpanListener.java
```diff
@@ -0,0 +1,106 @@
...
+      return "NoopSpanListener{}";
+    }
+  };
+
+  protected SpanListener() {
+  }
+
+  /**
+   * This is called when a span is allocated, but before it is started. An allocation here will
+   * result in one of:
```
anuraaga, line 52: One of the following will be called later when the created span is no longer in use.

anuraaga: Currently I was surprised since it reads like these are called right away

brave/src/main/java/brave/handler/SpanListener.java
```diff
@@ -0,0 +1,106 @@
...
+   *
+   * <ol>
+   *   <li>{@link #onAbandon} if this was a speculative context</li>
```
anuraaga, line 55: I guess it instead of this

brave/src/main/java/brave/internal/handler/SafeSpanListener.java
```diff
@@ -0,0 +1,165 @@
...
+    @Override void doOnFinish(TraceContext context, MutableSpan span) {
+      delegate.onFinish(context, span);
+    }
+
+    @Override public String toString() {
+      return delegate.toString();
+    }
+  }
+
+  static final class Multiple extends SafeSpanListener {
```
anuraaga, line 123: Not sure if it's necessarily better, but I was sort of expecting there to just be one implementation of `SafeSpanListener` and this would actually be a `SpanListener`. We could possibly make it public too since it could be helpful for users to be able to combine multiple listeners but don't have to right now.

brave/src/main/java/brave/internal/recorder/OrphanTracker.java
adriancole: I started to do this, then realized we had hacked this quite a bit in PendingSpans, and also shading adds a lot of complexity and also the class is quite bigger than ours anyway. I cleaned up our variant instead here https://github.com/openzipkin/brave/pull/1152

brave/src/main/java/brave/internal/handler/SafeSpanListener.java
adriancole: I see what you mean.. good idea on only one safe thing.. I can go back and address the others.

I don't think we're ready to support user-side compositing as it is unlikely they will do a better job than we would. Probably better for us to just say "don't composite on your own" then wait for a scream? (this would apply to the several other places we do similar compositing). Not doing user-level compositing also keeps the api sprawl down.. brave's already collecting quite a number of public types...

brave/src/main/java/brave/handler/SpanListener.java
adriancole: sgtm

brave/src/main/java/brave/internal/handler/SafeSpanListener.java
anuraaga: Not exposing a compositing listener to public unless we find a use for it in the future sgtm

adriancole:
@anuraaga I just pushed clean to work over latest HEAD. I've not incorporated your feedback, yet, but will.

brave/src/test/java/brave/features/handler/FinishedChildrenListener.java
```diff
@@ -0,0 +1,88 @@
...
+import java.util.Set;
+
+import static java.util.Collections.emptyIterator;
+
+public abstract class FinishedChildrenListener extends SpanListener {
+
+  protected abstract void onFinish(MutableSpan parent, Iterator<MutableSpan> children);
+
+  /** This holds the children of the current parent until the former is finished or abandoned. */
+  final WeakConcurrentMap<TraceContext, TraceContext> childToParent = new WeakConcurrentMap<>();
```
adriancole, line 31: note this now uses weak maps

adriancole: oops forgot to switch this back to https://github.com/raphw/weak-lock-free as an example shouldn't use our internal types..

brave/src/main/java/brave/handler/MutableSpan.java
```diff
@@ -72,7 +72,7 @@
   ArrayList<String> tags;
   /** Also use pair indexing for annotations, but type object to store (startTimestamp, value). */
   ArrayList<Object> annotations;
-  Throwable error;
+  Throwable caller, error;
```
adriancole, line 75:
```suggestion
  Throwable error;
```

brave/src/main/java/brave/RealSpan.java
```diff
@@ -140,13 +140,12 @@
   }
 
   @Override public void finish() {
-    finish(clock.currentTimeMicroseconds());
+    finish(0L);
```
adriancole, line 143: TODO move these changes to master as they aren't specific to this feature

brave/src/main/java/brave/RealScopedSpan.java
```diff
@@ -74,8 +74,7 @@
 
   @Override public void finish() {
     scope.close();
-    if (pendingSpans.remove(context) == null) return; // don't double-report
-    state.finishTimestamp(clock.currentTimeMicroseconds());
+    if (!pendingSpans.finish(context, 0L)) return; // don't double-report
```
adriancole, line 77: TODO move this change to master as it isn't specific to this feature

brave/RATIONALE.md
```diff
@@ -1,5 +1,32 @@
...
+For example, we need to manipulate Baggage without overriding `PropagationFactory`. For example,
+some baggage are derived (parsed from other headers). Given a create hook, one can post process
+baggage before the span is visible to the user.
+
+Another example is parent-child relationships. With a create hook someone can do things like count
+children (needed by [stackdriver model](https://cloud.google.com/trace/docs/reference/v2/rpc/google.devtools.cloudtrace.v2#google.devtools.cloudtrace.v2.Span)).
+
+Finally, some services require the entire "local root" to be reported at the same time. The only
+known example of this is [DataDog](https://github.com/DataDog/dd-trace-java/blob/406b324a82b482d7d8ad3faa5f9ccdd307c72308/dd-trace-ot/src/main/java/datadog/trace/common/writer/Writer.java#L20).
+A create hook can be used track a local root, allowing buffering of all descendants.
```
jeqo, line 15: nit:

```suggestion
There are a few reasons why `SpanListener` exists. For example:

* We need to manipulate Baggage without overriding `PropagationFactory`, e.g. when
some baggage are derived (parsed from other headers): Given a create hook, one can post process
baggage before the span is visible to the user.
* Another example is parent-child relationships: With a create hook someone can do things like count
children (needed by [stackdriver model](https://cloud.google.com/trace/docs/reference/v2/rpc/google.devtools.cloudtrace.v2#google.devtools.cloudtrace.v2.Span)).
* Some services require the entire "local root" to be reported at the same time. The only
known example of this is [DataDog](https://github.com/DataDog/dd-trace-java/blob/406b324a82b482d7d8ad3faa5f9ccdd307c72308/dd-trace-ot/src/main/java/datadog/trace/common/writer/Writer.java#L20): A create hook can be used track a local root, allowing buffering of all descendants.
```

brave/src/main/java/brave/RealSpan.java
jeqo, line 143: @adriancole would you mind sharing the reasoning behind `0L` instead of `#currentTimeMicroseconds()`

brave/src/main/java/brave/handler/SpanListener.java
```diff
@@ -0,0 +1,106 @@
...
+   * <p>Even though the span here will is incomplete (missing {@link MutableSpan#finishTimestamp()},
+   * it is reported to the tracing system unless a {@link FinishedSpanHandler} returns false.
+   */
+  public void onFlush(TraceContext context, MutableSpan span) {
+  }
+
+  /**
+   * Called when the trace context was garbage collected prior to completion.
+   *
+   * <p>Unlike {@link FinishedSpanHandler#supportsOrphans()}, this is called even if {@linkplain
```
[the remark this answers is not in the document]
adriancole: your questions are helpful as I also feel uncomfortable.

I'm wondering if the type shouldn't be called something else..

We are mostly talking about a binding between a trace context and a state object for the purpose of collection and when that binding is complete (for any reason). When there are
only two states it is simpler and no diagrams are needed.

Plus we don't need to describe nuance (though we can in rationale), such as that spans are sometimes never started or their start timestamp overwritten (ex "start" is not a state in a state diagram, just field manipulation)

Ex
```java
SampledSpanListener {
  enum Reason {
    FINISH,
    FLUSH,
    GC;
  }
  void onBind(TraceContext, MutableSpan)
  void onRelease(TraceContext, MutableSpan, Reason)
}
```

brave/src/main/java/brave/handler/SpanListener.java
```diff
@@ -0,0 +1,106 @@
...
+  };
+
+  protected SpanListener() {
+  }
+
+  /**
+   * This is called when a span is allocated, but before it is started. An allocation here will
+   * result in one of:
+   *
+   * <ol>
```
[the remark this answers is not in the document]
adriancole: spans are sometimes never started or their start timestamp overwritten (ex "start" is not a state in a state diagram, just field manipulation)

TODO: add this to rationale

brave/src/main/java/brave/RealSpan.java
adriancole: to prevent overhead of using the clock when the span is already finished.

adriancole:
@anuraaga please look at https://github.com/openzipkin/brave/pull/1148#discussion_r407617993 as I think we should rework this interface to be simpler vs try to explain its current state. wdyt?

brave/src/main/java/brave/RealSpan.java
adriancole: TODO: comment

adriancole: fixed upstream

brave/src/main/java/brave/handler/SpanListener.java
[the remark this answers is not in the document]
anuraaga: This interface does seem nicer since many use cases won't even care about the reason and it's more explicit that only one of those is called. That said, I don't know if conceptually it's so different, docs that would apply to the individual interface methods seem to just move into the Reason enum.

Also unsure of bind, the word isn't very intuitive for me. `onStart` / `onEnd` seem clear, with the caveat of `end` looking a lot like `finish` (but then again `release` I guess looks like `GC` 🤣 ).

Wondering what happens if the methods return `boolean`? `onBind` can return secondary sampling decision, `onRelease` can replace `FinishedSpanHandler` so don't have to reason about that?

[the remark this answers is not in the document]
adriancole: I think the biggest difference the enum brings is that the state diagram becomes unambiguous.. whatever the begin hook has only one end hook and they are both always called. Those who care about the reason of the end can look at it. I agree docs will just move, but this obviates docs about a state diagram I think.

naming is not my favorite either.. was trying to avoid the word finish. I tried thinking through various things that suggest an association.
ex

associate/disassociate
allocate/deallocate
bind/unbind

random words that just say start/stop
begin/end
start/finish (dangerous as conflated with span ops)

I'm currently thinking..
onSampled
onCollected

wrt to `onCollected` returning false, I think we could I guess yeah, just it would also imply that any finishedSpanListeners would need to be converted first to these in order to perform things like redaction. Another way is we can make this able to internally dispatch to finishedSpanHandlers... ex Whatever this is called exposes a factory that accepts a single (potentially composite) FinishedSpanHandler instance we give to it...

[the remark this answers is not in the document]
adriancole: one possibility especially with the design as it is, is that we can deprecate FinishedSpanHandler. I feel one thing this design does is makes the relationships a whole lot more clear. So what would happen in this case is that we make a special SampledSpanListener that does a dispatch first to any FinishedSpanHandlers before calling its onCollected hook. that way we don't have to mention `FinishedSpanHandler` at all? It is a little sad that `FinishedSpanHandler` didn't live that long, but I feel this is becoming a much better conceptual model.

adriancole:
rebased.. next will finally go through feedback :)

adriancole:
In honor of tomorrow's 5 year anniversary of the gathering of OpenZipkin by @eirslett, I propose we re-instate the name `SpanCollector` for this. It was used in Brave 3.x when @kristofa was head honcho.

Later, I'll deprecate FinishedSpanHandler as an implementation of this.

thumbsup this comment if you like the word SpanCollector, comment with a different word.

brave-tests/src/test/java/brave/internal/recorder/PendingSpansClassLoaderTest.java
```diff
@@ -15,6 +15,7 @@
 
 import brave.handler.FinishedSpanHandler;
 import brave.handler.MutableSpan;
+import brave.handler.SpanCollector;
```
adriancole, line 18: eventhough nostalgia for `SpanCollector` `SpanHandler` may still be better name I admit as we can't move the `MutableSpan` out of this package without certain chaos...

brave/RATIONALE.md
```diff
@@ -1,5 +1,29 @@
 # brave rationale
 
+## SpanListener
```
adriancole, line 3: ignore drift here for now as I need to rewrite it anyway but don't want to rewrite it two or three times

brave-tests/src/test/java/brave/internal/recorder/PendingSpansClassLoaderTest.java
jeqo: Also `collector` can cause confusion with server-side collectors.

Just for reference, and to add one more alternative naming, in Kafka Clients [_interceptors_](https://cwiki.apache.org/confluence/display/KAFKA/KIP-42%3A+Add+Producer+and+Consumer+Interceptors) have a similar function.

adriancole: good point on collector being conflated (I think  I recall now we dodged the word collector in the first place due to this..)

personally I like handler better, as usually interceptor means getting in the middle of something. ex

```java
void intercept(thing, chain){
  return chain.proceed(thing);
}
```

jeqo: Fair enough. I also prefer Handler instead of Listener..--naming is hard :)

brave/src/main/java/brave/handler/SpanCollector.java
```diff
@@ -0,0 +1,134 @@
...
+ *
+ * <p>If caching the context or span parameters between callbacks, consider a {@link WeakReference}
+ * to avoid holding up garbage collection.
+ *
+ * <p>The {@link #begin} callback primarily supports tracking of children, or partitioning of
+ * data for backend that needs to see an entire {@linkplain TraceContext#localRootId() local root}.
+ */
+public interface SpanCollector {
+  /** What ended the data collection? */
+  enum Cause {
```
jorgheymans, line 54: for an enum `Cause` i would linguistically expect past-tense constants, `ABANDONED` - `FINISHED` etc ?

brave/src/main/java/brave/handler/SpanCollector.java
```diff
@@ -0,0 +1,134 @@
...
+    /**
+     * Called when the trace context was garbage collected prior to completion.
+     *
+     * <p>Unlike {@link FinishedSpanHandler#supportsOrphans()}, this is called even if empty.
+     * Non-empty spans are reported to the tracing system unless a {@link FinishedSpanHandler}
+     * returns false.
+     *
+     * @see FinishedSpanHandler#supportsOrphans()
+     */
+    ORPHAN
```
jorgheymans, line 87: the other constants are named linear to their `Span` methods, why `ORPHAN` for gc collected, are there other use cases leading to `ORPHAN` besides gc ?

brave/src/main/java/brave/handler/SpanCollector.java
```diff
@@ -0,0 +1,134 @@
...
+  void begin(TraceContext context, MutableSpan span, @Nullable TraceContext parent);
+
+  /**
+   * Called when data collection completes for one of the following reasons:
+   *
+   * <p><ol>
+   * <li>{@link Cause#ABANDON} if it was a speculative context</li>
+   * <li>{@link Cause#FINISH} if it was reported complete</li>
+   * <li>{@link Cause#FLUSH} if it was intentionally reported incomplete</li>
+   * <li>{@link Cause#ORPHAN} if it was reported incomplete due to garbage collection</li>
```
jorgheymans, line 124: does this need to be duplicated from `Cause` ? 

`Called when data collection completes` on its own is enough, and maybe link in the `cause` parameter to the enum.

brave/src/test/java/brave/features/handler/FinishedChildrenCollector.java
```diff
@@ -0,0 +1,85 @@
...
+  @Override
+  public void begin(TraceContext context, MutableSpan span, @Nullable TraceContext parent) {
+    if (!context.isLocalRoot()) { // a child
+      childToParent.putIfProbablyAbsent(context, parent);
+      parentToChildren.add(parent, span);
+    }
+  }
+
+  @Override public boolean end(TraceContext context, MutableSpan span, Cause cause) {
+    // Kick-out if the this was not a normal finish
```
jorgheymans, line 45:
```suggestion
    // Kick-out if this was not a normal finish
```

brave/src/test/java/brave/features/handler/FinishedChildrenCollector.java
```diff
@@ -0,0 +1,85 @@
...
+import brave.internal.Nullable;
+import brave.propagation.TraceContext;
+import com.blogspot.mydailyjava.weaklockfree.WeakConcurrentMap;
+import java.util.Iterator;
+import java.util.LinkedHashSet;
+import java.util.Set;
+
+import static java.util.Collections.emptyIterator;
+
+public abstract class FinishedChildrenCollector implements SpanCollector {
```
jorgheymans, line 27: does this need a type level javadoc ?

brave/src/main/java/brave/handler/SpanCollector.java
adriancole: well GC is the side affect of orphaning your span (ex by creating it and dropping a reference to it). I think that's why we named it orphan back then, but I can look up notes if you like.

Main thing is the fact that it is GC'ed is only because someone orphaned it. If they held a reference it would not become GC'ed. Possibly I can clarify the description somewhere.

brave/src/main/java/brave/handler/SpanCollector.java
adriancole: sgtm I can put this on the type docs of Cause, as I do think an overview helps, but possibly not here..

brave/src/test/java/brave/features/handler/FinishedChildrenCollector.java
adriancole: test class, but sure

brave/src/main/java/brave/handler/SpanCollector.java
adriancole: sgtm thanks

adriancole:
ok last changed it back to SpanHandler and notably folded FinishedSpanHandler underneath. Please ignore javadoc and markdown drift which still talks about FinishedSpanHandler for now.

We have to handle (hehe) "alwaysSampleLocal" anyway in order to support FinishedSpanHandler, but I'd like to see if we can change the method slightly to support parameterized decisions as needed by https://github.com/openzipkin-contrib/zipkin-secondary-sampling

After that, the docs need to be shimmied up and FinishedSpanHandler should be deprecated

adriancole:
fyi I'm going to finish this up now, no pun intended.

The above commit includes all resolved comments and no changes based on a rebase. Unresolved comments are considered unresolved. Docs need to be done as well things like spring-beans.

adriancole:
ok cleanups are in.. I will rebase #1167 over this and prepare for release!

adriancole:
Thanks @jeqo @jorgheymans and @anuraaga !

PR #1149, opened 2020-04-11
Ensures currentSpan, toSpan and joinSpan use existing context references

adriancole:
Before, `Tracer.currentSpan`, `toSpan` and `joinSpan` either didn't
decorate external contexts (in no-op) or needlessly recreated context
instances when a span was already present in the pending map.i

This not only adds overhead, but interferes with our ability to track
the difference between a newly allocated span vs incidental changes to
an existing one due to use of `currentSpan`, `toSpan` or `joinSpan`.

This change is necessary to support `SpanHandler` when powered by
identity maps.

See #1148

### DIFF
--- a/brave/src/main/java/brave/NoopSpan.java
+++ b/brave/src/main/java/brave/NoopSpan.java
@@ -101,6 +101,8 @@ final class NoopSpan extends Span {
     return isEqualToNoopOrLazySpan(context, o);
   }
 
+  // We don't compare a RealSpan vs a NoopSpan as they can never equal each other.
+  // RealSpan's are always locally sampled and Noop ones are always not.
   static boolean isEqualToNoopOrLazySpan(TraceContext context, Object o) {
     if (o instanceof LazySpan) {
       return context.equals(((LazySpan) o).context);
--- a/brave/src/main/java/brave/NoopSpan.java
+++ b/brave/src/main/java/brave/NoopSpan.java
@@ -1,5 +1,5 @@
 /*
- * Copyright 2013-2019 The OpenZipkin Authors
+ * Copyright 2013-2020 The OpenZipkin Authors
  *
  * Licensed under the Apache License, Version 2.0 (the "License"); you may not use this file except
  * in compliance with the License. You may obtain a copy of the License at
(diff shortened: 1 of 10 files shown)


brave/src/main/java/brave/Tracer.java
```diff
@@ -369,27 +365,48 @@ public Span nextSpan(TraceContextOrSamplingFlags extracted) {
...
-    if (isDecorated(context)) return _toSpan(context);
-
-    return _toSpan(decorateContext(
+    // Re-use a pending context if present: This ensures reference consistency on Span.context()
+    TraceContext pendingContext = swapForPendingContext(context);
+    if (pendingContext != null) return _toSpan(pendingContext);
+
+    // There are a few known scenarios for the context to be absent from the pending map:
+    // * Created by a separate tracer (localRootId set)
+    // * Resurrected span from a garbage collection (localRootId set)
```
adriancole, line 381: this one is very hard to grok, but basically if the input context is the same ids as the one garbage collected, but a different instance, it is effectively resurrection from the POV of the user (as they wouldn't know about ref swapping anyway). I can remove this or revise this comment if confusing

brave/src/main/java/brave/internal/recorder/PendingSpan.java
```diff
@@ -15,17 +15,32 @@
...
+import java.lang.ref.WeakReference;
 
-public final class PendingSpan {
+/**
+ * This includes a weak reference of the trace context, which allows externalized forms of the trace
+ * context to be swapped for the one in use.
+ *
+ * <p>This is a weak reference to ensure that {@link PendingSpans} can clean up on GC.
+ */
+public final class PendingSpan extends WeakReference<TraceContext> {
```
anuraaga, line 28: I don't think this type is exposed to users and can be leaked by them. Can't we use a normal reference?

brave/src/main/java/brave/Tracer.java
adriancole: sgtm thanks

brave/src/main/java/brave/internal/recorder/PendingSpan.java
adriancole: if we use a hard reference, it breaks PendingSpans. The keys used in PendingSpans include a weak reference to the trace context, if the value of the same entry has a hard reference, the weak keys are made useless as they can never be collected. Is there a good way to rephrase this?

brave/src/main/java/brave/NoopSpan.java
```diff
@@ -92,10 +92,22 @@
...
+   * code should not act differently given an instance of lazy or {@link NoopSpan}.
+   */
   @Override public boolean equals(Object o) {
     if (o == this) return true;
-    if (!(o instanceof NoopSpan)) return false;
-    return context.equals(((NoopSpan) o).context);
+    return isEqualToNoopOrLazySpan(context, o);
+  }
+
+  static boolean isEqualToNoopOrLazySpan(TraceContext context, Object o) {
```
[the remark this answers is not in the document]
adriancole: added this for the two places it applies (here and over in real)

```
  // We don't compare a RealSpan vs a NoopSpan as they can never equal eachother.
  // RealSpan's are always locally sampled and Noop ones are always not.
```

brave/src/main/java/brave/internal/recorder/PendingSpan.java
adriancole: I'll rephrase as it isn't clear that this type is a map entry value anyway..

adriancole: PTAL! thanks

PR #1152, opened 2020-04-12
Extracts WeakConcurrentMap from PendingSpans and uses for trackOrphans

adriancole:
This formalizes that PendingSpans was derived from weaklockfree, pulling
it up so that we can decouple orphan handling in the future.

This also fixes attribution which was in the source, but not NOTICE
before. Sorry @raphw (though not sure you noticed.. pun intended)

Finally, this does a fair amount of simplification on the orphan path.
Most importantly, we pre-emptively fork the context as opposed to
copying most of its values for the orphan path.

### DIFF
--- a/brave/src/main/java/brave/internal/weaklockfree/WeakConcurrentMap.java
+++ b/brave/src/main/java/brave/internal/weaklockfree/WeakConcurrentMap.java
@@ -1,0 +1,152 @@
+/*
+ * Copyright 2013-2020 The OpenZipkin Authors
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License"); you may not use this file except
+ * in compliance with the License. You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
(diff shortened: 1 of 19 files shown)


adriancole:
The jar ends up a smidge smaller than it was before, still rounding to 216KiB

We haven't looked at size in a while, so probably owe a sweep for duplicated or unnecessary bytecode.

brave/src/main/java/brave/internal/weaklockfree/WeakConcurrentMap.java
```diff
@@ -0,0 +1,152 @@
...
+      return hashCode;
+    }
+
+    @Override public String toString() {
+      T value = get();
+      return value != null ? value.toString() : "ClearedReference()";
+    }
+
+    /**
+     * While a lookup key will invoke equals against this, the visa versa is not true. This methos
```
adriancole, line 130:
```suggestion
     * While a lookup key will invoke equals against this, the visa versa is not true. This method
```

PR #1153, opened 2020-04-12
Updates from feedback on #1142

adriancole:


brave/src/main/java/brave/internal/baggage/BaggageHandler.java
```diff
@@ -56,12 +56,18 @@
   S newState(BaggageField field, String value);
 
   /**
-   * Updates a state object to handle a field value change.
+   * Updates a state object to include a value change.
+   *
+   * <p>When the value is {@code null} and the input only includes the given field, either {@code
```
adriancole, line 61: personally I think this is a lot simpler if empty coerces to null. A lot less to explain, no chance of accidental mutable empty types, easier tests.. do you see what I mean @anuraaga ?

brave/src/main/java/brave/internal/baggage/BaggageHandler.java
```diff
@@ -56,12 +56,18 @@
...
+   * null} or an appropriate empty sentinel value should be returned.
+   *
+   * <p>When {@code null} is returned, the next non-null update results in a call to {@link
+   * #newState}. Otherwise, the empty state object will become the next input parameter here.
    *
    * @see BaggageField#updateValue(TraceContext, String)
    * @see BaggageField#updateValue(TraceContextOrSamplingFlags, String)
    */
-  S mergeState(S state, BaggageField field, @Nullable String value);
+  @Nullable S updateState(S state, BaggageField field, @Nullable String value);
```
adriancole, line 70: normal baggage S will be string, so this has to be nullable, lack of annotation before was a mistake!

PR #1154, opened 2020-04-12
align javax.annotation-api 1.3.1, 1.3.2 to 1.3.2 

calvinhkf:
align library versions to avoid inconsistent API behaviors. (Or unify library version by removing explicitly declared version tag in _instrumentation/jaxrs2/pom.xml_


adriancole:
odd as this is a provided dep, it is strange there would be any impact. what build system are you using? (maven compile deps pin versions, but provided does not..)

calvinhkf:
@adriancole just to make sure it will not introduce any inconsistency...

adriancole:
works for me. Thanks!

calvinhkf:
Thanks!

PR #1155, opened 2020-04-13
Starts making zipkin2.Span optional for external reporting

adriancole:
This migrates data formerly handled only by the zipkin reporter into
`MutableSpan`. To complete decoupling will take future pull requests
that add trace IDs to `MutableSpan` and make the logging reporter use
`MutableSpan` by default instead of zipkin types.

### DIFF
--- a/brave/src/main/java/brave/Tracing.java
+++ b/brave/src/main/java/brave/Tracing.java
@@ -35,6 +35,7 @@ import java.util.Set;
 import java.util.concurrent.atomic.AtomicBoolean;
 import java.util.concurrent.atomic.AtomicReference;
 import java.util.logging.Level;
+import zipkin2.Endpoint;
 import zipkin2.reporter.Reporter;
 
 /**
@@ -137,7 +138,7 @@ public abstract class Tracing implements Closeable {
   public static final class Builder {
     String localServiceName = "unknown", localIp;
     int localPort; // zero means null
-    zipkin2.reporter.Reporter<zipkin2.Span> spanReporter;
+    Reporter<zipkin2.Span> spanReporter;
     Clock clock;
     Sampler sampler = Sampler.ALWAYS_SAMPLE;
     CurrentTraceContext currentTraceContext = CurrentTraceContext.Default.inheritable();
@@ -199,7 +200,7 @@ public abstract class Tracing implements Closeable {
      * #localPort(int)}. Will be removed in Brave v6.
      */
     @Deprecated
-    public Builder endpoint(zipkin2.Endpoint endpoint) {
+    public Builder endpoint(Endpoint endpoint) {
       if (endpoint == null) throw new NullPointerException("endpoint == null");
       this.localServiceName = endpoint.serviceName();
       this.localIp = endpoint.ipv6() != null ? endpoint.ipv6() : endpoint.ipv4();
@@ -208,9 +209,10 @@ public abstract class Tracing implements Closeable {
     }
 
     /**
-     * Controls how {@linkplain TraceContext#sampled() remote sampled} spans report to a {@code
-     * io.zipkin.reporter2:zipkin-reporter} destination. This input is usually a {@link
-     * zipkin2.reporter.AsyncReporter} which batches spans before reporting them.
+     * Controls how {@linkplain TraceContext#sampled() remote sampled} spans report to a
+     * <a href="https://github.com/openzipkin/zipkin-reporter-java">io.zipkin.reporter2:zipkin-reporter</a>destination.
+     * This input is usually a {@link zipkin2.reporter.AsyncReporter} which batches spans before
+     * reporting them.
      *
      * <p>For example, here's how to batch send spans via http:
      *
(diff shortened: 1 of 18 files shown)


brave/src/main/java/brave/Tracing.java
```diff
@@ -34,10 +35,7 @@
 import java.util.concurrent.atomic.AtomicBoolean;
 import java.util.concurrent.atomic.AtomicReference;
 import java.util.logging.Level;
-import java.util.logging.Logger;
-import zipkin2.reporter.AsyncReporter;
 import zipkin2.reporter.Reporter;
```
anuraaga, line 38: Should this import be gone too?

brave/src/main/java/brave/Tracing.java
```diff
@@ -210,11 +208,9 @@ public Builder endpoint(zipkin2.Endpoint endpoint) {
     }
 
     /**
-     * Controls how spans are reported. Defaults to logging, but often an {@link AsyncReporter}
-     * which batches spans before sending to Zipkin.
-     *
-     * The {@link AsyncReporter} includes a {@link Sender}, which is a driver for transports like
-     * http, kafka and scribe.
+     * Controls how {@linkplain TraceContext#sampled() remote sampled} spans report to a {@code
+     * io.zipkin.reporter2:zipkin-reporter} destination. This input is usually a {@link
```
anuraaga, line 212: Maybe an `<a>` tag linking to the module or something. Currently, it's hard to know this is a maven artifact.

brave/src/main/java/brave/Tracing.java
anuraaga: That being said, I think it's still fairly standard to import even from optional dependencies.

brave/src/main/java/brave/internal/handler/MutableSpanConverter.java
```diff
@@ -65,38 +76,30 @@ void convert(MutableSpan span, Span.Builder result) {
...
-    if (serviceName == null) serviceName = localServiceName;
-    if (ip == null) ip = localIp;
-    if (port <= 0) port = localPort;
-    if (localServiceName.equals(serviceName)
-      && (localIp == null ? ip == null : localIp.equals(ip))
-      && localPort == port) {
-      span.localEndpoint(localEndpoint);
+  void addLocalEndpoint(@Nullable String serviceName, @Nullable String ip, int port, Builder span) {
+    if (serviceName != null) serviceName = serviceName.toLowerCase(Locale.ROOT);
+    if (hashEndpointParameters(serviceName, ip, port) == defaultEndpointHashCode) {
```
anuraaga, line 88: Think you still need to compare the values in case of hash collision

brave/src/main/java/brave/internal/handler/MutableSpanConverter.java
```diff
@@ -65,38 +76,30 @@ void convert(MutableSpan span, Span.Builder result) {
...
+      span.localEndpoint(defaultEndpoint);
     } else {
       span.localEndpoint(Endpoint.newBuilder().serviceName(serviceName).ip(ip).port(port).build());
     }
   }
 
-  enum Consumer implements TagConsumer<Span.Builder>, AnnotationConsumer<Span.Builder> {
+  enum Consumer implements TagConsumer<Builder>, AnnotationConsumer<Builder> {
```
anuraaga, line 95: Possibly was due to refactoring - suggest putting Span back here

brave/src/main/java/brave/handler/MutableSpan.java
```diff
@@ -74,113 +83,204 @@
...
-      && tags.isEmpty()
-      && annotations == null
-      && error == null;
-  }
-
-  /** Returns the {@link brave.Span#name(String) span name} or null */
+  }
+
+  /** @since 5.12 */
+  public MutableSpan(MutableSpan toCopy) {
```
anuraaga, line 91: If not mistaken, this has been added mainly to support `defaultSpan`. Since `defaultSpan` only has three fields specified, it seems like it might be a net-loss to copy all these other default fields when creating a span.

brave/src/main/java/brave/internal/handler/MutableSpanConverter.java
adriancole: fair enough :)

adriancole: I didn't think about this. luckily somewhere I have a string that has a different value with same hashcode.. /me digs

brave/src/main/java/brave/Tracing.java
adriancole: no worries I'll revert it. this was accidental as I was trying to find where everything was called.

brave/src/main/java/brave/Tracing.java
adriancole: good idea

brave/src/main/java/brave/internal/handler/MutableSpanConverter.java
adriancole: it was line wrap, but yeah ok.

brave/src/main/java/brave/handler/MutableSpan.java
adriancole: not exactly but close. empty is checked, so we have to cycle through all state anyway. seems higher leverage to implement equals vs do maintenance on 3 fields that we expect to be the comparison state, yet could drift later for any reason (ex if we implement default tags with MutableSpan later)

brave/src/main/java/brave/internal/handler/MutableSpanConverter.java
adriancole: thanks for the tip on gitter https://dzone.com/articles/what-is-wrong-with-hashcode-in-javalangstring

adriancole:
ok I think this is done

PR #1156, opened 2020-04-13
Adds trace IDs to MutableSpan

adriancole:
This allows features only possible prior with zipkin model:
* re-parenting span to the local root (skeletal spans)
* resetting trace ID to an external one (secondary sampling)

One simplifying change is that this only tracks empty span allocation
when `Tracing.Builder.trackOrphans()` is set. This is because we no
longer have an accidental way of creating them with
`Tracer.currentSpan()`


brave/src/main/java/brave/handler/MutableSpan.java
```diff
@@ -90,6 +91,9 @@ public MutableSpan() {
   /** @since 5.12 */
   public MutableSpan(MutableSpan toCopy) {
     if (toCopy == null) throw new NullPointerException("toCopy == null");
+    traceId = toCopy.traceId;
+    parentId = toCopy.parentId;
```
anuraaga, line 95: Missing `localRootId`

adriancole: ooohh I can use the copy constructor in the existing tests! thanks for the keen eye meanwhile!

PR #1157, opened 2020-04-14
Moves construct from trace context responsibility to MutableSpan

adriancole:
This moves initializing from a given trace context from the internal
type `PendingSpans` to `MutableSpan` where it is easier to test, and
also can be used when we separate out orphan tracking to its own type.


adriancole:
merging as this helps me rebase the span listener thing

PR #1158, opened 2020-04-14
Fixes error parsing of anonymous throwables

adriancole:
IIRC we use anonymous throwables in Armeria (don't recall if all have a message or not)

cc @trustin @minwoox


brave/src/main/java/brave/ErrorParser.java
```diff
@@ -23,6 +23,17 @@
 // This implementation works with SpanCustomizer, MutableSpan and ScopedSpan, which don't share a
 // common interface, yet both support tag and annotations.
 public class ErrorParser {
+  static final ErrorParser INSTANCE = new ErrorParser();
+
+  /**
+   * Returns a singleton default instance.
+   *
+   * @since 5.11
```
adriancole, line 31:
```suggestion
   * @since 5.12
```

PR #1159, opened 2020-04-14
Removes zipkin dependency from Brave core

adriancole:
This implements json directly in MutableSpan in order to decouple Brave
from Zipkin types. Future work can decouple ancillary modules such as
tests or make a high-performance codec for zipkin-reporter-java (or even
a non zipkin-reporter-java variant).

This allows vendors to implement interfaces such as `FinishedSpanHandler`
without pulling zipkin types (reducing the size of their agents)


NOTICE
```diff
@@ -11,6 +11,13 @@ by Google in the Guava Library:
   * License: Apache License v2.0
   * Homepage: https://github.com/google/guava
 
+This product contains a modified portion of 'JsonUtf8Writer', distributed
+by Square, Inc in the Moshi Library:
+
+  * Copyright (C) 2010 Google Inc.
```
adriancole, line 17: this is actually correct even if it seems like copy/paste error https://github.com/square/moshi/blob/2cf3f0204c38d62f14d258bc671c010ae6f0ea83/moshi/src/main/java/com/squareup/moshi/JsonUtf8Writer.java

adriancole:
Eventhough we are directly implementing json, the jar is still less than 250KiB (220) and with zero deps and a floor version of Java 1.6

brave/src/main/java/brave/handler/MutableSpan.java
```diff
@@ -746,6 +748,129 @@ static boolean updateOrRemove(ArrayList list, int i, Object value, @Nullable Obj
       && equal(error, that.error);
   }
 
+  /** Writes this span in Zipkin V2 format */
+  // Ported from zipkin2.internal.V2SpanWriter and may eventually move to a separate codec type
+  @Override public String toString() {
+    StringBuilder b = new StringBuilder();
```
anuraaga, line 754: Forget if we have a ThreadLocal StringBuilder buffer, it would probably help here. But can do it later if we don't have it conveniently available.

brave/src/main/java/brave/internal/JsonEscaper.java
```diff
@@ -0,0 +1,83 @@
...
+ * Unless required by applicable law or agreed to in writing, software distributed under the License
+ * is distributed on an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express
+ * or implied. See the License for the specific language governing permissions and limitations under
+ * the License.
+ */
+package brave.internal;
+
+// Initially, a copy of zipkin2.internal.JsonEscaper
+public final class JsonEscaper {
+  public static CharSequence jsonEscape(CharSequence v) {
```
anuraaga, line 18: Should be a pretty simple win to accept `StringBuilder` as argument here instead of allocating inside.

adriancole: good point!

brave/src/main/java/brave/handler/MutableSpan.java
adriancole: if we bound it I suppose we could? This is currently only used in logging and I'm not sure if we want to leave a stringbuilder on each thread just in case toString is called...

adriancole: in a reporter totally agree (as the reporter thread is likely only 1 or few)

adriancole: by logging I mean if no-one sets a handler or a spanReporter

anuraaga: Ah yeah that sounds good

adriancole:
before closing this out, I'll make a more complete codec thing (that calculates size) in the test tree, only to ensure we have no accidental reliance on package-private stuff to make a real codec.

brave/src/main/java/brave/Tracing.java
```diff
@@ -403,8 +406,19 @@ public Builder trackOrphans() {
...
       return new Default(this);
     }
+  }
 
-    Builder() {
+  static final class LogFinishedSpanHandler extends FinishedSpanHandler {
+    final Logger logger = Logger.getLogger(Tracer.class.getName());
+
+    @Override public boolean handle(TraceContext context, MutableSpan span) {
+      if (!logger.isLoggable(Level.INFO)) return false;
```
usulkies, line 415: @adriancole This logging created SO much noise in logs.
I suggest that it should be a lower level of logging (not INFO) or (and maybe in addition) let the logger be logger of `LogFinishedSpanHandler` and not the main `Tracer.class` logger so we can configure this logger specifically.
Currently the only way to reduce the noise is to change the `Tracer` logger to log only WARN and above.

adriancole: moving this question to here https://github.com/openzipkin/zipkin-support/issues/22 to avoid spamming here. you can also use https://gitter.im/openzipkin/zipkin next time.

PR #1161, opened 2020-04-15
Adds javadoc publication to gh-pages

### LINKED ISSUE #1160: publish javadoc to gh_pages like we do in zipkin
we have such nice javadoc, shame it isn't published! we can copy/pasta from travis/publish.sh in zipkin

adriancole:
I routinely have to point users to source code in order to reference
Javadoc. This is a sub-par rendering experience vs a pointer to
something nicely rendered like

https://zipkin.io/zipkin/2.11.12/zipkin/index.html

Fixes #1160

### DIFF
--- a/travis/publish.sh
+++ b/travis/publish.sh
@@ -1,6 +1,6 @@
 #!/usr/bin/env bash
 #
-# Copyright 2013-2019 The OpenZipkin Authors
+# Copyright 2015-2020 The OpenZipkin Authors
 #
 # Licensed under the Apache License, Version 2.0 (the "License"); you may not use this file except
 # in compliance with the License. You may obtain a copy of the License at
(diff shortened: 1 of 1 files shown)


adriancole:
ps duly noted that people can use this while it runs, it is easier https://javadoc.io/doc/io.zipkin.brave/brave/latest/index.html

The main rationale here is permalinks under the premise that gh-pages is more stable than other things (though we can also make a cname for brave under zipkin.io which would give more flexibility in the future..)

travis/publish.sh
```diff
@@ -104,6 +104,49 @@ safe_checkout_master() {
...
+  # Update simple version-level index
+  if ! grep "$version" index.html 2>/dev/null; then
+    echo "<li><a href=\"${version}/index.html\">${version}</a></li>" >> index.html
+  fi
+
+  # Ensure links are ordered by versions, latest on top
+  sort -rV index.html > index.html.sorted
+  mv index.html.sorted index.html
+
+  git add "$version"
```
adriancole, line 144: @abesto I'm wondering if here and also zipkin, if it wouldn't make sense to add the similar logic we do in docker tags to link latest, major, and major.minor

travis/publish.sh
```diff
@@ -104,6 +104,49 @@ safe_checkout_master() {
...
+  # Update gh-pages
+  git fetch origin gh-pages:gh-pages
+  git checkout gh-pages
+  rm -rf "$version"
+  mv "javadoc-builddir/$version" ./
+  rm -rf "javadoc-builddir"
+
+  # Update simple version-level index
+  if ! grep "$version" index.html 2>/dev/null; then
+    echo "<li><a href=\"${version}/index.html\">${version}</a></li>" >> index.html
```
adriancole, line 137: in looking at these, it might be a little much to have the entire version as we don't typically change api on patch release (it is a bit wordy to say "brave core api 5.11.2" in the header of each index.html)

adriancole:
setup the orphan gh-pages branch https://zipkin.io/brave/

PR #1162, opened 2020-04-19
Can be customized Clock.

wangliang1986:
Can be customized Clock.
For example: using database TickClock.

### DIFF
--- a/brave/src/main/java/brave/internal/recorder/ClockFactory.java
+++ b/brave/src/main/java/brave/internal/recorder/ClockFactory.java
@@ -1,0 +1,30 @@
+package brave.internal.recorder;
+
+import brave.Clock;
+
+/**
+ * Clock factory. Can be customized the clock instant
+ *
+ * @author wangliang <841369634@qq.com
+ * @date 2020/4/19 14:45
+ */
+public class ClockFactory {
+
+  private static Clock instant;
+
+  public static Clock getInstant() {
+    return instant;
+  }
+
+  public static void setInstant(Clock clock) {
+    instant = clock;
+  }
+
+  public static Clock build(long baseEpochMicros, long baseTickNanos) {
+    if (instant != null) {
+      return instant;
+    } else {
+      return new TickClock(baseEpochMicros, baseTickNanos);
+    }
+  }
+}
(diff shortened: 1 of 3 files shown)


adriancole:
how is this different than Tracing.Builder.clock() for you?

ex this exists:
```java
    /**
     * Assigns microsecond-resolution timestamp source for operations like {@link Span#start()}.
     * Defaults to JRE-specific platform time.
     *
     * <p>Note: timestamps are read once per trace, then {@link System#nanoTime() ticks}
     * thereafter. This ensures there's no clock skew problems inside a single trace.
     *
     * See {@link Tracing#clock(TraceContext)}
     */
    public Builder clock(Clock clock) {
      if (clock == null) throw new NullPointerException("clock == null");
      this.clock = clock;
      return this;
    }
```

wangliang1986:
> 
> 
> how is this different than Tracing.Builder.clock() for you?
> 
> ex this exists:
> 
> ```java
>     /**
>      * Assigns microsecond-resolution timestamp source for operations like {@link Span#start()}.
>      * Defaults to JRE-specific platform time.
>      *
>      * <p>Note: timestamps are read once per trace, then {@link System#nanoTime() ticks}
>      * thereafter. This ensures there's no clock skew problems inside a single trace.
>      *
>      * See {@link Tracing#clock(TraceContext)}
>      */
>     public Builder clock(Clock clock) {
>       if (clock == null) throw new NullPointerException("clock == null");
>       this.clock = clock;
>       return this;
>     }
> ```

Oh, I don't know it exists. How do I use it?

adriancole:
If you have a different clock you can specify it. to change timestamps externally is also possible (Span.start(long)) just, this can create skew in a trace. Mostly you don't need to change this unless your operating system has a better clock than java can provide.

(this is implemented in brave.internal.Platform)

hope these notes help. if you have more questions, try gitter https://gitter.im/openzipkin/zipkin

wangliang1986:
> 
> 
> If you have a different clock you can specify it. to change timestamps externally is also possible (Span.start(long)) just, this can create skew in a trace. Mostly you don't need to change this unless your operating system has a better clock than java can provide.
> 
> (this is implemented in brave.internal.Platform)
> 
> hope these notes help. if you have more questions, try gitter https://gitter.im/openzipkin/zipkin

In spring cloud, different hosts will have different times, and I am lazy to do synchronization, so I hope to achieve the same time standard through the same time source.

adriancole:
please use gitter to talk more. typically people use a service like NTP to change host clock instead of trying to change just the tracing clock (and not logs etc)

PR #1163, opened 2020-04-19
Replaces "carrier" with "request"

adriancole:
We prematurely adopted the confusing term "carrier" when trying to work
with OpenTracing years ago. This removes that term for request
everywhere. This does not break any api, as the only thing besides
parameter renames are test methods.

This is in preparation of a future Brave 6 world where ideally no one
will ever need to be confused about what "carrier" is, only for us to
say "an awkward word for request".

### DIFF
--- a/brave/src/main/java/brave/internal/baggage/BaggageHandler.java
+++ b/brave/src/main/java/brave/internal/baggage/BaggageHandler.java
@@ -70,22 +70,22 @@ public interface BaggageHandler<S> {
   @Nullable S updateState(S state, BaggageField field, @Nullable String value);
 
   /**
-   * Extracts any state from a remote value received by {@link Propagation.Getter#get(Object,
+   * Extracts any state from a request value received by {@link Propagation.Getter#get(Object,
    * Object)}.
    *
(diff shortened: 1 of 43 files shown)


brave/src/main/java/brave/internal/baggage/BaggageHandler.java
```diff
@@ -70,22 +70,22 @@
...
+   * Extracts any state from a request value received by {@link Propagation.Getter#get(Object,
    * Object)}.
    *
-   * <p>Ex. When the state is a simple string, this will just use the remote value directly.
+   * <p>Ex. When the state is a simple string, this will just use the request value directly.
    * {@linkplain #isDynamic() Dynamic values} will need to perform some decoding, such as splitting
    * on comma and equals.
    */
-  @Nullable S fromRemoteValue(String remoteValue);
+  @Nullable S fromRequestValue(Object request, String value);
```
adriancole, line 80: we aren't currently using this, but we will need this for secondary sampling. This was what I was changing and realized.. wow, we should really stop using so many weird terms!

adriancole:
merging this as I highly doubt any remaining reviewers disagree :) anyway do comment if otherwise

PR #1166, opened 2020-04-20
Moves FinishedSpanHandler.alwaysSampleLocal to Tracing.Builder

adriancole:
When FinishedSpanHandler.alwaysSampleLocal was defined, we didn't have
TracingCustomizer which allowed looser coupling of configuration. This
moves `FinishedSpanHandler.alwaysSampleLocal()` to `Tracing.Builder` so
we don't have to inherit this burden in the upcoming `SpanHandler`.


brave/src/main/java/brave/Tracing.java
```diff
@@ -455,6 +478,10 @@ public Tracing build() {
       // Make sure any exceptions caused by handlers don't crash callers
       FinishedSpanHandler finishedSpanHandler =
         NoopAwareFinishedSpanHandler.create(spanHandlers, noop);
+      boolean alwaysSampleLocal = builder.alwaysSampleLocal;
+      for (FinishedSpanHandler handler : spanHandlers) {
+        if (handler.alwaysSampleLocal()) alwaysSampleLocal = true;
```
anuraaga, line 483: `// Handle deprecated FinishedSpanHandler.alwaysSampleLocal`

PR #1167, opened 2020-04-20
Deprecates Tracing.Builder.spanReporter

adriancole:
Notably this uses zipkin-reporter-brave directly, and deprecates ErrorParser for Tags.ERROR


brave/README.md
```diff
@@ -56,7 +59,7 @@ The tracer creates and joins spans that model the latency of potentially
 distributed work. It can employ sampling to reduce overhead in process
 or to reduce the amount of data sent to Zipkin.
 
-Spans returned by a tracer report data to Zipkin when finished, or do
+Spans returned by a tracer report data to Zipkin when , or do
```
adriancole, line 62:
```suggestion
Spans returned by a tracer report data to Zipkin when finished, or do
```

brave/README.md
```diff
@@ -590,6 +593,69 @@ customizer.customize(tracingBuilder);
...
+  nonZipkinReporter.report(converted);
+  return true;
+}
+```
+
+### Child Counting Example
+Some data formats desire knowing how many spans a parent created. Below is an
+example of how to do that, using [WeakConcurrentMap](https://github.com/raphw/weak-lock-free).
+
+```
```
adriancole, line 625:
```suggestion
```java
```

trajano:
HI I saw this deprecation when I updated, but I can't locate the release note on what the replacement should be.  The @Deprecated does not document what the replacement should be.

PR #1168, opened 2020-04-21
Adds more rationale and fixes some fuzz in HTTP

adriancole:


brave/RATIONALE.md
```diff
@@ -43,6 +43,72 @@ Brave 4.19 added `ErrorParser`, internally used by `Span.error(Throwable)`. This
 design incubated in [Spring Cloud Sleuth](https://github.com/spring-cloud/spring-cloud-sleuth) prior to becoming a primary type here.
 Our care and docs around error handling in general is credit to [Nike Wingtips](https://github.com/Nike-Inc/wingtips#warning-about-error-handling-when-using-try-with-resources-to-autoclose-spans).
 
+### Why we don't tag "error=false"
+"error=false" is almost always the result of mis-mapping from OpenTracing who
```
adriancole, line 47: Added this section due to a chat with @jcchavezs. Basically the first thing we should focus on is that "error=true" or "error=false".. this is usually a modeling mistake. Before getting into the nuance of booleans, we should help people out of mistakes.

PR #1170, opened 2020-04-22
Makes Dubbo Error code words not numbers

adriancole:
This makes "dubbo.error_code" use the code name, like "TIMEOUT_EXCEPTION", as
opposed to the code number like 2.

This primarily matches conventions being defined in #999, where we prefer error
code names over numbers. The below rationale is a copy/paste from the new
RATIONALE.md for Dubbo, and helps explain the subtle points here vs HTTP status.

---

HTTP status codes are grouped by classification, and have existed so long that
support teams can usually identify meaning by looking at a number like 401.
Being triple digits, it is relatively easy to search for what an HTTP status
means.

Dubbo code numbers are more like enum ordinals. There's no grouping and it is
not easy to identify a problem quickly by seeing a number like 2 vs the code
name "TIMEOUT_EXCEPTION". There is no generic documentation on Dubbo errors. If
given only the number 2, a user unfamiliar with how Dubbo works internally will
have a hard time. For example, searching Dubbo's code base for "2" will return
less relevant results than searching for "TIMEOUT_EXCEPTION".

It may seem that exception messages can overcome this problem, and they
certainly can when present. Also, exception messages can be localized. However,
exception messages are not good for trace search because they are long and
contain variables.

For all these reasons, we use code names, not numbers, for Dubbo, as defined in
`RpcException`'s constants (so that they are easy to search).

### DIFF
--- a/instrumentation/dubbo-rpc/pom.xml
+++ b/instrumentation/dubbo-rpc/pom.xml
@@ -32,7 +32,7 @@
     <main.java.version>1.6</main.java.version>
     <main.signature.artifact>java16</main.signature.artifact>
     <!-- Use brave-instrumentation-dubbo for Apache Dubbo 2.7+, not this module. -->
-    <dubbo.version>2.6.7</dubbo.version>
+    <dubbo.version>2.6.8</dubbo.version>
   </properties>
 
   <dependencies>
--- a/instrumentation/dubbo-rpc/src/main/java/brave/dubbo/rpc/DubboParser.java
+++ b/instrumentation/dubbo-rpc/src/main/java/brave/dubbo/rpc/DubboParser.java
@@ -69,19 +69,28 @@ final class DubboParser {
   }
 
   /**
-   * We decided to not map Dubbo codes to human readable names like {@link
-   * RpcException#BIZ_EXCEPTION} even though we defined "rpc.error_code" as a human readable name.
-   *
-   * <p>The reason was a comparison with HTTP status codes, and the choice was between returning
-   * just numbers or reusing "UNKNOWN_EXCEPTION" which is defined in Dubbo for code "0" for any
-   * unknown code. Returning numbers was the less bad option as it doesn't conflate code words.
-   *
-   * <p>Later, we can revert this back to code words, but once this gets into the RPC mapping for
-   * Dubbo it will be hard to change.
+   * This library is no-longer being released, so it should not have any maintenance on error codes.
+   * The error codes here were defined in 2012.
    */
   @Nullable static String errorCode(Throwable error) {
     if (error instanceof RpcException) {
-      return String.valueOf(((RpcException) error).getCode());
+      int code = ((RpcException) error).getCode();
+      switch (code) {
+        case RpcException.UNKNOWN_EXCEPTION:
+          return "UNKNOWN_EXCEPTION";
+        case RpcException.NETWORK_EXCEPTION:
+          return "NETWORK_EXCEPTION";
+        case RpcException.TIMEOUT_EXCEPTION:
+          return "TIMEOUT_EXCEPTION";
+        case RpcException.BIZ_EXCEPTION:
+          return "BIZ_EXCEPTION";
+        case RpcException.FORBIDDEN_EXCEPTION:
+          return "FORBIDDEN_EXCEPTION";
+        case RpcException.SERIALIZATION_EXCEPTION:
+          return "SERIALIZATION_EXCEPTION";
+        default:
+          return String.valueOf(code);
+      }
     }
     return null;
   }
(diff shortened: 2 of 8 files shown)


instrumentation/dubbo-rpc/pom.xml
```diff
@@ -32,7 +32,7 @@
     <main.java.version>1.6</main.java.version>
     <main.signature.artifact>java16</main.signature.artifact>
     <!-- Use brave-instrumentation-dubbo for Apache Dubbo 2.7+, not this module. -->
-    <dubbo.version>2.6.7</dubbo.version>
+    <dubbo.version>2.6.8</dubbo.version>
```
adriancole, line 35: I hadn't realized Dubbo are still releasing the Alibaba library.. I bumped just to make sure they are still not adding error codes.

instrumentation/dubbo-rpc/src/main/java/brave/dubbo/rpc/DubboParser.java
```diff
@@ -69,19 +69,28 @@ static boolean parseRemoteIpAndPort(Span span) {
...
+        case RpcException.TIMEOUT_EXCEPTION:
+          return "TIMEOUT_EXCEPTION";
+        case RpcException.BIZ_EXCEPTION:
+          return "BIZ_EXCEPTION";
+        case RpcException.FORBIDDEN_EXCEPTION:
+          return "FORBIDDEN_EXCEPTION";
+        case RpcException.SERIALIZATION_EXCEPTION:
+          return "SERIALIZATION_EXCEPTION";
+        default:
+          return String.valueOf(code);
```
anuraaga, line 92: I get where this is coming from. Just to share only one point I'm concerned about - if some servers in a fleet use a new version of instrumentation and others don't, the same exception shows up as name in some, int in others. It does seem quite rare though since why not bump instrumentation when bumping Dubbo? So all good here.

adriancole: the only other option I can think of is using reflection in to lookup the symbols. Ex search for public static ints whose name ends in _EXCEPTION

that would make it no maintenance and since they are all public, unlikely to cause a weird crash.. wdyt?

adriancole: I will give it a try. we use one-time reflection in a lot of code, like servlet etc. we would have heard complaints by now if someone is unable to use reflection at all..

anuraaga: Thanks, that sounds like a nice approach!

adriancole: couldn't use _EXCEPTION as they are not consistent :) we blindly trust int constants for now. It is unlikely to cause an issue I think and I agree this is better. thanks, rag!

adriancole:
PS I didn't use reflection on the alibaba side as eventhough they are still patching it, I highly doubt they will add error codes. If they do, we can revisit, but that's low risk.

instrumentation/dubbo/src/main/java/brave/dubbo/DubboParser.java
```diff
@@ -25,6 +29,36 @@
...
+   * constants.
+   */
+  static Map<Integer, String> errorCodeNumberToName() {
+    Map<Integer, String> result = new LinkedHashMap<>();
+    for (Field field : RpcException.class.getDeclaredFields()) {
+      if (Modifier.isStatic(field.getModifiers()) && field.getType() == int.class) {
+        try {
+          result.put((Integer) field.get(null), field.getName());
+        } catch (Exception e) {
+          assert false : e.getMessage(); // make all unit tests fail
```
anuraaga, line 50: Nice

instrumentation/dubbo/src/main/java/brave/dubbo/DubboParser.java
```diff
@@ -25,6 +29,36 @@
...
+   * would not be seen until an upgrade of Brave. That, or we'd have to fall back to a code number
+   * until the upgrade of Brave converged.
+   *
+   * <p>This uses reflection instead, which is unlikely to fail as Dubbo's error codes are public
+   * constants.
+   */
+  static Map<Integer, String> errorCodeNumberToName() {
+    Map<Integer, String> result = new LinkedHashMap<>();
+    for (Field field : RpcException.class.getDeclaredFields()) {
+      if (Modifier.isStatic(field.getModifiers()) && field.getType() == int.class) {
```
anuraaga, line 46: How about checking `Modifier.isPublic` too?

adriancole: yeah good idea. it makes more sense reading it now, also.

PR #1171, opened 2020-04-24
Adds BaggagePropagation.allKeyNames and deprecates non-string propagation

adriancole:
In 5.11, we partially replaced functionality of ExtraFieldPropagation.extraKeys().
This completes the work after I noticed it was needed for several use cases, not
just OpenTracing. `BaggagePropagation.allKeyNames()` is also more robust when
propagation is wrapped by proxy or otherwise.

The Propagation `<K>` parameter allowed integration with requests that had
headers, but didn't use String types. This was over-generalized, as in practice
only gRPC instrumentation ever used this key type (`Metadata.Key`).

Removing this `<K>` parameter dramatically simplifies the model as it removes
the need to explain the key factory and the edge case it supported, which can
be accomplished differently.

In Brave 6, we will need to keep the generic parameter `Propagation<String>`
(to avoid API breaks). However, the internals will be far less complex once we
can remove KeyFactory etc that are no longer needed.


brave/src/main/java/brave/baggage/BaggagePropagation.java
```diff
@@ -259,6 +240,45 @@ public FactoryBuilder add(BaggagePropagationConfig config) {
...
+  // On OpenTracing TextMap: https://github.com/opentracing/opentracing-java/issues/305
+  public static List<String> allKeyNames(Propagation<String> propagation) {
+    if (propagation == null) throw new NullPointerException("propagation == null");
+    // When baggage or similar is in use, the result != TraceContextOrSamplingFlags.EMPTY
+    TraceContextOrSamplingFlags emptyExtraction =
+      propagation.extractor((c, k) -> null).extract(Boolean.TRUE);
+    List<String> baggageKeyNames = ExtraBaggageContext.getAllKeyNames(emptyExtraction);
+    if (baggageKeyNames.isEmpty()) return propagation.keys();
+
+    List<String> result = new ArrayList<>(propagation.keys());
```
anuraaga, line 277: It might be a tiny bit more performant to presize instead of precopy

```java
List<String> result = new ArrayList<>(propagation.keys().size()+ baggageKeyNames.size());
result.addAll(propataion.keys();
result.addAll(baggageKeyNames);
return unmodifiable(result);
```

adriancole: sure thing

adriancole:
@anuraaga fyi the first time you looked at this, I accidentally didn't push half the change. (git messup)

brave/src/main/java/brave/baggage/BaggagePropagation.java
```diff
@@ -138,92 +139,72 @@ public FactoryBuilder add(BaggagePropagationConfig config) {
...
       }
       configs.add(field);
-      Set<String> lcKeyNames = new LinkedHashSet<>();
+      if (field.keyNames().isEmpty()) {
+        fieldToHandler.put(field.field, BaggageHandlers.string(field.field));
+        return this;
+      }
+
       for (String keyName : field.keyNames) {
-        String lcName = validateName(keyName).toLowerCase(Locale.ROOT);
```
adriancole, line 147: ps this was redundant as config keyNames are by definition lc

brave/src/main/java/brave/baggage/BaggagePropagation.java
```diff
@@ -259,6 +240,46 @@ public FactoryBuilder add(BaggagePropagationConfig config) {
...
+   * </ul>
+   *
+   * <h3>Details</h3>
+   * The {@code propagation} parameter is required because there may be multiple tracers with
+   * different baggage configuration. Also, {@link Propagation} instances can be wrapped, so you
+   * cannot use {@code instanceof} to identify if baggage is internally supported. For example,
+   * {@link ExtraFieldPropagation} internally wraps {@link BaggagePropagation}.
+   *
+   * <p>This is different than {@link BaggageField#getAll(TraceContext)}, as propagation keys may be
+   * different than {@link BaggageField#name() baggage field names}.
```
adriancole, line 262:
```suggestion
   * different than {@linkplain BaggageField#name() baggage field names}.
```

adriancole:
thanks for the look! nearing the RPC finish line now!

PR #1172, opened 2020-04-25
Refreshes main README about Request tracing

adriancole:
This attempts to clarify the main README to reduce grief for folks who read it.
Particularly, sometimes we get people modeling different things, such as local
activity or long-lived connections as CLIENT spans.

See https://github.com/openzipkin/zipkin-support/issues/2

### DIFF
--- a/brave/README.md
+++ b/brave/README.md
@@ -167,21 +167,14 @@ void userCode() {
 }
 ```
 
-### Request tracing
-Check for [instrumentation written here](../instrumentation/) and [Zipkin's list](https://zipkin.io/pages/existing_instrumentations.html)
-before rolling your own Request instrumentation!
+### Remote spans
+Check for [instrumentation and abstractions](../instrumentation/) and [Zipkin's list](https://zipkin.io/pages/existing_instrumentations.html)
+for common patterns like client-server or messaging communication. Please at
+least read [this doc](../instrumentation/README.md) before deciding to write
+your own code to represent remote communication.
 
-Brave includes two fundamental types to support tracing interprocess
-communication `brave.Request` and `brave.Response`. These types represent the
-following communication patterns, defined by the [Zipkin Api][https://zipkin.io/zipkin-api/#/default/post_spans]
-and returned by `spanKind()`:
-
- * CLIENT
- * SERVER
- * PRODUCER
- * CONSUMER
-
-Once you model your request, you generally need to...
+If you really think that the existing abstractions do not match your need, and
+you want to model your request yourself, you generally need to...
 1. Start the span and add trace headers to the request
 2. Put the span in scope so things like log integration works
 3. Invoke the request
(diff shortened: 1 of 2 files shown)


brave/README.md
```diff
@@ -167,37 +167,59 @@ void userCode() {
...
-// when the response is complete, finish the span
-span.finish();
-```
+You should use these abstractions instead of modeling your own, as there are
+modeling gotchas that might not be intuitive at first. For example, a common
+mistake is using CLIENT kind for long lived connections such as database pools
+or chat sessions. Doing so, however, can result in confusing data, such as
+incorrect service diagrams or server calls being attached to the wrong trace.
+If you have doubts, please chat with a maintainer on [Gitter](https://gitter.im/openzipkin/zipkin]
+as it might save you grief later!
```
jorgheymans, line 222: i would move this block above 'once you model your request'

brave/README.md
```diff
@@ -167,37 +167,59 @@ void userCode() {
...
+communication `brave.Request` and `brave.Response`. These types represent the
+following communication patterns, defined by the [Zipkin Api][https://zipkin.io/zipkin-api/#/default/post_spans]
+and returned by `spanKind()`:
+
+ * CLIENT
+ * SERVER
+ * PRODUCER
+ * CONSUMER
+
+Once you model your request, you generally need to...
```
jorgheymans, line 184: "If you really think that the existing abstractions do not match your need, and you want to model your request yourself, you generally ...."

brave/README.md
```diff
@@ -167,37 +167,59 @@ void userCode() {
...
+  span.finish(); // 5.
+}
+```
 
-// if there is an error, tag the span
-span.tag("error", error.getCode());
+The above code is example only. Abstractions exist that handle aspects
+including sampling, header processing and data parsing:
+ * [HTTP](../instrumentation/http/README.md) is CLIENT/SERVER
+ * [Messaging](../instrumentation/rpc/README.md) is PRODUCER/CONSUMER
```
anuraaga, line 213: Wrong href

adriancole: thx fixed in latest.  I realized we repeated a lot of this in the instrumentation README!

instrumentation/README.md
```diff
@@ -95,14 +95,52 @@ We worked very hard to make writing new instrumentation easy and efficient.
...
+
+Using abstractions also helps avoid modeling gotchas that might not be
+intuitive at first.
+
+For example, a common mistake is using CLIENT kind for long lived connections
+such as database pools or chat sessions. Doing so, however, can result in
+confusing data, such as incorrect service diagrams or server calls being
+attached to the wrong trace.
+
+It is also a common misconception that a grouping of CLIENT spans should itself
```
adriancole, line 127: @anuraaga this bud's for you

instrumentation/README.md
```diff
@@ -95,14 +95,52 @@ We worked very hard to make writing new instrumentation easy and efficient.
...
+to look for tests in the "features" package before deciding something doesn't
+work.
+* [Core feature tests](../brave/src/test/java/brave/features)
+* [HTTP feature tests](http-tests/src/test/java/brave/http/features)
+* [Messaging feature tests](messaging/src/test/java/brave/rpc/features)
+* [RPC feature tests](rpc/src/test/java/brave/rpc/features)
+
+If you end up in a position where you still believe you need a custom model,
+please contact us before committing to it. There may be an alternative not yet
+documented, that might prefer. Moreover, folks usually want support and we
```
adriancole, line 143:
```suggestion
documented, you might prefer. Moreover, folks usually want support and we
```

adriancole:
thanks for the feedback @jorgheymans and @anuraaga!

PR #1173, opened 2020-04-26
Adds rationale for client response callback context

adriancole:
I'm raising this as we've had many PRs to fix this over time, but no centralized rationale to discuss it. My motivation is to link gRPC rationale to this, but that can be decoupled.

This is a very long explanation derived from ideas sparked by @eirslett's initial rationale over 2 years ago https://github.com/openzipkin/zipkin-js/issues/176#issuecomment-355636603

It was later expanded upon during an [Armeria](https://github.com/line/armeria) pow-wow with @anuraag @kojilin and @trustin.
https://github.com/openzipkin/openzipkin.github.io/wiki/2018-08-06-Zipkin-and-Armeria-at-LINE-Fukuoka#teaching-brave-to-use-armerias-requestcontext

cc @openzipkin/core 

#1014
#1095
#1067
https://github.com/openzipkin/brave/commit/cf3b13836b81895500d8a8cf0ee2484aae304c6a

### DIFF
--- a/instrumentation/RATIONALE.md
+++ b/instrumentation/RATIONALE.md
@@ -87,6 +87,9 @@ rate, if a data dependency is important, you can consider mapping it as a tag.
 This rationale was first mentioned in the below javascript issue.
 https://github.com/openzipkin/zipkin-js/issues/176#issuecomment-355636603
 
+It was later built upon during an [Armeria](https://github.com/line/armeria) pow-wow:
+https://github.com/openzipkin/openzipkin.github.io/wiki/2018-08-06-Zipkin-and-Armeria-at-LINE-Fukuoka#teaching-brave-to-use-armerias-requestcontext
+
 ## Calling `Span.finish()` while the context is in scope
 Instrumentation should call `Span.finish()` with the same context in scope as
 opposed to clearing or using a different span context.
--- a/instrumentation/RATIONALE.md
+++ b/instrumentation/RATIONALE.md
@@ -2,6 +2,91 @@
 Rationale here applies to common decisions made in this directory. See
 [Brave's RATIONALE](../brave/RATIONALE.md) for internal rationale.
 
+## Why does the client response callback run in the invocation context?
+This rationale applies equally to CLIENT and PRODUCER spans.
+
+Asynchronous code is often modeled in terms of callbacks. For example, the
+following pseudo code represents a chain of 3 client calls.
+```java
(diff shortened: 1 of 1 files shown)


instrumentation/RATIONALE.md
```diff
@@ -2,6 +2,94 @@
...
+}
+```
+
+It might be surprising that calls "2" and "3" execute in the "parent" trace
+context, as opposed to the preceding client call. Put another way, response
+callbacks run in the invocation context, which cause new spans to appear as
+as a siblings, as opposed to children of the previous callback (in this case
+a client).
+
+This may sound intuitive to those thinking in terms of callback nesting depth,
```
adriancole, line 27:
```suggestion
This may sound unintuitive to those thinking in terms of callback nesting depth,
```

adriancole:
I switched the example to psuedo-reactive as that's increasingly common. I can switch back to something else, like future.whenComplete() if someone says to.

instrumentation/RATIONALE.md
```diff
@@ -2,6 +2,100 @@
...
+real parent in a typical trace UI!
+
+We acknowledge that using the invocation context as the parent of follow-up
+requests (response callback) is imperfect. It means any data dependency between
+one response and the next request is not represented in the hierarchy. It also
+means callback depth with not manifest in the trace hierarchy. That said,
+follow-up requests still share not just the same trace, but also the same local
+root, and also direct parent. As the clocks are the same (and in fact locked
+against skew), the happens after relationship manifests in span timing. At any
+rate, if a data dependency is important, you can consider mapping it as a tag.
```
anuraaga, line 91: How about also talking about local spans (span without kind)? I guess some people want to model grouped client spans and intuitively think the group should also be a client span, though in our model, where we want to describe inter-service links, this should be a local span.

adriancole: sounds good will add

adriancole: I think it is a different topic, so I'll add a different section.

Is there a coherent link between this and that or just something that caught your mind. I'm not sure where to join the points as this is about the response callback and not the span kind vs no span kind topic..

I know we have another #1172 about that..

adriancole: moved to https://github.com/openzipkin/brave/pull/1172

adriancole:
thanks for the feedback @anuraaga! If any others have feedback, feel free to give it or edit the file in a PR

PR #1174, opened 2020-04-26
Deprecates gRPC message processing via GrpcParser

adriancole:
Currently, there is no evidence of usage on GitHub, and also the author,
@jorgheymans confirmed they aren't using these hooks internally at the
moment.

After discussion with Jorg and @anuraaga, we decided that message
processing is better left to normal gRPC interceptors. This allows the
RPC project to finish (#999) without adding complexity for a possibly
unused feature. Meanwhile, this gives us a chance to mention the
alternative (in the README).

See also #1173 about the callback context (though same as before)


adriancole:
thanks for the fast turnaround @anuraaga!

PR #1175, opened 2020-04-27
Renovates gRPC instrumentation in preparation of standard parsing

adriancole:
This fixes some glitches in preparation of #999

* Bumps gRPC to 1.29 (there was no impact)
* GrpcClientRequest and GrpcServerRequest unwrapped `this`, not the call.
* Anonymous GrpcClientParser and GrpcServerParser used ErrorParser unnecessarily
  * This is implicitly done in the Zipkin span handler
* Status.cause was ignored
  * This led to a less precise "error" tag
* Client headers were unnecessarily null during sampling. This was caused by
  too eager allocation of a span.
* ClientCall.Listener was wrapped twice
* Most ClientCall hooks were not placed in scope
* ServerCall's span was started too late


adriancole:
FYI continuous-integration/travis-ci/pr rarely passes when you change things in the pom. ex it fails in invoker tests due to drift between master and whatever is in the PR. I'm not sure what this is about, but seems to explain why we get winky (one red one green) builds on changes sometimes.

instrumentation/grpc/src/it/grpc12/pom.xml
```diff
@@ -47,6 +47,11 @@
       <artifactId>junit</artifactId>
       <version>@junit.version@</version>
     </dependency>
+    <dependency>
+      <groupId>org.mockito</groupId>
+      <artifactId>mockito-core</artifactId>
+      <version>@mockito.version@</version>
```
anuraaga, line 53: Don't know if I've seen `@` before and not sure how to search for it. Quick pointer to what it is?

instrumentation/grpc/src/it/grpc12/src/test/resources/mockito-extensions/org.mockito.plugins.MockMaker
```diff
@@ -0,0 +1 @@
+mock-maker-inline
```
anuraaga, line 1: Maybe not since it's exposed out of tree, but could consider putting these resources in `brave-test`

instrumentation/grpc/src/main/java/brave/grpc/GrpcClientResponse.java
```diff
@@ -0,0 +1,84 @@
...
+
+// intentionally not yet public until we add tag parsing functionality
+final class GrpcClientResponse extends Response {
+  final GrpcClientRequest request;
+  @Nullable final Status status;
+  @Nullable final Metadata trailers;
+  @Nullable final Throwable error;
+
+  GrpcClientResponse(GrpcClientRequest request,
+    @Nullable Status status, @Nullable Metadata trailers, @Nullable Throwable error) {
```
anuraaga, line 31: I think `status` or `trailers` are never `null`

instrumentation/grpc/src/main/java/brave/grpc/TracingClientInterceptor.java
```diff
@@ -127,38 +192,23 @@
...
-        delegate().onMessage(message);
-      }
-    }
-
-    @Override public void onClose(Status status, Metadata trailers) {
-      try (Scope scope = currentTraceContext.maybeScope(span.context())) {
-        super.onClose(status, trailers);
-        parser.onClose(status, trailers, span.customizer());
+      } catch (Throwable e) {
+        error = e;
```
anuraaga, line 207: Not sure but I think this would only happen if there is an error in the business logic on the client side when processing the closure, even if the request itself succeeded. So this might not warrant marking the span as error.

instrumentation/grpc/src/main/java/brave/grpc/TracingServerInterceptor.java
```diff
@@ -91,96 +99,119 @@ Span nextSpan(TraceContextOrSamplingFlags extracted, GrpcServerRequest request)
...
     @Override public void close(Status status, Metadata trailers) {
-      try (Scope scope = currentTraceContext.maybeScope(span.context())) {
-        super.close(status, trailers);
-        parser.onClose(status, trailers, span.customizer());
+      Throwable error = null;
+      try (Scope scope = currentTraceContext.maybeScope(context)) {
+        delegate().close(status, trailers);
       } catch (Throwable e) {
-        span.error(e);
+        error = e;
```
anuraaga, line 155: I'm less clear about this one since closure on the server side is handled by the framework, so an error may or may not indicate the response failed. Ideally, our interceptor is in a location where the only possibility of an exception here is a grpc-java bug.

instrumentation/grpc/src/main/java/brave/grpc/GrpcClientResponse.java
```diff
@@ -0,0 +1,84 @@
...
+    if (request == null) throw new NullPointerException("request == null");
+    this.request = request;
+    this.status = status;
+    this.trailers = trailers;
+    this.error = error != null ? error : status != null ? status.getCause() : null;
```
anuraaga, line 36: Related to my other comments, instead of passing in an error, for gRPC it's probably more correct to just always use `status.getCause()`. The framework is supposed to convert any business logic exception into a `Status` with the cause set to that exception.

instrumentation/grpc/src/main/java/brave/grpc/TracingClientInterceptor.java
```diff
@@ -30,95 +32,158 @@
...
+        responseListener,
+        invocationContext,
+        spanRef,
+        request
+      );
+
+      try (Scope scope = currentTraceContext.maybeScope(span.context())) {
+        super.start(responseListener, headers);
+      } catch (Throwable e) {
+        finishWithError(spanRef.getAndSet(null), e);
```
anuraaga, line 120: I think close will be called with this error packed into the `Status` and we probably don't need to specially handle it.

instrumentation/grpc/src/it/grpc12/pom.xml
adriancole: maven-invoker-plugin isolates the build environment, so you have to opt-in to inherit anything. This syntax means the property is lifted from the invoking environment

http://maven.apache.org/plugins/maven-invoker-plugin/examples/filtering.html

instrumentation/grpc/src/it/grpc12/src/test/resources/mockito-extensions/org.mockito.plugins.MockMaker
adriancole: brave-test doesn't currently depend on mockito, and mockito versions have been painful in the past.. good idea in general, but I feel I don't want to move anything mockito related to a wider scope.. make sense?

instrumentation/grpc/src/main/java/brave/grpc/GrpcClientResponse.java
adriancole: I think what you are saying in general is that we should assume grpc will always call close() even if client interceptors/handlers etc fail.

I'm cool with that. We do have code that is in other libraries that we should still be careful about. In the case that we find grpc has bugs what will happen is that we will leak spans. This is imho ok as such leaks we can address in a non-api affecting patch release.

instrumentation/grpc/src/main/java/brave/grpc/GrpcClientResponse.java
adriancole: when we assume close is always called, I agree.. this ties to your other point, which I'm ok documenting in RATIONALE so in case the assumption proves wrong, we can update and move do this again

Just note if it does prove wrong we will have to mark something we said was not nullable nullable, which would surprise people.

Please ack if you feel confident enough that we should assume grpc always calls close, as that's not an assumption we make elsewhere, and frankly I would be surprised if that's provable.

anuraaga: Thanks - after writing some tests, realized that indeed `close` isn't always called when interceptors are badly behaving.

For the nullability in particular, I noticed we only call `new Grpc*Response` in `close` though, elsewhere we just finish the span directly without using this class. I guess it makes sense, that a GrpcResponse is a gRPC response, where gRPC responses always have status / trailers as per the wire format, and any other case there is no response and this wouldn't be initialized, but maybe we're anticipating such a case so keeping it loose is fine too.

adriancole: appreciate the help with the tests as without them this is all folklore!

FWIW the code with "finish" is more forward thinking for #999 which would actually use the status stuff. more of this was re-jigging in anticipation of that being used... admittedly makes less sense until that it merged also.

adriancole: since the GrpcResponse types are hidden anyway, let's move this to #999 where you can see them used (and also where they are made public)

instrumentation/grpc/src/main/java/brave/grpc/TracingClientInterceptor.java
adriancole: agree. made a note in the rationale

instrumentation/grpc/src/main/java/brave/grpc/GrpcClientResponse.java
adriancole: I think this comment drifted as we are catching things which don't incur a grpc status, so resolving it.

instrumentation/grpc/src/main/java/brave/grpc/TracingClientInterceptor.java
adriancole: userInterceptor_throwsOnStart now exists to explain this.

instrumentation/grpc/src/main/java/brave/grpc/TracingServerInterceptor.java
adriancole: agreed and in rationale now. thanks!

adriancole:
Thanks for the thoughtful consideration @anuraaga. This is much better than before

adriancole:
force pushed as I have no other good ideas on how to get the mockito change to pass in travis.

adriancole:
I'm going to special case the mockito depedendency as I cannot get travis to pass in invoker mode in any way

adriancole:
force pushed again, with specific includes on the grpc invoker tests. At least locally, this doesn't use mockito at all now.

adriancole:
I am removing the mockito flag @anuraaga. It is unsupportable and there are so many other things worth time than flailing on this 

```
Mockito cannot mock this class: class io.grpc.MethodDescriptor.

Can not mock final classes with the following settings :

 - explicit serialization (e.g. withSettings().serializable())

 - extra interfaces (e.g. withSettings().extraInterfaces(...))

You are seeing this disclaimer because Mockito is configured to create inlined mocks.

You can learn about inline mocks and their limitations under item #39 of the Mockito class javadoc.

Underlying exception : org.mockito.exceptions.base.MockitoException: Could not modify all classes [class io.grpc.MethodDescriptor, class java.lang.Object]

```

adriancole:
fingers crossed.

PR #1176, opened 2020-04-27
Makes Http adapters finishable with only (response, span)

adriancole:
Similar to #999, HTTP should also be able to derive the error directly
from `HttpResponse` vs a second argument. This also allows exceptions
that nest status codes to become parsable into span tags.

This also updates places where we still used Tracer.withSpanInScope
in HTTP. This api has a problem that it cannot detect if a span is already
in scope.

### DIFF
--- a/instrumentation/http/src/main/java/brave/http/HttpClientHandler.java
+++ b/instrumentation/http/src/main/java/brave/http/HttpClientHandler.java
@@ -97,7 +97,7 @@ public final class HttpClientHandler<Req, Resp> extends HttpHandler {
     this.tracer = httpTracing.tracing().tracer();
     this.sampler = httpTracing.tracing().sampler();
     this.httpSampler = httpTracing.clientRequestSampler();
-    this.serverName = !"" .equals(httpTracing.serverName()) ? httpTracing.serverName() : null;
+    this.serverName = !"".equals(httpTracing.serverName()) ? httpTracing.serverName() : null;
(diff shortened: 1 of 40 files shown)


instrumentation/http/src/main/java/brave/http/HttpHandler.java
```diff
@@ -15,6 +15,9 @@
 
 import brave.Span;
 import brave.internal.Nullable;
+import brave.internal.Platform;
+
+import static brave.internal.Throwables.propagateIfFatal;
 
 abstract class HttpHandler {
```
adriancole, line 22: note: changes to this file are mainly to it almost identical to RpcHandler in #999

instrumentation/http/src/main/java/brave/http/HttpClientHandler.java
```diff
@@ -97,7 +97,7 @@
     this.tracer = httpTracing.tracing().tracer();
     this.sampler = httpTracing.tracing().sampler();
     this.httpSampler = httpTracing.clientRequestSampler();
-    this.serverName = !"".equals(httpTracing.serverName()) ? httpTracing.serverName() : null;
+    this.serverName = !"" .equals(httpTracing.serverName()) ? httpTracing.serverName() : null;
```
anuraaga, line 100: The space is back!

adriancole: haha grrrr

PR #1184, opened 2020-05-02
Corrects syntax for markdown link in instrumentation README

dangoslen:
noticed that a link was not created properly in markdown syntax while reading the instrumentation guides for creating new instrumentation


adriancole:
appreciated!

PR #1186, opened 2020-05-03
Adds UnsafeArrayMap and deprecated BaggageField.getAll for getAllValues

adriancole:
In analyzing the usage of the (recently added BaggageField.getAll), I
noticed in every case the keys are coersed to strings and all non-null
values are read.

This provides a more efficient alternative, via `UnsafeArrayMap`, pulled
in from #1183

### DIFF
--- a/brave/src/main/java/brave/internal/baggage/UnsafeArrayMap.java
+++ b/brave/src/main/java/brave/internal/baggage/UnsafeArrayMap.java
@@ -1,0 +1,388 @@
+/*
+ * Copyright 2013-2020 The OpenZipkin Authors
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License"); you may not use this file except
+ * in compliance with the License. You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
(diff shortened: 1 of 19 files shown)


brave/src/main/java/brave/internal/baggage/UnsafeArrayMap.java
```diff
@@ -0,0 +1,388 @@
...
+    @Override public boolean removeAll(Collection<?> c) {
+      throw new UnsupportedOperationException();
+    }
+
+    @Override public void clear() {
+      throw new UnsupportedOperationException();
+    }
+  }
+
+  static final class Entry<K, V> implements Map.Entry<K, V> {
```
anuraaga, line 329: Can we use `SimpleImmutableEntry` instead of this class?

adriancole: sure can. I didn't know that was public! also it is the same as are lowest java level. score!

adriancole: 80 less lines thanks to you!

adriancole:
you're a champ @anuraaga !

PR #1187, opened 2020-05-03
Extracts ValueUpdater from ExtraBaggageFields for use in decoding

adriancole:
This addresses a design problem where we formerly leaked an internal
type for baggage decoding. Even though BaggageCodec is not yet public,
it is in progress to become so. Cleaning this up paves the way for us
to expose it.

See https://github.com/openzipkin/brave/pull/1182#issuecomment-621575987

Once this is in, I'll rebase #1183

### DIFF
--- a/brave/RATIONALE.md
+++ b/brave/RATIONALE.md
@@ -297,7 +297,7 @@ propagation. There are some subtle points easy to miss about the design.
 * Decode is called before a trace context exists, hence no `context` parameter.
   * First successful decode returns `true`.
     * This allows us to prioritize and leniently read, multiple header names.
-  * The `ValueUpdater parameter allows us to safely collect values from
+  * The `ValueUpdater` parameter allows us to safely collect values from
     multiple contributors.
   * The `request` parameter allows secondary sampling to use `SamplerFunction`
     on inbound requests.
--- a/brave/RATIONALE.md
+++ b/brave/RATIONALE.md
@@ -335,7 +335,7 @@ The reverse isn't true. Decoding has different concerns entirely.
 * `BaggageField` instances are typically constant, so updates should be a
   function of `BaggageField`, not String.
 * There may be multiple decoders contributing to the same baggage state. It is
-  more safe underneath to not expose read-ops, such as iterators.
+  more safe underneath to prevent read-ops, such as iterators.
 * If we used a `Map`, implementations who ignore the documentation and hold a
   reference to it could later corrupt the state, violating copy-on-write.
   * defending against this is possible, but inefficient and distracting code.
--- a/brave/RATIONALE.md
+++ b/brave/RATIONALE.md
@@ -383,7 +383,7 @@ a design weakness in not choosing map.
 In practice, encoding formats typically have rules that splitter libraries
 weren't written for. These include limits, character set constraints, etc.
 Decoding a format is always harder than encoding due to rules about how to
-handle bad data. In general, it is better to assume a format-specific encoder
+handle bad data. In general, it is better to assume a format-specific decoder
 will be needed.
 
 Moreover, decoding through a map is inherently inefficient vs index-based
--- a/brave/RATIONALE.md
+++ b/brave/RATIONALE.md
@@ -277,27 +277,127 @@ In practice, we use a constant `HashMap` to find the index of a predefined
 fields. This is how we achieve constant time index lookup, while also keeping
 iteration performance of an array!
 
+### `BaggageField.ValueUpdater`
+`BaggageField.ValueUpdater` is a functional interface that allows field
+updates, possibly to `null`.
+
+It returns a `boolean` to inform callers an update was ignored, which can
+happen when the value didn't change or the field is disallowed. We use this to
+avoid unnecessary overhead when flushing changes to `MDC`.
+
+Being a functional interface, it is also safer to use when exposing state to
+outside callers. For example, we can provide a write view over a mutable array,
+and decouple that connection when decoding completes.
+
 ### Encoding and decoding baggage from remote fields
 
 `BaggageCodec` handles encoding and decoding of baggage fields for remote
 propagation. There are some subtle points easy to miss about the design.
 
-* Decode is called before a trace context exists, so there's no `context` parameter.
+* Decode is called before a trace context exists, hence no `context` parameter.
   * First successful decode returns `true`.
     * This allows us to prioritize and leniently read, multiple header names.
-  * TODO: updater parameter type.. should we do a write-through map?
-  * The `request` parameter allows secondary sampling to use `SamplerFunction` on inbound requests.
+  * The `ValueUpdater parameter allows us to safely collect values from
+    multiple contributors.
+  * The `request` parameter allows secondary sampling to use `SamplerFunction`
+    on inbound requests.
   * The `value` parameter could be a delimited string, or a plain value to set.
 
 * Encode is called only when a baggage exists and isn't all redacted
   * The `Map<String, String>` parameter allows implementations to encode all
     baggage in one string.
+    * Not `Map<BaggageField, String>` as we checked all implementations only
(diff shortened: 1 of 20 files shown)


brave/RATIONALE.md
```diff
@@ -228,6 +228,101 @@ implementations proposed in these papers are different to the implementation
...
+if `BaggageConfiguration` is initialized with fields countryCode and deviceId,
+the initial state array is `[countryCode, null, deviceId, null]`
+
+Retaining fields even when `null` allows among many things, the following:
+* Queries for all potential field names (remote propagation)
+* Constant time index lookup of predefined fields.
+
+In practice, we use a constant `HashMap` to find the index of a predefined
+fields. This is how we achieve constant time index lookup, while also keeping
+iteration performance of an array!
```
adriancole, line 270: this was just moved

brave/RATIONALE.md
```diff
@@ -228,6 +228,101 @@ implementations proposed in these papers are different to the implementation
...
+the array into it, and also pay implicit cost of iterator allocation for
+operations such as `entrySet()` even if the entry desired was not present. This
+holds present even extending `AbstractMap`, as most operations, even `get()`
+allocate iterators.
+
+#### `UnsafeArrayMap`
+We created `UnsafeArrayMap` as a view over our pair-wise internal state array.
+The implementation is very simple except that we have a redaction use case to
+address, which implies filtering keys. To address that, we keep a bitmap of all
+filtered keys and consider that when performing any scan operations.
```
adriancole, line 324: This also was just moved (I moved them as I noticed we already had a baggage heading!

brave/src/main/java/brave/internal/baggage/ExtraBaggageFields.java
```diff
@@ -69,16 +69,7 @@ public boolean isDynamic() {
...
-   * Updates a state object to include a value change.
-   *
-   * @param field the field that was updated
-   * @param value {@code null} means remove the mapping to this field.
-   * @return true implies a a change in the underlying state
-   * @see BaggageField#updateValue(TraceContext, String)
-   * @see BaggageField#updateValue(TraceContextOrSamplingFlags, String)
-   */
-  public boolean updateValue(BaggageField field, @Nullable String value) {
+  @Override public boolean updateValue(BaggageField field, @Nullable String value) {
```
adriancole, line 72: notice only change is `@Override`

adriancole:
I'm re-thinking this after looking deeper about how we apply value update changes. I don't think the user needs to be concerned with the same things we are. Particularly, the boolean return aspect is just an optimization on our side and matters not how users set values.

adriancole:
currently thinking to make a write-through map for this part of the code. Here, we don't hop threads, and haven't released the extra fields instance to a caller. so, there's no synchronization or other concerns like that. I took a rough stab and the mutable side is a few hundred lines, but they aren't bad especially as we can re-use them elsewhere. will post a change tomorrow.

adriancole:
changed mind  about the mutable map over an array because it can cause a security problem or concurrency problem later.

unlike the read-side which is against a COW array, the array behind this would be mutable. Eventhough we document to not hold a reference later, what if someone did? they could write through something that we mark as constant. or we have to allocate around the problem.

I think the original idea of a ValueUpdater is still the best plan. It narrows the api, allows for us to implement writes however we want. Since there's no read-back api on ValueUpdater, there's no risk of leaking data later. Since ValueUpdater is a functional interface it would be easy to make a shield that prevents those who hold a reference too long from invalidating the baggage state later.

iotw, I'm doing a git reset --hard on local changes and plan to continue with this one @anuraaga :D

adriancole:
completed rationale (one of the hardest to think through) @anuraaga PTAL!

brave/RATIONALE.md
```diff
@@ -228,6 +228,199 @@ implementations proposed in these papers are different to the implementation
...
+
+### Encoding and decoding baggage from remote fields
+
+`BaggageCodec` handles encoding and decoding of baggage fields for remote
+propagation. There are some subtle points easy to miss about the design.
+
+* Decode is called before a trace context exists, hence no `context` parameter.
+  * First successful decode returns `true`.
+    * This allows us to prioritize and leniently read, multiple header names.
+  * The `ValueUpdater parameter allows us to safely collect values from
```
adriancole, line 300:
```suggestion
  * The `ValueUpdater` parameter allows us to safely collect values from
```

brave/RATIONALE.md
```diff
@@ -228,6 +228,199 @@ implementations proposed in these papers are different to the implementation
...
+On the read side, the above iteration is stable because the underlying state is
+immutable. There's no chance of a concurrent modification or otherwise
+corrupting the encoded value. This applies even when the implementation ignores
+the documentation and holds a reference to the map after encoding.
+
+The reverse isn't true. Decoding has different concerns entirely.
+* `BaggageField` instances are typically constant, so updates should be a
+  function of `BaggageField`, not String.
+* There may be multiple decoders contributing to the same baggage state. It is
+  more safe underneath to not expose read-ops, such as iterators.
```
adriancole, line 338:
```suggestion
  more safe underneath to prevent read-ops, such as iterators.
```

brave/RATIONALE.md
```diff
@@ -228,6 +228,199 @@ implementations proposed in these papers are different to the implementation
...
+for (Entry<String, String> entry: splitter.split(encoded).entrySet()) {
+  BaggageField field = lookupOrCreateField(entry.getKey());
+  valueUpdater.updateValue(field, entry.getValue());
+}
+```
+
+A closing point is that the above splitter example is contrived, mainly to show
+a design weakness in not choosing map.
+
+In practice, encoding formats typically have rules that splitters libraries
```
adriancole, line 383:
```suggestion
In practice, encoding formats typically have rules that splitter libraries
```

brave/RATIONALE.md
```diff
@@ -228,6 +228,199 @@ implementations proposed in these papers are different to the implementation
...
+}
+```
+
+A closing point is that the above splitter example is contrived, mainly to show
+a design weakness in not choosing map.
+
+In practice, encoding formats typically have rules that splitter libraries
+weren't written for. These include limits, character set constraints, etc.
+Decoding a format is always harder than encoding due to rules about how to
+handle bad data. In general, it is better to assume a format-specific encoder
```
adriancole, line 386:
```suggestion
handle bad data. In general, it is better to assume a format-specific decoder
```

anuraaga:
Just random note, I wonder if we should be adding note about functional interfaces to javadoc. We can't add the annotation unfortunately, but that means that it's not technically safe for a user that has to worry about version compatibility to use lambdas.

brave/src/main/java/brave/baggage/BaggageField.java
```diff
@@ -101,6 +101,31 @@
...
 public final class BaggageField {
+  /**
+   * Used to decouple baggage value updates from {@link TraceContext} or {@link
+   * TraceContextOrSamplingFlags} storage.
+   *
+   * @since 5.12
+   */
+  public interface ValueUpdater {
+    /** @since 5.12 */
+    ValueUpdater NOOP = new ValueUpdater() {
```
anuraaga, line 112: Override toString too?

adriancole:
good note, @anuraaga Actually I think we do have a note like this elsewhere. Will do.

adriancole:
oh you said to javadoc. Yeah I suppose we can. I think the javadoc link will be bad if we use FunctionalInterface due to language level(?), so maybe we can use `{@code FunctionalInterface}` in those notes? wdyt?

anuraaga:
Yeah code or even just words is fine. Think it's important since it's a signal both to us not to add methods and to users that it's safe to use with a lambda.

adriancole:
thanks for all the feedback @anuraaga

PR #1190, opened 2020-05-06
Renames internal ExtraHandler back to ExtraFactory and clarifies

adriancole:
When integrating secondary sampling, I found the new name (I caused)
more distracting then helpful. The defensive api which led to renaming
was a cure worse than the disease of a preventable programming error.

Instead, this improves Javadoc, notably mentioning how to use the
factory properly (and not crashing on mistake).


adriancole:
here's a side effect which is I think, not just more like the code was before, but more intuitive (or.. maybe less unituitive?)

https://github.com/openzipkin-contrib/zipkin-secondary-sampling/pull/33/commits/b4221511cda5209145386cad449684823135029d

after this, next step is to make a base type for map views over arrays as we need it both for secondary sampling and w3c tracestate

brave/src/main/java/brave/internal/baggage/BaggageFieldsFactory.java
```diff
@@ -51,7 +51,7 @@ public static BaggageFieldsHandler create(List<BaggageField> fields, boolean dyn
     this.initialArrayLength = initialState.length;
   }
 
-  @Override protected BaggageFields provisionExtra() {
+  @Override public BaggageFields create() {
```
adriancole, line 54: note: this is public just because the type itself lives in a different package (brave.internal.baggage) than the call site (brave.baggage). In most cases, they will be in the same package, so implementations can leave it protected.

brave/src/main/java/brave/internal/baggage/Extra.java
```diff
@@ -21,23 +21,19 @@
 /**
  * Holds extended state in {@link TraceContext#extra()} or {@link TraceContextOrSamplingFlags#extra()}.
  *
- * <p>We need to retain propagation state extracted from headers. However, we don't know the trace
```
adriancole, line 24: moved to the factory as the relevant notes around this are there

brave/src/main/java/brave/internal/baggage/ExtraFactory.java
```diff
@@ -0,0 +1,188 @@
...
+ *
+ * <p>Even though {@link #<E>} is mutable, it must be copy-on-write internally. This type manages
+ * state forking that ensures that updates to child spans are invisible to their parents or
+ * siblings.
+ *
+ * @param <E> They type of {@link Extra} managed by this factory. Must be a final class. The type is
+ * typically package private to avoid accidental interference.
+ * @param <F> An instance of this factory. {@link #<E>} should be associated with only one factory.
+ */
+public abstract class ExtraFactory<E extends Extra<E, F>, F extends ExtraFactory<E, F>> {
```
adriancole, line 72: because I added so much javadoc, this became a new file. The logic hasn't changed except coersing bugs to the input context

PR #1193, opened 2020-05-07
Adds EntrySplitter for use in w3c specs and secondary-sampling

adriancole:
This adds `EntrySplitter` to help handle complex formats such as W3C
trace-context efficiently. For example, this has flexible whitespace
rules to comply with standards like 'tracestate' which require retaining
leading whitespace.

See #693


brave/src/test/java/brave/internal/codec/EntrySplitterTest.java
```diff
@@ -0,0 +1,279 @@
...
+      Map<String, String> attributes = new LinkedHashMap<>();
+      if (beginValue == endValue) { // no string allocation at all
+        attributes = Collections.emptyMap();
+      } else { // no string allocation to pass to the inner parser
+        attributes = new LinkedHashMap<>();
+        innerSplitter.parse(parseIntoMap, attributes, input, beginValue, endValue);
+      }
+      target.put(key, attributes);
+      return true;
+    }, keyToAttributes, "authcache;ttl=1;spanId=19f84f102048e047,gateway");
```
adriancole, line 138: @narayaruna look familiar? :)

brave/src/test/java/brave/internal/codec/EntrySplitterTest.java
```diff
@@ -0,0 +1,344 @@
...
+          .isInstanceOf(IllegalArgumentException.class)
+          .hasMessage("Invalid input: no key before '='");
+    }
+  }
+
+  /**
+   * This is an example of how to parse without allocating strings. This is based on
+   * https://github.com/openzipkin/zipkin-aws/blob/master/brave-propagation-aws/src/main/java/brave/propagation/aws/AWSPropagation.java
+   */
+  @Test public void example_parseAWSTraceId() {
```
adriancole, line 162: added allocation free trace ID example from AWS

brave/src/main/java/brave/internal/codec/EntrySplitter.java
```diff
@@ -0,0 +1,245 @@
...
+  }
+
+  public static final class Builder {
+    int maxEntries = Integer.MAX_VALUE;
+    char entrySeparator = ',', keyValueSeparator = '=';
+    boolean trimOWSAroundEntrySeparator = true, trimOWSAroundKeyValueSeparator = true;
+    boolean keyValueSeparatorRequired = true, shouldThrow = false;
+
+    public Builder maxEntries(int maxEntries) {
+      if (maxEntries == 0) throw new NullPointerException("maxEntries == 0");
```
samukce, line 37: Does it look more sense to use `IllegalArgumentException` instead of `NullPointerException`?

adriancole: good eye :)

brave/src/main/java/brave/internal/codec/EntrySplitter.java
```diff
@@ -0,0 +1,302 @@
...
+    /**
+     * The separator to use between a key and value. For example, given a string "k1=v1,k2=v2", the
+     * {@code keyValueSeparator} should be '='. Default: '='
+     *
+     * <p><em>Note:</em> Only the first {@code keyValueSeparator} identifies the end of the key
+     * until the next {@link #entrySeparator(char)}. This means values can include the {@code
+     * keyValueSeparator} character.
+     *
+     * <p>For example, the string "authcache;ttl=1;spanId=19f84f102048e047", with
+     * {@code keyValueSeparator=;} parses {@code [("authcache", "ttl=1;spanId=19f84f102048e047)]}
```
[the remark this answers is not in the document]
adriancole: heh it is real :) https://github.com/openzipkin-contrib/zipkin-secondary-sampling/blob/d3b6c1fc5f84269afb3785394cfd909ee37215bd/docs/design.md#the-sampling-field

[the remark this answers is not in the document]
adriancole: but yeah for the sake of example, here it is fine to use like you said. less to think about

brave/src/main/java/brave/internal/codec/EntrySplitter.java
```diff
@@ -0,0 +1,302 @@
...
+     * On validation fail, should this throw an exception or log?. The use case to throw is when
+     * validating input (ex into a builder), or in unit tests.
+     */
+    public Builder shouldThrow(boolean shouldThrow) {
+      this.shouldThrow = shouldThrow;
+      return this;
+    }
+
+    public EntrySplitter build() {
+      return new EntrySplitter(this);
```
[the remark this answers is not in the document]
adriancole: yep good eye

brave/src/main/java/brave/internal/codec/EntrySplitter.java
```diff
@@ -0,0 +1,302 @@
...
+  public <T> boolean parse(
+      Handler<T> handler, T target, String input, int beginIndex, int endIndex) {
+    int remainingEntries = maxEntries, beginKey = -1, endKey = -1, beginValue = -1;
+    for (int i = beginIndex; i < endIndex; i++) {
+      char c = input.charAt(i);
+
+      boolean nextIsEnd = i + 1 == endIndex;
+      if (c == entrySeparator || nextIsEnd) { // finished an entry
+        if (c == keyValueSeparator) {
+          beginValue = i; // empty value: ex "key=" "k1 ="
```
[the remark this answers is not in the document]
adriancole: will check.. worth it if true!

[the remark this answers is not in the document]
adriancole: nextIsEnd entrySeparator can both be true. I updated the comment to clarify! thanks

brave/src/main/java/brave/internal/codec/EntrySplitter.java
```diff
@@ -0,0 +1,302 @@
...
+        }
+
+        if (beginKey == -1 && beginValue == -1) {
+          continue; // ignore empty entries, like ",,"
+        } else if (beginKey == -1) {
+          return logOrThrow(missingKey, shouldThrow); // ex. "=" ",="
+        } else if (nextIsEnd && beginValue == -1) { // ex "k1" "k1 " "a=b" "..=,"
+          // We reached the end of a key-only entry, a single character entry or an empty entry
```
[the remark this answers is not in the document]
adriancole: to be consistent with param names I went with end of input (as the string length can be longer than endIndex)

adriancole:
Thanks for the look @anuraaga PTAL

brave/src/main/java/brave/internal/codec/EntrySplitter.java
```diff
@@ -0,0 +1,315 @@
...
+    /**
+     * The separator to use between a key and value. For example, the input "k1=v1,k2=v2" should
+     * have {@code keyValueSeparator} '='. Default: '='
+     *
+     * <p><em>Note:</em> Only the first {@code keyValueSeparator} identifies the end of the key
+     * until the next {@link #entrySeparator(char)}. This means values can include the {@code
+     * keyValueSeparator} character.
+     *
+     * <p>For example, the input "condition=animal=cat" with {@code keyValueSeparator} '=' parses
+     * {@code [("condition", "animal=cat7")]}
```
adriancole, line 76:
```suggestion
     * {@code [("condition", "animal=cat")]}
```

brave/src/main/java/brave/internal/codec/EntrySplitter.java
```diff
@@ -0,0 +1,315 @@
...
+     *
+     * <p>For example, the input "condition=animal=cat" with {@code keyValueSeparator} '=' parses
+     * {@code [("condition", "animal=cat")]}
+     *
+     * @see #keyValueSeparator(char)
+     */
+    public Builder keyValueSeparator(char keyValueSeparator) {
+      if (keyValueSeparator == 0) throw new IllegalArgumentException("keyValueSeparator == 0");
+      if (keyValueSeparator == entrySeparator) {
+        throw new IllegalArgumentException("keyValueSeparator == entrySeparator");
```
anuraaga, line 83: Validating in the setter is nicer but I guess it means we should default the values to `0` and replace with the actual defaults in `build`. Otherwise, it requires hoops to do the horrible, but not inconceivable, using `=` as entry separator and `,` as keyvalue separator.

adriancole: I'm convinced deferring is better :D

brave/src/main/java/brave/internal/codec/EntrySplitter.java
```diff
@@ -0,0 +1,315 @@
...
+   * @param input string including data to parse
+   * @param beginIndex begin index of the {@code input}, inclusive
+   * @param endIndex end index of the {@code input}, exclusive
+   * @return true if we reached the {@code endIndex} without failures.
+   */
+  public <T> boolean parse(
+      Handler<T> handler, T target, String input, int beginIndex, int endIndex) {
+    if (handler == null) throw new NullPointerException("handler == null");
+    if (target == null) throw new NullPointerException("target == null");
+    if (input == null) throw new NullPointerException("input == null");
```
samukce, line 216: Should we also use it as `IllegalArgumentException` in those argument validations here? When I pass an invalid argument to a method I expect the method raises some validation error about those instead off give feedback about the invalid arguments wrapped in an NPE error.

brave/src/main/java/brave/internal/codec/EntrySplitter.java
```diff
@@ -0,0 +1,315 @@
...
+public final class EntrySplitter {
+  public static Builder newBuilder() {
+    return new Builder();
+  }
+
+  public static final class Builder {
+    int maxEntries = Integer.MAX_VALUE;
+    char entrySeparator = ',', keyValueSeparator = '=';
+    boolean trimOWSAroundEntrySeparator = true, trimOWSAroundKeyValueSeparator = true;
+    boolean keyValueSeparatorRequired = true, shouldThrow = false;
```
samukce, line 35: We are implementing the builder pattern here but those fields are accessible externally in the package. Doing this we could use the builder in the wrong way as this usage for instance `EntrySplitter.newBuilder().maxEntries = 0;`. Do we have any motivation for this? Otherwise, we could set as private.

brave/src/main/java/brave/internal/codec/EntrySplitter.java
adriancole: On this one I don't agree. the only reason we explicitly raise them here is to prevent confusion later when java would do it for you, but without a message.

If it were different, ex two Nullable params and one must be non-null, yeah that would be an IAE.

Make sense?

brave/src/main/java/brave/internal/codec/EntrySplitter.java
adriancole: in Brave we try to be conscious of bytecode size as we cause bloat otherwise. That's why you'll see us almost never use the private modifier. We don't support other tools sharing the same package as we do, so it isn't a problem in practice.

adriancole:
@samukce I now realize we don't have rationale for our coding conventions including things you just asked.. I'll add it so it is easier to link to later. 

We only lightly say a few things here https://github.com/openzipkin/brave/blob/master/HACKING.md but there have been issues in the past particularly about bytecode size. If you aren't careful the jar can grow really big as it is more often people add code without removing code.

here are a few related changes which help control size and stay dep free:
https://github.com/openzipkin/brave/pull/730
https://github.com/openzipkin/brave/pull/1159
https://github.com/openzipkin/zipkin/pull/1236
https://github.com/openzipkin/zipkin/pull/2062

brave/src/main/java/brave/internal/codec/EntrySplitter.java
samukce: it makes senses but at the same time, it looks like we are translating the NPE to a more readable NPE. But in the end, the method doesn't work with one of those arguments is not null. So there is a dependency on this method works that is why I think the IAE makes more sense. Imagining to add `@NotNull` to the arguments so we would see a sort of `Argument for @NotNull parameter 'b'`.

brave/src/main/java/brave/internal/codec/EntrySplitter.java
samukce: hmm, interesting. I will take a look at the readme that you sent. thanks for share. :)

brave/src/main/java/brave/internal/codec/EntrySplitter.java
anuraaga: I found it a bit weird a long time ago but think it's general convention to use NPE for these, codified into Java 8 with `requireNonNull` which throws NPE.

samukce: argg.... there is not consistency between `dosomething(@NotNull String a)` x `requireNonNull(a)` oO
So, it's fine either way. no argument for now anymore. :)

brave/RATIONALE.md
```diff
@@ -18,10 +18,50 @@ thought as it would betray productivity and make this document unreadable.
 Rationale here should be limited to impactful designs, and aspects non-obvious,
 non-conventional or subtle.
 
-## Public namespace
+## Java conventions
```
samukce, line 21: 👍

brave/RATIONALE.md
```diff
@@ -18,10 +18,57 @@ thought as it would betray productivity and make this document unreadable.
...
+
+Modifiers on fields and methods are distracting to read and increase the size
+of the bytecode generated during compilation. By recovering the size otherwise
+spent on private modifiers, we not only avoid hitting limits, but we are also
+able to add more code with the same jar size.
+
+For example, Brave 5.12 remains less than 250KiB, with no dependencies, all
+features including deprecation bridges, and an embedded json serializer.
+
+This means we don not support sharing our packages with third parties, but we
```
adriancole, line 51:
```suggestion
This means we do not support sharing our packages with third parties, but we
```

brave/src/main/java/brave/internal/codec/EntrySplitter.java
```diff
@@ -0,0 +1,312 @@
...
+        if (endKey == -1) {
+          if (keyValueSeparatorRequired && c != keyValueSeparator) {
+            return logOrThrow(missingKeyValueSeparator, shouldThrow); // throw on "k1" "k1=v2,k2"
+          }
+
+          // Even though we have an empty value, we need to handle whitespace and
+          // boundary conditions.
+          //
+          // For example, using entry separator ',' and KV separator '=':
+          // "...,k1" and input[i] == 'y', we want i + 1, so that the key includes the 'y'
```
adriancole, line 247:
```suggestion
          // "...,k1"  and input[i] == 'y', we want i + 1, so that the key includes the 'y'
```

adriancole:
Thanks for the help @anuraaga and @samukce

PR #1194, opened 2020-05-08
Reduces internal exposure and coupling

adriancole:
* We no longer need InternalBaggage in secondary-sampling so we should delete it
* ExtraFactory is less complex to shade by avoiding InternalPropagation
  * adding TraceContext.Builder.clearExtra() gets us within 1 allocation of prior


adriancole:
@anuraaga fyi but you don't need to review this unless you feel like it.

PR #1196, opened 2020-05-09
Makes MutableSpan easier with tags() and annotations() method

adriancole:
This makes `MutableSpan` easier, especially in tests, by adding methods
for tags and annotations similar to `zipkin2.Span`. Unlike
`zipkin2.Span`, this presents annotations as a collection of `Map.Entry`.

Doing so keeps us from having to define a new type mostly used in tests.
The allocation-free consumer functions are left in place.

One unit test was converted in the process. We can convert the rest of
them following this PR.

Under the scenes, this uses the same `UnsafeArrayMap` we use in baggage.
To port to this model, we had to use arrays directly instead of arraylist.
However, we act internally the same as arraylist does.

### DIFF
--- a/brave/src/main/java/brave/handler/MutableSpan.java
+++ b/brave/src/main/java/brave/handler/MutableSpan.java
@@ -96,15 +96,12 @@ public final class MutableSpan implements Cloneable {
 
   //
   // The below use object arrays instead of ArrayList. The intent is not for safe sharing
-  // (copy-on-write), as this type is externally synchronized. In other words, this isn't copy-on-write.
+  // (copy-on-write), as this type is externally synchronized. In other words, this isn't
+  // copy-on-write. We just grow arrays as we need to similar to how ArrayList does it.
   //
-  // rather just resizing
-  // the array similar to how ArrayList does internally.
-  //
-  /** To reduce the amount of allocation use a pair-indexed list for tag (key, value). */
-  Object[] tags = EMPTY_ARRAY;
-  /** Also use pair indexing for annotations, but type object to store (timestamp, value). */
-  Object[] annotations = EMPTY_ARRAY;
+  // tags [(key, value)] annotations [(timestamp, value)]
+  Object[] tags = EMPTY_ARRAY, annotations = EMPTY_ARRAY;
+  int tagCount, annotationCount;
 
   /** @since 5.4 */
   public MutableSpan() {
@@ -151,7 +148,9 @@ public final class MutableSpan implements Cloneable {
     remotePort = toCopy.remotePort;
     // In case this is a default span, don't hold a reference to the same array!
     tags = copy(toCopy.tags);
+    tagCount = toCopy.tagCount;
     annotations = copy(toCopy.annotations);
+    annotationCount = toCopy.annotationCount;
     error = toCopy.error;
   }
 
@@ -532,7 +531,7 @@ public final class MutableSpan implements Cloneable {
 
   /** @since 5.12 */
   public int annotationCount() {
-    return annotations.length / 2;
+    return annotationCount;
   }
 
   /**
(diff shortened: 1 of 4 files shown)


brave/src/test/java/brave/features/handler/RedactingSpanHandlerTest.java
```diff
@@ -28,9 +28,8 @@
 import java.util.regex.Pattern;
 import org.junit.After;
 import org.junit.Test;
-import zipkin2.Annotation;
```
adriancole, line 31: I switched only one test to show decoupling from zipkin types

brave/src/test/java/brave/handler/MutableSpanTest.java
```diff
@@ -91,7 +89,7 @@
     // When exporting into a list, a lambda would usually need to close over the list, which results
     // in a new instance per invocation. Since there's a target type parameter, the lambda for this
     // style of conversion can be constant, reducing overhead.
-    List<Tag> listTarget = new ArrayList<>(span.tagCount());
```
adriancole, line 94: in hindsight maybe it is worth putting these back.. they are allocation free ways to get the count..

brave/src/main/java/brave/handler/MutableSpan.java
```diff
@@ -81,9 +85,9 @@
   int localPort, remotePort;
 
   /** To reduce the amount of allocation use a pair-indexed list for tag (key, value). */
-  ArrayList<String> tags;
-  /** Also use pair indexing for annotations, but type object to store (startTimestamp, value). */
-  ArrayList<Object> annotations;
+  Object[] tags = EMPTY_ARRAY;
```
anuraaga, line 88: Just to confirm it's ok for this to not be volatile. I guess an `ArrayList` would have already had issues when resizing though if there is a problem, but I'm not sure that memory is synced if users allowed to mutate without the lock.

adriancole: yeah this type isn't thread safe it is externally synchronized. the only reason doing copy-on-write is for the map views, to avoid having issues with people sharing them. I'll make a note

adriancole: copy-on-write for update I mean

adriancole: I think I added confusion by even saying copy-on-write.. my bad. PTAL at latest!

adriancole:
I removed my distracting stuff about copy-on-write. The main thing is we are implementing the same thing arraylist does.

I also fixed the bug where I prevented the copy constructor from doing what it should. Basically we should copy only in the copy constructor. There's no problem updating exisitng values as these are mutable types, externally synchronized until control is passed to the handlers.

anuraaga:
LGTM

brave/src/main/java/brave/handler/MutableSpan.java
```diff
@@ -883,6 +928,32 @@ static void writeEndpoint(StringBuilder b,
...
+  }
+
+  static Object[] copy(Object[] input) {
+    return input.length > 0 ? Arrays.copyOf(input, input.length) : EMPTY_ARRAY;
+  }
+
+  static Object[] remove(Object[] input, int i) {
+    if (input.length == 2 && i == 0) return EMPTY_ARRAY;
+    Object[] result = Arrays.copyOf(input, input.length - 2);
+    System.arraycopy(input, i + 2, result, i, input.length - i - 2);
```
adriancole, line 953: oops we don't need to do this.. all we need to do is shift and set null. Otherwise dropping all but one tag (like skeletal spans do) will thrash. will fix

adriancole:
noticed we were caching the hashCode which is a bug in a mutable type (unless we had a mod counter, which we don't)

PR #1197, opened 2020-05-09
Clean up integration test

Guchman:
I want to add my case test example to kafka-streams instrumentation, but IT tests were complicate and I've done some clean up. It should be easier to read now, at least I hope so :)

### DIFF
--- a/instrumentation/kafka-streams/src/test/java/brave/kafka/streams/ITKafkaStreamsTracing.java
+++ b/instrumentation/kafka-streams/src/test/java/brave/kafka/streams/ITKafkaStreamsTracing.java
@@ -18,12 +18,6 @@ import brave.messaging.MessagingTracing;
 import brave.propagation.TraceContext;
 import com.github.charithe.kafka.EphemeralKafkaBroker;
 import com.github.charithe.kafka.KafkaJunitRule;
-import java.io.FileNotFoundException;
-import java.util.ArrayList;
-import java.util.Arrays;
-import java.util.List;
-import java.util.Objects;
-import java.util.Properties;
 import org.apache.kafka.clients.CommonClientConfigs;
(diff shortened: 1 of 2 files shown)


Guchman:
Also, I wondering maybe it would be a good idea to commit Square Code Style with .idea/codeStyles directory? It's common to ignore whole .idea folder, but it's not the best practice. It's would be easier for contribution. I could open another PR with it.

jorgheymans:
Thanks @Guchman , care to elaborate a bit exactly what you cleaned up ? Because there are so many changes in the file, the diff is hard to read for anyone assessing this PR. I would suggest that maybe you stop by in the gitter channel first and discuss what exactly you don't like on this class before doing this. Not saying it cannot be improved, just the way you did it makes it hard to vet.

Guchman:
@jorgheymans I've move copy-paste code to methods with meaningful names, just to make test-cases shorter and more readable without any changes to assertions or behavior. Just to make them more "given/when/then"-like. I think the easier way to review that PR is to read raw test cases first not diffs because there is a lot of method-extraction refactoring and stuff.

adriancole:
here's why your PR was perfectly not timed :) https://github.com/openzipkin/brave/pull/1202

Guchman:
Thanks for explanation anyway:) I'm agree that a clean test base is not very useful from users perspective, but I believe it's useful to have more declarative approach in tests for developers experience. Anyway, I will discuss first. Just didn't want to bother you with that kind of discussion, sorry

adriancole:
PR doesnt remove the discussion it forces it :)

right now as we are all beyond busy let's try to suggest minimal changes
especially when unfamiliar with project style.  small change first is a
nice way to get started and ease in anyway. catch you later

PR #1199, opened 2020-05-10
Adds Tracing.Builder.spanHandlers() and clearSpanHandlers()

adriancole:
Unlike span reporter, span handlers are additive. Unless there's a way
to readback the current state, you cannot replace a handler of a
specific type or configuration. By adding the same methods as we did on
`BaggagePropagation` for `configs()`, span handlers can be re-orderered
or pruned.

### DIFF
--- a/brave/src/test/java/brave/features/handler/DefaultTagsTest.java
+++ b/brave/src/test/java/brave/features/handler/DefaultTagsTest.java
@@ -17,7 +17,6 @@ import brave.ScopedSpan;
 import brave.Tracing;
 import brave.handler.MutableSpan;
 import brave.handler.SpanHandler;
-import brave.propagation.StrictCurrentTraceContext;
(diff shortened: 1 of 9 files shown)


adriancole:
@anuraaga I know we talked about this the night before last. I'm pinging you for posterity but no need to action this.

brave/src/test/java/brave/features/handler/DefaultTagsTest.java
```diff
@@ -33,7 +32,6 @@
 public class DefaultTagsTest {
   TestSpanHandler spans = new TestSpanHandler();
   Tracing tracing = Tracing.newBuilder()
-    .currentTraceContext(StrictCurrentTraceContext.create())
```
adriancole, line 36: this test used no scoping ops

PR #1202, opened 2020-05-10
Migrates all tests to MutableSpan

adriancole:
This removes TestSpanReporter, which used zipkin dependencies.

It adds two types:

* `TestSpanHandler` for normal unit tests
* `IntegrationTestSpanHandler` for integration tests (blocking queue etc)

This also cleans up some tests that extended `ITRemote` even though they
were just unit tests.


adriancole:
@jeqo @jorgheymans @anuraaga I don't expect you to look at thousand line diff. I just want you to know finally I've migrated this.

We should still cut 5.12 pinning zipkin-reporter to use the right version (which includes reporter-brave). However, we are internally ready for Brave 6, as we no longer depend on types even in tests.

The only remnants are deprecated methods (ex. Span.remoteEndpoint and a couple Tracing.Builder methods)

PR #1203, opened 2020-05-10
Adds RemoteGetter and RemoteSetter used for kind-based injection in B3Propagation

adriancole:
This adds `RemoteGetter` and `RemoteSetter` used for kind-based injection in `B3Propagation`. What this means is that `Propagation` instances who vary format based on a request type can make decisions before constructing the `Injector` as opposed to at runtime, assuming they extend it. In any case if `Request` is implemented, the decision is still made at runtime.

This also clarifies the purpose of non-remote inject/extract. It is almost always for message processors. These could be truly local, eg in the same process. It could be remote, but more from a serialization POV, vs from a producer/consumer relationship. Most notably Kafka Streams is like this.

Finally, this makes B3Propagation.Factory return a singleton on no change.  This allows configuration frameworks to tell if the configuration is default or not. This also applies for any comparison, as equals and hashCode are now implemented.
change, at least tests are easier in sleuth now.


brave/src/main/java/brave/propagation/B3Propagation.java
```diff
@@ -153,8 +165,20 @@ public FactoryBuilder injectFormats(Span.Kind kind, Format format1, Format forma
     @Override public void inject(TraceContext context, R request) {
       Format[] formats = factory.injectFormats;
       if (request instanceof Request) {
-        Span.Kind kind = ((Request) request).spanKind();
-        formats = factory.kindToInjectFormats.get(kind);
+        switch (((Request) request).spanKind()) {
+          case CLIENT:
+            formats = factory.clientInjectFormats;
```
adriancole, line 170: in a future change we can actually return a constant injector, as the formats won't change per call to inject. The advantage would be similar to other places where we have constant vs composite... even though the iteration across an array of size two is fast, it shouldn't be quicker than something with no array. By default, there's only one inject format per kind.

brave/src/main/java/brave/propagation/B3Propagation.java
```diff
@@ -153,8 +165,20 @@ public FactoryBuilder injectFormats(Span.Kind kind, Format format1, Format forma
     @Override public void inject(TraceContext context, R request) {
       Format[] formats = factory.injectFormats;
       if (request instanceof Request) {
-        Span.Kind kind = ((Request) request).spanKind();
-        formats = factory.kindToInjectFormats.get(kind);
+        switch (((Request) request).spanKind()) {
```
adriancole, line 168: I'm tempted to do generic foo in `<R> TraceContext.Injector<R> injector(Setter<R, K> setter);` to find out the bound of `R`.. (reflection look up the TypeVariable of the setter) if we could, and found it was a request, then we wouldn't have to do this runtime check!

not tempted now. just a nag for future.

brave/src/main/java/brave/propagation/B3Propagation.java
adriancole: oh... I have an idea that's much easier! Add a subtype in request!

Request.Setter extends Propagation.Setter
exposes spanKind()

With this, we can hardwire B3 without needing reflection at runtime except the slow path for things that haven't extended Request.Setter yet.

adriancole: ok I did this :)

brave/src/main/java/brave/propagation/B3Propagation.java
adriancole: done but not with reflection

brave/src/main/java/brave/internal/propagation/InjectorFactory.java
```diff
@@ -0,0 +1,302 @@
...
+import brave.propagation.Propagation.Setter;
+import brave.propagation.TraceContext;
+import java.util.ArrayList;
+import java.util.Arrays;
+import java.util.Collections;
+import java.util.LinkedHashSet;
+import java.util.List;
+import java.util.Set;
+
+public final class InjectorFactory {
```
adriancole, line 28: This is an internal type exctracted from B3. I don't plan to do a huge amount of polishing on this in this PR as I'm at a severe time deficit. I will puts with it post merge though and any significant changes needed pre-merge

adriancole:
@anuraaga as I changed the scope of this PR to fix the underlying glitch in the model, might wanna PTAL and see if you still approve or not :D

adriancole:
going to back fill unit tests for merge

brave/src/main/java/brave/internal/propagation/InjectorFactory.java
```diff
@@ -0,0 +1,351 @@
...
+      return setter.equals(that.setter) && injectorFunction.equals(that.injectorFunction);
+    }
+
+    @Override public String toString() {
+      return "Injector{setter=" + setter + ", injectorFunction=" + injectorFunction + "}";
+    }
+  }
+
+  static InjectorFunction injectorFunction(InjectorFunction... injectorFunctions) {
+    if (injectorFunctions == null) throw new NullPointerException("injectorFunctions == null");
```
anuraaga, line 303: Don't think `Arrays.asList` does so probably want to check all the elements aren't null too

adriancole: sure thing

brave/src/main/java/brave/internal/propagation/InjectorFactory.java
adriancole: I lied.. polish is here now

PR #1204, opened 2020-05-10
Makes default propagation for non-request spans B3 single

adriancole:
As described in the referenced issue. In-process messaging sometimes
uses propagation, and the default of B3 multi makes the overhead quite
bad. This switches to B3 single which significantly redudes that burden.

https://github.com/spring-cloud/spring-cloud-sleuth/pull/1607
^^ already changed this in sleuth. this is just defaulting here

We should note that B3 single is widely supported at this point, and all
our instrumentation extend Request when injecting.

See #1020 as this was the change that allowed requests to differentiate

### DIFF
--- a/brave/src/main/java/brave/propagation/B3Propagation.java
+++ b/brave/src/main/java/brave/propagation/B3Propagation.java
@@ -63,10 +63,10 @@ public abstract class B3Propagation<K> implements Propagation<K> {
 
   /**
    * Defaults to {@link Format#MULTI} for client/server spans and {@link Format#SINGLE_NO_PARENT}
-   * for messaging. Non-request spans default to {@link Format#MULTI}.
+   * for messaging. Non-request spans default to {@link Format#SINGLE_NO_PARENT}.
    */
   public static final class FactoryBuilder {
-    Format injectFormat = Format.MULTI;
+    Format injectFormat = Format.SINGLE_NO_PARENT;
     final EnumMap<Span.Kind, Format[]> kindToInjectFormats = new EnumMap<>(Span.Kind.class);
 
     FactoryBuilder() {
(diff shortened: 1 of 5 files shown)


adriancole:
It is a fair question. I guess it will be a sooner or later thing, and I don't expect us to do a lot of releases of 5.x... so if no-one addresses this they will pin themselves on something we aren't maintaining if we wait until 6.x...

adriancole:
My thought is that if someone is updating brave, and would update it to 5.12, then they likely update other things, and will be either using our instrumentation or other instrumentation that can parse single by now. It is hard to "deprecate" default configuration.. this is the dillemma of the issue. I suppose we can add a log-once to warn level now and defer it to another minor at the very least?

anuraaga:
Yeah having a log to help transition at least seems nice. I'm less worried about Brave/Zipkin tooling, but I can imagine at least one case out there of user code itself reading the headers to do something with it that would stop seeing it

adriancole:
it is a good point. since this is default I have another thing in progress
RemoteGetter (including span kind) which can allow a more relevant means to
warn people. otherwise we will only get warned at startup vs when it is
read.

adriancole:
I thought a lot about this and thing that having a WARN message, especially
as we have to use JUL, is a case where the cure is worse than the disease.
Just so we are clear, only B3 do we have the ability to separately control
local propagation format (at the moment), and all variants of B3 read all
header formats.

I'm inclined to just let this bomb on 6, I guess.. but we really must fix
this by then for sure. This is not just a high overhead problem, but also
leaking a hyphenated format exactly the place it is most likely to have
problems with it (messaging processors).

Maybe we can update Kafka Streams and the messaging README to warn people
about this configuration problem until we feel safe to fix it, in worst
case Brave 6 which might happen this year.

adriancole:
meanwhile in preparation I'll search all github for http
instrumentation not yet using our request types and try to help them
update. This will reduce the surface area ideally to the "really
local" so that when we go 6, this is less a concern and inventoried to
some extent.

thanks for sounding the alarm!

adriancole:
https://github.com/openzipkin/brave/issues/1211

Issue #1207, opened 2020-05-11
brave.spring.webmvc.ITSpanCustomizingHandlerInterceptor is Flakey


adriancole:
[INFO] Running brave.spring.webmvc.ITSpanCustomizingHandlerInterceptor

11:58:24,790 WARN  [qtp2057313197-409] servlet.DispatcherServlet (DispatcherServlet.java:1116) - No mapping found for HTTP request with URI [/foo/bark] in DispatcherServlet with name 'org.springframework.web.servlet.DispatcherServlet-4f5df8c7'

[ERROR] Tests run: 33, Failures: 1, Errors: 0, Skipped: 0, Time elapsed: 8.759 s <<< FAILURE! - in brave.spring.webmvc.ITSpanCustomizingHandlerInterceptor

[ERROR] brave.spring.webmvc.ITSpanCustomizingHandlerInterceptor.spanHandlerSeesException_async  Time elapsed: 0.167 s  <<< FAILURE!

java.lang.AssertionError: 

Expecting actual not to be null

	at brave.test.http.ITHttpServer.spanHandlerSeesError(ITHttpServer.java:533)

	at brave.test.http.ITHttpServer.spanHandlerSeesException_async(ITHttpServer.java:512)

	at java.base/jdk.internal.reflect.NativeMethodAccessorImpl.invoke0(Native Method)

	at java.base/jdk.internal.reflect.NativeMethodAccessorImpl.invoke(NativeMethodAccessorImpl.java:62)

	at java.base/jdk.internal.reflect.DelegatingMethodAccessorImpl.invoke(DelegatingMethodAccessorImpl.java:43)

	at java.base/java.lang.reflect.Method.invoke(Method.java:567)

	at org.junit.runners.model.FrameworkMethod$1.runReflectiveCall(FrameworkMethod.java:59)

	at org.junit.internal.runners.model.ReflectiveCallable.run(ReflectiveCallable.java:12)

	at org.junit.runners.model.FrameworkMethod.invokeExplosively(FrameworkMethod.java:56)

	at org.junit.internal.runners.statements.InvokeMethod.evaluate(InvokeMethod.java:17)

	at org.junit.internal.runners.statements.RunBefores.evaluate(RunBefores.java:26)

	at org.junit.internal.runners.statements.RunAfters.evaluate(RunAfters.java:27)

	at org.junit.internal.runners.statements.FailOnTimeout$CallableStatement.call(FailOnTimeout.java:288)

[...]

PR #1208, opened 2020-05-12
Attempts to deflake servlet after Jetty 9 update

adriancole:
After moving to Jetty 9, we get occasional flakes seeing errors in servlet
requests. This moves the atomic boolean around to help rule out redundnant calls
as a problem. Possibly, though it could be that something is eating the error
or that the request is completing prior to the error being visible somehow.

See #1207


adriancole:
ahah!

```
ITSpanCustomizingAsyncHandlerInterceptor>ITHttpServer.spanHandlerSeesException_async:513->ITHttpServer.spanHandlerSeesError:538 Span finished multiple times
```

PR #1209, opened 2020-05-12
Adds SpanHandler.handesAbandoned to emphasize it is an edge case

adriancole:
ABANDONED is an edge case which literally means don't send data. This
makes it opt in instead of the inverse which is work and confusion
causing.


brave/src/main/java/brave/internal/handler/NoopAwareSpanHandler.java
```diff
@@ -75,10 +85,18 @@ public static SpanHandler create(SpanHandler[] handlers,
   }
 
   static final class CompositeSpanHandler extends SpanHandler {
+    final long handlesAbandonedBitset;
     final SpanHandler[] handlers;
 
     CompositeSpanHandler(SpanHandler[] handlers) {
       this.handlers = handlers;
+      long handlesAbandonedBitset = 0;
```
anuraaga, line 93: Is the bitset to allow `handlesAbandoned` to be somewhat slow (maybe reading from a config value). Pushing that to the user may not be a big deal since I think it's pretty common for frameworks to read these sort of methods inline, even in JDK.

brave/src/main/java/brave/internal/handler/NoopAwareSpanHandler.java
```diff
@@ -89,12 +107,25 @@ public static SpanHandler create(SpanHandler[] handlers,
     }
 
     @Override public boolean end(TraceContext context, MutableSpan span, Cause cause) {
-      for (SpanHandler handler : handlers) {
-        if (!handler.end(context, span, cause)) return false;
+      if (cause != Cause.ABANDONED) {
+        for (SpanHandler handler : handlers) {
```
anuraaga, line 111: If not bitset think this becomes just

```java
for (SpanHandler handler : handlers) {
  if (cause != ABANDONING || handler.handlesAbandon()) {
    handle()
```

brave/src/main/java/brave/internal/handler/NoopAwareSpanHandler.java
adriancole: sounds fine to me

brave/src/main/java/brave/internal/handler/NoopAwareSpanHandler.java
adriancole: much better thanks! eb394a63e

PR #1210, opened 2020-05-12
Adds MutableSpan.removeTag for common case conversion of tags to fields

adriancole:
This allows an easier syntax for re-mapping tags as fields.

Ex.
```java
String remoteServiceName = span.removeTag("peer.service");
if (remoteServiceName != null) span.remoteServiceName(remoteServiceName);
```

See https://github.com/spring-cloud/spring-cloud-sleuth/issues/1620

### DIFF
--- a/brave/src/test/java/brave/handler/MutableSpanTest.java
+++ b/brave/src/test/java/brave/handler/MutableSpanTest.java
@@ -100,6 +100,18 @@ public class MutableSpanTest {
   }
 
   /** This is a compile test to show how the signature is intended to be used */
+  @Test public void removeTag_usageExplained() {
+    MutableSpan span = new MutableSpan();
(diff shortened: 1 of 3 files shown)


brave/src/test/java/brave/handler/MutableSpanTest.java
```diff
@@ -99,6 +99,18 @@
     );
   }
 
+  /** This is a compile test to show how the signature is intended to be used */
```
anuraaga, line 102: This comment seems to be drifted on multiple tests so no need to fix now, but realize it doesn't seem correct (we have assertions it's not just a compile test).

adriancole: agreed!

adriancole:
thanks for the look @anuraaga!

PR #1213, opened 2020-05-14
Demonstrates how SpanHandler can now set root baggage

adriancole:
This is something we couldn't do before without depending on HTTP. With
SpanHandler, we can now set default baggage even based on root by heuristically
looking to see if the value was formerly set. (you can't depend on lack of
parent ID as some formats remove this).

https://gitter.im/spring-cloud/spring-cloud-sleuth?at=5ebaa80f7c04b92f535b5844

### DIFF
--- a/brave/src/test/java/brave/features/propagation/SetOnceBaggageTest.java
+++ b/brave/src/test/java/brave/features/propagation/SetOnceBaggageTest.java
@@ -1,0 +1,109 @@
+/*
+ * Copyright 2013-2020 The OpenZipkin Authors
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License"); you may not use this file except
+ * in compliance with the License. You may obtain a copy of the License at
+ *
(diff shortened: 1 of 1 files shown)


brave/src/test/java/brave/features/propagation/SetOnceBaggageTest.java
```diff
@@ -0,0 +1,109 @@
...
+ */
+public class SetOnceBaggageTest {
+  static final BaggageField EPOCH_SECONDS = BaggageField.create("epoch_seconds");
+
+  static final class RootOnlyBaggage extends SpanHandler {
+    @Override
+    public boolean begin(TraceContext context, MutableSpan span, @Nullable TraceContext parent) {
+      if (EPOCH_SECONDS.getValue(context) == null) { // only set at the first span
+        long epochSeconds = System.currentTimeMillis() / 1000;
+        sleepSlightlyOverASecond();
```
adriancole, line 49: this is only to make the value change for the two traces :P

PR #1214, opened 2020-05-15
Test maintenance

adriancole:
* Fixes bug where ITHttpClient used a pattern where it shouldn't
* Fixes ITRemote which should use a sensible service name (as spans end up in logs)
* Adds some basic tests of the IT fixtures

### DIFF
--- a/brave/src/test/java/brave/internal/recorder/PendingSpansTest.java
+++ b/brave/src/test/java/brave/internal/recorder/PendingSpansTest.java
@@ -13,15 +13,19 @@
  */
 package brave.internal.recorder;
 
+import brave.Clock;
 import brave.GarbageCollectors;
 import brave.handler.MutableSpan;
 import brave.handler.SpanHandler;
 import brave.internal.InternalPropagation;
+import brave.internal.Nullable;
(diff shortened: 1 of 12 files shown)


adriancole:
polished up orphan tracker and backfilled tests as otherwise it is a guessing game in CI what caused the orphan (such as https://github.com/spring-cloud/spring-cloud-sleuth/issues/1635)

brave/src/test/java/brave/internal/recorder/PendingSpansTest.java
```diff
@@ -158,12 +166,12 @@ public void reportOrphanedSpans_afterGC() {
     assertThat(spans).hasSize(2);
     // orphaned without data
     assertThat(spans.get(0).id()).isEqualTo("0000000000000002");
-    assertThat(spans.get(0).containsAnnotation("brave.flush")).isTrue();
+    assertThat(spans.get(0).containsAnnotation("brave.flush")).isFalse();
```
adriancole, line 169: this is fixed now also

adriancole:
thanks for the look @anuraaga

PR #1216, opened 2020-05-17
Polish based on use in secondary sampling and sleuth

adriancole:
* notably secondary sampling needs to undo the shared flag
* NPE due to forgotten decrement on tag
* SpringBootTest timeout

### DIFF
--- a/brave-tests/src/main/java/brave/test/IntegrationTestSpanHandler.java
+++ b/brave-tests/src/main/java/brave/test/IntegrationTestSpanHandler.java
@@ -147,7 +147,7 @@ public final class IntegrationTestSpanHandler extends SpanHandler implements Tes
     MutableSpan result;
     try {
       // Same rationale as ITRemote: tests should never take 5s in normal circumstances.
-      // In practice, @SpringBootTest tests overrun 3-5 seconds in CI.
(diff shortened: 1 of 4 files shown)


brave-tests/src/main/java/brave/test/IntegrationTestSpanHandler.java
```diff
@@ -146,7 +146,9 @@ public MutableSpan takeLocalSpan() {
   MutableSpan doTakeSpan(boolean flushed) {
     MutableSpan result;
     try {
-      result = spans.poll(3, TimeUnit.SECONDS);
+      // Same rationale as ITRemote: tests should never take 5s in normal circumstances.
+      // In practice, @SpringBootTest tests overrun 3-5 seconds in CI.
+      result = spans.poll(20, TimeUnit.SECONDS);
```
adriancole, line 151: I always wanted this to be less, but I feel like moving it to 5s will just result in another PR to move it to 10

anuraaga: So how about set to 10? ;)

adriancole:
> So how about set to 10? ;)
11.5 and that's my final offer!

PR #1218, opened 2020-05-18
Internally synchronizes TestSpanHandler

adriancole:
This synchronizes `TestSpanHandler` internally to help people migrate
off tools like Sleuth's `ArrayListSpanReporter` which is pinned to
Zipkin types.

### DIFF
--- a/brave-tests/src/main/java/brave/test/TestSpanHandler.java
+++ b/brave-tests/src/main/java/brave/test/TestSpanHandler.java
@@ -13,12 +13,14 @@
  */
 package brave.test;
 
+import brave.Span;
 import brave.handler.MutableSpan;
 import brave.handler.SpanHandler;
 import brave.propagation.TraceContext;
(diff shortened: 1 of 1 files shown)


brave-tests/src/main/java/brave/test/TestSpanHandler.java
```diff
@@ -41,34 +43,62 @@
  * }
  * }</pre>
  *
+ * <h3>Comparison with {@link IntegrationTestSpanHandler}</h3>
+ * It is possible to use this type in multi-threaded tests, but there are usually problems that
+ * arise better solved by {@link IntegrationTestSpanHandler}. Here's a few examples.
```
adriancole, line 48:
```suggestion
 * arise better solved by {@link IntegrationTestSpanHandler}. Here are a few examples.
```

adriancole:
thanks for the look rag!

PR #1219, opened 2020-05-18
Lets those using IntegrationTestSpanHandler know the difference between orphans and spans never started

adriancole:
This integrates orphan handling so it is easier to tell if a span was
orphaned, or simply never started. This came up in Sleuth where some
integration tests never started a span and I didn't know which bug it
was.

### DIFF
--- a/brave-tests/src/main/java/brave/test/IntegrationTestSpanHandler.java
+++ b/brave-tests/src/main/java/brave/test/IntegrationTestSpanHandler.java
@@ -134,7 +134,7 @@ public final class IntegrationTestSpanHandler extends SpanHandler implements Tes
     intentionallyWrongDefaultSpan.tag("not", "me");
     orphanTracker = OrphanTracker.newBuilder()
         .defaultSpan(intentionallyWrongDefaultSpan)
-        // This will not be exactly the right timestamp Tracing.Builder.clock is overridden.
+        // This will not be exactly the right timestamp when Tracing.Builder.clock is overridden.
         // However, it is rare to override this in an integration test and complex to pass it.
         .clock(Platform.get().clock())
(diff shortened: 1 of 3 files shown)


brave-tests/src/test/java/brave/test/IntegrationTestSpanHandlerTest.java
```diff
@@ -28,6 +29,25 @@
   TraceContext context = TraceContext.newBuilder().traceId(1L).spanId(2L).build();
   MutableSpan span = new MutableSpan(context, null);
 
+  @Test public void goodMessageForUnstartedSpan() {
+    spanHandler.end(context, span, FINISHED); // NOT ABANDONED!
```
adriancole, line 33:
```suggestion
    spanHandler.end(context, span, FINISHED); // NOT ORPHANED!
```

brave-tests/src/main/java/brave/test/IntegrationTestSpanHandler.java
```diff
@@ -90,14 +116,31 @@
...
 
+  public IntegrationTestSpanHandler() {
+    // OrphanTracker detects to see if it should add "brave.flushed" or not, as it is used in
+    // production some times and avoiding this could be helpful. This forces a failed match,
+    // so that we can detect orphans even when no data was added.
+    MutableSpan intentionallyWrongDefaultSpan = new MutableSpan();
+    intentionallyWrongDefaultSpan.tag("not", "me");
+    orphanTracker = OrphanTracker.newBuilder()
+        .defaultSpan(intentionallyWrongDefaultSpan)
+        // This will not be exactly the right timestamp Tracing.Builder.clock is overridden.
```
adriancole, line 137:
```suggestion
        // This will not be exactly the right timestamp when Tracing.Builder.clock is overridden.
```

PR #1220, opened 2020-05-18
Fixes test order in ITHttpClient and ITHttpServer.spanHandlerSeesError

adriancole:
This fixes order where a callback that checked for duplicate errors
happened after span reporting. Depending on how fast the verification
of the span went, the error checker could fail.

This also renames the base test handler so it is more clear which is
which.

https://github.com/line/armeria/pull/2695#issuecomment-630102560

### DIFF
--- a/instrumentation/http-tests/src/main/java/brave/test/http/ITHttpServer.java
+++ b/instrumentation/http-tests/src/main/java/brave/test/http/ITHttpServer.java
@@ -36,8 +36,6 @@ import java.util.Arrays;
 import java.util.LinkedHashSet;
 import java.util.Set;
 import java.util.concurrent.ConcurrentLinkedDeque;
-import java.util.concurrent.atomic.AtomicInteger;
-import java.util.concurrent.atomic.AtomicReference;
 import java.util.logging.Level;
 import java.util.logging.Logger;
 import javax.servlet.UnavailableException;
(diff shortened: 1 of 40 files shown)


instrumentation/http-tests/src/main/java/brave/test/http/ITHttpServer.java
```diff
@@ -530,13 +529,15 @@ void spanHandlerSeesError(String path) throws IOException {
             return true;
           }
         })
+        .addSpanHandler(testSpanHandler)
```
adriancole, line 532: This is the part that should fix the bug

instrumentation/http-tests/src/main/java/brave/test/http/ITHttpServer.java
```diff
@@ -530,13 +529,15 @@ void spanHandlerSeesError(String path) throws IOException {
...
         .build());
     init();
 
+    // If this passes, a span was reported with an error
     httpStatusCodeTagMatchesResponse_onUncaughtException(path, ".*not ready");
 
     assertThat(caughtThrowables)
-        .withFailMessage("Span didn't finish")
+        .withFailMessage("Span finished with error, but caughtThrowables empty")
```
adriancole, line 540: horrible error.. my fault

PR #1222, opened 2020-05-19
Adds break-glass release instructions

adriancole:


adriancole:
@jorgheymans @anuraaga If you are up to being able to do releases when things are busted, you'll need to add a sonatype account if you don't already have one, and ask to be added here https://issues.sonatype.org/browse/OSSRH-16669

RELEASE.md
```diff
@@ -84,14 +84,19 @@ SONATYPE_PASSWORD=your_sonatype_password
 VERSION=xx-version-to-release-xx
 
 # now from latest master, prepare the release. We are intentionally deferring pushing commits
-./mvnw --batch-mode -s ./.settings.xml -Prelease -nsu -DreleaseVersion=$VERSION -Darguments="-DskipTests -Dlicense.skip=true" release:prepare  -DpushChanges=false
```
jorgheymans, line 87: curious as to why batch mode is being removed here ?

adriancole: because it is manual anyway. that was my thinking. we have batch mode for the release scripts and this was copy/paste from that.

jorgheymans: okay, only thing is that you'll have these annoying download progress indicators in the log etc but i guess that's not a big issue.

adriancole: no problem I will put it back, just half the things were one and half the other. batch mode works for me

adriancole:
thanks for the feedback @jorgheymans helpful!

PR #1224, opened 2020-06-01
Adds MessagingRequest.messageId

adriancole:
This adds MessagingRequest.messageId typically visible during response processing on the producer, or request processing side on the consumer

### DIFF
--- a/instrumentation/messaging/RATIONALE.md
+++ b/instrumentation/messaging/RATIONALE.md
@@ -78,3 +78,19 @@ continue incoming state.
 Finally, this approach has been in use since late 2017, when we refined our
 only messaging instrumentation, `kafka-clients`, to support message processing.
 By re-using known working practice, we have less risk in abstraction.
+
+## Why don't we define a message ID for tools lacking one, such as Kafka?
+
+Typical message ID formats encode multiple components such as a broker ID or network address,
+destination, timestamp or offset. It may be tempting to compose a format to include the dimensions
+that likely pinpoint a message when there's no format defined by the library. For example, in Kafka,
+we could compose a format like `topic-partition-offset` to ensure `MessagingRequest.id()` would not
+be `null`, and people can access not-yet-standard fields such as partition or offset.
+
+If we did that, we'd add overhead with the only consumer being tracing itself. It would fail as a
+correlation field with other libraries as by definition our format would be bespoke. Moreover,
+higher layers of abstraction which might have a defined message ID format could be confused with
+ours. Later, if that same tool creates a message ID format, it would likely be different than ours.
+
+For reasons including these, if there's no canonical format, we opt out of synthesizing a message ID
+and just return `null`.
--- a/instrumentation/messaging/RATIONALE.md
+++ b/instrumentation/messaging/RATIONALE.md
@@ -80,46 +80,50 @@ only messaging instrumentation, `kafka-clients`, to support message processing.
 By re-using known working practice, we have less risk in abstraction.
 
 ## Message ID
-In practice, the message ID is used for retrieval, correlation, duplicate detection or a combination
-of these. We derive semantics by looking at multiple open source projects and cloud services, as
-well the special case of JMS.
+In practice, the message ID is used for deletion, visibility control, retrieval, correlation,
+duplicate detection or a combination of these. We derive semantics by looking at multiple open
+source projects and cloud services, as well the special case of JMS.
 
-| System   | Kind     | Operation      | Direction | Field                  | Owner  | Scope      | Format
-|----------|----------|----------------|-----------|------------------------|--------|------------|--------
-| AMQP     | PRODUCER | publish        | Request   | message-id             | Local  | Global     | 1-255 characters
-| AMQP     | CONSUMER | consume        | Request   | message-id             | Remote | Global     | 1-255 characters
-| Artemis  | CONSUMER | receive        | Request   | messageId              | Remote | Global     | random uint64
-| Kafka    | PRODUCER | send           | Response  | offset                 | Remote | Topic      | uint64
-| Kafka    | CONSUMER | poll           | Request   | offset                 | Remote | Topic      | uint64
-| Kinesis  | PRODUCER | Publish        | Response  | SequenceNumber         | Remote | Stream     | 1-128 digits
-| Kinesis  | CONSUMER | Lambda         | Request   | sequenceNumber         | Remote | Global     | 1-128 digits
-| SQS      | PRODUCER | SendMessage    | Request   | MessageDeduplicationId | Local  | Queue      | SHA-256(body)
-| SQS      | CONSUMER | ReceiveMessage | Request   | MessageDeduplicationId | Remote | Queue      | SHA-256(body)
-| SQS      | PRODUCER | SendMessage    | Response  | MessageId              | Remote | Global     | UUID
-| SQS      | CONSUMER | ReceiveMessage | Response  | MessageId              | Remote | Global     | UUID
-| SNS      | PRODUCER | Publish        | Response  | MessageId              | Remote | Global     | UUID
-| SNS      | CONSUMER | POST           | Request   | x-amz-sns-message-id   | Remote | Global     | UUID
-| PubSub   | PRODUCER | Publish        | Response  | message_id             | Remote | Topic      | Integer
-| PubSub   | CONSUMER | Push           | Request   | message_id             | Remote | Topic      | Integer
-| PubSub   | CONSUMER | Pull           | Response  | message_id             | Remote | Topic      | Integer
-| JMS      | PRODUCER | Send           | Response  | JMSMessageId           | Remote | Global     | ID:opaque string
-| JMS      | CONSUMER | Receive        | Request   | JMSMessageId           | Remote | Global     | ID:opaque string
-| MQTT     | PRODUCER | PUBLISH        | Request   | Packet Identifier      | Local  | Connection | uint16
-| MQTT     | PRODUCER | PUBACK/PUBREC  | Response  | Packet Identifier      | Local  | Connection | uint16
-| MQTT     | CONSUMER | PUBLISH        | Request   | Packet Identifier      | Remote | Connection | uint16
-| MQTT     | CONSUMER | PUBACK/PUBREC  | Response  | Packet Identifier      | Remote | Connection | uint16
-| Pulsar   | PRODUCER | Send           | Response  | MessageId              | Remote | Topic      | bytes(ledger|entry|parition)
-| Pulsar   | CONSUMER | Receive        | Request   | MessageId              | Remote | Topic      | bytes(ledger|entry|parition)
-| RocketMQ | PRODUCER | send           | Response  | msgId                  | Remote | Topic      | HEX(ip|port|offset)
-| RocketMQ | CONSUMER | consumeMessage | Request   | msgId                  | Remote | Topic      | HEX(ip|port|offset)
-| STOMP    | PRODUCER | SEND           | Request   | receipt Header         | Local  | Connection | arbitrary
-| STOMP    | PRODUCER | RECEIPT/ERROR  | Response  | receipt-id Header      | Local  | Connection | arbitrary
-| STOMP    | CONSUMER | SEND           | Request   | receipt Header         | Remote | Connection | arbitrary
-| STOMP    | CONSUMER | RECEIPT/ERROR  | Response  | receipt-id Header      | Remote | Connection | arbitrary
-| STOMP    | PRODUCER | MESSAGE        | Request   | message-id Header      | Local  | Connection | arbitrary
-| STOMP    | PRODUCER | ACK/NACK       | Response  | id Header              | Local  | Connection | arbitrary
-| STOMP    | CONSUMER | MESSAGE        | Request   | message-id Header      | Remote | Connection | arbitrary
-| STOMP    | CONSUMER | ACK/NACK       | Response  | id Header              | Remote | Connection | arbitrary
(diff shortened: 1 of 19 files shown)


adriancole:
cc @JaidenAshmore

instrumentation/messaging/src/main/java/brave/messaging/MessagingResponse.java
```diff
@@ -0,0 +1,35 @@
...
+/**
+ * Abstract response type used for parsing and sampling of messaging clients and servers.
+ *
+ * @see ProducerResponse
+ * @see ConsumerResponse
+ * @since 5.13
+ */
+public abstract class MessagingResponse extends Response {
+  /**
+   * The request that initiated this messaging response or {@code null} if unknown.
```
anuraaga, line 28: Curious when this could be set - if the MQ is run in-process?

instrumentation/messaging/src/main/java/brave/messaging/MessagingRequest.java
```diff
@@ -59,13 +57,26 @@
   /**
    * Messaging channel name, e.g. "hooks" or "complaints". {@code null} if unreadable.
    *
-   * <p>Conventionally associated with the key "messaging.channel_name"
```
adriancole, line 62: ps sometime around july last year we decided to replace message.X with messaging.X I don't recall the rationale but it should have been written down. I think it was to do with how the other properties would sound if prefixed "message" also to match the abstraction name.

The only reason I mention this is that we should decide if the tag name "messaging.id" is ok for the message id or we make "messaging.message_id" which is also awkward for the same. I don't have a strong feeling either way, but happy to add to the RATIONALE whatever we choose.

instrumentation/messaging/src/main/java/brave/messaging/MessagingResponse.java
adriancole: probably not the best sentence. So, for example the messageId field is only readable during the response processing of AWS SQS.

Different than RPC I think here it isn't important that the request is what initiated the response, rather it is simply the start of the operation, available here so you can late process data like message Id

instrumentation/messaging/src/main/java/brave/messaging/MessagingRequest.java
anuraaga: Lean towards `messaging.id` since when in doubt, shorter seems better

instrumentation/messaging/src/main/java/brave/messaging/MessagingResponse.java
anuraaga: Ah I think it could be `Information about the request that...`

adriancole:
thanks for the feedback @anuraaga

instrumentation/messaging/RATIONALE.md
```diff
@@ -78,3 +78,19 @@ continue incoming state.
...
+we could compose a format like `topic-partition-offset` to ensure `MessagingRequest.id()` would not
+be `null`, and people can access not-yet-standard fields such as partition or offset.
+
+If we did that, we'd add overhead with the only consumer being tracing itself. It would fail as a
+correlation field with other libraries as by definition our format would be bespoke. Moreover,
+higher layers of abstraction which might have a defined message ID format could be confused with
+ours. Later, if that same tool creates a message ID format, it would likely be different than ours.
+
+For reasons including these, if there's no canonical format, we opt out of synthesizing a message ID
+and just return `null`.
```
jorgheymans, line 96: +1, chances are very high that the trace or span contains enough contextual information to pinpoint to a specific message anyway if ever needed.

instrumentation/messaging/src/main/java/brave/messaging/MessagingRequest.java
```diff
@@ -66,6 +66,23 @@
...
 
+  /**
+   * The possibly system generated value that uniquely identifies this message. Return {@code null} if
+   * the ID was unreadable or the transport has no canonical message ID format.
+   *
+   * <p>Examples:
+   * <pre><ul>
+   *   <li>Amazon SQS - "MessageId" response field. ex "5fea7756-0ea4-451a-a703-a558b933e274"</li>
+   *   <li>JMS - "JMSMessageID" header. Ex "ID:10.77.42.209-4280-1477454185311-1:1:1391:1:1"</li>
+   *   <li>RabbitMQ - AMQP "message-id" (max 256 char)</li>
```
jorgheymans, line 77: Maby add `<li>Kakfa - null</li>` here once we're decided on this

adriancole:
ok pared this down to only messageId

Needs unit tests and also integration in JMSProducer which will only sometimes be able to read it (ex when async is used and also when messageID isn't turned off).

adriancole:
OK I added a huge table we can use for other fields later also, cc particularly @jeqo as this is similar to what's in the wiki https://github.com/openzipkin/openzipkin.github.io/wiki/Messaging-instrumentation-abstraction

adriancole:
@jorgheymans @jeqo given we've decided not to materialize a global ID when one doesn't exist, this doesn't mean we can't use the offset as a messageId. For example, many formats in the table have no global message ID, and I've hopefully clarified that there's no expectation for messageId to be global. I think the closest thing to messageId in kafka is offset, and since we'll have a field for partition later, it should be pretty good right? wdyt?

jorgheymans:
Should we then even populate messageId in the Kafka scenario as yes it's "the closest thing" but on its own rather meaningless.

adriancole:
@jorgheymans once we have partitionId as a normal field, it isn't meaningless is it? unless you mean to say offset is usually not needed. Either way I'm ok leaving it out for kafka.

adriancole:
bear in mind that the value of messageId in activemq is also meaningless unless you know the broker info. Most message IDs require you to at least also know the topic. What I meant to add in the RATIONALE was that message IDs do not imply global scope. Given that, I feel like it is a question indeed if we should choose to have no messageId in Kafka. I don't feel strongly, but I'd like to put a reason down.

adriancole:
All major services are now in. We are at a crossroads though.. Do we try to ensure there's a messageID field? or do we just point to other properties that serve similar functions.

Personally, I feel it is too messy to call something else messageID than messageId since we agreed to have partitionId/partitionOffset. Literally say messageID is for things like AMQP and SaaS offerings who define a field named such. If a messageId field doesn't exist, even if something is similar, return `null`.

In the case of something similar existing, we can have a `@see` link saying that partitioned brokers like kafka and rocketmq should look at paritionId and partitionOffset. This will become more even more clear when we define protocol field.

It is true that Azure servicebus has both messageId and paritition related fields, but I think this doesn't need mentioning apart from saying these fields aren't mutually exclusive, even though often are.


so the run down of fields we'd eventually add to our abstraction: 
messageId, correlationId (both of these are strings in AMQP and JMS, some of these in other things)
partitionId, partitionOffset (both of these numeric, the former -1 if not applicable the later interpreted based on partitionId so that it can be an unboxed unsigned long)

the special case is azure bus has all 4. I don't think any other service has all 4 fields, and even in azure it would be rare to see all 4 fields at the same time

while 4 fields seem like a lot of "ids" it is actually not all the things... it doesn't cover ack ids for example. However, it is the parts list to identify a message whether it breaks trace across a link or not.

What do you think?

anuraaga:
I agree with just matching other definitions of message id and not being required since there doesn't seem to be a consistent or useful common definitions for it. Though it does sort of make me wonder whether it deserves a special field than and not just be a tag - presumably the most value from the ID will be from linking back to another cli to debug the message and not sometimes within brave. If I recall correctly, we use things like route for sampling decisions, and fields like topic seem to provide similar utility. Not sure what a user will actually do with a message ID.

instrumentation/messaging/RATIONALE.md
```diff
@@ -78,3 +78,134 @@ continue incoming state.
...
+we could compose a format like `topic-partition-offset` to ensure `MessagingRequest.id()` would have
+all identifying information.
+
+If we did that, we'd add overhead with the only consumer being tracing itself. It would fail as a
+correlation field with other libraries as by definition our format would be bespoke. Moreover,
+higher layers of abstraction which might have a defined message ID format could be confused with
+ours. Later, if that same tool creates a message ID format, it would likely be different from ours.
+
+For reasons including these, if there's no message ID field, we fall back to a numeric property like
+offset or sequence, and failing that just return `null`.
```
jeqo, line 211: If we agree with keeping messageId only when this field exists, we should update this line.
```suggestion
For reasons including these, if there's no message ID field, and failing that just return `null`.
```

adriancole: thanks for the help!

adriancole:
@anuraaga good point. I added the why section

instrumentation/messaging/RATIONALE.md
```diff
@@ -78,3 +78,155 @@ continue incoming state.
...
+combination of these use cases.
+
+The purpose of the Message ID, as defined here, is a correlation between one "send" event and any
+number of "receive" events. Not all operations can bet tracked in one trace. Some consumers log, but
+don't trace. Others break traces, or trace in different systems. Moreover, this project doesn't
+trace all operations. With the message ID, processes curious about an ACK or NACK can find logs
+related to those operations, even if the logs do not include trace IDs.
+
+### What are some examples of message IDs?
+We derive semantics by looking at multiple open source projects and cloud services, as well the
```
jorgheymans, line 95:
```suggestion
We derive semantics by looking at multiple open source projects and cloud services, as well as the
```

instrumentation/messaging/RATIONALE.md
```diff
@@ -78,3 +78,155 @@ continue incoming state.
...
 By re-using known working practice, we have less risk in abstraction.
+
+## Message ID
+
+### Why would someone tag message ID?
+The message ID allows deletion, visibility control, retrieval, correlation, duplicate detection or a
+combination of these use cases.
+
+The purpose of the Message ID, as defined here, is a correlation between one "send" event and any
+number of "receive" events. Not all operations can bet tracked in one trace. Some consumers log, but
```
jorgheymans, line 89:
```suggestion
number of "receive" events. Not all operations can be tracked in one trace. Some consumers log, but
```

instrumentation/messaging/RATIONALE.md
```diff
@@ -78,3 +78,155 @@ continue incoming state.
...
+| STOMP       | CONSUMER | MESSAGE        | Request   | message-id Header           | Remote | Connection | arbitrary
+| STOMP       | CONSUMER | ACK/NACK       | Response  | id Header                   | Remote | Connection | arbitrary
+
+### What are some examples of IDs that aren't message IDs?
+TODO: move some rows above into here
+
+### Isn't correlation ID the same as a message ID?
+A correlation ID is a system-wide lookup value that possibly can pass multiple steps. A message ID
+can have a scope as small as one segment (Ex. flow from producer to broker), and typically an
+implementation detail. It can bet the case that they are the same, but it is not commonly the case.
```
jorgheymans, line 145: you writing this from a hotel room in Vegas ?
```suggestion
implementation detail. It can be the case that they are the same, but it is not commonly the case.
```

instrumentation/messaging/RATIONALE.md
```diff
@@ -78,3 +78,155 @@ continue incoming state.
...
+message ID field borders on arbitrary. Here are some examples to reinforce this.
+
+Amazon SQS includes a service-generated `MessageId` that can identify a message later consumed.
+However, to delete an instance of that message you need one of potentially many `ReceiptHandle`s
+associated with the `MessageId`. The client can also set certain IDs. For example, a client sets
+`MessageDeduplicationId` before sending a message to a FIFO queue to suppress redundant sends. In
+other words there are at least 3 identifiers for a single message, in different formats, depending
+on the use case.
+
+Azure Queue a service-generated `MessageId`, but also a client-generated `x-ms-client-request-id`
```
jorgheymans, line 179:
```suggestion
Azure Queue includes a service-generated `MessageId`, but also a client-generated `x-ms-client-request-id`
```

instrumentation/messaging/RATIONALE.md
```diff
@@ -78,3 +78,155 @@ continue incoming state.
...
+that likely pinpoint a message when there's no format defined by the library. For example, in Kafka,
+we could compose a format like `topic-partition-offset` to ensure `MessagingRequest.id()` would have
+all identifying information.
+
+If we did that, we'd add overhead with the only consumer being tracing itself. It would fail as a
+correlation field with other libraries as by definition our format would be bespoke. Moreover,
+higher layers of abstraction which might have a defined message ID format could be confused with
+ours. Later, if that same tool creates a message ID format, it would likely be different from ours.
+
+For reasons including these, if there's no message ID field, and failing that just return `null`.
```
jorgheymans, line 232: :thinking:  ... ThisSentenceCannotBeGrokkedException 
```suggestion
For reasons including these, if there's no message ID field, and failing that just return `null`.
```
the `and failing that` part confuses me, what do you mean

adriancole:
ok I think the doc side is done. There's not much to do on the code. I'll do that now.

adriancole:
@jorgheymans 

>Should we then even populate messageId in the Kafka scenario as yes it's "the closest thing" but on its own rather meaningless.

What I did was stub a comment so implementors know what's going on. A soon PR will add offset and in the process javadoc link to suggest offset/sequence could be present

instrumentation/messaging/RATIONALE.md
adriancole: hopefully fixed

adriancole:
@anuraaga @jeqo @jorgheymans @llinder many thanks for the input on this.. 2 weeks well spent! I'll raise the partition related things next, following that the tags impls

PR #1229, opened 2020-06-20
Consistently only clear trace ID headers in messaging

adriancole:
Before, we cleared only trace ID headers in messaging, but cleared all
headers in Kafka and Rabbit. The problem in the latter was that it would
clear baggage like "user_id" and "country_code". These don't interfere
with the trace tree and worse, can prevent users from seeing correlation
IDs in third party code that doesn't use Brave.

The fix here is to clarify code and docs, moving implementation of Kafka
and RabbitMQ to match JMS.

### DIFF
--- a/brave/src/main/java/brave/propagation/Propagation.java
+++ b/brave/src/main/java/brave/propagation/Propagation.java
@@ -156,26 +156,42 @@ public interface Propagation<K> {
   }
 
   /**
-   * The propagation fields defined. If your request is reused, you should delete the fields here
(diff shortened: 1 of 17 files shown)


adriancole:
I've not completed this as I need unit tests etc. Just raising as I think the main points are important to think about sooner vs later. This will also drift #1047

adriancole:
Thanks for the look @anuraaga! @jeqo @jorgheymans feel free to give feedback post-merge and I'll follow-up.

PR #1231, opened 2020-06-21
Adds client side of HTTP OPTIONS and empty path tests

adriancole:
This makes the client tests have parity with server ones. Notably, this
tests OPTIONS requests work, and that "" coerses to "/".


jcchavezs:
Question: do we also (or remove) add trailing slash for /mypath or /mypath/
?

adriancole:
> do we also (or remove) add trailing slash for /mypath or /mypath/

I've noticed in micrometer that duplicate slashes are removed, leading slashes are asserted always and trailing slashes are removed. This is a slightly bigger change than asserting no empty path, but we can chat about it... maybe in a new issue?

PR #1234, opened 2020-07-07
Annotate Response#unwrap with @Nullable

okue:
Annotate `unwrap()` with `@Nullable` like other functions returning `null` such as `error()`, `request()`.


adriancole:
thanks! lucky pull request number, too!

Issue #1236, opened 2020-07-12
Add support for batch consumer (Receiving messages as arraylist)


jamesdube:
I have a queue listener where I receive messages as an array list (see method signature) and the tracing listener is trying to cast that to a message hence throwing an exception. Is there a way i can override this or anything, i'd want to add tracing to my application but this is becoming a stumbling block The offending lines of code are highlighted below:

https://github.com/openzipkin/brave/blob/6c44ccd5f545013369059e2223b48caef476c889/instrumentation/spring-rabbit/src/main/java/brave/spring/rabbit/TracingRabbitListenerAdvice.java#L75

Spring Boot Version: ` 2.2.7.RLEASE`

Spring Cloud Version: `Hoxton.RELEASE`

Rabbit queue listener:
```java
    @Override
    @RabbitListener(
            queues = "successful-transactions",
            containerFactory = "simpleRabbitListenerContainerFactory"
    )
    public void dequeue(List<Message> transactions) {
        logger.info("Receiving -> size {}",transactions.size());
    }
```

Stacktrace:
```java
java.lang.ClassCastException: java.util.ArrayList cannot be cast to org.springframework.amqp.core.Message
	at brave.spring.rabbit.TracingRabbitListenerAdvice.invoke(TracingRabbitListenerAdvice.java:76) ~[brave-instrumentation-spring-rabbit-5.9.0.jar:na]
	at org.springframework.aop.framework.ReflectiveMethodInvocation.proceed(ReflectiveMethodInvocation.java:186) ~[spring-aop-5.2.6.RELEASE.jar:5.2.6.RELEASE]
	at org.springframework.aop.framework.JdkDynamicAopProxy.invoke(JdkDynamicAopProxy.java:212) ~[spring-aop-5.2.6.RELEASE.jar:5.2.6.RELEASE]
	at org.springframework.amqp.rabbit.listener.$Proxy94.invokeListener(Unknown Source) ~[na:2.2.6.RELEASE]
	at org.springframework.amqp.rabbit.listener.AbstractMessageListenerContainer.invokeListener(AbstractMessageListenerContainer.java:1486) ~[spring-rabbit-2.2.6.RELEASE.jar:2.2.6.RELEASE]
	at org.springframework.amqp.rabbit.listener.AbstractMessageListenerContainer.doExecuteListener(AbstractMessageListenerContainer.java:1481) ~[spring-rabbit-2.2.6.RELEASE.jar:2.2.6.RELEASE]
[...]

PR #1237, opened 2020-07-13
Add support for rabbitmq batch consume

dengliming:
for [https://github.com/openzipkin/brave/issues/1236](https://github.com/openzipkin/brave/issues/1236)

@adriancole  I try fixed this. Please see if that makes sense.

### DIFF
--- a/instrumentation/spring-rabbit/src/main/java/brave/spring/rabbit/TracingRabbitListenerAdvice.java
+++ b/instrumentation/spring-rabbit/src/main/java/brave/spring/rabbit/TracingRabbitListenerAdvice.java
@@ -31,6 +31,10 @@ import org.springframework.amqp.core.MessageProperties;
 import org.springframework.amqp.rabbit.annotation.RabbitListener;
 import org.springframework.amqp.rabbit.listener.SimpleMessageListenerContainer;
 
+import java.util.ArrayList;
+import java.util.Collection;
+import java.util.List;
+
 import static brave.Span.Kind.CONSUMER;
 import static brave.spring.rabbit.SpringRabbitTracing.RABBIT_EXCHANGE;
 import static brave.spring.rabbit.SpringRabbitTracing.RABBIT_QUEUE;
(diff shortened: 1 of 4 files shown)


instrumentation/spring-rabbit/src/main/java/brave/spring/rabbit/TracingRabbitListenerAdvice.java
```diff
@@ -72,40 +76,53 @@
...
-    TraceContextOrSamplingFlags extracted =
-      springRabbitTracing.extractAndClearTraceIdHeaders(extractor, request, message);
+      TraceContextOrSamplingFlags extracted =
+        springRabbitTracing.extractAndClearTraceIdHeaders(extractor, request, message);
 
-    // named for BlockingQueueConsumer.nextMessage, which we can't currently see
-    Span consumerSpan = springRabbitTracing.nextMessagingSpan(sampler, request, extracted);
-    Span listenerSpan = tracer.newChild(consumerSpan.context());
+      // named for BlockingQueueConsumer.nextMessage, which we can't currently see
+      Span consumerSpan = springRabbitTracing.nextMessagingSpan(sampler, request, extracted);
```
adriancole, line 94: This is an interesting situation as we want one consumer span for the batch I think.. also, there's only one listener, so should only be one listener span!

The tricky bit is that we have an ambiguous situation. For example, we may have different trace contexts in the batch of messages. Consider 5 messages, and two of them have "b3" header in them.

This "join" type operation forces us to choose which of these possibly multiple traces should be the trace of the listener.

Another option is like you've done, which is to start a span for every message, but since only one can be the "current trace" we still I think should pick one. We don't yet have a link model, but meanwhile yeah I feel like we should have only 1 span started for the listener, because only one child started in a listener could be attached to a trace.

cc @jorgheymans @anuraaga @jeqo for feedback on this.

anuraaga: I'd generally go with span per message since batching is an implementation detail of how the messages are handled. Many user requests to buy a sticker result in wanting to write a money transaction to a database, queues are great for supporting retries idempotently to make that reliable but each transaction is still independent from each other. To me, each message is very much like the request to an API server, so the messages without any header in them would just be new traces.

Note, there might be other cases where this isn't really the cases (the messages are all related and part of some larger type of processing, maybe analytics?) in which case single span could make sense too.

anuraaga: I realized, actually this change is probably related to my second paragraph - if it's a configuration change to switch from one message per listener to one batch per listener, that means a user is sort of opting into that behavior, presumably since there's a semantic relationship among the batch. So just one span does seem to make sense, but picking the trace is problematic.

dengliming: So the solution is keep only one consumer span and one listener span  (maybe from the first message)？

adriancole: so far, I don't see a better way. for example, if we see any messages we can start and finish a consumer span, but we keep a reference to only the first consumer span, if exists use that as the listener's parent

dengliming: @adriancole Please re-review. Thanks.

adriancole:
I assume rabbit will never call using empty list.. your approach is pretty simple and might well be the best start here


how about adding to the IT the annotated method for a list receiver to make sure it works?

dengliming:
> I assume rabbit will never call using empty list.. your approach is pretty simple and might well be the best start here
> 
> how about adding to the IT the annotated method for a list receiver to make sure it works?

I will add testcase this weekend. It seems that just add config 
`rabbitListenerContainerFactory.setBatchListener(true);`
and add a  list receiver to test?

adriancole:
> I will add testcase this weekend. It seems that just add config
rabbitListenerContainerFactory.setBatchListener(true);
and add a list receiver to test?

sounds correct to me!

dengliming:
> > I will add testcase this weekend. It seems that just add config
> > rabbitListenerContainerFactory.setBatchListener(true);
> > and add a list receiver to test?
> 
> sounds correct to me!

Hi~ I'm having problems adding test cases right now. Should I add a new queue to test batch consume and create a new class? If I put it in ITSpringRabbit, it's seems to large. Can you give me some ideas? Thanks.

adriancole:
you can choose to put in same IT class or refactor. some of our test
classes are large. we will later do refactoring anyway when we implement
abstract message parsing (similar to http requestparser)

dengliming:
@adriancole I've already added testcase. Would you please check if there are any other thing i missed?

instrumentation/spring-rabbit/src/test/java/brave/spring/rabbit/ITSpringRabbitTracing.java
```diff
@@ -150,4 +152,16 @@
     awaitMessageConsumed();
     // reporter rules verify nothing was reported
   }
+
+  @Test public void batchConsumerTest() {
+    produceUntracedMessage(exchange_batch.getName(), binding_batch);
```
adriancole, line 157: please add a test for two messages in all combinations as it shows a problem to solve later

non traced that listener is a new trace
first traced that listener continues that trace
two traced that listener continues first  trace but TODO we aren't handling the second.
last trace that listener is new trace and TODO about the second

dengliming: fixed at last commit.

instrumentation/spring-rabbit/src/test/java/brave/spring/rabbit/TracingRabbitListenerAdviceTest.java
```diff
@@ -196,6 +203,72 @@
...
+    MessageProperties props2 = new MessageProperties();
+    props2.setHeader("X-B3-TraceId", TRACE_ID_2);
+    props2.setHeader("X-B3-SpanId", SPAN_ID_2);
+    props2.setHeader("X-B3-ParentSpanId", PARENT_ID_2);
+    props2.setHeader("X-B3-Sampled", SAMPLED);
+    Message message2 = MessageBuilder.withBody(new byte[0]).andProperties(props2).build();
+    onBatchMessageConsumed(Arrays.asList(message, message2));
+
+    // two traced that listener continues first trace but TODO we aren't handling the second.
+    assertThat(spans.get(0))
```
adriancole, line 233: last thing from me on this.. I see you are only checking spans.get(0) the reported trace.

though I don't see any problems.. maybe ensure the B3 headers are cleared in all messages in each test after onBatchMessageConsumed similar to other tests

```
    // cleared the headers to later work doesn't try to use the old parent
    assertThat(message.getMessageProperties().getHeaders()).isEmpty();
```

dengliming: Thanks. fixed at last commit.

PR #1242, opened 2020-07-30
Fix comment; s/HttpTracing.serverName/HttpTracing.clientOf

okue:
Now, `HttpTracing#serverName` is not public method, thus replace it with `clientOf`.


adriancole:
Thank you! this was incorrect as you noticed. Appreciate that you went directly to fix without an issue first. It is easy process for all 🥇

Issue #1243, opened 2020-08-18
Brave 5.12.x OSGi bundles not resolvable


dkulp:
## Describe the Bug
The io.zipkin.brave:brave bundle is not resolvable in Karaf (and other OSGi containers) as it has a hard "Import-Package" of jdk.internal.vm.annotation.    This seems to be some sort of artifact due to building with JDK14 I think.  Not really sure.  This 5.11 versions of brave seem OK.

## Steps to Reproduce
Try resolving the bundle in karaf.

Most likely, the simple fix would be to change the brave/bnd.bnd to have:
'''
Import-Package: \
  jdk.internal.vm.annotation;optional=true,
  *
'''
which should allow it to resolve.

adriancole:
thanks for noticing this!

adriancole:
@cschneider @reta can either of you ack this way out is ok? or mention a different one?

adriancole:
one thing is we have to do this in many modules as we have manifest info here and also zipkin and zipkin-reporter projects.. ideally we can change plugin config to avoid 20 file changes.

reta:
Should be fixed by https://github.com/openzipkin/brave/pull/1247

PR #1247, opened 2020-08-19
Fixing OSGi resolution issues caused by jdk.internal.vm.annotation

reta:
Addressed https://github.com/openzipkin/brave/issues/1243, checked with 5.12.5-SNAPSHOT builds (local), the issue is gone. Thanks to  @dkulp and @adriancole .

### DIFF
--- a/brave/bnd.bnd
+++ b/brave/bnd.bnd
@@ -1,4 +1,5 @@
 Import-Package: \
+  jdk.internal.vm.annotation;resolution:=optional,\
   *
 Export-Package: \
   brave,\
--- a/pom.xml
+++ b/pom.xml
@@ -79,7 +79,7 @@
 
     <!-- use the same values in bom/pom.xml -->
     <zipkin.version>2.21.7</zipkin.version>
-    <zipkin-reporter.version>2.15.1-SNAPSHOT</zipkin-reporter.version>
+    <zipkin-reporter.version>2.15.1</zipkin-reporter.version>
 
     <!-- Ensure older versions of spring still work -->
     <spring5.version>5.2.7.RELEASE</spring5.version>
--- a/pom.xml
+++ b/pom.xml
@@ -78,8 +78,8 @@
     <errorprone.version>2.4.0</errorprone.version>
 
     <!-- use the same values in bom/pom.xml -->
-    <zipkin.version>2.21.1</zipkin.version>
-    <zipkin-reporter.version>2.15.0</zipkin-reporter.version>
+    <zipkin.version>2.21.7</zipkin.version>
+    <zipkin-reporter.version>2.15.1-SNAPSHOT</zipkin-reporter.version>
 
     <!-- Ensure older versions of spring still work -->
     <spring5.version>5.2.7.RELEASE</spring5.version>
@@ -129,7 +129,7 @@
     <maven-failsafe-plugin.version>3.0.0-M4</maven-failsafe-plugin.version>
     <maven-invoker-plugin.version>3.2.1</maven-invoker-plugin.version>
     <maven-enforcer-plugin.version>3.0.0-M3</maven-enforcer-plugin.version>
-    <maven-bundle-plugin.version>4.2.1</maven-bundle-plugin.version>
+    <maven-bundle-plugin.version>5.1.1</maven-bundle-plugin.version>
     <license-maven-plugin.version>3.0</license-maven-plugin.version>
   </properties>
 
--- a/brave-bom/pom.xml
+++ b/brave-bom/pom.xml
@@ -35,7 +35,7 @@
 
     <!-- use the same values in ../pom.xml -->
     <zipkin.version>2.21.7</zipkin.version>
-    <zipkin-reporter.version>2.15.1-SNAPSHOT</zipkin-reporter.version>
+    <zipkin-reporter.version>2.15.1</zipkin-reporter.version>
   </properties>
 
   <organization>
--- a/brave-bom/pom.xml
+++ b/brave-bom/pom.xml
@@ -34,8 +34,8 @@
     <main.basedir>${project.basedir}/..</main.basedir>
 
     <!-- use the same values in ../pom.xml -->
-    <zipkin.version>2.21.1</zipkin.version>
-    <zipkin-reporter.version>2.15.0</zipkin-reporter.version>
+    <zipkin.version>2.21.7</zipkin.version>
+    <zipkin-reporter.version>2.15.1-SNAPSHOT</zipkin-reporter.version>
   </properties>
 
   <organization>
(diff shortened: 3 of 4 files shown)


adriancole:
@reta thanks! can you verify that this doesn't recurse? ex the jars that depend on brave don't end up with the same problem. you can search for bnd files here or in console https://github.com/openzipkin/brave/find/master

this was one reason I was wondering if this could be handled at the plugin level.. possibly we should raise a JIRA on it? wdyt

reta:
@adriancole yes, sure, so far `brave` seems to be the only module which needs this dependency. CXF uses

```
io.zipkin.zipkin2:zipkin
io.zipkin.reporter2:zipkin-reporter
io.zipkin.reporter2:zipkin-reporter-brave
io.zipkin.brave:brave
io.zipkin.brave:brave-instrumentation-http
```

I have checked also

```
io.zipkin.brave:brave-instrumentation-messaging
io.zipkin.brave:brave-instrumentation-rpc
```

Others require dedicated JARs but none of them required `jdk.internal.vm.annotation`

adriancole:
super weird.. I wonder then what is causing the dep on `jdk.internal.vm.annotation`...

reta:
@adriancole I found it, it is `@Hidden` annotation which comes from lambda- generated code for `zipkin2.reporter.okhttp3.OkHttpSender` class, it is in `zipkin-sender-okhttp3` module

adriancole:
@reta cool. we can un-lambda that. great detective work!

adriancole:
So we only had that one lambda in all of the src/main of zipkin-reporter. I used intellij's "structural search" to verify the one you said was it (outside tests and benchmarks).

Screen Shot 2020-08-21 at 8 08 30 AM
Screen Shot 2020-08-21 at 8 08 43 AM
Screen Shot 2020-08-21 at 8 09 08 AM

adriancole:
It is still strange, though... isn't it?.. There's no compile dep on okhttp from io.zipkin.brave:brave

reta:
@adriancole that's correct, so if you disassemble it using `javap`, you will see the `jdk/internal/vm/annotation/Hidden` in there (at the end of the snippet, this class seems to be generated by `retrolambda`), for example (this is taken from 2.15.0 version):

```
$ javap -v OkHttpSender\$\$Lambda\$1.class                                                         
Classfile /C:/Users/reta/Downloads/OkHttpSender$$Lambda$1.class                                                
  Last modified May 17, 2020; size 709 bytes                                                                   
  MD5 checksum 10eb8be2fbfee9f83f6a2b3b5b0193dd                                                                
  Compiled from "OkHttpSender.java"                                                                            
final class zipkin2.reporter.okhttp3.OkHttpSender$$Lambda$1 implements java.util.concurrent.ThreadFactory      
  minor version: 0                                                                                             
  major version: 51                                                                                            
  flags: (0x1030) ACC_FINAL, ACC_SUPER, ACC_SYNTHETIC                                                          
  this_class: #2                          // zipkin2/reporter/okhttp3/OkHttpSender$$Lambda$1                   
  super_class: #4                         // java/lang/Object                                                  
  interfaces: 1, fields: 1, methods: 4, attributes: 1                                                          
Constant pool:                                                                                                 
   #1 = Utf8               zipkin2/reporter/okhttp3/OkHttpSender$$Lambda$1                                     
   #2 = Class              #1             // zipkin2/reporter/okhttp3/OkHttpSender$$Lambda$1                   
   #3 = Utf8               java/lang/Object                                                                    
   #4 = Class              #3             // java/lang/Object                                                  
   #5 = Utf8               java/util/concurrent/ThreadFactory                                                  
   #6 = Class              #5             // java/util/concurrent/ThreadFactory                                
   #7 = Utf8               OkHttpSender.java                                                                   
   #8 = Utf8               instance                                                                            
   #9 = Utf8               Lzipkin2/reporter/okhttp3/OkHttpSender$$Lambda$1;                                   
  #10 = Utf8               <init>                                                                              
  #11 = Utf8               ()V                                                                                 
  #12 = NameAndType        #10:#11        // "<init>":()V                                                      
  #13 = Methodref          #4.#12         // java/lang/Object."<init>":()V                                     
  #14 = Utf8               newThread                                                                           
  #15 = Utf8               (Ljava/lang/Runnable;)Ljava/lang/Thread;                                            
  #16 = Utf8               Ljdk/internal/vm/annotation/Hidden;                                                 
```
What is mysterious though ... I was able to reproduce it only once locally by building the `2.15.1-SNAPSHOT` ... I am still looking why it happened with `2.15.0` but not (anymore) with `2.15.1-SNAPSHOT`. Building out of release tag (`release-2.15.0`) does not generate same bytecode either ....

reta:
> 
> 
> It is still strange, though... isn't it?.. There's no compile dep on okhttp from io.zipkin.brave:brave

Not exactly but in `test` scope only:
```
$ mvn dependency:tree -pl :brave | grep okhttp3
[INFO] +- com.squareup.okhttp3:mockwebserver:jar:4.7.2:test
[INFO] |  \- com.squareup.okhttp3:okhttp:jar:4.7.2:test
[INFO] +- io.zipkin.reporter2:zipkin-sender-okhttp3:jar:2.15.0:test
```

adriancole:
opened https://github.com/luontola/retrolambda/issues/160 actually I found the culprit... we also had one here in BaggagePropagation

adriancole:
@reta I added a commit, that after you build https://github.com/openzipkin/zipkin-reporter-java/pull/179 can verify if all clear?

reta:
awesome, thanks a lot @adriancole , I think we can close this pull request

adriancole:
nah can't close this as there's one here :D

brave/bnd.bnd
```diff
@@ -1,4 +1,5 @@
 Import-Package: \
+  jdk.internal.vm.annotation;resolution:=optional,\
```
adriancole, line 2:
```suggestion
```

reta:
@adriancole the MANIFEST.MF (for `brave` and `zipkin-sender-okhttp3`) are good, CXF build passes against the snapshots with no issues, thanks!

pom.xml
```diff
@@ -129,7 +129,7 @@
     <maven-failsafe-plugin.version>3.0.0-M4</maven-failsafe-plugin.version>
     <maven-invoker-plugin.version>3.2.1</maven-invoker-plugin.version>
     <maven-enforcer-plugin.version>3.0.0-M3</maven-enforcer-plugin.version>
-    <maven-bundle-plugin.version>4.2.1</maven-bundle-plugin.version>
+    <maven-bundle-plugin.version>5.1.1</maven-bundle-plugin.version>
```
reta, line 132: :+1:

adriancole:
thanks.. I'll go ahead and ship this as don't want people pinned.

adriancole:
Mind having a look? https://github.com/openzipkin/brave/pull/1252

PR #1252, opened 2020-09-11
Tries to hone OSGi imports according to use

### LINKED ISSUE #1248: OSGi package import missing in context-slf4j
When deploying `brave-context-slf4j-5.12.5.jar` into an OSGi runtime (in our case Equinox/Liferay), the packages resolve successfully. However, when actually using it (`MDCScopeDecorator`), I get this exception:

`Caused by: java.lang.ClassNotFoundException: brave.internal.CorrelationContext cannot be found by io.zipkin.brave.context-slf4j_5.12.5`

I re-osgified the bundle via bnd-maven-plugin which generated the following `Import-Package` instruction:

`Import-Package: brave;version="[5.12,6)",brave.baggage;version="[5.12,6)",brave.internal;version="[5.12,6)";braveinternal=true,brave.propagation;version="[5.12,6)",org.slf4j`

This fixes the issue.

adriancole:
It is my understanding that unless we shade something, we have to add
internal import for the few utilities we use.

Fixes #1248

### DIFF
--- a/brave-bom/pom.xml
+++ b/brave-bom/pom.xml
@@ -35,7 +35,7 @@
 
     <!-- use the same values in ../pom.xml -->
     <zipkin.version>2.21.7</zipkin.version>
-    <zipkin-reporter.version>2.15.1</zipkin-reporter.version>
+    <zipkin-reporter.version>2.15.2</zipkin-reporter.version>
   </properties>
 
   <organization>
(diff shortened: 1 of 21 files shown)


reta (approved):
LGTM, `brave` exports `brave.internal` with mandatory `braveinternal` attribute, the imports look correct.

reta:
Just checked the recent build, looks perfect

```
Import-Package: brave.baggage;version="[5.12,6)",brave.internal;braveint
 ernal=true;version="[5.12,6)",brave.propagation;version="[5.12,6)",org.
 slf4j;version="[1.7,2)"

```

adriancole:
Thanks also to @helgoboss for identifying and closely looking at this. Appreciate the second, well third eyes @reta!

PR #1259, opened 2020-10-13
Updates versions and notes why JDK 14

adriancole:


pom.xml
```diff
@@ -581,6 +580,8 @@
         <configuration>
           <!-- Try to prevent flakes in CI -->
           <reuseForks>false</reuseForks>
+          <!-- workaround to SUREFIRE-1831 -->
+          <useModulePath>false</useModulePath>
```
adriancole, line 584: cc @jorgheymans

Issue #1258, opened 2020-10-06
RabbitMQ: tracing headers not propagated to reply of @RabbitListener


pzapasni:
**This is a copy of [spring-cloud-sleuth/issues/1752](https://github.com/spring-cloud/spring-cloud-sleuth/issues/1752), since I was asked to report this issue here**

In my app I'm using ``spring-cloud-starter-sleuth:2.2.5`` with ``spring-boot-starter-amqp:2.2.7``. I have a ``@RabbitListener`` annotated method which accepts a message, processes it, and returns a response:

```java
@Service
@RequiredArgsConstructor
@Slf4j
@Transactional
public class MyListener {

    private final MyService service;

    @RabbitListener(queues = "${amqp.queue.name}")
    public MyEvent handleCommand(@Payload @Valid MyCommand command) {
        log.info("Received command {}", command);
        return service.handleCommand(command);
    }
}
```

If the incoming message has the ``b3`` header, the tracing information is present throughout the processing of the message. However, when the response message from that processing is being returned to RabbitMQ, it does not contain the ``b3`` header, which breaks the tracing.

I've investigated this issue a bit, and it seems to me that the only instrumentation for the listener (or rather ``SimpleRabbitListenerContainerFactory``) is being done through the ``SleuthRabbitBeanPostProcessor`` and ``TracingRabbitListenerAdvice``, however I couldn't find any code there that would take care of injecting the header into the response message. Is my assumption correct and this is a missing feature, or did I miss something?

adriancole:
We typically use https://gitter.im/openzipkin/zipkin for starting issues, but this is fine as I think it is likely to result in a code change.

Here's the part about this from [rabbit docs](https://docs.spring.io/spring-amqp/api/org/springframework/amqp/rabbit/annotation/RabbitListener.html)
>  Annotated methods may have a non void return type. When they do, the result of the method invocation is sent as a reply to the queue defined by the ReplyTo header of the incoming message. When this value is not set, a default queue can be provided by adding @SendTo to the method declaration.

As you noticed, right now this reply-to feature isn't implemented in a way that would continue the trace. There's no test for this and no one had asked for it until now. Depending on the code involved, it may be easy or hard to wrap the producer code in spring-rabbit in context of the "listener span" executing this method.

Zipkin is a completely volunteer run project, so if you can help implement this, it would be grand.

[...]

PR #1260, opened 2020-10-13
Add trace headers to the reply from a @RabbitListener

pzapasni:
This is an implementation draft, rather than a complete solution, as I do not know the internal workings of Brave. Any additional insight would be greatly appreciated.

This is a fix for #1258. The aim here is to create an additional tracing span for the producer that sends the reply from ``@RabbitListener``-annotated method back to RabbitMQ. While I'm not sure if that's correct, I created an integration test that assumes that the span hierarchy should look like so:

```
producer span (initial message) -> remote consumer span (listener) -> local consumer span (message processing)
producer span (initial message) -> producer remote span (response message) -> remote consumer span (response listener) -> local consumer span (response message processing)
```

To achieve that, I instrumented the listener container factory using the ``SimpleRabbitListenerContainerFactory#setBeforeSendReplyPostProcessors`` method, which seems to be the correct way of setting up a before-reply hook. However, there are two problems with this approach:

1. The post processor interferes with the ``TracingRabbitListenerAdvice``, because the before-reply hook is ran before the invocation of the advised method completes (after [this](https://github.com/openzipkin/brave/blob/master/instrumentation/spring-rabbit/src/main/java/brave/spring/rabbit/TracingRabbitListenerAdvice.java#L108) method is invoked, but before the ``return`` happens). This causes the response message producer span to be a child of the local consumer span (message processing span), which I believe is incorrect.
2. The usage of ``SimpleRabbitListenerContainerFactory#setBeforeSendReplyPostProcessors`` is completely fine when instrumenting a completely new ``SimpleRabbitListenerContainerFactory`` instance, however it might cause issues when instrumenting an existing one, as it overwrites any previously added post processors. There is no ``getBeforeSendReplyPostProcessors`` or ``addBeforeSendReplyPostProcessors`` method, which makes adding an additional post processor impossible, unless we'd want to resort to some reflection hacks.

As I mentioned before, I'm a novice to Brave, so any help in solving the abovementioned problems is welcome.

### DIFF
--- a/instrumentation/spring-rabbit/src/main/java/brave/spring/rabbit/SpringRabbitTracing.java
+++ b/instrumentation/spring-rabbit/src/main/java/brave/spring/rabbit/SpringRabbitTracing.java
@@ -16,7 +16,6 @@ package brave.spring.rabbit;
 import brave.Span;
 import brave.Tracer;
 import brave.Tracing;
-import brave.baggage.BaggagePropagation;
 import brave.messaging.MessagingRequest;
 import brave.messaging.MessagingTracing;
 import brave.propagation.B3Propagation;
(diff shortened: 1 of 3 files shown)


adriancole:
thanks I'll take a look and add a polishing commit

adriancole:
great start. Thanks! I'll merge your commit in and close in https://github.com/openzipkin/brave/compare/rabbit-response-trace-headers on green

adriancole:
merged manually in latest master. thanks!

PR #1261, opened 2020-10-17
Accept any string as zipkinServiceName parameter.

rgamez:
While `remoteServiceName` accepts any string, the jdbc url parameter which provides a way to override it only accepts characters from the word group `[a-zA-Z0-9_]` this change remove that assumption and capture the complete value of the parameter and it solves a problem that I'm experiencing where in my naming convention (kebab case) is not possible to have the correct value due the presence of dash (`-`), for example `my-database-override` results in `my` as the `remoteServiceName`.

### DIFF
--- a/instrumentation/p6spy/src/main/java/brave/p6spy/TracingJdbcEventListener.java
+++ b/instrumentation/p6spy/src/main/java/brave/p6spy/TracingJdbcEventListener.java
@@ -29,8 +29,9 @@ import static brave.Span.Kind.CLIENT;
 
 final class TracingJdbcEventListener extends SimpleJdbcEventListener {
 
+  // Captures all the characters between = and either the next & or the end of the string.
   private static final Pattern URL_SERVICE_NAME_FINDER =
-    Pattern.compile("zipkinServiceName=(\\w*)");
+    Pattern.compile("zipkinServiceName=(.*?)(?:&|$)");
 
   @Nullable final String remoteServiceName;
   final boolean includeParameterValues;
--- a/instrumentation/p6spy/src/test/java/brave/p6spy/TracingJdbcEventListenerTest.java
+++ b/instrumentation/p6spy/src/test/java/brave/p6spy/TracingJdbcEventListenerTest.java
@@ -25,20 +25,27 @@ import com.p6spy.engine.spy.option.P6OptionsRepository;
 import java.sql.Connection;
 import java.sql.DatabaseMetaData;
 import java.sql.SQLException;
+
+import junitparams.JUnitParamsRunner;
+import junitparams.Parameters;
+import junitparams.naming.TestCaseName;
 import org.junit.After;
 import org.junit.Before;
+import org.junit.Rule;
 import org.junit.Test;
 import org.junit.runner.RunWith;
 import org.mockito.Mock;
-import org.mockito.junit.MockitoJUnitRunner;
+import org.mockito.junit.MockitoJUnit;
+import org.mockito.junit.MockitoRule;
 
 import static org.assertj.core.api.Assertions.assertThat;
 import static org.mockito.Mockito.verify;
 import static org.mockito.Mockito.verifyNoMoreInteractions;
 import static org.mockito.Mockito.when;
 
-@RunWith(MockitoJUnitRunner.class)
+@RunWith(JUnitParamsRunner.class)
 public class TracingJdbcEventListenerTest {
+  @Rule public MockitoRule rule = MockitoJUnit.rule();
   @Mock Connection connection;
   @Mock DatabaseMetaData metaData;
   @Mock StatementInformation statementInformation;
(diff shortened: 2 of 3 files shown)


instrumentation/p6spy/src/main/java/brave/p6spy/TracingJdbcEventListener.java
```diff
@@ -30,7 +30,7 @@
 final class TracingJdbcEventListener extends SimpleJdbcEventListener {
 
   private static final Pattern URL_SERVICE_NAME_FINDER =
-    Pattern.compile("zipkinServiceName=(\\w*)");
+    Pattern.compile("zipkinServiceName=(.*?)(?:&|$)");
```
jorgheymans, line 33: Could you maybe add a small comment what the regex is aiming to match ? Not everyone can work it out from the top of their head.

jorgheymans (changes requested):
Thanks for this ! Could you maybe add a dedicated test with a few boundary testcases for the regex so it becomes clear what will match and what not ?

instrumentation/p6spy/src/main/java/brave/p6spy/TracingJdbcEventListener.java
rgamez: Comment included.

rgamez:
Thanks @jorgheymans, I replaced the tests with some parametrized ones and included a few examples of some likely naming conventions and a few edge cases.

instrumentation/p6spy/src/test/java/brave/p6spy/TracingJdbcEventListenerTest.java
```diff
@@ -121,6 +128,34 @@
     verify(span).remoteIpAndPort("1.2.3.4", 5555);
   }
 
+  @Test
+  @TestCaseName("remoteServiceName should be '{1}'")
+  @Parameters({
```
adriancole, line 133: this dep hasn't been updated since 2017, so might haunt us later. you mind switching to just using a map and calling a method not marked test for each pair here?

instrumentation/p6spy/src/test/java/brave/p6spy/TracingJdbcEventListenerTest.java
```diff
@@ -121,6 +128,34 @@
...
+    "?zipkinServiceName=my_database&foo=bar, my_database",
+    "?zipkinServiceName=my-database&foo=bar, my-database",
+    "?zipkinServiceName=my-database-1&foo=bar, my-database-1",
+    "?zipkinServiceName=my.database&foo=bar, my.database",
+    "?zipkinServiceName=my.database&foo=bar, my.database",
+    "?zipkinServiceName=my-database:5432&foo=bar, my-database:5432",
+    "?zipkinServiceName=my-database@localhost&foo=bar, my-database@localhost",
+    "?zipkinServiceName=my-database, my-database",
+    "?zipkinServiceName=my-database-1&zipkinServiceName=my-database-2, my-database-1",
+    "?zipkinServiceName=, default",
```
adriancole, line 144: empty should coerce to null, not "default" IMHO

instrumentation/p6spy/src/test/java/brave/p6spy/TracingJdbcEventListenerTest.java
```diff
@@ -121,6 +128,34 @@
...
+    "?zipkinServiceName=myDatabase&foo=bar, myDatabase",
+    "?zipkinServiceName=my_database&foo=bar, my_database",
+    "?zipkinServiceName=my-database&foo=bar, my-database",
+    "?zipkinServiceName=my-database-1&foo=bar, my-database-1",
+    "?zipkinServiceName=my.database&foo=bar, my.database",
+    "?zipkinServiceName=my.database&foo=bar, my.database",
+    "?zipkinServiceName=my-database:5432&foo=bar, my-database:5432",
+    "?zipkinServiceName=my-database@localhost&foo=bar, my-database@localhost",
+    "?zipkinServiceName=my-database, my-database",
+    "?zipkinServiceName=my-database-1&zipkinServiceName=my-database-2, my-database-1",
```
adriancole, line 143: am I reading this carefully? the service name would have a comma in it? that seems odd.. I think we should pick last instead.

adriancole (commented):
thanks, the tests are telling!

instrumentation/p6spy/src/test/java/brave/p6spy/TracingJdbcEventListenerTest.java
rgamez: bear in mind that this parameter is an override for `remoteServiceName`, not providing it just results in the default (the database name) used, that's the pre-existing behaviour.

instrumentation/p6spy/src/test/java/brave/p6spy/TracingJdbcEventListenerTest.java
rgamez: that's just the separator between the two parameters queryString and the expected name.

instrumentation/p6spy/src/test/java/brave/p6spy/TracingJdbcEventListenerTest.java
rgamez: unfortunately for junit4 there are no better alternatives, junit5 has support for this style of tests out of the box, let me see what I can do.

adriancole: Arrays.asList in a for loop is better than a dead dependency. can you do add this? I'm worried about best being enemy of good..

adriancole:
ps we ended up forced into jupiter, so after #1267 we will have it if you want to use it here

adriancole:
I'll buff out the test dep after merge. thanks for the help!

rgamez:
Great, came to volunteer for the jupiter port but I see that you already did it, thanks.

adriancole:
incidentally rolled back the blanket jupiter as it would imply brave-tests which is a packaged jar, ending up staining classpaths of modules that import it. However, feel free to add jupiter (using the pinned version) to the "test" scope of brave-test if you like and port away!

what I did actually was use the junit4 experimental extension for the time being.

PR #1262, opened 2020-10-26
Adds RpcTracing.propagation() and fixes grpc-trace-bin propagation

adriancole:
When gRPC dropped census, it broke our integration tests. We removed
them, and in doing so lost coverage of the deprecated `grpc-trace-bin`
format, which drifted and broke.

@dimi-nk's company are still using that format. The cleanest way to
support it is to allow overriding of `RpcTracing.propagation()`. This
allows us the most flexibility including backporting the feature
without breaking API.

While I've done this only for RpcTracing, it might make sense for
HTTP and Messaging, as some library specific formats might exist
elsewhere. If no one asks for that, I'll leave it to just RpcTracing.

cc @anuraaga @jeqo @jorgheymans @jcchavezs @saturnism

### DIFF
--- a/instrumentation/grpc/src/main/java/brave/grpc/GrpcPropagation.java
+++ b/instrumentation/grpc/src/main/java/brave/grpc/GrpcPropagation.java
@@ -100,9 +100,9 @@ final class GrpcPropagation implements Propagation<String> {
     }
 
     @Override public void inject(TraceContext context, R request) {
-      if (request instanceof GrpcClientRequest) {
+      if (request instanceof GrpcRequest) {
         byte[] serialized = TraceContextBinaryFormat.toBytes(context);
-        Metadata metadata = ((GrpcClientRequest) request).headers;
+        Metadata metadata = ((GrpcRequest) request).headers();
         metadata.removeAll(GRPC_TRACE_BIN);
         metadata.put(GRPC_TRACE_BIN, serialized);
         TagsBin tags = context.findExtra(TagsBin.class);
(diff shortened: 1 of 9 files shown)


adriancole:
FWIW [Finagle](https://github.com/twitter/finagle/blob/develop/finagle-thrift/src/main/java/com/twitter/finagle/thrift/thrift/RequestHeader.java) also has a custom format for RPC, and so does [RSocket](https://github.com/rsocket/rsocket/blob/master/Extensions/Tracing-Zipkin.md) (cc @OlegDokuka), so they could use this also. However, we don't have implementation in this repo for those libraries so there's nothing to cite here.

adriancole:
actually I think we'll need this for Messaging also, as I know one reason we never merged spring-messaging was because it was really odd in terms of propagation (some different headers used etc). There's probably a case for Http not thinking of yet, but it would be odd to add this for Rpc and Messaging, but not Http. So, inclined to add for all 3.

anuraaga (approved):

adriancole:
merging so @dimi-nk can more easily test this (via snapshot version). in the next day or two we should decide about also doing this for Http and MessagingTracing. I think so.. then, we can cut a version and fix the broken gRPC trace-bin propagation.

jcchavezs:
LGTM

dimi-nk:
@adriancole do I need to use a specific repository to get the snapshot builds?

adriancole:
you can use https://oss.jfrog.org/artifactory/oss-snapshot-local

dimi-nk:
@adriancole I was able to test this and it now works 👍 

```java
  @Bean
  @Primary
  GrpcTracing getGrpcTracing(RpcTracing rpcTracing) {
    return GrpcTracing.newBuilder(rpcTracing).grpcPropagationFormatEnabled(true).build();
  }
```

adriancole:
thanks for the feedback!

adriancole:
follow-up for messaging https://github.com/openzipkin/brave/pull/1263

adriancole:
https://github.com/openzipkin/brave/pull/1264 for HTTP

PR #1263, opened 2020-10-27
Adds MessagingTracing.propagation()

adriancole:
This allows messaging-specific, and messaging library-specific
propagation, similar to what was done in #1262 for RPC.


adriancole:
cc @WillemJiang @davsclaus as I made some notes about how this could be used for interop in camel

PR #1264, opened 2020-10-27
Adds HttpTracing.propagation()

adriancole:
This allows HTTP-specific, and HTTP library-specific
propagation, similar to what was done in #1262 for RPC.


jcchavezs (approved):
LGTM

PR #1265, opened 2020-10-31
Palindrome

sushree007:
To check if the string is palindrome in nature, i.e. upon reversing the string, we get the same string again

### DIFF
--- a/palindrome
+++ b/palindrome
@@ -1,0 +1,18 @@
+import java.io.*;
+import java.util.*;
+
+public class Solution {
+
+    public static void main(String[] args) {
+        
+        Scanner sc=new Scanner(System.in);
+        String A=sc.next();
+        String a="";
+        for(int i=A.length()-1;i>=0;i--)
+        a+=A.charAt(i);
+        if(A.compareTo(a)==0)
+        System.out.println("Yes");
+        else
+        System.out.println("No");
+    }
+}


palindrome
```diff
@@ -0,0 +1,18 @@
+import java.io.*;
```
anuraaga, line 1: Filename needs to be `Solution.java`

palindrome
```diff
@@ -0,0 +1,18 @@
...
+import java.util.*;
+
+public class Solution {
+
+    public static void main(String[] args) {
+        
+        Scanner sc=new Scanner(System.in);
```
anuraaga, line 8:
```suggestion
        Scanner sc = new Scanner(System.in);
```

palindrome
```diff
@@ -0,0 +1,18 @@
...
+        String A=sc.next();
```
anuraaga, line 9:
```suggestion
        String A = sc.next();
```

palindrome
```diff
@@ -0,0 +1,18 @@
...
+        String a="";
+        for(int i=A.length()-1;i>=0;i--)
+        a+=A.charAt(i);
```
anuraaga, line 12:
```suggestion
        String a = new StringBuilder(A).reverse().toString();
```

Less code and should have better performance. Best would be to iterate from both ends of A instead of allocating any extra strings though.

palindrome
```diff
@@ -0,0 +1,18 @@
...
+        if(A.compareTo(a)==0)
+        System.out.println("Yes");
+        else
+        System.out.println("No");
```
anuraaga, line 16:
```suggestion
        if(A.equals(a)) {
          System.out.println("Yes");
        } else {
          System.out.println("No");
        }
```

anuraaga (commented):
Think we're close!

jorgheymans:
Can you add some tests ?

anuraaga:
Think we had our fun with this so going to go ahead and close

PR #1268, opened 2020-11-02
Add error handling for `IllegalStateException`s from Kafka `Headers`

### LINKED ISSUE #1266: (Kafka) TracingProducer errors without sending record if setting headers fails
## Describe the Bug
TracingProducer tries to inject headers to propagate the current trace context. If an exception occurs during this then that exception is allowed to propagate and the record does not get sent to Kafka.

## Steps to Reproduce
https://github.com/m50d/brave/commit/8e9b256f8f52219b5edd0accb41ca3c1eae15b6a has a test reproduction with the case I encountered: `setReadOnly` had already been called on the relevant record's headers, and so attempting to change them failed.

## Expected Behaviour
I appreciate that this is arguable, but I would expect tracing to swallow and log exceptions as far as possible rather than allowing an error in tracing to interfere with the application's main functionality.

m50d:
Fixes #1266 

Since `Headers` actually defines these specific methods to `throw IllegalStateException` it seemed more appropriate to catch the specific exception rather than a general `Throwable` as suggested in #1266 ? But I'd be happy with either approach.

### DIFF
--- a/instrumentation/kafka-clients/src/main/java/brave/kafka/clients/KafkaHeaders.java
+++ b/instrumentation/kafka-clients/src/main/java/brave/kafka/clients/KafkaHeaders.java
@@ -17,12 +17,17 @@ import brave.internal.Nullable;
 import org.apache.kafka.common.header.Header;
 import org.apache.kafka.common.header.Headers;
 
+import static brave.kafka.clients.KafkaTracing.log;
 import static java.nio.charset.StandardCharsets.UTF_8;
 
 final class KafkaHeaders {
   static void replaceHeader(Headers headers, String key, String value) {
-    headers.remove(key);
-    headers.add(key, value.getBytes(UTF_8));
+    try {
+      headers.remove(key);
+      headers.add(key, value.getBytes(UTF_8));
+    } catch (IllegalStateException e) {
+      log(e, "error setting header {0} in headers {1}", key, headers);
+    }
   }
 
   @Nullable static String lastStringHeader(Headers headers, String key) {
--- a/instrumentation/kafka-clients/src/main/java/brave/kafka/clients/KafkaTracing.java
+++ b/instrumentation/kafka-clients/src/main/java/brave/kafka/clients/KafkaTracing.java
@@ -17,6 +17,7 @@ import brave.Span;
 import brave.SpanCustomizer;
 import brave.Tracer;
 import brave.Tracing;
+import brave.internal.Nullable;
 import brave.messaging.MessagingRequest;
 import brave.messaging.MessagingTracing;
 import brave.propagation.B3Propagation;
@@ -29,6 +30,10 @@ import brave.sampler.SamplerFunction;
 import java.util.Iterator;
 import java.util.LinkedHashSet;
 import java.util.Set;
+import java.util.logging.Level;
+import java.util.logging.LogRecord;
+import java.util.logging.Logger;
+
 import org.apache.kafka.clients.consumer.Consumer;
 import org.apache.kafka.clients.consumer.ConsumerRecord;
 import org.apache.kafka.clients.producer.Producer;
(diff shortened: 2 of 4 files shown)


instrumentation/kafka-clients/src/main/java/brave/kafka/clients/KafkaHeaders.java
```diff
@@ -17,12 +17,17 @@
...
 import static java.nio.charset.StandardCharsets.UTF_8;
 
 final class KafkaHeaders {
   static void replaceHeader(Headers headers, String key, String value) {
-    headers.remove(key);
-    headers.add(key, value.getBytes(UTF_8));
+    try {
+      headers.remove(key);
+      headers.add(key, value.getBytes(UTF_8));
+    } catch (IllegalStateException e) {
```
adriancole, line 28: let's catch throwable vs guess which exceptions might be raised during this? less maintenance that way.

```suggestion
    } catch (Throwable t) {
      propagateIfFatal(t);
```

instrumentation/kafka-clients/src/main/java/brave/kafka/clients/KafkaTracing.java
```diff
@@ -243,4 +252,35 @@ static void addTags(ConsumerRecord<?, ?> record, SpanCustomizer result) {
...
+
+
+  /**
+   * Avoids array allocation when logging a parameterized message when fine level is disabled. The
+   * second parameter is optional.
+   *
+   * <p>Ex.
+   * <pre>{@code
+   * try {
+   *    return message.getStringProperty(name);
```
adriancole, line 264: update this comment as it drifted both in the content here and the api used from logging (JUL doesn't have parameterizations)

instrumentation/kafka-clients/src/main/java/brave/kafka/clients/KafkaHeaders.java
```diff
@@ -17,12 +17,17 @@
 import org.apache.kafka.common.header.Header;
 import org.apache.kafka.common.header.Headers;
 
+import static brave.kafka.clients.KafkaTracing.log;
 import static java.nio.charset.StandardCharsets.UTF_8;
 
 final class KafkaHeaders {
```
adriancole, line 23: add a test for this class that has the read-only nature that triggered all of this?

instrumentation/kafka-clients/src/main/java/brave/kafka/clients/KafkaHeaders.java
m50d: You did see that `headers.remove` and `headers.add` are both declared as `throws IllegalStateException`? Though since that's not a checked exception I can see the argument that it's not actually easy to maintain.

instrumentation/kafka-clients/src/main/java/brave/kafka/clients/KafkaTracing.java
m50d: I don't quite follow? I copied this from `JmsTracing`, it's invoking the same API and I assume the parameterization is working (if not then it must be broken in `JmsTracing` as well).

adriancole: sorry I mean "message.getStringProperty" is not relevant in kafka as the name is not message and there's no method getStringProperty. basically the javadoc needs a fixup.

similarly the catch in the javadoc here has drifted.

make sense?

instrumentation/kafka-clients/src/main/java/brave/kafka/clients/KafkaHeaders.java
m50d: Will do

instrumentation/kafka-clients/src/main/java/brave/kafka/clients/KafkaTracing.java
m50d: I see - I've made the javadoc more generic (I don't think it makes sense to use an extremely specific example here as this method is general-purpose?)

adriancole: I don't even mind no javadoc, just don't want it wrong :P

adriancole (approved):
looks great

Issue #1228, opened 2020-06-18
[feature request] instrument Apache HttpClient 5+


cvmocanu:
First, thanks for the great work you're doing!

Now to the ticket.

Apache HttpClient [was released in February](https://hc.apache.org/news.html). The API of the ``HttpClientBuilder`` changed (in particular, the package name is now ``org.apache.hc.client5``, and the methods ``decorateProtocolExec`` and ``decorateMainExec`` on which Brave relies in ``TracingHttpClientBuilder``, are not there anymore, so the current instrumentation cannot be used for HttpClient 5+.

Please add a module that instruments HttpClient 5+.

PR #1274, opened 2020-11-04
Support Apache HttpClient 5.2+

kojilin:
* Support classic/async/h2async client
* Compare to 4.x, 5.x exposes ExecInterceptor. So we can avoid making a builder class and just add ExecInterceptor/RequestInterceptor/ResponseInterceptor to achieve the tracing.

--
### Why 5.2?

* 5.1.x has issues which makes our IT for async failed by open/close in different thread.(Not investigate yet but 5.0.4 & 5.2 doesn't have this issue)
```
java.io.IOException: java.util.concurrent.ExecutionException: java.lang.IllegalStateException: Thread [Time-limited test] opened scope, but thread [httpclient-dispatch-1] closed it
	at brave.httpclient5.ITTracingH2AsyncClientBuilder.blockOnFuture(ITTracingH2AsyncClientBuilder.java:54)
	at brave.httpclient5.ITTracingH2AsyncClientBuilder.invoke(ITTracingH2AsyncClientBuilder.java:47)
	at brave.httpclient5.ITTracingH2AsyncClientBuilder.get(ITTracingH2AsyncClientBuilder.java:87)
	at brave.httpclient5.ITTracingH2AsyncClientBuilder.get(ITTracingH2AsyncClientBuilder.java:44)
	at brave.test.http.ITHttpClient.httpPathTagExcludesQueryParams(ITHttpClient.java:387)
```

* 5.0.4 has different that ITTracingH2AsyncClientBuilder#failedRequestInterceptorRemovesScope throws RuntimeException directly instead of `java.io.IOException: java.util.concurrent.ExecutionException: java.lang.RuntimeException: Test`. 5.0.3, 5.2 doesn't have this issue.

* 5.0.3 has issues which make H2Async & CachingH2Async not work(test will be failed)
  * https://issues.apache.org/jira/browse/HTTPCLIENT-2123
  * https://issues.apache.org/jira/browse/HTTPCLIENT-2126 

--
Solve https://github.com/openzipkin/brave/issues/1228

### DIFF
--- a/instrumentation/httpclient5/pom.xml
+++ b/instrumentation/httpclient5/pom.xml
@@ -1,0 +1,73 @@
+<?xml version="1.0"?>
+<!--
+
+    Copyright 2013-2020 The OpenZipkin Authors
+
+    Licensed under the Apache License, Version 2.0 (the "License"); you may not use this file except
+    in compliance with the License. You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+    Unless required by applicable law or agreed to in writing, software distributed under the License
+    is distributed on an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express
+    or implied. See the License for the specific language governing permissions and limitations under
+    the License.
+
+-->
+<project xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
+  xmlns="http://maven.apache.org/POM/4.0.0"
+  xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 http://maven.apache.org/xsd/maven-4.0.0.xsd">
+  <parent>
+    <groupId>io.zipkin.brave</groupId>
+    <artifactId>brave-instrumentation-parent</artifactId>
+    <version>5.12.8-SNAPSHOT</version>
+  </parent>
+  <modelVersion>4.0.0</modelVersion>
+
+  <artifactId>brave-instrumentation-httpclient5</artifactId>
+  <name>Brave Instrumentation: Apache HttpClient v5.0+</name>
+
+  <properties>
+    <!-- Matches Export-Package in bnd.bnd -->
+    <module.name>brave.httpclient</module.name>
+
+    <main.basedir>${project.basedir}/../..</main.basedir>
+    <main.java.version>1.6</main.java.version>
+    <main.signature.artifact>java16</main.signature.artifact>
+  </properties>
+
+  <dependencies>
+    <dependency>
+      <groupId>${project.groupId}</groupId>
+      <artifactId>brave-instrumentation-http</artifactId>
+    </dependency>
+    <dependency>
+      <groupId>org.apache.httpcomponents.client5</groupId>
+      <artifactId>httpclient5</artifactId>
+      <scope>provided</scope>
+    </dependency>
+    <dependency>
+      <groupId>org.apache.httpcomponents.client5</groupId>
+      <artifactId>httpclient5-cache</artifactId>
+      <optional>true</optional>
+    </dependency>
+
+    <dependency>
+      <groupId>${project.groupId}</groupId>
+      <artifactId>brave-tests</artifactId>
+      <scope>test</scope>
+    </dependency>
+    <dependency>
+      <groupId>${project.groupId}</groupId>
+      <artifactId>brave-instrumentation-http-tests</artifactId>
+      <scope>test</scope>
+    </dependency>
+  </dependencies>
+  <build>
+    <plugins>
+      <plugin>
+        <artifactId>maven-invoker-plugin</artifactId>
+      </plugin>
+    </plugins>
+  </build>
+</project>
--- a/instrumentation/httpclient5/pom.xml
+++ b/instrumentation/httpclient5/pom.xml
@@ -1,17 +1,19 @@
 <?xml version="1.0"?>
 <!--
-  Copyright 2013-2023 The OpenZipkin Authors
 
-  Licensed under the Apache License, Version 2.0 (the "License"); you may not use this file except
-  in compliance with the License. You may obtain a copy of the License at
+    Copyright 2013-2023 The OpenZipkin Authors
 
-  http://www.apache.org/licenses/LICENSE-2.0
+    Licensed under the Apache License, Version 2.0 (the "License"); you may not use this file except
+    in compliance with the License. You may obtain a copy of the License at
 
-  Unless required by applicable law or agreed to in writing, software distributed under the License
-  is distributed on an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express
-  or implied. See the License for the specific language governing permissions and limitations under
-  the License.
-  -->
+    http://www.apache.org/licenses/LICENSE-2.0
+
+    Unless required by applicable law or agreed to in writing, software distributed under the License
+    is distributed on an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express
+    or implied. See the License for the specific language governing permissions and limitations under
+    the License.
+
+-->
 <project xmlns="http://maven.apache.org/POM/4.0.0" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 http://maven.apache.org/xsd/maven-4.0.0.xsd">
   <parent>
     <groupId>io.zipkin.brave</groupId>
--- a/instrumentation/httpclient5/pom.xml
+++ b/instrumentation/httpclient5/pom.xml
@@ -1,19 +1,17 @@
 <?xml version="1.0"?>
 <!--
+  Copyright 2013-2023 The OpenZipkin Authors
 
-    Copyright 2013-2020 The OpenZipkin Authors
+  Licensed under the Apache License, Version 2.0 (the "License"); you may not use this file except
+  in compliance with the License. You may obtain a copy of the License at
 
-    Licensed under the Apache License, Version 2.0 (the "License"); you may not use this file except
-    in compliance with the License. You may obtain a copy of the License at
+  http://www.apache.org/licenses/LICENSE-2.0
 
-    http://www.apache.org/licenses/LICENSE-2.0
-
-    Unless required by applicable law or agreed to in writing, software distributed under the License
-    is distributed on an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express
-    or implied. See the License for the specific language governing permissions and limitations under
-    the License.
-
--->
+  Unless required by applicable law or agreed to in writing, software distributed under the License
+  is distributed on an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express
+  or implied. See the License for the specific language governing permissions and limitations under
+  the License.
+  -->
 <project xmlns="http://maven.apache.org/POM/4.0.0" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 http://maven.apache.org/xsd/maven-4.0.0.xsd">
   <parent>
     <groupId>io.zipkin.brave</groupId>
@@ -32,7 +30,8 @@
     <main.basedir>${project.basedir}/../..</main.basedir>
     <main.java.version>1.7</main.java.version>
     <main.signature.artifact>java17</main.signature.artifact>
-    <httpclient5.version>5.2.1</httpclient5.version>
+    <!-- To obey the integration tests, we need to use the version >= 5.2 -->
+    <httpclient5.version>5.2</httpclient5.version>
   </properties>
 
   <dependencies>
--- a/instrumentation/httpclient5/pom.xml
+++ b/instrumentation/httpclient5/pom.xml
@@ -29,11 +29,12 @@
 
   <properties>
     <!-- Matches Export-Package in bnd.bnd -->
-    <module.name>brave.httpclient</module.name>
+    <module.name>brave.httpclient5</module.name>
 
     <main.basedir>${project.basedir}/../..</main.basedir>
-    <main.java.version>1.6</main.java.version>
-    <main.signature.artifact>java16</main.signature.artifact>
+    <main.java.version>1.7</main.java.version>
+    <main.signature.artifact>java17</main.signature.artifact>
+    <httpclient5.version>5.0.3</httpclient5.version>
   </properties>
 
   <dependencies>
@@ -44,14 +45,15 @@
     <dependency>
       <groupId>org.apache.httpcomponents.client5</groupId>
       <artifactId>httpclient5</artifactId>
+      <version>${httpclient5.version}</version>
       <scope>provided</scope>
     </dependency>
     <dependency>
       <groupId>org.apache.httpcomponents.client5</groupId>
       <artifactId>httpclient5-cache</artifactId>
+      <version>${httpclient5.version}</version>
       <optional>true</optional>
     </dependency>
-
     <dependency>
       <groupId>${project.groupId}</groupId>
       <artifactId>brave-tests</artifactId>
--- a/instrumentation/httpclient5/pom.xml
+++ b/instrumentation/httpclient5/pom.xml
@@ -18,12 +18,12 @@
   <parent>
     <groupId>io.zipkin.brave</groupId>
     <artifactId>brave-instrumentation-parent</artifactId>
-    <version>5.13.3-SNAPSHOT</version>
+    <version>5.15.1-SNAPSHOT</version>
   </parent>
   <modelVersion>4.0.0</modelVersion>
 
   <artifactId>brave-instrumentation-httpclient5</artifactId>
-  <name>Brave Instrumentation: Apache HttpClient v5.0+</name>
+  <name>Brave Instrumentation: Apache HttpClient v5.2+</name>
 
   <properties>
     <!-- Matches Export-Package in bnd.bnd -->
--- a/instrumentation/httpclient5/pom.xml
+++ b/instrumentation/httpclient5/pom.xml
@@ -32,6 +32,11 @@
     <main.basedir>${project.basedir}/../..</main.basedir>
     <main.java.version>1.7</main.java.version>
     <main.signature.artifact>java17</main.signature.artifact>
+    <!--
+      Need to wait until 5.0.4.
+      https://issues.apache.org/jira/browse/HTTPCLIENT-2123
+      https://issues.apache.org/jira/browse/HTTPCLIENT-2126
+    -->
     <httpclient5.version>5.0.3</httpclient5.version>
   </properties>
 
--- a/instrumentation/httpclient5/pom.xml
+++ b/instrumentation/httpclient5/pom.xml
@@ -14,13 +14,11 @@
     the License.
 
 -->
-<project xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
-  xmlns="http://maven.apache.org/POM/4.0.0"
-  xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 http://maven.apache.org/xsd/maven-4.0.0.xsd">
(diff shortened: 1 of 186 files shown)


instrumentation/httpclient5/src/main/java/brave/httpclient5/AsyncHandleSendHandler.java
```diff
@@ -0,0 +1,60 @@
...
+
+import static brave.httpclient5.HttpClientUtils.parseTargetAddress;
+
+class AsyncHandleSendHandler implements AsyncExecChainHandler {
+  private final HttpClientHandler<HttpClientRequest, HttpClientResponse> handler;
+  @Nullable
+  private final String serverName;
+
+  AsyncHandleSendHandler(HttpTracing httpTracing) {
+    this.serverName = "".equals(httpTracing.serverName()) ? null : httpTracing.serverName();
```
adriancole, line 27: this seems we should fix httpTracing.serverName() to not permit empty string.. mind moving it there? (ex builder coerce empty to null)

instrumentation/httpclient5/pom.xml
```diff
@@ -0,0 +1,73 @@
...
+    <version>5.12.8-SNAPSHOT</version>
+  </parent>
+  <modelVersion>4.0.0</modelVersion>
+
+  <artifactId>brave-instrumentation-httpclient5</artifactId>
+  <name>Brave Instrumentation: Apache HttpClient v5.0+</name>
+
+  <properties>
+    <!-- Matches Export-Package in bnd.bnd -->
+    <module.name>brave.httpclient</module.name>
```
adriancole, line 32:
```suggestion
    <module.name>brave.httpclient5</module.name>
```

instrumentation/httpclient5/pom.xml
```diff
@@ -0,0 +1,73 @@
...
+
+    <main.basedir>${project.basedir}/../..</main.basedir>
+    <main.java.version>1.6</main.java.version>
```
adriancole, line 35: double check if this is the minimum version of httpclient5?

instrumentation/httpclient5/src/main/java/brave/httpclient5/HandleReceiveHandler.java
```diff
@@ -0,0 +1,75 @@
...
+import org.apache.hc.client5.http.classic.ExecChain;
+import org.apache.hc.client5.http.classic.ExecChain.Scope;
+import org.apache.hc.client5.http.classic.ExecChainHandler;
+import org.apache.hc.client5.http.protocol.HttpClientContext;
+import org.apache.hc.core5.http.ClassicHttpRequest;
+import org.apache.hc.core5.http.ClassicHttpResponse;
+import org.apache.hc.core5.http.HttpException;
+import org.apache.hc.core5.http.HttpHost;
+
+public class HandleReceiveHandler implements ExecChainHandler {
```
adriancole, line 36:
```suggestion
class HandleReceiveHandler implements ExecChainHandler {
```

instrumentation/httpclient5/src/main/java/brave/httpclient5/HandleReceiveHandler.java
```diff
@@ -0,0 +1,75 @@
...
+  private final Tracer tracer;
```
adriancole, line 37: make all these package private as bytecode is smaller and it is more legible

instrumentation/httpclient5/src/main/java/brave/httpclient5/HandleSendHandler.java
```diff
@@ -0,0 +1,66 @@
...
+import org.apache.hc.client5.http.classic.ExecChain;
+import org.apache.hc.client5.http.classic.ExecChain.Scope;
+import org.apache.hc.client5.http.classic.ExecChainHandler;
+import org.apache.hc.core5.http.ClassicHttpRequest;
+import org.apache.hc.core5.http.ClassicHttpResponse;
+import org.apache.hc.core5.http.HttpException;
+
+import static brave.httpclient5.HttpClientUtils.parseTargetAddress;
+
+public class HandleSendHandler implements ExecChainHandler {
```
adriancole, line 33:
```suggestion
class HandleSendHandler implements ExecChainHandler {
```

instrumentation/httpclient5/src/main/java/brave/httpclient5/HandleSendHandler.java
```diff
@@ -0,0 +1,66 @@
...
+
+  private final HttpClientHandler<HttpClientRequest, HttpClientResponse> handler;
+  @Nullable
+  private final String serverName;
+
+  public HandleSendHandler(HttpTracing httpTracing) {
```
adriancole, line 39:
```suggestion
  HandleSendHandler(HttpTracing httpTracing) {
```

instrumentation/httpclient5/src/main/java/brave/httpclient5/HttpClientUtils.java
```diff
@@ -0,0 +1,50 @@
...
+
+import brave.Span;
+import brave.propagation.CurrentTraceContext;
+import java.net.InetAddress;
+import org.apache.hc.client5.http.cache.CacheResponseStatus;
+import org.apache.hc.client5.http.cache.HttpCacheContext;
+import org.apache.hc.core5.http.HttpHost;
+import org.apache.hc.core5.http.protocol.HttpContext;
+
+public class HttpClientUtils {
```
adriancole, line 11:
```suggestion
class HttpClientUtils {
```

instrumentation/httpclient5/src/main/java/brave/httpclient5/TracingHttpClients.java
```diff
@@ -0,0 +1,93 @@
...
+import org.apache.hc.client5.http.impl.ChainElement;
+import org.apache.hc.client5.http.impl.async.CloseableHttpAsyncClient;
+import org.apache.hc.client5.http.impl.async.H2AsyncClientBuilder;
+import org.apache.hc.client5.http.impl.async.HttpAsyncClientBuilder;
+import org.apache.hc.client5.http.impl.classic.CloseableHttpClient;
+import org.apache.hc.client5.http.impl.classic.HttpClientBuilder;
+
+import static org.apache.hc.client5.http.impl.ChainElement.PROTOCOL;
+
+public class TracingHttpClients {
```
adriancole, line 30:
```suggestion
public class HttpClient5Tracing {
```

instrumentation/httpclient5/src/main/java/brave/httpclient5/TracingHttpClients.java
```diff
@@ -0,0 +1,93 @@
...
+  public static CloseableHttpClient create(Tracing tracing, HttpClientBuilder httpClientBuilder) {
```
adriancole, line 31: separate duty of creating HttpClient5Tracing component from making a CloseableHttpClient. this allows one-time configuration even if using multiple components.

ex.
```java
HttpClient5Tracing httpClient5Tracing = HttpClient5Tracing.newBuilder(httpTracing);

httpAsyncClient = httpClient5Tracing.build(httpAsyncClientBuilder);
```

instrumentation/httpclient5/src/test/java/brave/httpclient5/ITTracingCachingH2ClientBuilder.java
```diff
@@ -0,0 +1,67 @@
...
+import java.util.Arrays;
+import okhttp3.mockwebserver.MockResponse;
+import org.apache.hc.client5.http.impl.async.CloseableHttpAsyncClient;
+import org.apache.hc.client5.http.impl.cache.CachingH2AsyncClientBuilder;
+import org.junit.Test;
+
+import static brave.Span.Kind.CLIENT;
+import static org.assertj.core.api.Assertions.assertThat;
+
+public class ITTracingCachingH2ClientBuilder extends ITTracingH2AsyncClientBuilder {
```
adriancole, line 30:
```suggestion
class ITTracingCachingH2ClientBuilder extends ITTracingH2AsyncClientBuilder {
```

instrumentation/httpclient5/src/test/java/brave/httpclient5/ITTracingH2AsyncClientBuilder.java
```diff
@@ -0,0 +1,216 @@
...
+public class ITTracingH2AsyncClientBuilder extends ITHttpAsyncClient<CloseableHttpAsyncClient> {
+
+  static void invoke(CloseableHttpAsyncClient client, SimpleHttpRequest req) throws IOException {
+    Future<SimpleHttpResponse> future = client.execute(req, null);
+    blockOnFuture(future);
+  }
+
+  static <V> V blockOnFuture(Future<V> future) throws IOException {
+    try {
+      return future.get(3, TimeUnit.SECONDS);
```
adriancole, line 42: just be lazy and throw new IOException(e) for wide catch Exception as it is too much distraction in the test file and outcome would be similar as long as surefire is set to include stacktrace (which it is)

instrumentation/httpclient5/src/test/java/brave/httpclient5/ITTracingHttpAsyncClientBuilder.java
```diff
@@ -0,0 +1,218 @@
...
+    blockOnFuture(future);
+  }
+
+  /**
+   * Ensures we don't wrap exception messages.
+   */
+  static <V> V blockOnFuture(Future<V> future) throws IOException {
+    try {
+      return future.get(3, TimeUnit.SECONDS);
+    } catch (InterruptedException e) {
```
adriancole, line 53: same about wide catch

pom.xml
```diff
@@ -254,6 +255,16 @@
         <artifactId>httpasyncclient</artifactId>
         <version>4.1.4</version>
       </dependency>
+      <dependency>
```
adriancole, line 258: move these into httpclient5 pom as we ran into trouble sometimes controlling versions in parent. you can leave the property here or move it to httpclient5

adriancole (commented):
Thanks for the big work koji! I made a first pass on some style things some maybe you might remember how we are weird..

instrumentation/httpclient5/src/main/java/brave/httpclient5/AsyncHandleSendHandler.java
kojilin: copy from httpclient(4.x) one, but looks like HttpClientHandler handles this well. Maybe I can add another PR to change the httpclient(4.x) later.

instrumentation/httpclient5/src/main/java/brave/httpclient5/TracingHttpClients.java
kojilin: Curious usually `HttpTracing` has the same server name, so the same instance should be used for the same endpoint? But separate to `newBuilder` & `build` still better than a single call.

instrumentation/httpclient5/src/test/java/brave/httpclient5/ITTracingCachingH2ClientBuilder.java
kojilin: This is a test code, so I should keep it as public?

instrumentation/httpclient5/pom.xml
kojilin: it's 1.7, https://github.com/apache/httpcomponents-client/blob/master/README.txt thanks.

adriancole:
do you mind to rebase this on latest master (note current version changed)

travis hasn't worked for almost 2 weeks. master now uses GitHub actions

kojilin:
Updated, but 5.0.4 is not released yet. So I think we can merge after its release.

instrumentation/httpclient5/pom.xml
```diff
@@ -0,0 +1,63 @@
...
+  <name>Brave Instrumentation: Apache HttpClient v5.0+</name>
+
+  <properties>
+    <!-- Matches Export-Package in bnd.bnd -->
+    <module.name>brave.httpclient5</module.name>
+
+    <main.basedir>${project.basedir}/../..</main.basedir>
+    <main.java.version>1.7</main.java.version>
+    <main.signature.artifact>java17</main.signature.artifact>
+    <httpclient5.version>5.0.3</httpclient5.version>
```
adriancole, line 35: Add XML comment here saying 5.0.4 is needed and ref the two issues. That comment should stay after we update so that people know (in case someone changes version and has a problem and no one remember)

instrumentation/httpclient5/src/main/java/brave/httpclient5/HttpClient5Tracing.java
```diff
@@ -0,0 +1,90 @@
...
+      (httpRequest, entityDetails, httpContext) -> HttpClientUtils.closeScope(httpContext));
+    httpAsyncClientBuilder.addResponseInterceptorFirst(
+      (httpResponse, entityDetails, httpContext) -> HttpClientUtils.openScope(httpContext,
+        currentTraceContext));
+    httpAsyncClientBuilder.addResponseInterceptorLast(
+      (httpResponse, entityDetails, httpContext) -> HttpClientUtils.closeScope(httpContext));
+    return new TracingHttpAsyncClient(httpAsyncClientBuilder.build(), currentTraceContext);
+  }
+
+  public CloseableHttpAsyncClient create(H2AsyncClientBuilder h2AsyncClientBuilder) {
```
adriancole, line 73: I would suggest name build for this method and above as code you are changing would just be calling builder.build() instead. also simplify by naming the param builder, (also in method above this)

instrumentation/httpclient5/src/main/java/brave/httpclient5/HttpClient5Tracing.java
```diff
@@ -0,0 +1,90 @@
...
+    h2AsyncClientBuilder.addRequestInterceptorFirst(
+      (httpRequest, entityDetails, httpContext) -> HttpClientUtils.openScope(httpContext,
+        currentTraceContext));
+    h2AsyncClientBuilder.addRequestInterceptorLast(
+      (httpRequest, entityDetails, httpContext) -> HttpClientUtils.closeScope(httpContext));
+    h2AsyncClientBuilder.addResponseInterceptorFirst(
+      (httpResponse, entityDetails, httpContext) -> HttpClientUtils.openScope(httpContext,
+        currentTraceContext));
+    h2AsyncClientBuilder.addResponseInterceptorLast(
+      (httpResponse, entityDetails, httpContext) -> HttpClientUtils.closeScope(httpContext));
```
adriancole, line 87: instead of this, make HttpClientUtils.closeScope/openScope a singleton. This also allows you to give the class a name which can be helpful in troubleshooting vs a lambda

I don't know the function addResponseInterceptorLast accepts, but say..

```java
// top-level file, but package private
final class TraceContextInterceptor implements RequestInterceptor, ResponseInterceptor {
  final CurrentTraceContext currentTraceContext;
 --snip--
```

Then, when you make new instance of HttpClient5Tracing, you save one of these as a field, and use it everywhere. it makes the instantiation look cleaner also.

instrumentation/httpclient5/src/main/java/brave/httpclient5/TracingHttpAsyncClient.java
```diff
@@ -0,0 +1,94 @@
...
+    HttpHost target, AsyncRequestProducer requestProducer,
+    AsyncResponseConsumer<T> responseConsumer,
+    HandlerFactory<AsyncPushConsumer> pushHandlerFactory, HttpContext context,
+    FutureCallback<T> callback) {
+    TraceContext invocationContext = currentTraceContext.get();
+    if (invocationContext != null) {
+      context.setAttribute(TraceContext.class.getName(), invocationContext);
+    }
+    return delegate.execute(target, requestProducer, responseConsumer, pushHandlerFactory, context,
+      callback != null && invocationContext != null
```
adriancole, line 75: maybe swap this reference like this above call to `delegate.execute` as it is a little complicated. this allows you to ask why.. another option is to do `TraceContextFutureCallback.create` and make that function exit with input callback based on condition. also clean and comment can go there.

instrumentation/httpclient5/src/test/java/brave/httpclient5/ITTracingCachingH2ClientBuilder.java
adriancole: sorry yeah we are still junit 4. my mistake

instrumentation/httpclient5/src/main/java/brave/httpclient5/HttpResponseWrapper.java
```diff
@@ -0,0 +1,57 @@
...
+ * the License.
+ */
+
+package brave.httpclient5;
+
+import brave.http.HttpClientResponse;
+import brave.internal.Nullable;
+import org.apache.hc.core5.http.HttpResponse;
+
+final class HttpResponseWrapper extends HttpClientResponse {
```
adriancole, line 21: you can find some wrapper tests in brave to copy/paste find/replace for these..  in async it is better to have unit tests for conversions as NPEs later in integration tests can be nasty

adriancole (commented):
Looks very good Koji!

adriancole:
fyi it seems both issues are marked fixed in 5.0.4, but I looked and there is no vote thread for that version. Hence, it might be until next year https://mail-archives.apache.org/mod_mbox/hc-dev/202012.mbox/browser

skastenholz:
I am looking for a way to integrate Spring 5 Webclient and think it would be feasible with this "new" instrumentation. Any reason it did not yet get merged? Any alternative to instrument Spring 5 Webclient? Thank you.

ParaSait:
Can this be merged, please? Is there anything still holding this back?

jcchavezs:
Ping @shakuzen

laurensdeb:
Has there been any progress towards merging this PR?

kojilin:
Updated, I think all the tests are passed. Maybe need to check more due to I made it long before.

laurensdeb:
@kojilin Any updates on this? We would love to integrate with HC 5 in one of our projects.

kojilin:
I'm not the maintainer of the project, so need the other people to check and merge.

jeqo:
I can add a review over the week. 

If some lone can earlier, please go ahead.
Sorry this has take this long

instrumentation/httpclient5/pom.xml
jeqo: @kojilin do you think the comments suggested here are worth adding?

instrumentation/httpclient5/src/main/java/brave/httpclient5/HttpClientUtils.java
```diff
@@ -0,0 +1,64 @@
...
+    InetAddress address = target.getAddress();
+    if (address != null) {
+      if (span.remoteIpAndPort(address.getHostAddress(), target.getPort())) {
+        return;
+      }
+    }
+    span.remoteIpAndPort(target.getHostName(), target.getPort());
+  }
+
+  static boolean isLocalCached(HttpContext context, Span span) {
```
jeqo, line 56: Hit the following issue when testing this instrumentation: https://github.com/openzipkin/brave-example/pull/91/files#r1159109251

I guess not all users have cache dependency in the classpath and will fail when this is evaluated on the response path. Could we handle this exception?

kojilin: Thanks, will handle this.

instrumentation/httpclient5/pom.xml
kojilin: I added the comment we need to use >= 5.2

jeqo (approved):
LGTM, thanks @kojilin !

jeqo:
Seems license headers are up again, please run: `./mvnw com.mycila:license-maven-plugin:format` to fix it

jcchavezs:
Any feedback here @shakuzen ?

shakuzen (approved):

jcchavezs (approved):

jeqo:
Beautiful, thanks @kojilin !! 🎉

JochemKuijpers:
Amazing work! 👏
This is the last blocker for an upgrade to Apache HTTP 5.x for me. 

@jeqo I noticed that after the merge, the CI tests failed: https://github.com/openzipkin/brave/actions/runs/4724094424
Seems something related to the license header that's not up to spec, but I'm not familiar with how this is supposed to run.

(how) can I help fix these issues so we can get a new artifact with the added HttpClient 5.2+ support?

PR #1277, opened 2020-11-07
Changes testcontainers to not pull docker.io and switches to Arm64 builds

adriancole:
This follows advice from testcontainers to opt-out of incidental images
used until they are republished on a non rate-limited registry.

https://www.testcontainers.org/supported_docker_environment/image_registry_rate_limiting/

This also moves the build to the much faster arm64 config

Same as https://github.com/openzipkin/zipkin-reporter-java/pull/192

### gRPC 1.2's protoc doesn't support arm64
The version of google's protoc coupled with gRPC doesn't support arm64 until gRPC 1.9. Instead of interfering with our floor version, this checks in generated code. The generator works fine on amd64 (macbooks).

### arm64 runs seems more sensitive to classpath glitches
We have had weird log4j2 -> jul bridge glitches in the past, but in arm64 it is more consistently flakey. We only need the bridge during integration tests, so there's now a classpath exclusion of the jul bridge otherwise.

### mongodb needs an arm64 capable image
I had to build and publish an arm64-capable image of mongodb, using the last version packaged with alpine (so that the image can be smaller)

```Dockerfile
FROM alpine:3.9.6
ARG maintainer="OpenZipkin https://gitter.im/openzipkin/zipkin"
LABEL maintainer=$maintainer
LABEL org.opencontainers.image.authors=$maintainer
LABEL org.opencontainers.image.description="MongoDB on Alpine Linux"
LABEL alpine-version=3.9.6
LABEL mongodb-version=4.0.5

WORKDIR /mongodb
RUN apk add --update --no-cache mongodb=~4.0.5 && chown mongodb .

USER mongodb

EXPOSE 27017
CMD ["mongod", "--quiet", "--bind_ip", "0.0.0.0", "--dbpath", "/mongodb"]
```

```bash
docker buildx build --platform linux/amd64,linux/arm64 -t ghcr.io/openzipkin/mongodb-alpine:4.0.5 . --push
```


adriancole:
gRPC 1.2's version of protoc doesn't support ARM64. Rather than raise the floor, this checks in the generated sources. One advantage of java, afterall, is not needing to know about the native platform.

adriancole:
mongo docker tests are running on arm finally

```
[INFO] Running brave.mongodb.ITMongoDBTracing

[INFO] Tests run: 9, Failures: 0, Errors: 0, Skipped: 0, Time elapsed: 18.918 s - in brave.mongodb.ITMongoDBTracing

[INFO] 
```

PR #1281, opened 2020-11-23
Have StrictScopeDecorator also detect leaked to GC scopes.

anuraaga:
With all the thought I've had to put into context leakage lately, I thought it could be nice to have `StrictScopeDecorator` detect GC leaked scopes too, for enabling on a staging server. What do you think?

### DIFF
--- a/brave/src/main/java/brave/internal/collect/WeakConcurrentMap.java
+++ b/brave/src/main/java/brave/internal/collect/WeakConcurrentMap.java
@@ -43,7 +43,7 @@ import java.util.concurrent.ConcurrentMap;
  * <p>See https://github.com/raphw/weak-lock-free
  */
 public class WeakConcurrentMap<K, V> extends ReferenceQueue<K> implements Iterable<Map.Entry<K, V>> {
-  protected final ConcurrentMap<WeakKey<K>, V> target = new ConcurrentHashMap<>();
+  final ConcurrentMap<WeakKey<K>, V> target = new ConcurrentHashMap<>();
 
   @Nullable public V getIfPresent(K key) {
     if (key == null) throw new NullPointerException("key == null");
--- a/brave/src/main/java/brave/internal/collect/WeakConcurrentMap.java
+++ b/brave/src/main/java/brave/internal/collect/WeakConcurrentMap.java
@@ -17,6 +17,10 @@ import brave.internal.Nullable;
 import java.lang.ref.Reference;
 import java.lang.ref.ReferenceQueue;
 import java.lang.ref.WeakReference;
+import java.util.AbstractMap;
+import java.util.Iterator;
+import java.util.Map;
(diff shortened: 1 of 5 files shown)


brave/src/main/java/brave/propagation/StrictScopeDecorator.java
```diff
@@ -40,13 +40,11 @@
  * }</pre>
  */
 // Closeable so things like Spring will automatically execute it on shutdown and expose leaks!
-public final class StrictScopeDecorator implements ScopeDecorator, Closeable {
+public final class StrictScopeDecorator extends WeakConcurrentMap<Scope, CallerStackTrace> implements ScopeDecorator, Closeable {
```
adriancole, line 43: I might suggest having the WeakConcurrentMap type be an inner class vs leaking it in the signature

If you have more time, bonus points to see if upstream works now (including shading it). I think raphael changed it to allow our features to work.

adriancole: what I meant here was a package-private inner static class like `ScopeToCallerStackTrace extends WeakConcurrentMap`. That would allow you to call methods you wanted exposed. It is another ref, but StrictScopeDecorator is a singleton, so shouldn't matter. The class will be slightly larger, that's it!

anuraaga: Yup good call, changed to an inner class. Looks like the changes for the upstream library haven't been released yet so after that we can try cleaning up.

brave/src/main/java/brave/internal/collect/WeakConcurrentMap.java
```diff
@@ -66,6 +69,12 @@
     return target.remove(key);
   }
 
+  /** Iterates over the entries in this map. */
+  @Override
+  public Iterator<Map.Entry<K, V>> iterator() {
```
anuraaga, line 74: Added an iterator method to match upstream (instead of my previous approach of the `protected` field which stopped being feasible)

brave/src/test/java/brave/TracingTest.java
```diff
@@ -112,7 +112,7 @@
 
   @Test public void spanHandler_loggingByDefault() {
     try (Tracing tracing = Tracing.newBuilder().build()) {
-      assertThat(tracing.tracer().pendingSpans).extracting("spanHandler.delegate")
+      assertThat((Object) tracing.tracer().pendingSpans).extracting("spanHandler.delegate")
```
anuraaga, line 115: Assertion changed from `ObjectAssert` to `ListAssert`. Other ways to fix

- Don't implement `Iterable` in `WeakConcurrentMap`
- Stop extending `WeakConcurrentMap` in `PendingSpans`

Could do now or in the future when integrating upstream's library in.

adriancole (approved):
looks good!

PR #1282, opened 2020-11-24
Moves to GitHub actions build

adriancole:
* Travis is still cutoff despite over a week of email.
 * We have to switch to GitHub actions to avoid paralyzing the project


adriancole:
this first run will be somewhat longer in execution as there's no maven cache in GH actions, yet. brave has a lot of modules pulling various things.

anuraaga (approved):

PR #1284, opened 2020-12-23
Makes entry splitter work with CharSequence

adriancole:
This allows us to split and parse CharSequence instead of String, which
reduces allocations when parsing W3C tracestate.

This adds a utility `CharSequences` to help with tasks.

### DIFF
--- a/brave/src/main/java/brave/internal/codec/CharSequences.java
+++ b/brave/src/main/java/brave/internal/codec/CharSequences.java
@@ -1,0 +1,134 @@
+/*
+ * Copyright 2013-2020 The OpenZipkin Authors
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License"); you may not use this file except
+ * in compliance with the License. You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software distributed under the License
+ * is distributed on an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express
+ * or implied. See the License for the specific language governing permissions and limitations under
+ * the License.
+ */
+package brave.internal.codec;
+
+/**
+ * Most of our parsing tools accept {@link CharSequence} instead of {@link String} to avoid
+ * unnecessary allocation. This contains common functions available on {@link String}, preferring
+ * signatures that match our utilities such as {@link EntrySplitter}.
+ */
+public final class CharSequences {
+
+  /**
+   * Joins two character sequences together. Inputs must be immutable. This is useful when the
+   * result is only scanned. It does not implement {@link #toString()} or {@link #equals(Object)}.
+   */
+  public static CharSequence concat(CharSequence left, CharSequence right) {
+    if (left == null) throw new NullPointerException("left == null");
+    if (right == null) throw new NullPointerException("right == null");
+    if (left.length() == 0) return right; // isEmpty is Java 15+
+    if (right.length() == 0) return left;
+    return new ConcatCharSequence(left, right);
+  }
+
+  static final class ConcatCharSequence implements CharSequence {
+    final CharSequence left, right;
+    final int length;
+
+    ConcatCharSequence(CharSequence left, CharSequence right) {
+      this.left = left;
+      this.right = right;
+      this.length = left.length() + right.length();
+    }
+
+    @Override public int length() {
+      return length;
+    }
+
+    @Override public char charAt(int index) {
+      if (index < 0) throw new IndexOutOfBoundsException("index < 0");
+      if (index >= length) throw new IndexOutOfBoundsException("index >= length");
+      int leftLength = left.length();
+      if (index < leftLength) return left.charAt(index);
+      return right.charAt(index - leftLength);
+    }
+
+    @Override public CharSequence subSequence(int beginIndex, int endIndex) {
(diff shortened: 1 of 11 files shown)


brave/src/main/java/brave/internal/codec/CharSequences.java
```diff
@@ -0,0 +1,134 @@
...
+ * unnecessary allocation. This contains common functions available on {@link String}, preferring
+ * signatures that match our utilities such as {@link EntrySplitter}.
+ */
+public final class CharSequences {
+
+  /**
+   * Joins two character sequences together. Inputs must be immutable. This is useful when the
+   * result is only scanned. It does not implement {@link #toString()} or {@link #equals(Object)}.
+   */
+  public static CharSequence concat(CharSequence left, CharSequence right) {
```
anuraaga, line 27: Don't think I see this used in this PR so just curious what situations this would be used for.

anuraaga (approved):
Worth adding any jmh? `String` is workhorse of Java so it'd be nice to confirm it doesn't have any magic defeating `CharSequence` even at the expense of some alloc for a realish type of usage.

brave/src/main/java/brave/internal/codec/CharSequences.java
adriancole: Good point, I think initially I thought I would use concat for writing a combined header itself (ex "b3=1"+",otherthings"), but noticed StringBuilder does a better job when the result is a string. and string is required for Setter anyway.

The type used inside the method is a part of implementing withoutSubSequence, used to black hole b3's entry in tracestate, but you are right that the static method isn't used. I was also two minds on this, but no problem moving the javadoc down to the inner type and raising it if there were any later uses.

👍

adriancole:
will do on adding JMH

adriancole:
Ok rejigged after noticing String's behaviour on subsequence. Here are the bench results in my laptop for somewhat realistic data. The only place this is of real use is when splitting the middle of a string (ex to own an entry). This can happen at least in w3c tracestate and also secondary sampling.

```bash
$ egrep '(^Bench|p0.999 |gc.alloc.rate.norm)' /tmp/foo
Benchmark                                                                                                          Mode     Cnt     Score     Error   Units
CharSequencesBenchmarks.regionMatches_hit:regionMatches_hit·p0.999                                               sample             0.266             us/op
CharSequencesBenchmarks.regionMatches_hit:·gc.alloc.rate.norm                                                    sample      15     0.005 ±   0.001    B/op
CharSequencesBenchmarks.regionMatches_miss:regionMatches_miss·p0.999                                             sample             0.174             us/op
CharSequencesBenchmarks.regionMatches_miss:·gc.alloc.rate.norm                                                   sample      15     0.001 ±   0.001    B/op
CharSequencesBenchmarks.withoutSubSequence_middle_max:withoutSubSequence_middle_max·p0.999                       sample             9.600             us/op
CharSequencesBenchmarks.withoutSubSequence_middle_max:·gc.alloc.rate.norm                                        sample      15  3488.413 ±   0.022    B/op
CharSequencesBenchmarks.withoutSubSequence_middle_max_string:withoutSubSequence_middle_max_string·p0.999         sample            10.552             us/op
CharSequencesBenchmarks.withoutSubSequence_middle_max_string:·gc.alloc.rate.norm                                 sample      15  6040.687 ±   0.036    B/op
CharSequencesBenchmarks.withoutSubSequence_middle_normal:withoutSubSequence_middle_normal·p0.999                 sample             0.405             us/op
CharSequencesBenchmarks.withoutSubSequence_middle_normal:·gc.alloc.rate.norm                                     sample      15    40.005 ±   0.001    B/op
CharSequencesBenchmarks.withoutSubSequence_middle_normal_string:withoutSubSequence_middle_normal_string·p0.999   sample             1.646             us/op
CharSequencesBenchmarks.withoutSubSequence_middle_normal_string:·gc.alloc.rate.norm                              sample      15   424.042 ±   0.001    B/op

```

adriancole:
@jeqo one thing I mentioned to rag, but not you yet. The trace-context format is a performance bomb for messaging at the point anyone uses tracestate header. One thing this does is insulate us, such that if receiving 1000 messages doesn't mean allocating some factor N strings, some of which may never be used if the messages are filtered out.

adriancole:
went ahead and added bench missing for entry splitter. notably it shows nothing is allocated.

```
Benchmark                                                                                Mode     Cnt     Score     Error   Units
EntrySplitterBenchmarks.parse_middle_max:parse_middle_max·p0.999                       sample            15.904             us/op
EntrySplitterBenchmarks.parse_middle_max:·gc.alloc.rate.norm                           sample      15     0.460 ±   0.063    B/op
EntrySplitterBenchmarks.parse_middle_max_string:parse_middle_max_string·p0.999         sample            13.216             us/op
EntrySplitterBenchmarks.parse_middle_max_string:·gc.alloc.rate.norm                    sample      15  5169.177 ±   0.051    B/op
EntrySplitterBenchmarks.parse_middle_normal:parse_middle_normal·p0.999                 sample             9.216             us/op
EntrySplitterBenchmarks.parse_middle_normal:·gc.alloc.rate.norm                        sample      15     0.029 ±   0.004    B/op
EntrySplitterBenchmarks.parse_middle_normal_string:parse_middle_normal_string·p0.999   sample             9.600             us/op
EntrySplitterBenchmarks.parse_middle_normal_string:·gc.alloc.rate.norm                 sample      15   504.110 ±   0.007    B/op
```

anuraaga (approved):

PR #1291, opened 2021-02-25
Update broken Amazon X-Ray link in README.md

rewolf:
The Amazon X-Ray link was dead, as the project had been updated.  Fixed the link to point to the new location.


jorgheymans:
Thanks !

PR #1293, opened 2021-02-28
Fix Javadoc in SpanHandler

rewolf:
A link was missing the `@` marker and thus would not render correctly. Additionally, 'abrupt' was unconventionally used as a verb ; replaced it with 'disrupt'.


rewolf:
Please suggest a more appropriate word than `disrupt` if it doesn't fit the intended meaning. Maybe `truncate`, `interrupt`?

codefromthecrypt (approved):
language change works for me. Thanks!

PR #1306, opened 2021-09-07
Fixes race condition with toSpan and flush

### LINKED ISSUE #1295: Parent can only be null in a local root
**Describe the bug**

related issue: https://github.com/spring-cloud/spring-cloud-sleuth/issues/1814

I am using Spring Cloud Sleuth 3.0.1 and Brave 5.3.12
The bug is really sporadic but does cause the worfkflow engine (flowable, which I have instrumented manually) to not call the next tasks in the BPMN workflow.

The error is this:
```
Exception in thread "orders-flowable-async-job-executor-thread-4" java.lang.AssertionError: Bug (or unexpected call to internal code): parent can only be null in a local root!
	at brave.internal.recorder.PendingSpans.getOrCreate(PendingSpans.java:89)
	at brave.Tracer._toSpan(Tracer.java:410)
	at brave.Tracer.toSpan(Tracer.java:382)
	at brave.Tracer.toSpan(Tracer.java:376)
	at brave.LazySpan.span(LazySpan.java:141)
	at brave.LazySpan.context(LazySpan.java:40)
	at org.springframework.cloud.sleuth.brave.bridge.BraveSpan.context(BraveSpan.java:48)
	at org.springframework.cloud.sleuth.brave.bridge.BraveTracer.nextSpan(BraveTracer.java:52)
	at org.springframework.cloud.sleuth.instrument.async.TraceRunnable.run(TraceRunnable.java:61)
	at java.util.concurrent.ThreadPoolExecutor.runWorker(ThreadPoolExecutor.java:1149)
	at java.util.concurrent.ThreadPoolExecutor$Worker.run(ThreadPoolExecutor.java:624)
	at java.lang.Thread.run(Thread.java:748)
```

I think it is on this line in TraceRunable:
image

I am of the opinion that the TraceRunanble should not prevent the delegate Runnable in any way. We might benefit from a check before calling the `nextSpan` method or catching this exception somehow.

**Sample**

I can't reliably reproduce it, it is sporadic. We have on the order of 100 tests using flowable but I get one error every now and then. 

Flowable can be set up with a custom implementation of a JobExectuor. I have it set up like so:
```java
 private AsyncExecutor initAsyncExecutor(String tennantId, final int maxConcurrent) {
        final DefaultAsyncJobExecutor asyncExecutor = new DefaultAsyncJobExecutor();
        asyncExecutor.setAutoActivate(false);

        // other setup, irrelevant to the case

        // customized default async job executor initialization with tracing info
        // org.flowable.job.service.impl.asyncexecutor.DefaultAsyncJobExecutor.initAsyncJobExecutionThreadPool
        asyncExecutor.setThreadPoolQueue(new ArrayBlockingQueue<>(queueSize));
        asyncExecutor.setExecutorService(
            new TraceableExecutorService(
                this.applicationContext,
                new ThreadPoolExecutor(
                    asyncExecutor.getCorePoolSize(),
                    asyncExecutor.getMaxPoolSize(),
                    asyncExecutor.getKeepAliveTime(), TimeUnit.MILLISECONDS,
                    asyncExecutor.getThreadPoolQueue(),
                    new BasicThreadFactory.Builder()
                        .namingPattern(tennantId + "-flowable-async-job-executor-thread-%d")
                        .build()
                )
            )
        );

        return asyncExecutor;
    }
```

andylintner:
Fixes #1295
Previous to this fix, a call to Tracer::toSpan concurrent with a call
to flush the span from pendingSpans could result in an assertion error.

We now only fetch the pendingSpan once for a single call to toSpan.

### DIFF
--- a/brave/src/main/java/brave/Tracer.java
+++ b/brave/src/main/java/brave/Tracer.java
@@ -1,5 +1,5 @@
 /*
- * Copyright 2013-2020 The OpenZipkin Authors
+ * Copyright 2013-2021 The OpenZipkin Authors
  *
  * Licensed under the Apache License, Version 2.0 (the "License"); you may not use this file except
  * in compliance with the License. You may obtain a copy of the License at
@@ -193,9 +193,8 @@ public class Tracer {
   }
 
   /** Returns an equivalent context if exists in the pending map */
-  TraceContext swapForPendingContext(TraceContext context) {
-    PendingSpan pendingSpan = pendingSpans.get(context);
-    return pendingSpan != null ? pendingSpan.context() : null;
+  PendingSpan getPendingSpan(TraceContext context) {
+    return pendingSpans.get(context);
   }
 
   /**
@@ -377,9 +376,12 @@ public class Tracer {
   }
 
   Span toSpan(@Nullable TraceContext parent, TraceContext context) {
-    // Re-use a pending context if present: This ensures reference consistency on Span.context()
-    TraceContext pendingContext = swapForPendingContext(context);
-    if (pendingContext != null) return _toSpan(parent, pendingContext);
+    // Re-use a pending span if present: This ensures reference consistency on Span.context()
+    PendingSpan pendingSpan = getPendingSpan(context);
+    if (pendingSpan != null) {
+      if (isNoop(context)) return new NoopSpan(context);
+      return _toSpan(context, pendingSpan);
+    }
 
     // There are a few known scenarios for the context to be absent from the pending map:
     // * Created by a separate tracer (localRootId set)
@@ -408,6 +410,10 @@ public class Tracer {
 
     // allocate a mutable span in case multiple threads call this method.. they'll use the same data
     PendingSpan pendingSpan = pendingSpans.getOrCreate(parent, context, false);
+    return _toSpan(context, pendingSpan);
+  }
+
+  Span _toSpan(TraceContext context, PendingSpan pendingSpan) {
     TraceContext pendingContext = pendingSpan.context();
     // A lost race of Tracer.toSpan(context) is the only known situation where "context" won't be
     // the same as pendingSpan.context()
--- a/brave/src/main/java/brave/Tracer.java
+++ b/brave/src/main/java/brave/Tracer.java
@@ -192,11 +192,6 @@ public class Tracer {
     }
   }
 
-  /** Returns an equivalent context if exists in the pending map */
-  PendingSpan getPendingSpan(TraceContext context) {
-    return pendingSpans.get(context);
-  }
-
   /**
    * Explicitly creates a child within an existing trace. The result will be have its parent ID set
    * to the input's span ID. If a sampling decision has not yet been made, one will happen here.
@@ -377,7 +372,7 @@ public class Tracer {
 
   Span toSpan(@Nullable TraceContext parent, TraceContext context) {
     // Re-use a pending span if present: This ensures reference consistency on Span.context()
-    PendingSpan pendingSpan = getPendingSpan(context);
+    PendingSpan pendingSpan = pendingSpans.get(context);
     if (pendingSpan != null) {
       if (isNoop(context)) return new NoopSpan(context);
       return _toSpan(context, pendingSpan);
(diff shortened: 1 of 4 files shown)


jcchavezs:
Ping @anuraaga

andylintner:
@anuraaga - I believe the test failure from the automation is just an inconsistent test. The same passes locally. Let me know if you think otherwise and I can dig further.

jcchavezs:
@jeqo @jorgheymans any of you can give a check on this?

jcchavezs:
@andylintner test is green.

brave/src/main/java/brave/Tracer.java
```diff
@@ -377,9 +376,12 @@ public Span toSpan(TraceContext context) {
   }
 
   Span toSpan(@Nullable TraceContext parent, TraceContext context) {
-    // Re-use a pending context if present: This ensures reference consistency on Span.context()
-    TraceContext pendingContext = swapForPendingContext(context);
-    if (pendingContext != null) return _toSpan(parent, pendingContext);
+    // Re-use a pending span if present: This ensures reference consistency on Span.context()
+    PendingSpan pendingSpan = getPendingSpan(context);
```
jeqo, line 380:
```suggestion
    PendingSpan pendingSpan = pendingSpans.get(context);
```

brave/src/main/java/brave/Tracer.java
```diff
@@ -377,9 +376,12 @@ public Span toSpan(TraceContext context) {
...
+    if (pendingSpan != null) {
+      if (isNoop(context)) return new NoopSpan(context);
```
jeqo, line 382: Should this be moved into the `_toSpan` method?

jeqo (commented):
Overall seems like a sensible change, avoiding to get parent span twice to avoid race condition.

If we can make sure this is the only execution path that is affected by this issue, it looks good to me.

I wonder if instead of having an overloaded version of `_toSpan`, we could _switch_ into the new interface `TraceContext context, PendingSpan` and move `PendingSpan pendingSpan = pendingSpans.getOrCreate(parent, context, false);` into the other usages of `_toSpan`. wdyt?

nkonev:
Hi @jeqo 
I ran into this issue.
Can this PR be merged ?

jcchavezs:
Ping @llinder

søn. 29. mai 2022, 16:39 skrev Nikita Konev ***@***.***>:

nkonev:
Any updates ?

brave/src/main/java/brave/Tracer.java
```diff
@@ -192,12 +192,6 @@ public final Span joinSpan(TraceContext context) {
     }
   }
 
-  /** Returns an equivalent context if exists in the pending map */
-  TraceContext swapForPendingContext(TraceContext context) {
```
jcchavezs, line 196: I wonder why this was removed.

brave/src/test/java/brave/TracerTest.java
```diff
@@ -1,5 +1,5 @@
 /*
- * Copyright 2013-2020 The OpenZipkin Authors
+ * Copyright 2013-2021 The OpenZipkin Authors
```
jcchavezs, line 2: 2022

brave/src/main/java/brave/Tracer.java
```diff
@@ -1,5 +1,5 @@
 /*
- * Copyright 2013-2020 The OpenZipkin Authors
+ * Copyright 2013-2021 The OpenZipkin Authors
```
jcchavezs, line 2: 2022

jcchavezs:
Running this PR to see how the test fails without the fix https://github.com/openzipkin/brave/pull/1336

jcchavezs:
Hi @andylintner unfortunately the unit test does not fail when the fix is not in place, could you please rework it so we show the failure? see https://github.com/openzipkin/brave/pull/1336

jcchavezs:
We need a test showing the issue otherwise we can merge the fix.

codefromthecrypt (approved):
I'm ok with these changes. Thanks for the help.

codefromthecrypt:
according to #1336, this test passes before or after this change. I'll merge the latter, as the test needs to be improved in order to show what's wrong. When ready please raise a new PR!

PR #1318, opened 2021-12-17
Increase Sonatype OSS staging timeout.

llinder:
Staging servers have had an increased latency for days which
makes it impossible to do releases. This increases the timeout
above the average latency.


llinder:
@jcchavezs staging latencies are still really high which is preventing releases from happening. Hopefully this fixes it and once Sonatype reenables staging servers I can try to cut another release. As it is I have retried the current release like 15 times and always ends in timeouts.

jcchavezs:
+1

jcchavezs:
Do you mind recutting the release?

Issue #1316, opened 2021-12-16
Kafka Streams instrumentation not compatible with kafka-streams 3.0.0


codependent:
## Describe the Bug
A detailed description of this bug can be found on Stackoverflow: https://stackoverflow.com/questions/70382128/tracingkafkaclientsupplier-error-implementations-of-kafkaclientsupplier-should

## Steps to Reproduce
Create a Spring Boot 2.6.1 application with Spring Cloud 2021.0.0, adding spring-cloud-sleuth to instrument it.

## Expected Behaviour
The application should be instrumented correctly, not crashing at startup.

`TracingKafkaClientSupplier` should override this method from `KafkaClientSupplier`:

```
default Admin getAdmin(final Map<String, Object> config) {
        throw new UnsupportedOperationException("Implementations of KafkaClientSupplier should implement the getAdmin() method.");
}
```

However the current implementation returns `AdminClient` instead of `Admin`, thus, throwing the default method exception:

```
@Override public AdminClient getAdminClient(Map<String, Object> config) {
        return AdminClient.create(config);
}
```

jcchavezs:
@codependent is this solved by https://github.com/openzipkin/brave/pull/1312?

codependent:
@jcchavezs I don't think that'll solve it for kafka-streams 3.0.0. The commit leaves `TracingKafkaClientSupplier` with:

```
  @Deprecated public AdminClient getAdminClient(Map<String, Object> config) {
    return getAdmin(config);
  }

  @Override public AdminClient getAdmin(Map<String, Object> config) {
    return AdminClient.create(config);
  }
```

However, In 3.0.0 the signature that should be overriden is different from the last method (notice it should return `Admin`, not `AdminClient`:

```
public interface KafkaClientSupplier {
    default Admin getAdmin(Map<String, Object> config) {
        throw new UnsupportedOperationException("Implementations of KafkaClientSupplier should implement the getAdmin() method.");
    }
```

jcchavezs:
Are you able to work out a PR?

garyrussell:
`AdminClient` is a subclass of `Admin` so it's ok for an implementation to narrow the type.

garyrussell:
`AdminClient` is a subclass of `Admin` so it's ok for an implementation to narrow the type returned, but the return type of the method should be changed to `Admin` to match the interface - just remove the `@Override` to satisfy the compiler.

garyrussell:
[...]

PR #1320, opened 2021-12-22
Fixed TracingConsumer adding currentLag implementation

codependent:
This PR fixes the problem described in https://github.com/openzipkin/brave/issues/1316

In order to keep it backwards compatible I had to use reflection to call the delegate's `currentLag` method.

### DIFF
--- a/instrumentation/kafka-clients/src/main/java/brave/kafka/clients/TracingConsumer.java
+++ b/instrumentation/kafka-clients/src/main/java/brave/kafka/clients/TracingConsumer.java
@@ -23,7 +23,11 @@ import brave.sampler.SamplerFunction;
 
 import java.lang.reflect.Method;
 import java.time.Duration;
-import java.util.*;
+import java.util.List;
+import java.util.Map;
+import java.util.Set;
+import java.util.Collection;
+import java.util.LinkedHashMap;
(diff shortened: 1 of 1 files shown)


instrumentation/kafka-clients/src/main/java/brave/kafka/clients/TracingConsumer.java
```diff
@@ -20,12 +20,10 @@
...
 import brave.sampler.SamplerFunction;
+
+import java.lang.reflect.Method;
 import java.time.Duration;
-import java.util.Collection;
-import java.util.LinkedHashMap;
-import java.util.List;
-import java.util.Map;
-import java.util.Set;
+import java.util.*;
```
jcchavezs, line 26: Could we go back to explicit imports please? This is the pattern we follow in this repo.

jcchavezs:
Shall we cover the change with a unit test?

instrumentation/kafka-clients/src/main/java/brave/kafka/clients/TracingConsumer.java
codependent: Done

jcchavezs:
Ping @jeqo

jeqo:
Already fixed by #1322

PR #1323, opened 2022-01-12
Compatible with versions after dubbo-2.7.14

wuwen5:
Since version dubbo-2.7.14, the ```RpcContext.getContext().getAttachments()```  a copy map is returned,
so the put data will be lost.

https://github.com/apache/dubbo/blob/8240ea26cb00a52dd000d99e3997fe554134b9c3/dubbo-rpc/dubbo-rpc-api/src/main/java/org/apache/dubbo/rpc/RpcContext.java#L557

https://github.com/apache/dubbo/blob/8240ea26cb00a52dd000d99e3997fe554134b9c3/dubbo-common/src/main/java/org/apache/dubbo/common/utils/MapUtils.java#L35

### DIFF
--- a/instrumentation/dubbo/src/test/java/brave/dubbo/ITTracingFilter_Consumer.java
+++ b/instrumentation/dubbo/src/test/java/brave/dubbo/ITTracingFilter_Consumer.java
@@ -1,5 +1,5 @@
 /*
- * Copyright 2013-2020 The OpenZipkin Authors
+ * Copyright 2013-2022 The OpenZipkin Authors
  *
  * Licensed under the Apache License, Version 2.0 (the "License"); you may not use this file except
  * in compliance with the License. You may obtain a copy of the License at
@@ -48,7 +48,7 @@ public class ITTracingFilter_Consumer extends ITTracingFilter {
   @Before public void setup() {
     server.start();
 
-    String url = "dubbo://" + server.ip() + ":" + server.port() + "?scope=remote&generic=bean";
+    String url = "dubbo://" + server.ip() + ":" + server.port() + "?scope=remote";
     client = new ReferenceConfig<>();
     client.setGeneric("true");
     client.setFilter("tracing");
--- a/instrumentation/dubbo/pom.xml
+++ b/instrumentation/dubbo/pom.xml
@@ -1,6 +1,6 @@
 <!--
 
-    Copyright 2013-2021 The OpenZipkin Authors
+    Copyright 2013-2022 The OpenZipkin Authors
 
     Licensed under the Apache License, Version 2.0 (the "License"); you may not use this file except
     in compliance with the License. You may obtain a copy of the License at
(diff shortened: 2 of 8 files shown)


instrumentation/dubbo/src/test/java/brave/dubbo/ITTracingFilter_Consumer.java
```diff
@@ -48,7 +48,7 @@
   @Before public void setup() {
     server.start();
 
-    String url = "dubbo://" + server.ip() + ":" + server.port() + "?scope=remote&generic=bean";
```
jcchavezs, line 51: I wonder the reason for removing this query string.

jcchavezs:
Quick question: will this continue working fine with `dubbo 2.7.8`?

wuwen5:
> Quick question: will this continue working fine with `dubbo 2.7.8`?
Yes, it's fine

instrumentation/dubbo/src/test/java/brave/dubbo/ITTracingFilter_Consumer.java
wuwen5: It was used incorrectly, but did not take effect before 2.7.11，if  used ```generic=bean``` then the parameter type of ```GenericService#$invoke``` args[0] is ```JavaBeanDescriptor```

```java
public class ITTracingFilter_Provider extends ITTracingFilter {
  @Before public void setup() {
    server.service.setFilter("tracing");
    server.service.setGeneric("true");
    server.service.setInterface(GreeterService.class);
    server.service.setRef((method, parameterTypes, args) -> {
      //args[0] instanceof JavaBeanDescriptor
      String arg = (String) args[0];   // test fail
      if (arg.equals("bad")) throw new IllegalArgumentException("bad");
      return currentTraceContext.get() != null
          ? currentTraceContext.get().traceIdString()
          : "";
    });
```

see 
```GenericImplFilter#invoke```

**Review by shakuzen — APPROVED** [review]
Looks good to me. It looks like the `getObjectAttachments` method has been there for a while, so this should still work with older versions. The JavaDoc for `getAttachments` suggests using `getObjectAttachments` instead.

**Comment by wuwen5** [comment]
```java
@Experimental("Experiment api for supporting Object transmission")
public Map<String, Object> getObjectAttachments() {
    return attachments;
}
```

It looks like the ```getObjectAttachments``` method is unstable API, may get removed or changed in the next release, i think should be use ```RpcContext.getContext().setAttachment(keyName, value)```

```java
@Override protected void propagationField(String keyName, String value) {
    RpcContext.getContext().setAttachment(keyName, value);
}
```

shakuzen:
Good catch @wuwen5. `setAttachment` does sound better, then.

jcchavezs:
Are you OK with merging this @shakuzen ?

codefromthecrypt:
fixed by #1371 or open a new issue. thanks!

Issue #1297, opened 2021-04-07
Tracing.currentTracer().currentSpan().finish() does not report duration


ffahri:
## Describe the Bug
Tracing.currentTracer().currentSpan().finish() returns LazySpan and it does not contain duration, however it's available in the context inside LazySpan. When try to finish LazySpan it does not report duration. (duration=null in ZipkinV2 span json) 
For this reason we have to pass RealSpan object to our spanEnd methods

## Steps to Reproduce
Create a span using Tracing.currentTracer.next() and finish using Tracing.currentTracer().currentSpan().finish()

## Expected Behaviour

Expected the LazySpan finish reports the duration

khatkarrohit:
I am looking for something similar. I use Spring and using Brave OpenZipkin just to log trace ids. I wanted to log Duration or time taken by a span. Problem is that `Span` interface implementations like `RealSpan, NoopSpan & RealSpan` are package private and not straight forward to cast. 

It would be beneficial if `Span` interface itself has a methods that returns Duration or any other time unit or simply a long that tells whats the duration of `Span`.

jcchavezs:
Spans aren't meant to have readback functionality (if you see all methods, they all return void). This is very sensitive as we decouple the data from the model and you can turn easily a context into span without having any information about the data related to that span (see https://github.com/openzipkin/brave/blob/5be287b91c3d18da9fc7a597d71b12b27be6043c/brave/src/main/java/brave/Tracer.java#L375). If you truly need to read properties from span you can use finished span handler (which access to mutable span, see https://github.com/openzipkin/brave/blob/master/brave/src/main/java/brave/handler/MutableSpan.java)

khatkarrohit:
@jcchavezs I don't really need access to mutable span because I don't need to mutate any property. I just need read only info from Span for logging and other type of calculations. 

At present I have to create a same name `brave` package in my source code in order to avoid package private restriction from Java and cast `Span` to `RealSpan` and call that method when `Span.finish()` in order to calculate duration. Its a two step process now with type checking. 

This all can be easily done if Span allows to peek in its properties in read only way or immutable way. That would be sufficient.

jcchavezs:
Relevant rationale: https://github.com/openzipkin/brave/pull/588

José Carlos Chávez


tor. 3. feb. 2022 kl. 17:09 skrev khatkarrohit ***@***.***>:

jcchavezs:
Also relevant
[...]

PR #1324, opened 2022-02-01
#1297 Add finish with duration method support for Span.

khatkarrohit:
For issue https://github.com/openzipkin/brave/issues/1297


jcchavezs (changes requested):
Left some comments in the original issue.

codefromthecrypt:
sorry, this is against the design which doesn't read-back fields. you can use a `SpanHandler` to read-back the actual values via `MutableSpan` which was created for this purpose.

PR #1326, opened 2022-02-28
Feature/sql rows written

reftel:
Adds a tag specifying the number of rows written in an update, for correlating time spent versus rows written on SQLs and finding traces that performed large writes.

Having the corresponding tag for rows read would be very useful, but seems to be a lot more difficult, since you´d then have to count the number of .next calls on the result set - but P6Spy does not pass the ResultSetInformation to `onAfterExecuteQuery`.

### DIFF
--- a/instrumentation/p6spy/src/main/java/brave/p6spy/TracingJdbcEventListener.java
+++ b/instrumentation/p6spy/src/main/java/brave/p6spy/TracingJdbcEventListener.java
@@ -1,5 +1,5 @@
 /*
- * Copyright 2013-2020 The OpenZipkin Authors
+ * Copyright 2013-2022 The OpenZipkin Authors
  *
  * Licensed under the Apache License, Version 2.0 (the "License"); you may not use this file except
  * in compliance with the License. You may obtain a copy of the License at
@@ -16,6 +16,9 @@ package brave.p6spy;
 import brave.Span;
 import brave.internal.Nullable;
 import brave.propagation.ThreadLocalSpan;
+
+import com.p6spy.engine.common.PreparedStatementInformation;
+import com.p6spy.engine.common.ResultSetInformation;
 import com.p6spy.engine.common.StatementInformation;
 import com.p6spy.engine.event.SimpleJdbcEventListener;
 import com.p6spy.engine.logging.P6LogLoadableOptions;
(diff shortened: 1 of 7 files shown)


reftel:
Thank yo for your comments! I´ve added an `includeAffectedRowsCount` config option, renamed the tag to `sql.affected_rows`, and added documentation in the README.md. I also cleaned up the existing README a little as a separate commit.

jcchavezs:
Thanks a lot! Any input from others like @shakuzemn or @jeqo?

jeqo (approved):
LGTM

reftel:
Tests are failing due to 

`Error:  Failed to execute goal com.mycila:license-maven-plugin:4.0.rc2:check (default) on project brave-instrumentation-p6spy: Some files do not have the expected license header`

This is for files that are already there. Does the build really pass on master? Should I attempt to fix the files on this branch, or should I rebase on top of a correction somewhere?

reftel:
Updated copyright year ranges to make the test pass. I _believe_ this is actually incorrect, since it claims that changes were made to all the affected files in 2021, but specifying 2013-2020,2022 did not make the test pass.

instrumentation/p6spy/src/main/java/brave/p6spy/TracingJdbcEventListener.java
```diff
@@ -35,12 +38,14 @@
 
   @Nullable final String remoteServiceName;
   final boolean includeParameterValues;
+  final boolean includeAffectedRowsCount;
   final P6LogLoadableOptions logOptions;
 
   TracingJdbcEventListener(@Nullable String remoteServiceName, boolean includeParameterValues,
-    P6LogLoadableOptions logOptions) {
+    boolean includeAffectedRowsCount, P6LogLoadableOptions logOptions) {
```
jcchavezs, line 45: Is this a breaking change?

instrumentation/p6spy/src/main/java/brave/p6spy/TracingJdbcEventListener.java
```diff
@@ -66,10 +71,24 @@
...
 
+  @Override public void onAfterExecuteUpdate(
+    PreparedStatementInformation statementInformation, long timeElapsedNanos, int rowCount, SQLException e
+  ) {
+    Span span = ThreadLocalSpan.CURRENT_TRACER.remove();
+    if (span == null || span.isNoop()) return;
+    if (includeAffectedRowsCount) {
+      span.tag("sql.affected_rows", String.valueOf(rowCount));
+    }
+    finishSpan(span, e);
```
jcchavezs, line 82: Is this right? Do we need to finish Span given that this method did not exist before?

reftel: The span is started in `onBeforeAnyExecute` for both updates and other SQLs, but then we override `onAfterExecuteUpdate` for updates (instead of them being delegated to `onAfterAnyExecute` from `SimpleJdbcEventListener`). Therefore, we need to finish the span both in `onAfterExecuteUpdate` and `onAfterAnyExecute`.

instrumentation/p6spy/src/main/java/brave/p6spy/TracingJdbcEventListener.java
reftel: Thanks for pointing that out. I´ve added a version of the constructor that has the old signature, which delegates to the new one.

reftel:
Ping @jcchavezs

jcchavezs:
I will merge it today.

tir. 15. mar. 2022, 14:41 skrev Magnus Reftel ***@***.***>:

jcchavezs (approved):

PR #1333, opened 2022-06-18
Capture row counts for plain statements and batch

reftel:
Previously, only row counts for executeUpdate on prepared statements was
included as sql.affected_rows. With this change, row counts are also
included for executeUpdate on plain statements, and for executeBatch on
prepared and plain statements.

Capturing row counts on plain execute is a lot trickier, since it´s not
safe to call getUpdateCount behind the application´s back, as it is not
idempotent - we´d unexpectedly change state.

### DIFF
--- a/instrumentation/p6spy/src/test/java/brave/p6spy/ITTracingP6Factory.java
+++ b/instrumentation/p6spy/src/test/java/brave/p6spy/ITTracingP6Factory.java
@@ -113,12 +113,98 @@ public class ITTracingP6Factory {
   }
 
   @Test
-  public void addsRowsWrittenTag() throws Exception {
+  public void addsAffectedRowsTagToPreparedUpdateStatements() throws Exception {
     prepareExecuteUpdate("update t set c='x' where i=2");
 
     assertThat(spans)
       .flatExtracting(s -> s.tags().entrySet())
       .contains(entry("sql.affected_rows", "2"));
+  }
+
+  @Test
+  public void addsAffectedRowsTagToPlainUpdateStatements() throws Exception {
+    executeUpdate("update t set c='x' where i=2");
+
+    assertThat(spans)
+      .flatExtracting(s -> s.tags().entrySet())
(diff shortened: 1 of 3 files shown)


instrumentation/p6spy/src/test/java/brave/p6spy/ITTracingP6Factory.java
```diff
@@ -113,14 +113,63 @@ public void addsQueryTag() throws Exception {
...
+  @Test
+  public void addsEmptyAffectedRowsTagToEmptyPreparedBatchUpdates() throws Exception {
+    // In contrast to the plain statement case, this does produce loggable SQL, so a span is started. Since there are
+    // no entries in the batch, no updates are made, so there are no update counts. This is represented by an empty
+    // string as sql.affected_rows.
+    prepareExecuteBatchWithInts("update t set c='x' where i=?");
+
+    assertThat(spans)
+      .flatExtracting(s -> s.tags().entrySet())
+      .contains(entry("sql.affected_rows", ""));
```
jcchavezs, line 170: We usually avoid zero value tags because they are meaningless and also it is storage cost. I'd rather remove it.

reftel: OK, fixed in  889e1aa .

jcchavezs: i see, I wonder what happens when you have three batches and the affected rows are `2,0,3`? Can we cover that with a test?

reftel: As shown in bfac9c209, that one has "2,0,0", since there are no rows in the test data with i=3. Also added tests for i=0 ("0") and i=2,i=0,i=1 ("2,0,1").

jcchavezs (approved):

PR #1336, opened 2022-06-30
chore: test to prove #1306

jcchavezs:


codefromthecrypt (approved):
While not definitely representative of the intent of #1306, this proves the test used isn't fixed by it. Happy to merge.

PR #1343, opened 2022-08-22
Added support for Kafka 3.2.1

### LINKED ISSUE #1338: Add missing method in TracingConsumer to support Kafka 3.2.0+
To allow supporting newer versions of Kafka, would it be possible to upgrade the _TracingConsumer_ by adding the following method

`void enforceRebalance(final String reason);`

This method has been added in Kafka 3.2.0.

 Source code of the Consumer interface with the new method can be found here
https://github.com/apache/kafka/blob/3.2.0/clients/src/main/java/org/apache/kafka/clients/consumer/Consumer.java#L264

My current project using Spring Boot 2.7.2, Kafka 3.2.1 and Brave 5.13.9 had the issue which has been fixed by adding this new method locally. No other issues have been detected when using Kafka 3.2.1.

Any feedbacks are more than welcome on this issue. I may eventually contribute by dropping a PR.

frosiere:
Fixes  #1338

### DIFF
--- a/instrumentation/kafka-streams/src/test/java/brave/kafka/streams/ITKafkaStreamsTracing.java
+++ b/instrumentation/kafka-streams/src/test/java/brave/kafka/streams/ITKafkaStreamsTracing.java
@@ -1358,6 +1358,8 @@ public class ITKafkaStreamsTracing extends ITKafkaStreams {
     properties.put(StreamsConfig.APPLICATION_ID_CONFIG, testName.getMethodName());
     properties.put(StreamsConfig.consumerPrefix(ConsumerConfig.AUTO_OFFSET_RESET_CONFIG),
       Topology.AutoOffsetReset.EARLIEST.name().toLowerCase());
+    properties.put(StreamsConfig.DEFAULT_KEY_SERDE_CLASS_CONFIG, Serdes.StringSerde.class.getName());
+    properties.put(StreamsConfig.DEFAULT_VALUE_SERDE_CLASS_CONFIG, Serdes.StringSerde.class.getName());
     return properties;
   }
 
--- a/instrumentation/kafka-streams/src/test/java/brave/kafka/streams/ITKafkaStreamsTracing.java
+++ b/instrumentation/kafka-streams/src/test/java/brave/kafka/streams/ITKafkaStreamsTracing.java
@@ -1358,8 +1358,8 @@ public class ITKafkaStreamsTracing extends ITKafkaStreams {
     properties.put(StreamsConfig.APPLICATION_ID_CONFIG, testName.getMethodName());
     properties.put(StreamsConfig.consumerPrefix(ConsumerConfig.AUTO_OFFSET_RESET_CONFIG),
       Topology.AutoOffsetReset.EARLIEST.name().toLowerCase());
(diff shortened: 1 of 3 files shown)


jcchavezs:
Ping @jeqo

jeqo:
Thanks @frosiere ! Could you take a look at the failing tests? Seems something is broken on the Kafka client integration tests.

pom.xml
```diff
@@ -97,7 +97,7 @@
     <jetty-servlet25.version>7.6.21.v20160908</jetty-servlet25.version>
     <resteasy.version>3.14.0.Final</resteasy.version>
 
-    <kafka.version>3.0.0</kafka.version>
+    <kafka.version>3.2.0</kafka.version>
```
jeqo, line 100: @frosiere do you downgrade to `3.2.0` because of `kafka-junit` dependency?

frosiere: Yes, I had test issues when using 3.2.1 and there is no version of kafka-junit supporting Kafka 3.2.1 yet.
Hope the build is fixed now.

instrumentation/kafka-streams/src/test/java/brave/kafka/streams/ITKafkaStreamsTracing.java
```diff
@@ -1358,6 +1358,8 @@ Properties streamsProperties() {
     properties.put(StreamsConfig.APPLICATION_ID_CONFIG, testName.getMethodName());
     properties.put(StreamsConfig.consumerPrefix(ConsumerConfig.AUTO_OFFSET_RESET_CONFIG),
       Topology.AutoOffsetReset.EARLIEST.name().toLowerCase());
+    properties.put(StreamsConfig.DEFAULT_KEY_SERDE_CLASS_CONFIG, Serdes.StringSerde.class.getName());
+    properties.put(StreamsConfig.DEFAULT_VALUE_SERDE_CLASS_CONFIG, Serdes.StringSerde.class.getName());
```
jeqo, line 1362: It's including specific serdes on the DSL — though this doesn't hurt

frosiere: Good point, this has been fixed.

pom.xml
```diff
@@ -121,7 +121,7 @@
     <mockito.version>3.6.28</mockito.version>
     <jersey.version>2.33</jersey.version>
     <!-- must align with kafka version https://github.com/charithe/kafka-junit -->
-    <kafka-junit.version>4.2.1</kafka-junit.version>
+    <kafka-junit.version>4.2.2</kafka-junit.version>
```
jeqo, line 124: Check with this latest version. If it works, let's update kafka to 3.2.1
```suggestion
    <kafka-junit.version>4.2.3</kafka-junit.version>
```

jeqo:
It's failing on the license header, could you run `mvn license:format` to fix it? https://mycila.carbou.me/license-maven-plugin/#goals
I think it should go green after this.
Thanks!

pom.xml
frosiere: Kafka 3.2.1 can be used with Kafka Junit 4.2.3. Thanks.

jeqo (approved):
LGTM. Thank you @frosiere !

PR #1348, opened 2022-10-13
fix(sec): upgrade org.springframework:spring-web to 5.3.7

lxxawfl:
### What happened？
There are 1 security vulnerabilities found in org.springframework:spring-web 5.3.2
- [CVE-2021-22118](https://www.oscs1024.com/hd/CVE-2021-22118)

### What did I do？
Upgrade org.springframework:spring-web from 5.3.2 to 5.3.7 for vulnerability fix

### What did you expect to happen？
Ideally, no insecure libs should be used.

### How was this patch tested?
Run `mvn compile` failed locally, couldn't complete the build process.
Run `mvn clean test` failed locally, unit-test couldn't pass.

### The specification of the pull request
[PR Specification](https://www.oscs1024.com/docs/pr-specification/) from OSCS

### DIFF
--- a/instrumentation/spring-web/src/it/spring3/pom.xml
+++ b/instrumentation/spring-web/src/it/spring3/pom.xml
@@ -13,10 +13,7 @@
     or implied. See the License for the specific language governing permissions and limitations under
     the License.
 
--->
-<project xmlns="http://maven.apache.org/POM/4.0.0"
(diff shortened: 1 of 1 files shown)


codefromthecrypt:
this is an integration test, so it is intentionally an old version. If you are using trivy, you can use `--skip-files "**/src/it/*/pom.xml" ` to ignore these. here's the discussion https://github.com/aquasecurity/trivy/discussions/5787

PR #1370, opened 2023-05-02
Set JUL loggerName

mauhiz:
## Problem:
When using JUL to SLF4J bridge, the logger is misconfigured and logs are detected as being sent from "unknown.jul.logger".
As a result, it is not straightforward to configure logback to filter out brave debug logs.

## Solution:
This is apparently solvable but setting a logger name to each LogRecord produced, like in https://github.com/asciidoctor/asciidoctorj/pull/834

## Drawbacks
For people using JUL directly with the default format, this will change the log output slightly.

Before: 
```
Jan 01, 1970 9:00:00 AM null
FINE: Some message
```

After:

```
Jan 01, 1970 9:00:00 AM brave.Tracer
FINE: Some message
```

### DIFF
--- a/brave/src/main/java/brave/internal/Platform.java
+++ b/brave/src/main/java/brave/internal/Platform.java
@@ -14,6 +14,8 @@
 package brave.internal;
 
 import brave.Clock;
+import brave.Tracer;
+
 import java.net.InetAddress;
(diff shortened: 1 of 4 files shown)


codefromthecrypt (approved):
great idea

PR #1371, opened 2023-05-09
Dubbo3 support

ShenFeng312:
Hello! Due to Dubbo 2.7 no longer being maintained and an increasing number of users reporting that Brave cannot run on Dubbo 3, we have decided to upgrade the version of Dubbo. This version has smaller changes compared to the previous PR and is compatible with the features of Dubbo 2.7.

### DIFF
--- a/instrumentation/dubbo/README.md
+++ b/instrumentation/dubbo/README.md
@@ -1,5 +1,5 @@
 # brave-instrumentation-dubbo
-This is a tracing filter for RPC providers and consumers in [Apache Dubbo 2.7+](http://dubbo.apache.org/en-us/docs/dev/impls/filter.html)
+This is a tracing filter for RPC providers and consumers in Apache Dubbo3
 
 When used on a consumer, `TracingFilter` adds trace state as attachments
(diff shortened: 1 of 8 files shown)


ShenFeng312:
@jcchavezs

jcchavezs:
I am sorry for the delay. Running it.

ShenFeng312:
@jcchavezs  Hi, could you please help me rerun the unit tests?

jcchavezs:
This is a breaking change isn't? Will users in Dubbo 2.x able to consume this library? If not maybe we want to do a whole new package.

ShenFeng312:
> This is a breaking change isn't? Will users in Dubbo 2.x able to consume this library? If not maybe we want to do a whole new package.

 users in Dubbo 2.x able to consume this library

codefromthecrypt (approved):
I'm approving this, but we should add an invoker test to verify 2.7 still works. I hope you can raise a PR to do that. You can see examples of this in other projects which have a src/it directory. Thanks a lot!

PR #1372, opened 2023-05-09
Fixed the httpclient5 licenses.

JochemKuijpers:
The original PR (#1274) that introduced these classes originated from 2020, but the squash merge commit updated their git datetime to 2023, which makes the license formatting check fail on the master branch.

With this change, the headers are now correctly formatted according to the new commit's datetime, and a master branch build should hopefully succeed.

I'm making this PR because I'm waiting on the support for Apache HTTP 5.2+ to arrive, so I'm hoping with this contribution that a new release could happen soon.

PR #1373, opened 2023-05-10
fix: http5 parent version

jeqo:
With the headers version reported and fixed on #1372, seems http5 module got stuck on a previous version of the parent module when merging #1274 .

This fix should unblock the main workflows that are complaining about this issue.

### DIFF
--- a/instrumentation/httpclient5/pom.xml
+++ b/instrumentation/httpclient5/pom.xml
@@ -15,12 +15,13 @@
 
 -->
 <project xmlns="http://maven.apache.org/POM/4.0.0" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 http://maven.apache.org/xsd/maven-4.0.0.xsd">
+  <modelVersion>4.0.0</modelVersion>
+
   <parent>
     <groupId>io.zipkin.brave</groupId>
     <artifactId>brave-instrumentation-parent</artifactId>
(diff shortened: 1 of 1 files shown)


JochemKuijpers:
For what it's worth, I also ran into this locally when adjusting the license headers (and blamed my inexperience with setting up this project).

I managed to get it to compile by setting the project version for httpclient5 and it's dependencies back to `1.15.1` (no snapshot). I did not include this change in my commit of course. 

When attempting either `1.15.1-SNAPSHOT` or `1.15.2-SNAPSHOT`, Maven kept complaining about not being able to find the relevant dependency, despite it being present in the `oss.sonatype.org` repository. I didn't understand why..
That may be what you're running into now?

jeqo:
@JochemKuijpers maven is not configured to use sonatype for snapshots, so in this case as the http5 parent is not aligned with the project it's failing to find it.
If you build the whole project locally, then it should be working with 5.15.2-SNAPSHOT as it's the version of the parent pom: https://github.com/openzipkin/brave/blob/dac110162eb16abe755cffbc17ab2de5ec29f3ee/pom.xml#L22

jeqo:
Sure! I think once the build is fixed we can create a new release

jcchavezs (approved):

jeqo:
Thanks @jcchavezs! builds seem to be back on track (at least not failing because of wrong parent version). 
Will release 5.16.0 later today.

jeqo:
https://github.com/openzipkin/brave/releases/tag/5.16.0

PR #1374, opened 2023-05-11
Adds the missing instrumentation projects to brave-bom, fixes in README.md

JochemKuijpers:
These changes add the missing instrumentation projects to the brave-bom `pom.xml`, so users do not have to separate specify these project version constraints alongside the brave-bom project. This includes the recently added httpclient5 instrumentation.

I've maintained lexicographical ordering of the artifact ID in the XML file, so future additions or removals are easy to locate.

This PR includes some minor fixes to the README.md file, including:
* Fixing the build status link at the top to point to the right project page. It was pointing to another project while the image URL was reflecting the correct project.
* Reformatted the README file to wrap lines at column 76 where possible. This seemed to be the existing wrapping column already used in other paragraphs, so I've applied it to other paragraphs too.

### DIFF
--- a/brave-bom/pom.xml
+++ b/brave-bom/pom.xml
@@ -136,6 +136,11 @@
       </dependency>
       <dependency>
         <groupId>${project.groupId}</groupId>
+        <artifactId>brave-instrumentation-benchmarks</artifactId>
+        <version>${project.version}</version>
+      </dependency>
+      <dependency>
+        <groupId>${project.groupId}</groupId>
         <artifactId>brave-instrumentation-dubbo</artifactId>
         <version>${project.version}</version>
       </dependency>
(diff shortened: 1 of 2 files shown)


brave-bom/pom.xml
```diff
@@ -134,6 +134,11 @@
         <artifactId>brave-context-rxjava2</artifactId>
         <version>${project.version}</version>
       </dependency>
+      <dependency>
+        <groupId>${project.groupId}</groupId>
+        <artifactId>brave-instrumentation-benchmarks</artifactId>
```
shakuzen, line 139: I suspect it may be intentional that we don't include the benchmarks module in the bom. Is there a case where a third-party would depend on our benchmarks module?

JochemKuijpers: I don't know actually, I just added all of the projects. I'll leave this one up for regular contributers to decide.

JochemKuijpers: Personally I don't depend on it, but since it has its own package that's being published, I imagine getting the version alignment through the bom might be nice just in case.

jeqo: @JochemKuijpers I think it does not heart to have this modules here, as dependencies will have to be declared explicitly anyhow. I'm ok with this additions (also the tests module).

jeqo (approved):
LGTM

jcchavezs (approved):

PR #1378, opened 2023-09-08
Update zipkin to latest versions

strehle:
zipkin includes gson code internally but outdated version Bluckduck reports CVE-2022-25647


codefromthecrypt (approved):
thanks!

PR #1388, opened 2023-12-18
deps: bumps all main versions that compile

codefromthecrypt:
This bumps all main versions to current, except a few TODOs noted where there is drift. As this generally gets things years ahead and I'm out of time, good enough for now!


brave/src/main/java/brave/propagation/CurrentTraceContext.java
```diff
@@ -37,7 +37,7 @@
 public abstract class CurrentTraceContext {
   static {
     // ensure a reference to InternalPropagation exists
-    SamplingFlags.DEBUG.toString();
+    String unused = SamplingFlags.DEBUG.toString();
```
codefromthecrypt, line 40: errorprone

PR #1393, opened 2024-01-06
build: fixes bytecode level and updates build versions (#236)

codefromthecrypt:
This fixes a problem where we weren't releasing jars with the right bytecode level, due to a constant release=8. As we no longer can rely on RetroLambda to rewrite code, I've had to port things to the appropriate bytecode level manually. I also updated build deps that could be.

Special note: Projects that optionally use types above their bytecode level *unset* maven.compiler.release in the release profile. This allows the project to compile classes not at the source level. This is used in brave core and okhttp, for example.

### DIFF
--- a/brave/pom.xml
+++ b/brave/pom.xml
@@ -1,6 +1,6 @@
 <!--
 
-    Copyright 2013-2023 The OpenZipkin Authors
+    Copyright 2013-2024 The OpenZipkin Authors
 
     Licensed under the Apache License, Version 2.0 (the "License"); you may not use this file except
     in compliance with the License. You may obtain a copy of the License at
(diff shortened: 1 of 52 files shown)


brave/pom.xml
```diff
@@ -136,7 +126,11 @@
       <id>release</id>
       <properties>
         <!-- The brave jar needs to be Java 1.6 bytecode -->
-        <main.java.version>1.6</main.java.version>
+        <maven.compiler.source>1.6</maven.compiler.source>
+        <maven.compiler.target>1.6</maven.compiler.target>
+        <!-- Clear the release tag, so that we can compile Platform without
```
codefromthecrypt, line 131: example of clearing the tag.. this is what broke the bytecode version by accident last release, as the source/target were overridden by a hard-coded release property. Setting this to 6 breaks our ability to compile `brave.internal.Platform`, so clearing is the only way.

PR #1395, opened 2024-01-07
Removes all deprecated types and modules for Brave v6

codefromthecrypt:
Removes all deprecated types and modules for Brave v6. This also removes any internal symbols not used as a result of that.

This is the first release to have no Zipkin dependency, so users should move to manage their own zipkin-reporter deps before or at the time of upgrading to Brave 6.0.


brave-bom/pom.xml
```diff
@@ -34,8 +34,7 @@
     <main.basedir>${project.basedir}/..</main.basedir>
 
     <!-- use the same values in ../pom.xml -->
-    <zipkin.version>2.27.0</zipkin.version>
```
codefromthecrypt, line 37: no longer manage zipkin version as there is no code here that uses it.

brave-tests/src/main/java/brave/test/propagation/CurrentTraceContextTest.java
```diff
@@ -311,7 +311,7 @@ protected void is_inheritable(CurrentTraceContext inheritableCurrentTraceContext
   }
 
   @Test void restoresSpanAfterCallable() throws Exception {
-    try (Scope scope0 = currentTraceContext.newScope(context)) {
+    try (Scope scope = currentTraceContext.newScope(context)) {
```
codefromthecrypt, line 314: it was insane to use analysis to look for unused variables while this was called ws, scope, scope0, so forgive this change... but it really helped in the slog!

codefromthecrypt:
I will change the version of this PR to 6.0.0-SNAPSHOT once 5.18 is cut (shortly)

codefromthecrypt:
So brave no longer manages reporter versions, which is a good thing as then they cannot conflict. Brave is compatible with both zipkin-reporter 2.x and 3.x, but pinning the latter in the bom can make things complicated for old libraries. It is better to pin the reporter to whatever frameworks need. For example, otel pins 2.x at the moment.

codefromthecrypt:
will confirm in brave-example prior to release

PR #1396, opened 2024-01-07
Undeprecates propagation symbols for v6 interop

codefromthecrypt:
This undeprecates B3SinglePropagation and enforces at runtime that `Propagation.Factory#get()` must be implemented. I spot checked all the implementations on GitHub already do.

### DIFF
--- a/brave/src/main/java/brave/internal/propagation/StringPropagationAdapter.java
+++ b/brave/src/main/java/brave/internal/propagation/StringPropagationAdapter.java
@@ -1,5 +1,5 @@
 /*
- * Copyright 2013-2023 The OpenZipkin Authors
+ * Copyright 2013-2024 The OpenZipkin Authors
  *
  * Licensed under the Apache License, Version 2.0 (the "License"); you may not use this file except
  * in compliance with the License. You may obtain a copy of the License at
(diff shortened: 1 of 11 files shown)


codefromthecrypt:
ok I scraped github and raised PRs for all the known implementations of propagation. Most were things I did or copies of stuff @marcingrzejszczak made cc @shakuzen

After those projects merge, they will be safe to update into Brave 6 (#1395)

brave/src/main/java/brave/internal/propagation/StringPropagationAdapter.java
```diff
@@ -70,7 +70,10 @@
  * }</pre>
  *
  * @since 5.12
+ * @deprecated As of Brave 5.18, throw an {@link UnsupportedOperationException} in
```
codefromthecrypt, line 73: this is internal, but I put a deprecated message on it anyway as many have been using it without shading. The pull requests I mentioned move off it.

codefromthecrypt:
Also, all PRs I mentioned on this issue are opened to allow editors to write to my branch. Not saying they will, but they can as necessary.

codefromthecrypt:
I also released this to maven central as it seems there are a couple copies of brave w3c tracecontext (even if not the same impl as the one I wrote) https://central.sonatype.com/artifact/io.zipkin.contrib.brave-propagation-w3c/brave-propagation-tracecontext/versions https://github.com/openzipkin-contrib/brave-propagation-w3c

reta (approved):

PR #1403, opened 2024-01-08
deps: reverts to zipkin-reporter 2.x

codefromthecrypt:
After integrating with [Sleuth](https://github.com/spring-cloud/spring-cloud-sleuth/pull/2335) and [Otel](https://github.com/open-telemetry/opentelemetry-java/pull/6129), it seems hastey to manage the zipkin-reporter to 3.0. The reason is both projects end up sharing classpath between Brave and reporter types. Since Brave doesn't care about this version anyway, it can leave it at the prior and remove tension. Brave 6 will completely remove zipkin version management, anyway.

Following this change, I will cut a release-5.18.1 tag, then restore the Brave 6 commit (44e4081be625d5ff73c3862e2361ae92caf83e2a) and release it.

### DIFF
--- a/brave-bom/pom.xml
+++ b/brave-bom/pom.xml
@@ -35,7 +35,7 @@
 
     <!-- use the same values in ../pom.xml -->
     <zipkin.version>2.27.0</zipkin.version>
-    <zipkin-reporter.version>3.0.0</zipkin-reporter.version>
+    <zipkin-reporter.version>2.17.2</zipkin-reporter.version>
   </properties>
 
   <organization>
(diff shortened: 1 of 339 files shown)


brave-bom/pom.xml
```diff
@@ -35,7 +35,7 @@
 
     <!-- use the same values in ../pom.xml -->
     <zipkin.version>2.27.0</zipkin.version>
-    <zipkin-reporter.version>3.0.0</zipkin-reporter.version>
+    <zipkin-reporter.version>2.17.2</zipkin-reporter.version>
```
codefromthecrypt, line 38: this is the main change. the license thing was that reverting files touches them.

PR #1406, opened 2024-01-10
ci: deploys bom separately to allow re-do on partial failure

codefromthecrypt:
Before, the bom project deployed only after the parent succeeded. This meant a timeout would prevent it from being deployed, even if the timeout was not critical. This deploys the bom independently so that such a situation doesn't cause us to have to re-cut a version.

This also fixes where we didn't actually use the release args in tests.

### DIFF
--- a/.github/workflows/deploy.yml
+++ b/.github/workflows/deploy.yml
@@ -14,7 +14,17 @@ on:
 
 jobs:
   deploy:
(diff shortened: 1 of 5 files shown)


reta (approved):

codefromthecrypt:
this worked for reporter, proved by https://github.com/openzipkin/zipkin/pull/3683

PR #1409, opened 2024-01-15
kafka-tracing: injects init context on forward

### LINKED ISSUE #1382: Kafka streams processor tracing support on context.forward
In previous versions of Kafka Streams, it was possible to use a transformer to convert an input into an output after transformation.

Since Kafka Streams 3.4, these transformers are deprecated in favor of the new processor API. So, basically, a record can be forwarded to children processors using context.forward.

It's working fine in terms of Kafka flow but the tracing support seems incomplete (baggage values updated in a processor are not propagated) when using this method.

So, shouldn't we review the tracing to also take care of baggages propagation when using the forward method?
Maybe I missed something but I were not able to propagate any baggages.

Remark:

- The TracingContext contains the correct baggage values but this one is not consulted when doing a forward.
- The ProcessorContext contains the original baggage values

Maybe KafkaStreamsTracing should provide a forward method copying the tracing context on the record headers?

Any help or feedback is more than welcome.

codefromthecrypt:
Fixes #1382

### DIFF
--- a/instrumentation/kafka-streams/src/main/java/brave/kafka/streams/TracingProcessingContext.java
+++ b/instrumentation/kafka-streams/src/main/java/brave/kafka/streams/TracingProcessingContext.java
@@ -30,7 +30,7 @@ import org.apache.kafka.streams.processor.TaskId;
 import org.apache.kafka.streams.processor.api.ProcessingContext;
 import org.apache.kafka.streams.processor.api.RecordMetadata;
 
-class TracingProcessingContext<C extends ProcessingContext> implements ProcessingContext {
+abstract class TracingProcessingContext<C extends ProcessingContext> implements ProcessingContext {
   final C delegate;
   final Injector<Headers> injector;
   final TraceContext context;
--- a/instrumentation/kafka-streams/src/main/java/brave/kafka/streams/TracingProcessingContext.java
+++ b/instrumentation/kafka-streams/src/main/java/brave/kafka/streams/TracingProcessingContext.java
@@ -1,0 +1,101 @@
+/*
+ * Copyright 2013-2024 The OpenZipkin Authors
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License"); you may not use this file except
+ * in compliance with the License. You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software distributed under the License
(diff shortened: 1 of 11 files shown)


instrumentation/kafka-streams/src/test/java/brave/kafka/streams/ITKafkaStreamsTracing.java
```diff
@@ -263,6 +264,61 @@ record -> {
...
+    builder.<String, String>stream(inputTopic)
+      .process(kafkaStreamsTracing.process("forward",
+        () -> new ContextualProcessor<String, String, String, String>() {
+          @Override public void process(Record<String, String> record) {
+            context().forward(record);
+          }
+        }))
+      .process(kafkaStreamsTracing.process("check",
+        () -> (Processor<String, String, String, String>) record -> {
+          contexts.add(currentTraceContext.get());
```
codefromthecrypt, line 282: added this here as otherwise you don't know for sure that both it had the right parent and also the wrapped processor had the correct local span.

instrumentation/kafka-streams/src/main/java/brave/kafka/streams/TracingProcessingContext.java
```diff
@@ -0,0 +1,101 @@
...
+import org.apache.kafka.streams.StreamsMetrics;
+import org.apache.kafka.streams.processor.Cancellable;
+import org.apache.kafka.streams.processor.PunctuationType;
+import org.apache.kafka.streams.processor.Punctuator;
+import org.apache.kafka.streams.processor.StateStore;
+import org.apache.kafka.streams.processor.TaskId;
+import org.apache.kafka.streams.processor.api.ProcessingContext;
+import org.apache.kafka.streams.processor.api.RecordMetadata;
+
+class TracingProcessingContext<C extends ProcessingContext> implements ProcessingContext {
```
reta, line 33: Super  nit, just curious if we could mark this class `abstract`, it seems to be a building block but not used by itself

reta (approved):

instrumentation/kafka-streams/src/main/java/brave/kafka/streams/TracingProcessingContext.java
codefromthecrypt: good idea. thanks!

codefromthecrypt:
Thanks for the help @frosiere and @reta!

PR #1412, opened 2024-01-31
Support rocketmq

JoeKerouac:
support RocketMQ
see #1356


JoeKerouac:
@codefromthecrypt

codefromthecrypt:
ok I re-pushed into a single commit as we squash on merge anyway, and already spent the work to rebase this on master.

NEXT STEP:

The [docker image](https://hub.docker.com/r/apache/rocketmq/tags) for 5.1.4 is not multi-platform, so we need a multi-platform image or people like me cannot run docker tests (on apple silicon aka arm64). @JoeKerouac can you check on what is holding up https://github.com/apache/rocketmq-docker/issues/98?

codefromthecrypt:
so this is the current error until there's a multi-platform image. Our other images we make on own, because they are shared with a zipkin server transport. As there's no rocketmq server transport, this one is special case and we need to address the docker image in some way, at least upstream would be good to find out why not.

```
[INFO] Running brave.rocketmq.client.ITRocketMQTracingTest
08:31:48,591 WARN  [main] containers.GenericContainer (GenericContainer.java:488) - The architecture 'amd64' for image 'apache/rocketmq:5.1.4' (ID sha256:f14b643f5d86369b50362c0039938cf0346f451b559d95f4f2c530083f9e579f) does not match the Docker server architecture 'arm64'. This will cause the container to execute much more slowly due to emulation and may lead to timeout failures.
[ERROR] Tests run: 6, Failures: 1, Errors: 5, Skipped: 0, Time elapsed: 34.80 s <<< FAILURE! -- in brave.rocketmq.client.ITRocketMQTracingTest
[ERROR] brave.rocketmq.client.ITRocketMQTracingTest.tracingMessageListenerConcurrently -- Time elapsed: 6.291 s <<< ERROR!
```

codefromthecrypt:
@anuraaga @reta @shakuzen @jeqo I think this is a good example of us needing to add the new apple silicon runners and use them in the docker tests. If anyone feels up to task of github action slinging here and otherwise, please do! (can do an org search for testcontainers, but should be mostly zipkin, -dependencies, -reporter, here, brave-cassandra and the contrib storage things)

codefromthecrypt:
nevermind about CI catching this as there's not yet a free runner that works with arm64 *and docker* https://github.com/orgs/community/discussions/69211

JoeKerouac:
> ok I re-pushed into a single commit as we squash on merge anyway, and already spent the work to rebase this on master.
> 
> NEXT STEP:
> 
> The [docker image](https://hub.docker.com/r/apache/rocketmq/tags) for 5.1.4 is not multi-platform, so we need a multi-platform image or people like me cannot run docker tests (on apple silicon aka arm64). @JoeKerouac can you check on what is holding up [apache/rocketmq-docker#98](https://github.com/apache/rocketmq-docker/issues/98)?

I will do my best to resolve this issue; however, it may not be straightforward. I need some time, and there might not be a definite deadline for it.

JoeKerouac:
However, regarding the current test case failure, I will address it as soon as possible.

JoeKerouac:
@codefromthecrypt I don't know why the RocketMQ version in my code was upgraded from 4.6.0 to 5.1.4. Since there are significant differences between 5.1.4 and 4.x, the current goal is RocketMQ 4.6.0 instead of 5.1.4.

codefromthecrypt:
@JoeKerouac I don't think we have ever not used a current version as the first version of instrumentation. Can you mention why there is an issue with 5.x or why 5.x is not desired? When we don't use current versions, we get security (CVE) problems to deal with, and also usually eventually have to support latest anyway. We don't likely want to have something both for rocket 4.x and 5.x

codefromthecrypt:
for example, old and new dubbo caused a lot of work for the project. Not saying this is like dubbo, but dubbo was one of the most high maintenance and also slowest tests in the project. We don't want to repeat something like we had when we had 2 dubbos

JoeKerouac:
@codefromthecrypt Since that's the case, I'll go ahead and upgrade to version 5.1.4; it's already at version 5.1.4 now.

codefromthecrypt:
thanks @JoeKerouac. If there's a compatability issue with 5 against 4, we can add a maven-invoker-plugin test, but if it should work, let's just stick with 5.x.

instrumentation/rocketmq-client/src/test/java/brave/rocketmq/client/ITRocketMQTracingTest.java
```diff
@@ -226,17 +230,17 @@ protected ConsumeOrderlyStatus handleMessage(MessageExt messageExt,
     consumer.subscribe(topic, "*");
     CountDownLatch latch = new CountDownLatch(1);
     AtomicReference<Span> reference = new AtomicReference<>();
-    consumer.registerMessageListener(new TracingMessageListenerOrderly(0, consumerTracing) {
+    MessageListenerOrderly messageListenerOrderly = consumerTracing.wrap(new MessageListenerOrderly() {
```
codefromthecrypt, line 233: I think this test can be simpler using similar style as JMS. For example, all you need to do is prove that you can read back the tag like so:

```java
  @Test void should_complete_on_callback() {
    producer.setAsync(new CompletionListener() {
      @Override public void onCompletion(Message message) {
        tracing.tracer().currentSpanCustomizer().tag("onCompletion", "");
      }

      @Override public void onException(Message message, Exception exception) {
        tracing.tracer().currentSpanCustomizer().tag("onException", "");
      }
    });
    producer.send(jms.queue, "foo");

    assertThat(testSpanHandler.takeRemoteSpan(PRODUCER).tags())
      .containsKeys("onCompletion");
  }
```

In other words, in this test and similar ones that are testing callbacks, you can do like

```java
 tracing.tracer().currentSpanCustomizer().tag("consumeMessage", "");
```

or whatever the callback method name is, and don't worry about countdown latches, because the IntegrationTestSpanHandler has a blocking queue and it will wait for the span to arrive.

JoeKerouac: OK, I have already modified this part of the code.

codefromthecrypt: great. I like the amount of red in this diff!

brave-bom/pom.xml
```diff
@@ -226,6 +226,16 @@
         <artifactId>brave-instrumentation-okhttp3</artifactId>
         <version>${project.version}</version>
       </dependency>
+      <dependency>
+        <groupId>${project.groupId}</groupId>
+        <artifactId>brave-instrumentation-rocketmq-client</artifactId>
+        <version>${project.version}</version>
+      </dependency>
+      <dependency>
+        <groupId>${project.groupId}</groupId>
```
codefromthecrypt, line 235: this should be removed

instrumentation/rocketmq-client/src/main/java/brave/rocketmq/client/RocketMQTracing.java
```diff
@@ -0,0 +1,147 @@
...
+
+  public MessageListenerConcurrently wrap(MessageListenerConcurrently messageListenerConcurrently) {
+    return new TracingMessageListenerConcurrently(defaultDelayLevelWhenNextConsume, this, messageListenerConcurrently);
+  }
+
+  public MessageListenerConcurrently wrap(int delayLevelWhenNextConsume, MessageListenerConcurrently messageListenerConcurrently) {
+    return new TracingMessageListenerConcurrently(delayLevelWhenNextConsume, this, messageListenerConcurrently);
+  }
+
+  public MessageListenerOrderly unwrap(MessageListenerOrderly messageListenerOrderly) {
```
codefromthecrypt, line 133: I think this doesn't need to be public. either make package-private or remove

instrumentation/rocketmq-client/src/main/java/brave/rocketmq/client/RocketMQTracing.java
```diff
@@ -0,0 +1,147 @@
...
+    if (messageListenerOrderly instanceof TracingMessageListenerOrderly) {
+      return ((TracingMessageListenerOrderly)messageListenerOrderly).messageListenerOrderly;
+    }
+    return null;
+  }
+
+  public MessageListenerConcurrently unwrap(MessageListenerConcurrently messageListenerConcurrently) {
```
codefromthecrypt, line 140: I think this doesn't need to be public. either make package-private or remove

instrumentation/rocketmq-client/src/main/java/brave/rocketmq/client/RocketMQTracing.java
```diff
@@ -0,0 +1,147 @@
...
+
+  public Tracing tracing() {
+    return tracing;
+  }
+
+  public Tracer tracer() {
+    return tracer;
+  }
+
+  public MessageListenerOrderly wrap(MessageListenerOrderly messageListenerOrderly) {
```
codefromthecrypt, line 117: for these functions, follow existing convention which is to use a function name that is the same as the type being wrapped. Also needs javadoc. You can look at kafka-tracing for example, but make sure when following the conventions, the javadoc says correct. e.g. if it will produce a span or if it is just for adding tags:

```java
  /**
   * Extracts or creates a {@link Span.Kind#CONSUMER} span for each message received. This span is
   * injected onto each message so it becomes the parent when a processor later calls {@link
   * #nextSpan(ConsumerRecord)}.
   */
  public <K, V> Consumer<K, V> consumer(Consumer<K, V> consumer) {
    if (consumer == null) throw new NullPointerException("consumer == null");
    return new TracingConsumer<>(consumer, this);
  }
```

instrumentation/rocketmq-client/src/main/java/brave/rocketmq/client/RocketMQTracing.java
```diff
@@ -0,0 +1,147 @@
...
+  }
+
+  // We can't just skip clearing headers we use because we might inject B3 single, yet have stale B3
+  // multi, or vice versa.
+  void clearTraceIdHeaders(Map<String, String> headers) {
+    for (String traceIDHeader : traceIdHeaders)
+      headers.remove(traceIDHeader);
+  }
+
+  public Tracing tracing() {
```
codefromthecrypt, line 109: we don't need to expose this, as someone can use the same `MessagingTracing` to get this property used to construct this object, to get `tracing`. so remove it please.

instrumentation/rocketmq-client/src/main/java/brave/rocketmq/client/RocketMQTracing.java
```diff
@@ -0,0 +1,147 @@
...
+    return tracing;
+  }
+
+  public Tracer tracer() {
```
codefromthecrypt, line 113: per above remove this, too

instrumentation/rocketmq-client/src/main/java/brave/rocketmq/client/TracingMessageListenerConcurrently.java
```diff
@@ -0,0 +1,68 @@
...
+      Span span =
+        Util.createAndStartSpan(tracing, tracing.consumerExtractor, tracing.consumerSampler,
+          request, msg.getProperties());
+      span.name(RocketMQTags.FROM_PREFIX + msg.getTopic());
+
+      ConsumeConcurrentlyStatus result;
+      try (SpanInScope scope = tracing.tracer().withSpanInScope(span)) {
+        result = messageListenerConcurrently.consumeMessage(Collections.singletonList(msg), context);
+      } catch (Exception e) {
+        context.setDelayLevelWhenNextConsume(delayLevelWhenNextConsume);
```
codefromthecrypt, line 54: I don't think we should have the responsibility to decide when to retry messages, as we are only doing tracing. you'll see we don't do this in any other projects. Please remove this, as the delegate `messageListenerConcurrently` can do this if they like.



change the form here to like this here and everywhere and remove configuration like `delayLevelWhenNextConsume`

```java
    Tracer.SpanInScope scope = tracer.withSpanInScope(span);
    Throwable error = null;
    try {
      // do work
      error = e;
      throw e;
    } finally {
      // finish as an exception means the callback won't finish the span
      if (error != null) span.error(error).finish();
      scope.close();
    }
```

instrumentation/rocketmq-client/src/main/java/brave/rocketmq/client/TracingMessageListenerConcurrently.java
```diff
@@ -0,0 +1,68 @@
...
+        result = ConsumeConcurrentlyStatus.RECONSUME_LATER;
+      } finally {
+        long timestamp = tracing.tracing.clock(span.context()).currentTimeMicroseconds();
+        span.finish(timestamp);
```
codefromthecrypt, line 58: no need to manually set the timestamp 
```suggestion
        span.finish();
```

instrumentation/rocketmq-client/src/main/java/brave/rocketmq/client/Util.java
```diff
@@ -0,0 +1,50 @@
...
+      span = tracing.nextMessagingSpan(sampler, request, extracted);
+    } else {
+      span = tracer.newChild(traceContext);
+    }
+
+    span.kind(request.spanKind());
+    span.remoteServiceName(tracing.remoteServiceName);
+    span.tag(RocketMQTags.ROCKETMQ_TOPIC, request.channelName());
+    long timestamp = tracing.tracing.clock(span.context()).currentTimeMicroseconds();
+    span.start(timestamp);
```
codefromthecrypt, line 46: just use .start() instead of manually doing timestamp
```suggestion
    span.start();
```

codefromthecrypt (commented):
getting closer! After these changes will need another pass, and then I will start asking for and suggesting non IT unit tests, using examples from kafka or jms where we have them

codefromthecrypt:
nice to see passing tests @JoeKerouac!

JoeKerouac:
Sorry, I just finished the Chinese New Year holiday. I will handle it as soon as possible.

JoeKerouac:
@codefromthecrypt I have completed all the modifications suggested during the code review.

codefromthecrypt (commented):
Thanks for the progress! Besides notes mentioned, please audit the other messaging libraries and look for IT tests you may have missed. For example, custom sampling and continuing a trace. You can look at least at kafka-clients to see what is left out.

Also, has there been movement on a multi-platform image? Once you get the current docker setup simpler (ideally no custom shell script), we can look into making our own image, as we can't release this until it can run on arm64.

codefromthecrypt:
note: status is still the same since my last comments, except the license format has now changed and will need to be updated along with past comments here.

reta:
I think this one could be closed, delivered by https://github.com/openzipkin/brave/pull/1449

shakuzen:
Superseded by #1449

PR #1413, opened 2024-02-06
log4j12: uses version with CVEs only via invoker

codefromthecrypt:
snyk finally got me to do this ;) If folks are ok with this, I'll do similar for spring-beans which also shows up as a critical main CVE.


shakuzen (approved):

anuraaga (approved):

reta (approved):

codefromthecrypt:
cool. will do similar for spring-beans and also in zipkin-reporter, if we aren't already ..

PR #1414, opened 2024-02-13
removes IpLiteral dependency from ZipkinV2JsonWriter

codefromthecrypt:
This removes a dependency on `IpLiteral.detectFamily` from json writing, and adds a benchmark to make sure it doesn't hurt performance. While a nit, this clarifies that it has already happened, so that a proto encoder port will not think it needs to re-detect also.

Before
```
Benchmark                                                              Mode     Cnt    Score   Error   Units
ZipkinV2JsonWriterBenchmarks.write_bigClientSpan:gc.alloc.rate.norm  sample      15   24.104 ± 0.011    B/op
ZipkinV2JsonWriterBenchmarks.write_bigClientSpan:p0.99               sample            0.791           us/op
ZipkinV2JsonWriterBenchmarks.write_serverSpan:gc.alloc.rate.norm     sample      15   24.040 ± 0.006    B/op
ZipkinV2JsonWriterBenchmarks.write_serverSpan:p0.99                  sample            0.333           us/op

```

After
```
Benchmark                                                              Mode     Cnt    Score   Error   Units
ZipkinV2JsonWriterBenchmarks.write_bigClientSpan:gc.alloc.rate.norm  sample      15   24.101 ± 0.011    B/op
ZipkinV2JsonWriterBenchmarks.write_bigClientSpan:p0.99               sample            0.750           us/op
ZipkinV2JsonWriterBenchmarks.write_serverSpan:gc.alloc.rate.norm     sample      15   24.039 ± 0.007    B/op
ZipkinV2JsonWriterBenchmarks.write_serverSpan:p0.99                  sample            0.333           us/op
```

### DIFF
--- a/brave/src/test/java/brave/handler/MutableSpanTest.java
+++ b/brave/src/test/java/brave/handler/MutableSpanTest.java
@@ -652,6 +652,15 @@ public class MutableSpanTest {
       .isEqualTo(new MutableSpan(context, null));
   }
 
+  @Test void unwrapsIpv4() {
+    MutableSpan span = new MutableSpan();
+    span.localIp("::FFFF:43.0.192.2"); // mapped
+    span.remoteIp("::0000:43.0.192.2"); // compat
+
+    assertThat(span.localIp()).isEqualTo("43.0.192.2");
+    assertThat(span.remoteIp()).isEqualTo("43.0.192.2");
+  }
+
   @Test void tags() {
     MutableSpan span = new MutableSpan();
     assertThat(span.tagCount()).isZero();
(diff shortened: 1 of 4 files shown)


codefromthecrypt:
hmm this seems unrelated

```
 Error:  Failures: 
Error:    ThreadLocalSpanClassLoaderTest.currentTracer_basicUsage_unloadable:35 class brave.propagation.ThreadLocalSpanClassLoaderTest$ExplicitTracerBasicUsage includes state that couldn't be garbage collected
```

brave/src/test/java/brave/handler/MutableSpanTest.java
```diff
@@ -652,6 +652,15 @@ class Tag {
       .isEqualTo(new MutableSpan(context, null));
   }
 
+  @Test void unwrapsIpv4() {
```
codefromthecrypt, line 655: noticed we missed a test before on this

anuraaga (approved):

reta (approved):

codefromthecrypt:
Thanks for the look, folks!

PR #1415, opened 2024-02-13
json: fixes incorrect handling of error tag

codefromthecrypt:
Before, we didn't document that an existing tag with the same name as the `errorTag` parameter of the JSON writer would be retained. This documents it. It also fixes a bug where that key is not literally "error", for example, adding a tag named "exception".

I backfilled tests and adjusted benchmarks to make sure the result isn't worse than before.

Before
```
Benchmark                                                          Mode     Cnt    Score   Error   Units
ZipkinV2JsonWriterBenchmarks.sizeInBytes_bigClientSpan:gc.alloc.rate.norm  sample      15   0.042 ± 0.007    B/op
ZipkinV2JsonWriterBenchmarks.sizeInBytes_bigClientSpan:p0.99               sample           0.334           us/op
ZipkinV2JsonWriterBenchmarks.write_bigClientSpan:gc.alloc.rate.norm        sample      15  24.105 ± 0.008    B/op
ZipkinV2JsonWriterBenchmarks.write_bigClientSpan:p0.99                     sample           0.791           us/op
```

After
```
Benchmark                                                                    Mode     Cnt    Score   Error   Units
ZipkinV2JsonWriterBenchmarks.sizeInBytes_bigClientSpan:gc.alloc.rate.norm  sample      15    0.041 ± 0.006    B/op
ZipkinV2JsonWriterBenchmarks.sizeInBytes_bigClientSpan:p0.99               sample            0.334           us/op
ZipkinV2JsonWriterBenchmarks.write_bigClientSpan:gc.alloc.rate.norm        sample      15   24.101 ± 0.017    B/op
ZipkinV2JsonWriterBenchmarks.write_bigClientSpan:p0.99                     sample            0.750           us/op
```

### DIFF
--- a/brave/src/main/java/brave/handler/MutableSpanBytesEncoder.java
+++ b/brave/src/main/java/brave/handler/MutableSpanBytesEncoder.java
@@ -22,6 +22,12 @@ import java.util.List;
 /** Similar to {@code zipkin2.MutableSpan.SpanBytesEncoder} except no Zipkin dependency. */
 public abstract class MutableSpanBytesEncoder {
 
+  /**
+   * Encodes a {@linkplain MutableSpan} into Zipkin's V2 json format.
+   *
+   * @param errorTag sets the tag for a {@linkplain MutableSpan#error()}, if the corresponding key
+   *                 doesn't already exist.
+   */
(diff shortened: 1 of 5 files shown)


codefromthecrypt:
merging to unblock

PR #1420, opened 2024-02-27
trivy: adds maven-invoker-plugin ignoring config

codefromthecrypt:
I've given up trying to convince trivy to consider maven-invoker-plugin a test vs a deployment. This adds a config and a helper script to run it ad-hoc.


codefromthecrypt:
marking draft in case upstream accepts a plan forward which is to consider src/it deps "dev dependencies" which would have the same affect of not burdening folks with noise, unless they opt into it.

codefromthecrypt:
https://github.com/aquasecurity/trivy/pull/6213 obviates this

Issue #1421, opened 2024-02-29
Excessive allocations ExtraFactory.decorate


kilink:
I frequently see ExtraFactory.decorate at the top of the list of allocation hotspots when I profile our services. Here is an example from a 1 minute profile:

![image](https://github.com/openzipkin/brave/assets/835139/550a1384-4b40-477b-bc21-8833a2e029ac)

It looks to me like this is caused by excessive resizing of the extras ArrayList. If we look at ExtraFactory.decorate, I believe this happens because of [this line](https://github.com/openzipkin/brave/blob/987a6816cd4330b6a95e0c87c2077c9dde0f3f6a/brave/src/main/java/brave/internal/extra/ExtraFactory.java#L163):

```java
TraceContext.Builder builder = context.toBuilder().clearExtra().addExtra(claimed);
```

And then the subsequent loop that adds Extra instances one at a time; it's an interaction between the above code and [Lists.ensureMutable](https://github.com/openzipkin/brave/blob/987a6816cd4330b6a95e0c87c2077c9dde0f3f6a/brave/src/main/java/brave/internal/collect/Lists.java#L23); we always pass `Collections.emptyList()` to Lists.ensureMutable, which causes the ArrayList to get initialized with an initialCapacity of 0. Due to how Java's ArrayList calculates the new capacity when growing a list, for sizes this small we end up reallocating A LOT. Here is a simple example that demonstrates that:

```java
TraceContext.Builder builder = TraceContext.newBuilder().clearExtra();
builder.addExtra("foo"); // ArrayList.grow called, capacity = 0, minCapacity = 1, newCapacity = 1
builder.addExtra("bar"); // ArrayList.grow called, capacity = 1, minCapacity = 2, newCapacity = 2
builder.addExtra("baz"); // ArrayList.grow called, capacity = 2, minCapacity = 3, newCapacity = 3
builder.addExtra("quux"); // ArrayList.grow called, capacity = 3, minCapacity = 4, newCapacity = 4
builder.addExtra("blah"); // ArrayList.grow called, capacity = 4, minCapacity = 5, newCapacity = 6
builder.addExtra("blahblah"); // ArrayList.grow not called, capacity = 6
builder.addExtra("blahblahblah"); // ArrayList.grow called, capacity=6, minCapacity = 7, newCapacity = 9
```

[...]

PR #1424, opened 2024-03-12
Consolidates notes on "extra" data

codefromthecrypt:
Before this change, notes about state, particularly the "one instance per plugin" rule were not consolidated on the method that can cause problems `addExtra`. This migrates notes together with additional warnings that when we talk about this being used for propagation plugins, we also mean not for ad-hoc usage, as that can lead to excessive overhead.

See #1421

### DIFF
--- a/brave/src/main/java/brave/propagation/TraceContext.java
+++ b/brave/src/main/java/brave/propagation/TraceContext.java
@@ -14,9 +14,12 @@
 package brave.propagation;
 
 import brave.Span;
+import brave.baggage.BaggageFields;
+import brave.baggage.BaggagePropagation;
 import brave.internal.InternalPropagation;
 import brave.internal.Nullable;
 import brave.internal.Platform;
+import brave.internal.baggage.ExtraBaggageContext;
 import java.lang.ref.WeakReference;
 import java.util.Collections;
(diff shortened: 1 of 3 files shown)


brave/src/main/java/brave/propagation/TraceContextOrSamplingFlags.java
```diff
@@ -245,9 +246,10 @@ public TraceContextOrSamplingFlags sampled(boolean sampled) {
    * #context()} is {@code null}.
    *
    * @see TraceContext#extra()
+   * @see Builder#addExtra(Object) for notes on extra values.
    * @since 4.9
    */
-  public final List<Object> extra() {
```
codefromthecrypt, line 250: type is final, so this is redundant

brave/src/main/java/brave/propagation/TraceContext.java
```diff
@@ -372,6 +374,24 @@ public Builder clearExtra() {
     }
 
     /**
+     * This is an advanced function used for {@link Propagation} plugins, such as
```
reta, line 377: 👍

reta (approved):

PR #1431, opened 2024-04-10
Adds support for org.mongodb:mongodb-driver-core v5

### LINKED ISSUE #1430: Support MongoDB client version 5.x
## Describe the Bug
TraceMongoCommandListener calls connectionDescription.getServerAddress().getSocketAddress(). This throws a MethodNotFound exception in combination with the latest MongoDB client versions 5.x.

## Steps to Reproduce
Create a dummy project with brave MongoDB instrumentation + MongoDB client version 5.x and try to compile it.

## Expected Behaviour
The recommendation is to use "new InetSocketAddress(InetAddress.getByName(host), port);" to get the address.

codefromthecrypt:
This adds support for the new org.mongodb:mongodb-driver-core v5 driver via the following:
* move org.mongodb:mongodb-driver v3 to a maven invoker test
* change the provided artifact from org.mongodb:mongodb-driver -> org.mongodb:mongodb-driver-core
* change parsing of the remote IP and port to use reflection because there's no compatible alternative
* change tests to use mock instead of constructors, as the constructors are also incompatible

Fixes #1430

### DIFF
--- a/instrumentation/mongodb/pom.xml
+++ b/instrumentation/mongodb/pom.xml
@@ -31,15 +31,22 @@
 
     <main.basedir>${project.basedir}/../..</main.basedir>
 
-    <mongodb-driver.version>3.12.14</mongodb-driver.version>
+    <mongodb-driver.version>5.0.1</mongodb-driver.version>
+    <old-mongodb-driver.version>3.11.0</old-mongodb-driver.version>
   </properties>
 
   <dependencies>
     <dependency>
       <groupId>org.mongodb</groupId>
-      <artifactId>mongodb-driver</artifactId>
+      <artifactId>mongodb-driver-core</artifactId>
       <version>${mongodb-driver.version}</version>
       <scope>provided</scope>
+    </dependency>
+    <dependency>
+      <groupId>org.mongodb</groupId>
+      <artifactId>mongodb-driver-sync</artifactId>
+      <version>${mongodb-driver.version}</version>
+      <scope>test</scope>
     </dependency>
     <dependency>
       <groupId>${project.groupId}</groupId>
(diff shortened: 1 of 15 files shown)


instrumentation/mongodb/pom.xml
```diff
@@ -31,16 +31,23 @@
 
     <main.basedir>${project.basedir}/../..</main.basedir>
 
-    <mongodb-driver.version>3.12.14</mongodb-driver.version>
+    <mongodb-driver.version>5.0.1</mongodb-driver.version>
+    <old-mongodb-driver.version>3.11.0</old-mongodb-driver.version>
```
reta, line 35: Nit, may be `mongodb3-driver.version` instead of `old`?

```suggestion
    <mongodb3-driver.version>3.11.0</mongodb3-driver.version>
```

reta (approved):

instrumentation/mongodb/src/main/java/brave/mongodb/MongoDBDriver.java
```diff
@@ -0,0 +1,104 @@
...
+   * Attempt to match the driver version from two known patterns:
+   *
+   * <ol>
+   *   <li>org.mongodb:mongodb-driver - 3.x</li>
+   *   <li>org.mongodb:mongodb-driver-core - 5.x</li>
+   * </ol>
+   */
+  private static MongoDBDriver findMongoDBDriver() {
+    try {
+      Method getHost = ServerAddress.class.getMethod("getHost");
```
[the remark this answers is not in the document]
codefromthecrypt: I guess we can.. technically the old mongodb driver had a floor JRE of 1.6, but I don't think it is widely used enough in that config to matter. If someone complains, we can revert it.

instrumentation/mongodb/pom.xml
codefromthecrypt: I was using old- elsewhere, like old-okhttp old-amqp to make it obvious this isn't a prod dep ;)

version suffix is implicitly "old" like no longer maintained except in the case of spring, where two are maintained.. maybe I can update everything to a different word? anyway I don't want to be distracting, but that was the rationale. I can switch this, just there are other places where old is used.

maybe s/old/earliest or floor?

reta: All good, just a nit :D

codefromthecrypt: I think I like floor version better, so will go with that. because we don't want to have to rename everything each time the floor drops out.

codefromthecrypt:
thanks for the help @reta!

PR #1434, opened 2024-04-10
Fixes thread safety issue in Tag.tag

### LINKED ISSUE #1384: ArrayIndexOutOfBoundsException from MutableSpan
## Describe the Bug
When adding tags to an ongoing span, we occasionally see the following exception:

```
java.lang.ArrayIndexOutOfBoundsException: Index 4 out of bounds for length 4
  at brave.handler.MutableSpan.tag(MutableSpan.java:883)
  at brave.RealSpan.tag(RealSpan.java:112)
  at brave.SpanCustomizerShield.tag(SpanCustomizerShield.java:35)
  at brave.CurrentSpanCustomizer.tag(CurrentSpanCustomizer.java:41)
```

I believe this occurs in rare race-conditions in which some tags are removed while others are being added. I will attempt to reproduce this with a test, but I'm not having any luck so far. Wanted to create this issue and see if others are seeing anything similar.

codefromthecrypt:
The current codebase carefully synchronizes all access to MutableSpan where there can be concurrent use. We missed a spot where Tag.tag is not synchronizing access. The only other place we don't guard is ScopedSpan, which is documented explicitly as not thread safe.

Fixes #1384


brave/src/test/java/brave/TagTest.java
```diff
@@ -266,6 +271,27 @@ class TagTest {
...
+            return val;
+          }
+        };
+        service.submit(() -> tag.tag(input, context, mutableSpan));
+      }
+    } finally {
+      service.shutdown();
+      service.awaitTermination(1, TimeUnit.MINUTES);
+    }
+    assertThat(mutableSpan.tagCount()).isEqualTo(numThreads);
```
codefromthecrypt, line 292: if you take out synchronized, this test fails

anuraaga (approved):

reta (approved):

PR #1439, opened 2024-06-05
Remove mention of JDBC tracing

reftel:
Remove mention of JDBC tracing, since that used the P6Spy instrumentation, which was removed in 6.0.


codefromthecrypt (approved):
yep sounds right

codefromthecrypt:
will leave the commit, but after updating things for mysql this weekend, I think we do have some jdbc, just agree it is less since removing p6spy

rio123:
I am evaluating to use brave, but JDBC tracing is a must for me. I could not find anything on JDBC and brave besides this PR. Is it really not supported out of the box?

codefromthecrypt:
Right now, there are only mysql jdbc drivers in this repo. There may be other code elsewhere, just not in this repo.

Issue #1443, opened 2024-09-11
Expose http builder for Apache http5 implementation


gmellemstrand:
## Feature
With the addition of the `brave.httpclient5` package all the handling of the builder object is done internally, and the only option is to call the `build` method and get the finished client object. This happens in the `HttpClient5Tracing` class.

## Rationale
There can be many cases, when integrating with other libraries, where you want to add other options to the http builder besides what is being added by Brave. This option was also available in the http4 implementation, but was not added to the http5 implementation. Adding a new public method for this is a minimal change.

## Example Scenario
In our case we are in the process of upgrading `dropwizard-zipkin` to Dropwizard 3, which requires the use of Apache http5 (and in turn the Brave http5 implementation).

PR #1444, opened 2024-09-11
Expose builder for apache http5

gmellemstrand:
This relates to #1443 

In additon to the `build` method, I've added a new public method which adds the http tracing to the builder. This new method can be called instead of the `build` method in the builder is required in order to add additional options before building. The functionality should be identical for current users.

### DIFF
--- a/instrumentation/httpclient5/src/main/java/brave/httpclient5/HttpClient5Tracing.java
+++ b/instrumentation/httpclient5/src/main/java/brave/httpclient5/HttpClient5Tracing.java
@@ -43,15 +43,19 @@ public class HttpClient5Tracing {
     return new HttpClient5Tracing(httpTracing);
   }
 
+  public HttpClientBuilder addTracingToBuilder(HttpClientBuilder builder) {
+      if (builder == null) throw new NullPointerException("HttpClientBuilder == null");
+      builder.addExecInterceptorBefore(ChainElement.MAIN_TRANSPORT.name(),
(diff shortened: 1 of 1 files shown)


codefromthecrypt:
@reta @kojilin can you consider this when updating httpclient5 for the 5.4 break?

instrumentation/httpclient5/src/main/java/brave/httpclient5/HttpClient5Tracing.java
```diff
@@ -43,15 +43,19 @@ public static HttpClient5Tracing newBuilder(HttpTracing httpTracing) {
     return new HttpClient5Tracing(httpTracing);
   }
 
+  public HttpClientBuilder addTracingToBuilder(HttpClientBuilder builder) {
```
reta, line 46: @gmellemstrand I believe, adding additional method like this would introduce conflicting  (`addTracingToBuilder` vs `build`) instrumentation and confusion (when `addTracingToBuilder` should be used and why). I would recommend to have a factory method instead, fe:

```suggestion
    public static HttpClient5CustomTracing newBuilder(HttpTracing httpTracing, HttpClientBuilder builder) {
      return new HttpClient5CustomTracing(httpTracing, builder);
    }
```

In this case the `HttpClient5CustomTracing` type would take care of adding tracing to the provided `HttpClientBuilder` as well as having a `build()` method with no args. 

@codefromthecrypt does it make sense to you?

Issue #1445, opened 2024-10-02
Latest versions of brave-instrumentation-benchmarks are not present in maven central


Croway:
## Describe the Bug
Seems like that brave-instrumentation-benchmarks is not [released anymore](https://mvnrepository.com/artifact/io.zipkin.brave/brave-instrumentation-benchmarks), and the brave-bom cannot be built.

## Steps to Reproduce
Use the brave-bom in a plain maven project, declare brave-instrumentation-benchmarks as a dependency, and try to build the project.

brave-instrumentation-benchmarks:6.0.3 is part of brave-bom:6.0.3, but is not present on maven central.

## Expected Behaviour
All the dependencies declared in the BOM should exists.

reta:
Thank you @Croway !

PR #1446, opened 2024-10-03
Remove brave-instrumentation-benchmarks from brave-bom since this artifact is not published to Maven Central

reta:
Closes https://github.com/openzipkin/brave/issues/1445


codefromthecrypt (approved):
good eye

PR #1447, opened 2024-10-03
Routine dependency updates

reta:
Routine dependency updates


codefromthecrypt (approved):
handy no breaks.

I guess we're good until LTS Armageddon next year ;)

PR #1448, opened 2024-11-27
Note about potential deadlock with virtual threads

### LINKED ISSUE #1440: Potential deadlock between zipkin-reporter and Brave when used with Virtual Threads
## Describe the Bug

We observed a deadlock with Virtual Threads and Brave/Zipkin. In short, there are two paths to [CountBoundedQueue.offer](https://github.com/openzipkin/zipkin-reporter-java/blob/519133204bc5a56a5286c6f40facc10c76270b5f/core/src/main/java/zipkin2/reporter/internal/CountBoundedQueue.java#L71-L93) when finishing a span. [RealSpan.finish](https://github.com/openzipkin/brave/blob/69003dfc811418f0dbc42e9e17ff880ebe1f4b02/brave/src/main/java/brave/RealSpan.java#L132-L136) has a synchronized block, where [RealScopedSpan.finish](https://github.com/openzipkin/brave/blob/69003dfc811418f0dbc42e9e17ff880ebe1f4b02/brave/src/main/java/brave/RealScopedSpan.java#L62-L65
) does not. 

If an unmounted virtual thread using `RealScopedSpan` is next line for the lock, but all carriers are currently occupied by pinned VTs in `RealSpan.finish`, the application will deadlock:

https://gist.github.com/DanielThomas/dddd850f7e491cac3a2dd734978f4267

## Steps to Reproduce

See https://gist.github.com/DanielThomas/0b099c5f208d7deed8a83bf5fc03179e for a reduced example.

## Expected Behaviour

While the monitor pinning limitation will be addressed in future OpenJDK releases, in the meantime there's a good case for switching this class to `ReentrantLock` to guard the `MutableSpan` to ensure compatibility with virtual threads.

shakuzen:
Closes #1440

### DIFF
--- a/brave/README.md
+++ b/brave/README.md
@@ -948,6 +948,10 @@ Tracing brave4 = Tracing.newBuilder()...build();
 Brave brave3 = TracerAdapter.newBrave(brave4.tracer());
 ```
 
+## Virtual threads
+
+Virtual threads were added in Java 21 via [JEP 444](https://openjdk.org/jeps/444). While the Brave code makes no special consideration for virtual threads, you can generally use Brave with virtual threads. Prior to Java 24, which integrated [JEP 491](https://openjdk.org/jeps/491), it was possible for a deadlock to occur with specific conditions when finishing `Span`s and `ScopedSpan`s on virtual threads - see [#1440](https://github.com/openzipkin/brave/issues/1440).
+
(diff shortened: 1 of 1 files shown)


reta:
Thanks a lot @shakuzen !

PR #1450, opened 2025-02-12
modify logger class name for locating easily

xiaoduanayu:
modify logger class name for locating easily, it's helpful for locating the specific code with log.


codefromthecrypt (commented):
this sounds nice, but it will break folks who have used tracer as the config until now. So, I think we need more people to 👍 this one. Make sense?

PR #1453, opened 2025-02-16
Update Apache HttpClient5 to 5.4.1

reta:
Update Apache HttpClient5 to 5.4.2

### DIFF
--- a/instrumentation/httpclient5/pom.xml
+++ b/instrumentation/httpclient5/pom.xml
@@ -24,8 +24,7 @@
     <main.basedir>${project.basedir}/../..</main.basedir>
 
     <!-- To obey the integration tests, we need to use the version >= 5.2 -->
-    <!-- TODO: pinned due to timeouts on 5.4.x on brave.httpclient5.ITTracingCachingH2ClientBuilder.cacheControl -->
-    <httpclient5.version>5.3.1</httpclient5.version>
+    <httpclient5.version>5.4.1</httpclient5.version>
   </properties>
 
   <dependencies>
--- a/instrumentation/httpclient5/src/main/java/brave/httpclient5/HttpClientUtils.java
+++ b/instrumentation/httpclient5/src/main/java/brave/httpclient5/HttpClientUtils.java
@@ -58,8 +58,8 @@ class HttpClientUtils {
     if (!HAS_CLIENT_CACHE_SUPPORT) {
       return false;
     }
-    boolean cacheHit = CacheResponseStatus.CACHE_HIT == context.getAttribute(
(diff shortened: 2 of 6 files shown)


instrumentation/httpclient5/pom.xml
```diff
@@ -24,8 +24,7 @@
     <main.basedir>${project.basedir}/../..</main.basedir>
 
     <!-- To obey the integration tests, we need to use the version >= 5.2 -->
-    <!-- TODO: pinned due to timeouts on 5.4.x on brave.httpclient5.ITTracingCachingH2ClientBuilder.cacheControl -->
-    <httpclient5.version>5.3.1</httpclient5.version>
+    <httpclient5.version>5.4.1</httpclient5.version>
```
reta, line 27: @codefromthecrypt there was regression introduced into HttpCore5 5.3.3 [1] (that is used by HttpClient 5.4.2) and fixed by [2] (upcoming HttpCore5 5.3.4, I tested against the latest snapshots and things looks good). So I have pinned HttpClient5 to 5.4.1 (and we could update to 5.4.3 once released, no code changes are anticipated),
 

[1] https://github.com/apache/httpcomponents-core/commit/98e23cb82525d895691023f98ca5ebcfe43349e7
[2] https://github.com/apache/httpcomponents-core/commit/7274811b278641b46dead575f5ac5c15e052ea09

codefromthecrypt: Good research and sounds fine

instrumentation/httpclient5/src/main/java/brave/httpclient5/HttpClientUtils.java
```diff
@@ -58,8 +58,8 @@ static boolean isLocalCached(HttpContext context, Span span) {
     if (!HAS_CLIENT_CACHE_SUPPORT) {
       return false;
     }
-    boolean cacheHit = CacheResponseStatus.CACHE_HIT == context.getAttribute(
-      HttpCacheContext.CACHE_RESPONSE_STATUS);
+    final HttpCacheContext cacheContext = HttpCacheContext.cast(context);
```
codefromthecrypt, line 61: One q is.. is this type available in older versions than what we test? If not update the readme to say minimum version

codefromthecrypt (approved):

instrumentation/httpclient5/src/main/java/brave/httpclient5/HttpClientUtils.java
reta: Thanks @codefromthecrypt , the docs says `HttpCacheContext`  is` * @since 4.3` , we should be good here :)

PR #1455, opened 2025-03-12
log4j12: Automatically update snapshots in IT

reftel:
Otherwise, snapshot builds fail if one has not installed brave-tests first


reftel:
One still has to do `mvn install --projects=brave-tests` for builds to work, but without that patch, Maven will not retry fetching the `brave-tests` jar file even after it´s been put in place.

PR #1460, opened 2025-04-10
Add JDBI 3 plugin

reftel:
Adds a plugin for JDBI 3 (https://jdbi.org/ ) that sends spans via Brave for all SQL commands run. Tag names are chosen to match the old P6Spy instrumentation from the 5.x series.
Including both a Mocktio test and an integration test that uses H2. Not sure both are needed, as they test pretty much the same thing.


instrumentation/jdbi3/pom.xml
```diff
@@ -0,0 +1,79 @@
...
+
+    Copyright The OpenZipkin Authors
+    SPDX-License-Identifier: Apache-2.0
+
+-->
+<project xmlns="http://maven.apache.org/POM/4.0.0" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 http://maven.apache.org/xsd/maven-4.0.0.xsd">
+  <parent>
+    <groupId>io.zipkin.brave</groupId>
+    <artifactId>brave-instrumentation-parent</artifactId>
+    <version>6.1.1-SNAPSHOT</version>
```
reta, line 12:
```suggestion
    <version>6.2.0-SNAPSHOT</version>
```

instrumentation/jdbi3/README.md
```diff
@@ -0,0 +1,15 @@
...
+how long each statement takes, along with relevant tags like the query.
+
+To use it, call the `installPlugin` on the `Jdbi` instance you want do instrument,
+with a suitably configured `Jdbi3BRavePlugin`.
+
+By default, bind variable values are not included in the traces, for security
+reasons. If you want to include them, pass `true` as `includeBindVariables` in
+the `Jdbi3BravePlugin` constructor.
+
+Spans are never marked as error spans, even if JDBC throws an SQL Exception
```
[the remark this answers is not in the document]
reftel: In some cases, there will be an exception moment (but no access to the `Throwable`), but some errors result in neither exception moment nor completion moment.
I guess for contexts that lack a completion moment, we could mark the spans as errors, e.g. with `span.tag("error", "true")`. Would that be preferable?

reftel:
Rebased, bumped parent version, bumped jdbi version.

instrumentation/jdbi3/README.md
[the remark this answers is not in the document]
reftel: (e.g. something like https://github.com/reftel/brave/commit/e26911525cf67d0c7f9eefcfe05f7adf1225ac26#diff-61b852406fd97bfb23b48538d26e471b79214dc7af10ad6bc2a20004ca033f94R80-R82 )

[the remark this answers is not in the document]
reta: Yeah, thank you @reftel , I think that would be helpful for sure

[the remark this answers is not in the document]
reftel: Squashed that change into this one.

reta (approved):

reta:
@codefromthecrypt any concerns from you? thanks!

instrumentation/jdbi3/pom.xml
codefromthecrypt: 6.2.1 you meant 6.2.0 ;)

instrumentation/jdbi3/pom.xml
```diff
@@ -0,0 +1,79 @@
...
+        <executions>
+          <execution>
+            <phase>package</phase>
+            <goals>
+              <goal>resolve-dependencies</goal>
+            </goals>
+          </execution>
+        </executions>
+        <configuration>
+          <!-- Add dependencies specific to invoker tests so that they cache on go-offline -->
```
codefromthecrypt, line 65: remove until there is an invoker test

instrumentation/jdbi3/pom.xml
```diff
@@ -0,0 +1,79 @@
...
+
+    Copyright The OpenZipkin Authors
+    SPDX-License-Identifier: Apache-2.0
+
+-->
+<project xmlns="http://maven.apache.org/POM/4.0.0" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 http://maven.apache.org/xsd/maven-4.0.0.xsd">
+  <parent>
+    <groupId>io.zipkin.brave</groupId>
+    <artifactId>brave-instrumentation-parent</artifactId>
+    <version>6.2.1-SNAPSHOT</version>
```
codefromthecrypt, line 12: 6.2.0-SNAPSHOT, but should be universal across all files even the bom

```suggestion
    <version>6.2.0-SNAPSHOT</version>
```

instrumentation/jdbi3/src/main/java/brave/jdbi3/Jdbi3BravePlugin.java
```diff
@@ -0,0 +1,95 @@
...
+    @Override
+    public void customizeJdbi(Jdbi jdbi) {
+        if (tracing == null) {
+            return;
+        }
+        jdbi.getConfig(SqlStatements.class)
+            .addContextListener(new StatementContextListener() {
+                @Override
+                public void contextCreated(final StatementContext ctx) {
+                    String spanName = "jdbi." + ctx.describeJdbiStatementType();
```
codefromthecrypt, line 46: hmm try to match the data policy with mysql. I don't think we want to put jdbi in the span name.. https://github.com/openzipkin/brave/blob/master/instrumentation/mysql8/src/main/java/brave/mysql8/TracingQueryInterceptor.java

instrumentation/jdbi3/src/main/java/brave/jdbi3/Jdbi3BravePlugin.java
```diff
@@ -0,0 +1,95 @@
...
+                        if (includeBindVariables) {
+                            String bindVariables = ctx.getBinding()
+                                .describe(stmtConfig.getJfrParamMaxLength());
+                            span.tag("sql.binding", bindVariables);
+                        }
+
+                        span.tag("sql.rows", Long.toString(ctx.getMappedRows()));
+
+                        if (ctx.getCompletionMoment() == null) {
+                            span.tag("error", "true");
```
codefromthecrypt, line 81: we don't do error=true in zipkin. either the real error or empty string

instrumentation/jdbi3/src/main/java/brave/jdbi3/Jdbi3BravePlugin.java
```diff
@@ -0,0 +1,95 @@
...
+ */
+public class Jdbi3BravePlugin extends JdbiPlugin.Singleton {
+    private final Tracing tracing;
+    private final boolean includeBindVariables;
+
+    public Jdbi3BravePlugin(Tracing tracing) {
+            this(tracing, false);
+    }
+
+    public Jdbi3BravePlugin(Tracing tracing, boolean includeBindVariables) {
```
codefromthecrypt, line 32: if you look elsewhere you'll see types that have builders. Typically we don't add public ctors with options as it is constant maintenance. maybe find one to copy/paste find/replace

instrumentation/jdbi3/src/test/java/brave/jdbi3/Jdbi3BravePluginIT.java
```diff
@@ -0,0 +1,96 @@
...
+import org.junit.jupiter.api.BeforeEach;
+import org.junit.jupiter.api.Test;
+
+import brave.Span;
+import brave.Tracing;
+import brave.handler.MutableSpan;
+import brave.handler.SpanHandler;
+import brave.propagation.TraceContext;
+
+class Jdbi3BravePluginIT {
```
codefromthecrypt, line 24: try for naming conventions and other conventions in the project. IT prefix e.g. https://github.com/openzipkin/brave/blob/master/instrumentation/mysql8/src/test/java/brave/mysql8/ITTracingQueryInterceptor.java and also IT implies real DB. This might be a unit test? ack jdbc:h2:mem is grey area.

codefromthecrypt (commented):
I like the work, but I think we need to study a bit more other modules (or ask your favorite coding assistant to). Please use same conventions in coding and tests for the more recent similar code. you've made a good start though!

instrumentation/jdbi3/src/test/java/brave/jdbi3/Jdbi3BravePluginIT.java
reftel: It´s the use of an actual (though in-process and in-memory) database, along with using the third-party libraries as-is (as opposed to stubbing them, or parts of them, out) that made me call it an IT, but if you disagree, I´ll gladly change it.

reftel:
> I like the work, but I think we need to study a bit more other modules (or ask your favorite coding assistant to). Please use same conventions in coding and tests for the more recent similar code. you've made a good start though!

Thank you for your comments! I believe I´ve addressed them all in 17321f1db

instrumentation/jdbi3/pom.xml
codefromthecrypt: No this was my mistake. Whatever the next minor is would be the right version as we don't typically add new instrumentation in a patch

instrumentation/jdbi3/pom.xml
```diff
@@ -0,0 +1,47 @@
...
+
+    Copyright The OpenZipkin Authors
+    SPDX-License-Identifier: Apache-2.0
+
+-->
+<project xmlns="http://maven.apache.org/POM/4.0.0" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 http://maven.apache.org/xsd/maven-4.0.0.xsd">
+  <parent>
+    <groupId>io.zipkin.brave</groupId>
+    <artifactId>brave-instrumentation-parent</artifactId>
+    <version>6.2.0-SNAPSHOT</version>
```
reta, line 12: @reftel sincere apologies, the `6.2.1-SNAPSHOT` is the current version (6.2.0 already sailed), what @codefromthecrypt meant is to bump all POMs to `6.3.0-SNAPSHOT` since a new module being added. Thank you and apologies for some confusion here

reftel: No problem! =) I´ll include bumping all versions to `6.3.0-SNAPSHOT` that in the next push.

instrumentation/jdbi3/src/test/java/brave/jdbi3/Jdbi3BravePluginIT.java
reftel: I´ve changed the tests over to more-or-less a copy of the tests from the mysql8 instrumentation. I´ll make a stab at getting the tests run with both H2 and MySQL, as a way to show that the instrumentation is not tied to any particular database.

instrumentation/jdbi3/pom.xml
codefromthecrypt: cool. I'll be offline for a long weekend so if you manage before singapore time evening I can help merge. otherwise maybe @reta who is later in the globe

instrumentation/jdbi3/pom.xml
reftel: Bumped to 6.3.0-SNAPSHOT everywhere

instrumentation/jdbi3/src/test/java/brave/jdbi3/Jdbi3BravePluginIT.java
reftel: Made the test abstract, with concrete implementations that test against H2 in-memory, MySQL configured as in the mysql8 tests, Postgres via the Zonky.io extension, and Sqlite via the xerial.org extension.
In Junit 5.13, there will be a cleaner way of doing this, but it´s not released yet.

instrumentation/jdbi3/pom.xml
reftel: No stress! =) Enjoy your time off!

codefromthecrypt (commented):
last feedback!

Unless there is a main code difference, please test only one DB as an IT. We should save test time for things that improve coverage.

Finally, it is easy to miss, but brave-bom needs an update both version and also to include this module https://github.com/openzipkin/brave/blob/master/brave-bom/pom.xml#L12

reftel:
Thank you for your patience and prompt reviews! I hope all should be in order now.

> Unless there is a main code difference, please test only one DB as an IT. We should save test time for things that improve coverage.

OK, dropped everything but the MySQL one.

> Finally, it is easy to miss, but brave-bom needs an update both version and also to include this module https://github.com/openzipkin/brave/blob/master/brave-bom/pom.xml#L12

Done.

instrumentation/jdbi3/src/main/java/brave/jdbi3/Jdbi3BravePlugin.java
```diff
@@ -0,0 +1,173 @@
...
+
+        return null;
+    }
+
+    public static class Builder {
+        private final Tracing tracing;
+        private boolean includeBindVariables = false;
+
+        public Builder(Tracing tracing) {
+            this.tracing = tracing;
```
codefromthecrypt, line 161:
```suggestion
            if (tracing == null) throw new NullPointerException("tracing == null");
            this.tracing = tracing;
```

instrumentation/jdbi3/src/main/java/brave/jdbi3/Jdbi3BravePlugin.java
```diff
@@ -0,0 +1,173 @@
...
+            if (schemeSpecificUri.toString().length() < uri.toString().length()) {
+                return getRemoteServiceNameFromQueryParameter(schemeSpecificUri);
+            }
+        } catch (URISyntaxException ignored) {
+        }
+
+        return null;
+    }
+
+    public static class Builder {
```
codefromthecrypt, line 156:
```suggestion
    public static final class Builder {
```

instrumentation/jdbi3/src/main/java/brave/jdbi3/Jdbi3BravePlugin.java
```diff
@@ -0,0 +1,173 @@
...
+    }
+
+    static String getRemoteServiceName(StatementContext ctx) {
+        Connection connection = ctx.getConnection();
+        if (connection == null) {
+            return null;
+        }
+        URI uri;
+        try {
+            uri = new URI(connection.getMetaData().getURL().replaceAll("^jdbc:", ""));
```
codefromthecrypt, line 115: normally I would challenge you to make this similar to mysql code which is more efficient than regex. However, you've had a lot of back and forth so we can leave that for someone else to refine. Generally it is better to look at how we do things as often there is some performance mania going on, though not always.

https://github.com/openzipkin/brave/blob/ff4006230706a8ee357bc6ba18479faa2c3a86c2/instrumentation/mysql/src/main/java/brave/mysql/TracingStatementInterceptor.java#L76-L77

instrumentation/jdbi3/src/main/java/brave/jdbi3/Jdbi3BravePlugin.java
```diff
@@ -0,0 +1,173 @@
...
+import brave.Span;
+import brave.Tracing;
+
+/**
+ * Instrumentation for JDBI 3.
+ * <p>
+ * JDBI plugin that sends a trace span for every SQL command executed by Jdbi.
+ * Install by calling {@code jdbi.installPlugin(new Jdbi3BravePlugin(tracing));}
+ */
+public class Jdbi3BravePlugin extends JdbiPlugin.Singleton {
```
codefromthecrypt, line 30:
```suggestion
public final class Jdbi3BravePlugin extends JdbiPlugin.Singleton {
```

instrumentation/jdbi3/src/main/java/brave/jdbi3/Jdbi3BravePlugin.java
```diff
@@ -0,0 +1,173 @@
...
+        public Builder(Tracing tracing) {
+            this.tracing = tracing;
+        }
+
+        public Builder includeBindVariables(boolean includeBindVariables) {
+            this.includeBindVariables = includeBindVariables;
+            return this;
+        }
+
+        public Jdbi3BravePlugin build() {
```
codefromthecrypt, line 169: always return the most narrow type you can that works. Either plugin or plugin.singleton depending on what you think here. This is likely a place where since mysql plugin was made so long ago we didn't make do this.

```suggestion
        public JdbiPlugin.Singleton build() {
```

instrumentation/jdbi3/src/main/java/brave/jdbi3/Jdbi3BravePlugin.java
```diff
@@ -0,0 +1,173 @@
...
+
+    public static class Builder {
+        private final Tracing tracing;
+        private boolean includeBindVariables = false;
+
+        public Builder(Tracing tracing) {
+            this.tracing = tracing;
+        }
+
+        public Builder includeBindVariables(boolean includeBindVariables) {
```
codefromthecrypt, line 164: test or remove this. I get the point of it, but if you want it personally, you will want to add it to the IT to make sure someone doesn't break it later. If you don't use this personally, leave it out.

instrumentation/jdbi3/src/test/java/brave/jdbi3/ITJdbi3BravePlugin.java
```diff
@@ -0,0 +1,150 @@
...
+            .containsExactly(CLIENT);
+    }
+
+    @Test
+    void defaultSpanNameIsOperationName() {
+        prepareExecuteSelect(jdbi, QUERY);
+
+        assertThat(spans)
+            .extracting(MutableSpan::name)
+            .containsExactly("Query");
```
codefromthecrypt, line 125: I think you've misunderstood me. We shouldn't model differently than mysql. Please match your tests to what is here, as the span name shouldn't be "Query" that's not a SQL operator

https://github.com/openzipkin/brave/blob/ff4006230706a8ee357bc6ba18479faa2c3a86c2/instrumentation/mysql8/src/test/java/brave/mysql8/ITTracingQueryInterceptor.java#L109

instrumentation/jdbi3/src/main/java/brave/jdbi3/Jdbi3BravePlugin.java
```diff
@@ -0,0 +1,173 @@
...
+        this.includeBindVariables = includeBindVariables;
+    }
+
+    public static Builder newBuilder(Tracing tracing) {
+        return new Builder(tracing);
+    }
+
+    @Override
+    public void customizeJdbi(Jdbi jdbi) {
+        if (tracing == null) {
```
codefromthecrypt, line 46: remove tracing == null here to reduce branching, as it is a bug to have null here and we can guard against it.

instrumentation/jdbi3/src/main/java/brave/jdbi3/Jdbi3BravePlugin.java
```diff
@@ -0,0 +1,173 @@
...
+import org.jdbi.v3.core.statement.StatementContextListener;
+
+import brave.Span;
+import brave.Tracing;
+
+/**
+ * Instrumentation for JDBI 3.
+ * <p>
+ * JDBI plugin that sends a trace span for every SQL command executed by Jdbi.
+ * Install by calling {@code jdbi.installPlugin(new Jdbi3BravePlugin(tracing));}
```
codefromthecrypt, line 28: this drifted, also check  your README

instrumentation/jdbi3/src/main/java/brave/jdbi3/Jdbi3BravePlugin.java
```diff
@@ -0,0 +1,173 @@
...
+import org.jdbi.v3.core.statement.StatementContext;
+import org.jdbi.v3.core.statement.StatementContextListener;
+
+import brave.Span;
+import brave.Tracing;
+
+/**
+ * Instrumentation for JDBI 3.
+ * <p>
+ * JDBI plugin that sends a trace span for every SQL command executed by Jdbi.
```
codefromthecrypt, line 27: This sounds redundant. A lot of code here has had a lot of review, so you can avoid back and forth by copying it first, then revising. For example, this says what (a couple times) not why. e.g. if you look at the reference I mentioned, the javadoc says this `A MySQL query interceptor that will report to Zipkin how long each query takes` So the why is about the query. Also, we don't need to get specific to the version as that's already in the package name.

https://github.com/openzipkin/brave/blob/master/instrumentation/mysql8/src/main/java/brave/mysql8/TracingQueryInterceptor.java#L23-L31

instrumentation/jdbi3/pom.xml
```diff
@@ -0,0 +1,52 @@
...
+  <artifactId>brave-instrumentation-jdbi3</artifactId>
+  <name>Brave Instrumentation: JDBI3</name>
+
+  <properties>
+    <!-- Matches Export-Package in bnd.bnd -->
+    <module.name>brave.jdbi3</module.name>
+
+    <main.basedir>${project.basedir}/../..</main.basedir>
+
+    <embedded-postgres.version>2.1.0</embedded-postgres.version>
```
codefromthecrypt, line 25: I think there are a number of unused versions here. Also make sure you aren't declaring a version already declared elsewhere. I think you might be on `mysql-connector-java.version`. If that version isn't in the top-level pom due to lack of re-use, I would extract the one from mysql8 up to the top-level pom, then use the same variable name here.

When we don't take care to match other instrumentation, it leaves little bombs for maintainers later. For example, @reta and I have been dealing with versions for many years on code that sometimes was only polished once by the initial author (the original PR). This is why it is really ideal to look at things mentioned for review and look closely, as others are likely going to maintain it for you, and this is a perpetual task. Since it is, going the same way as other code lightens the load.

codefromthecrypt (commented):
I wanted to click approve, but really there are a lot of little things to clean up. clean up what you can. If you are out of time, then let's leave this PR open until it can be.

instrumentation/jdbi3/src/test/java/brave/jdbi3/ITJdbi3BravePlugin.java
reftel: That´s the reason I had the `jdbi.` prefix originally - the name is the Jdbi description of the statement type, not the first word of the statement. Extracting the first word of the statement seems very brittle (e.g. it fails when using comment prefixes, which is generally is good practice), and since the full SQL is in the `sql.query` tag, it is also redundant. Using the Jdbi description of the type of statement (e.g. Query, Update, or Call) seems to me to be a better way.
I can change it if you insist, but I do think it would result in a worse implementation.

instrumentation/jdbi3/src/main/java/brave/jdbi3/Jdbi3BravePlugin.java
reftel: Not sure the builder is still needed now that `includeBindVariables` is not there any longer, though.

instrumentation/jdbi3/src/main/java/brave/jdbi3/Jdbi3BravePlugin.java
reftel: Also added a simple cache, since a listener usually only ever sees one URL, and the remote service name for a given URL will always be the same.

instrumentation/jdbi3/src/main/java/brave/jdbi3/BraveStatementContextListener.java
```diff
@@ -0,0 +1,142 @@
...
+import org.jdbi.v3.core.statement.StatementContext;
+import org.jdbi.v3.core.statement.StatementContextListener;
+
+import brave.Span;
+import brave.Tracing;
+
+public class BraveStatementContextListener implements StatementContextListener {
+    private final Tracing tracing;
+    private static final Pattern serviceNamePattern = Pattern.compile("(^|&)zipkinServiceName=(?<serviceName>[^&]*)");
+    private Map<String, String> serviceNameCache = new HashMap<>();
```
reta, line 27: I suspect the `StatementContextListener` instance will be shared across many threads, should the cache be thread safe? 

```suggestion
    private Map<String, String> serviceNameCache = new ConcurrentHashMap<>();
```

reftel: Right, it probably will be shared. Switched it to a `ConcurrentHashMap`.

instrumentation/jdbi3/src/main/java/brave/jdbi3/BraveStatementContextListener.java
```diff
@@ -0,0 +1,142 @@
...
+import org.jdbi.v3.core.statement.StatementContext;
+import org.jdbi.v3.core.statement.StatementContextListener;
+
+import brave.Span;
+import brave.Tracing;
+
+public class BraveStatementContextListener implements StatementContextListener {
+    private final Tracing tracing;
+    private static final Pattern serviceNamePattern = Pattern.compile("(^|&)zipkinServiceName=(?<serviceName>[^&]*)");
+    private Map<String, String> serviceNameCache = new ConcurrentHashMap<String, String>() { };
```
reta, line 27:
```suggestion
    private final Map<String, String> serviceNameCache = new ConcurrentHashMap<String, String>() { };
```

instrumentation/jdbi3/src/main/java/brave/jdbi3/BraveStatementContextListener.java
```diff
@@ -0,0 +1,142 @@
...
+
+    String getRemoteServiceName(StatementContext ctx) {
+        Connection connection = ctx.getConnection();
+        if (connection == null) {
+            return null;
+        }
+        URI uri;
+        try {
+            String url = connection.getMetaData().getURL();
+            if (serviceNameCache.containsKey(url)) {
```
reta, line 88: Since we use `ConcurrentHashMap` now:
```suggestion
        return serviceNameCache.computeIfAbsent(url, key -> {
                if (url.startsWith("jdbc:")) {
                    // strip "jdbc:" prefix
                    url = url.substring(5);
                }
                final URI uri = new URI(url); // strip "jdbc:"
                return  extractRemoteServiceName(uri);
            });
```

instrumentation/jdbi3/src/main/java/brave/jdbi3/BraveStatementContextListener.java
```diff
@@ -0,0 +1,142 @@
...
+
+            String remoteServiceName = extractRemoteServiceName(uri);
+            serviceNameCache.put(url, remoteServiceName);
+            return remoteServiceName;
+        } catch (SQLException | URISyntaxException ignored) {
+            return null;
+        }
+    }
+
+    static String extractRemoteServiceName(URI uri) {
```
reta, line 105: Parsing JDBC URLs seems to be a large chunk of what this class is doing, probably worth extracting a separate `JdbcUrlParser `(so it could be also tested on different URLs).

reftel: There are already plenty of URL parsing tests in `shouldParseRemoteSeriveName` (which is now part of BraveStatementContextListenerTest). What other tests do you have in mind?

instrumentation/jdbi3/src/test/java/brave/jdbi3/BraveStatementContextListenerTest.java
```diff
@@ -0,0 +1,93 @@
...
+        "jdbc:postgresql://localhost:5432/testdb, localhost:5432",
+        "jdbc:postgresql://localhost:5432/testdb?zipkinServiceName=test, test",
+        "jdbc:h2:mem:testdb, h2",
+        "jdbc:sqlite:test.db, sqlite",
+        "jdbc:sqlite:test.db?zipkinServiceName=test, test",
+        "jdbc:oracle:thin:@localhost:1521:testdb, oracle",
+        "jdbc:oracle:thin:@(DESCRIPTION=(ADDRESS=(PROTOCOL=TCP)(HOST=myhost)(PORT=1521))(CONNECT_DATA=(SERVICE_NAME=myservice))), oracle",
+        "jdbc:oracle:oci:@//myhost.example.com:1521/MYSERVICE?zipkinServiceName=oracledb, oracledb"
+    })
+    public void shouldParseRemoteSeriveName(String url, String expected) throws SQLException {
```
reta, line 78:
```suggestion
    public void shouldParseRemoteServiceName(String url, String expected) throws SQLException {
```

instrumentation/jdbi3/src/main/java/brave/jdbi3/BraveStatementContextListener.java
reta: Yeah, I think my point is: `BraveStatementContextListenerTest::shouldParseRemoteSeriveName` does not test anything in `BraveStatementContextListener` besides JDBC URL parsing (for service name extraction) but needs like 5-6 mocks. The suggestion to have a dedicated parser would:
 - simplify `BraveStatementContextListener` since the large chunk of supporting code will go away
 - simplify testing by removing the need for mocking

What do you think?

instrumentation/jdbi3/src/main/java/brave/jdbi3/RemoteServiceNameResolver.java
```diff
@@ -0,0 +1,74 @@
...
+package brave.jdbi3;
+
+import java.net.URI;
+import java.net.URISyntaxException;
+import java.util.Map;
+import java.util.concurrent.ConcurrentHashMap;
+import java.util.regex.Matcher;
+import java.util.regex.Pattern;
+
+public class RemoteServiceNameResolver {
```
reta, line 14: Super minor, I think we shouldn't expose this class publicly:

```suggestion
final class RemoteServiceNameResolver {
```

instrumentation/jdbi3/src/main/java/brave/jdbi3/BraveStatementContextListener.java
```diff
@@ -0,0 +1,87 @@
...
+                span.finish(ctx.getCompletionMoment().toEpochMilli() * 1000);
+            } else if (ctx.getExceptionMoment() != null) {
+                span.finish(ctx.getExceptionMoment().toEpochMilli() * 1000);
+            } else {
+                span.finish();
+            }
+        });
+    }
+
+    String getRemoteServiceName(StatementContext ctx) {
```
reta, line 75: 👍 thank you

reta:
@codefromthecrypt mind taking another look? thank you!

PR #1464, opened 2025-05-28
jdbi3: converts to StatementLogger and follows existing style

codefromthecrypt:
This converts jdbi3 instrumentation to be more like MongoDB in implementation: configured by instance, not static like mysql connection driver. It also pares back the data policy to be the same as MySQL: this allows us a base case to move forward with. Finally, it switches implementation to a StatementLogger, so that exceptions are visible.

Infrastructure wise, this re-uses the more recent style from instrumentation here, including docker for the IT.

This is a follow-up to #1460 before we release a version, so API breaks are not an issue, yet.

Thanks to @reftel for the initial work on this, and @reta on the review of that. I did my best in RATIONALE.md here to explain why the changes are important.


instrumentation/jdbi3/src/main/java/brave/jdbi3/TracingSqlLogger.java
```diff
@@ -0,0 +1,93 @@
...
+  @Override public void logException(StatementContext ctx, SQLException ex) {
+    Span span = threadLocalSpan.remove();
+    if (span == null) return;
+    span.error(ex);
+    span.finish();
+  }
+
+  static void parseServerIpAndPort(String jdbcUrl, Span span, boolean addRemoteServiceName) {
+    // only allocate a URI once for both service name and remote address
+    URI url = URI.create(jdbcUrl.substring(5)); // strip "jdbc:"
```
anuraaga, line 74: Can consider "jdbc:".length instead of 5 since then I think don't need the comment.

Also just a double check that we don't need to actually check the prefix before trimming

anuraaga (approved):

